PR #246, opened 2014-05-25
date wrappers & translation functions

aMoniker:
Ran into these while converting underscores.

Date functions reflect wp builtins and run the same filters, but without expecting to be in the loop.

Translation functions are just useful to have as part of the twig environment.

### DIFF
--- a/functions/functions-twig.php
+++ b/functions/functions-twig.php
@@ -101,6 +101,16 @@ class TimberTwig {
 		$twig->addFunction('__', new Twig_SimpleFunction('__', function($text, $domain = 'default'){
 			return __($text, $domain);
 		}));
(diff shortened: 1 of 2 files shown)


lmartins:
Im also having issues with this.
I have a filter passing a translated string "Read More" to the post excerpts but despite being translated on my translation files it won't get picked up on the front-end. 

Any news about this?

gchtr:
Timber has added all relevant translation functions [quite some time ago](https://github.com/timber/timber/pull/918) and it looks like a part of this pull request has landed in core [in the 0.19.1 release](https://github.com/timber/timber/blame/e6e50d6ad1349913adcec4f2f542dfbb7d497415/functions/timber-post.php#L446-L475). Thanks for leading the way!

PR #281, opened 2014-07-03
Timber template hierarchy

mgmartel:
Now that the query is added to the context by default (#277), the WordPress template files often just repeat the same melody.. get context and render a twig file with pretty much the same name as the php template.

So I quickly put together this PR (which is a proof-of-concept for discussion), which adds a template loader to the default WP template loading that loads twig templates according to the same logic as the WP template hierarchy if no php template is found.

It checks what the query is for, and checks if an appropriate twig template exists in any of the Timber locations, exactly as the WP Template hierarchy does. If a twig template is found, it is rendered using
the default context (`Timber::get_context()`).

For example, if the query is for a single post of the `portfolio` post type, the following files will be looked for:
* `child-theme/single-portfolio.php`
* `parent-theme/single-portfolio.php`
* `child-theme/single.php`
* `parent-theme/single.php`

..and then..

* `child-theme/views/single-portfolio.twig`
* `parent-theme/views/single-portfolio.twig`
* `child-theme/views/single.twig`
* `parent-theme/views/single.twig`

...and then...

* `child-theme/index.php` (NB!)
* `parent-theme/index.php`
* `child-theme/views/index.twig`
* `parent-theme/views/index.twig`

As a proof-of-concept, I have included a naked version of the starter theme in this PR, so you can see that the sites can actually run without template files :)

A few notes:
* Themes are required to have an `index.php`, so the code in the PR checks the `index.php` in a hacky way: if `{template}.twig` is not found, it renders the found `index.php` and sees if there is any output. If there is output, it will use `index.php`. If there's no output, it will continue looking for `index.twig`. (there is a filter to override this) We may have to think this through.
* I added a filter on the context as loaded by the template loader. That way, developers can choose to do away with php template files while still adding contextual context (ahem.) by filtering template loader context and adding stuff based on the global query vars (for example, author info on author archives).
* The code filters on the `get_index_template` function. If a timber template is found, it is rendered and the filter returns `false`. WP will then simply not include a template file... unless a plugin has filtered `template_include` to use its own templates (WooCommerce? BuddyPress?). This is something test and perhaps find solution for...

Let me know if you agree this is a valid strategy, then I'll work it out some more and write tests.


lkraav:
Did any of this ever make it to master? Seems like groundbreaking stuff, but got stalled because of some difficulties in the end?

mgmartel:
@lkraav We got quite far on this, but I don't think this ever made it into master. Since my last message I haven't had much time to contribute to Timber, unfortunately.

@jarednova Shall we use this opportunity to pick up the thread again? As far as I can remember, my last commit was already fully functional and just needed some further testing and tweaking. How much has master changed since? Close enough to make a plan for merging this?

I'll be quite busy the next couple of weeks, so I won't be of much help until the last week of January. If you're still keen to get the template loader of the ground, how about making a plan to boot this again around that time?

mgmartel:
Hi @jarednova,

We have left this one hanging for much too long now! I have rebased `template-autoloader` on `master` and made sure everything is still working. All unit tests are still passing and all my Timber projects are still working as they should, so this is still good to go as far as I'm concerned.

It's been a while, so here is a recap of what's included in this PR:

* Template loader to automatically load the correct of `twig` template with default context according to template hierarchy (making `php` templates optional)
* Template autoloader is disabled by default, can be enabled by setting `Timber::$twig_template_hierarchy = true;`
* WooCommerce compatibility, without header.php / footer.php hack
* Manual render call for `php` templates with autoloader possible with simplified syntax: `Timber::render( $context );` (no need to pass in the `$filenames` paramter)
* Themes can include a `context.php` file, which will be used to load default theme context. The $context array is directly available in `context.php`.

At the time, I made an opinionated started theme for Timber that relied heavily on the template autoloader. I didn't push it to GitHub then because I figured I'd wait for this to be merged into `master`. I spent some time reviving it, and I think it's also pretty powerful stuff. If anything, it's a good demo of what the template autoloader can do. I finally pushed it to GitHub just now: https://github.com/mgmartel/timber-regions

I used it to build a proof-of-concept Foundation theme based on Timber, which I'll also push some time this week.

What are your thoughts about bringing this to `master`?

Mike

lkraav:
@jarednova is this door shut now? what triggered the close?

jarednova:
That was not intentional. I cleaned-up some stale branches this AM, but [this one](https://github.com/jarednova/timber/tree/mgmartel-template-autoloader) says it's still open. Hmm......

mgmartel:
Hi @jarednova,

Accidental close again? Just finished another Timber site using the template loader ([here](http://borneoadventure.com)) and still convinced that this is the way to go for theme development with Timber. The theme is mostly twig templates, as opposed to a mess of PHP files. 

I just added some more commits to this branch:
- Added support for page templates (this needs a PHP counterpart to show up in the page template selector, but it doesn't have to contain any code)
- WP 4.3 [introduced singular.php](https://make.wordpress.org/core/2015/07/06/singular-php-new-theme-template-in-wordpress-4-3/), this now has `singular.twig` support

Do you have any plans of merging this into master? If so, I'll rebase again!

Cheers,
Mike

Issue #217, opened 2014-05-06
Password Protected Posts / Pages


Jaace:
This functionality doesn't seem to be working. I believe it's tied into the loop in WordPress and because Timber avoids the loop it doesn't adhere to the normal password flow.

I'm not exactly sure how WordPress implements this functionality in the loop but I do see that when password is set on a post it is sent through post.post_password. This is only one part of the solution though as a form is usually shown and a password is asked for before the content is visible.

felipearosemena:
What's the status on this issue? Is there any plan to enable the default Wordpress password protected form?

connorjburton:
Hey,

We will most likely be looking at this post 1.0, which will land in the next couple of weeks.

jeremyzahner:
@connorjburton Any news on this?

felipearosemena:
@connorjburton +1

hrsetyono:
@connorjburton +1. Would appreciate this feature or temporary solution. Thanks

jarednova:
Hey @hrsetyono @felipearosemena @jeremyzahner @Jaace — check-out #1170

jarednova:
Closed by #1170 — release on Tuesday!

PR #285, opened 2014-07-09
Return password form if content requires password

aduth:
Mirroring the [default WordPress `get_the_content` behavior](https://core.trac.wordpress.org/browser/tags/3.9.1/src/wp-includes/post-template.php#L216), Timber's `get_content` and `content` should return the password form for unauthenticated users. There is more discussion about how this is currently handled in #217. I'd argue that it would be ideal for the developer to not concern themselves with conditionally handling password authentication in each template.

I had some issues getting VVV running in an attempt to verify the included unit test, so please verify on my behalf.

### DIFF
--- a/functions/timber-post.php
+++ b/functions/timber-post.php
@@ -672,6 +672,9 @@ class TimberPost extends TimberCore implements TimberCoreInterface {
      * @return string
      */
     function get_content($len = 0, $page = 0) {
+        if (post_password_required($this)) {
+            return get_the_password_form($this);
+        }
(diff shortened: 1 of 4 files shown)


orest22:
@jarednova I have the same problem {{post.content}} do not show password form when page is protected with password.

jarednova:
@aduth thanks for your work on this — I've picked this up at #1170 in a way that's backwards compatible but still covers the use case that you (and many others) have been seeking

PR #368, opened 2014-10-17
Simplyfing Twig Filter and Function setup

alexsomeoddpilot:
- Iterating over arrays instead of calling `new Twig_SimpleFunction` for each
- Moving helper functions out to TimberTwigHelper
- Using factories for TimberObjects instead of repeating same factory


gchtr:
A couple of years later we arrived at the same conclusion in #2408. Thanks for being ahead of the times!

PR #394, opened 2014-11-19
Image resize did not work on WPML sites

xavierpriour:
Hi,

tl;dr: Using the resize filter on a site with WPML translations resulted in broken resized image links, I fixed it (and added non-regression tests).

More details:
- WPML works by changing home_url() from 'example.com' to 'example.com?lang=en'
- timber resize used to build resized url by doing home_url().'...path...'
- mixing the 2 resulted in broken url of 'example.com **?lang=en** /wp-content/.../img-300x500-c-default.jpg'
- I updated resize to use content_url(), which WPML does not touch
- I very slightly refactored timber-image-helper in the process.

Thanks

### DIFF
--- a/functions/timber-image-helper.php
+++ b/functions/timber-image-helper.php
@@ -192,23 +192,6 @@ class TimberImageHelper {
      * @param string  $crop
      * @return string
      */
-    static function get_resize_file_rel( $src, $w, $h, $crop ) {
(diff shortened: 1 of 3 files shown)


jarednova:
Thanks much @xavierpriour. I finally got my hands on a dev copy of WPML to test against. The code and the tests were :+1:.

PR #421, opened 2014-12-19
Update timber-archives.php

marciojcoelho:
$show_year is always null, change to $show_year = $r['show_year'];


jarednova:
@marciojcoelho turns out the error was un-related to your change (rather, something introduced by WordPress 4.1 and the twentyfifteen theme). This change is clear!

Issue #441, opened 2015-01-09
Store cache data in the wp-content/cache directory


tmoitie:
This directory is typically better set up for permissions and would therefore be more appropriate a location.

aristath:
:+1: 
Or maybe even on `wp-content/uploads/timber-cache` 
Basically anything would be better than saving caches in the plugin folder. :smile:

aristath:
Fixed in https://github.com/jarednova/timber/pull/450

Now you can use the `timber/cache/location` filter to change the location where cached data is stored.

example:
```php
/**
 * Change Timber's cache folder.
 * We want to use wp-content/cache/timber
 */
function custom_change_twig_cache_dir() {
  return WP_CONTENT_DIR . '/cache/timber';
}
add_filter( 'timber/cache/location', 'custom_change_twig_cache_dir' );
```

connorjburton:
This is solved so I'm gonna close this.

PR #450, opened 2015-01-22
added a timber_cache_loc filter

aristath:
Add a filter that will allow us to change the location of twig cached files.

This way if someone wants to change the default location and store the caches say on wp-content/cache (see #441) or anywhere else they'll be able to do it.

### DIFF
--- a/functions/timber-loader.php
+++ b/functions/timber-loader.php
@@ -237,7 +237,7 @@ class TimberLoader {
             Timber::$twig_cache = true;
         }
         if (Timber::$twig_cache) {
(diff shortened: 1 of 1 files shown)


jarednova:
Thanks @aristath! I've been going through and updating WP hooks to follow `timber/this/format` which I've found much cleaner in dealing with other plugins. Thus, I've changed `timber_cache_loc` to `timber/cache/location` -- let the hooks begin!

aristath:
Awesome, thanks!

And yes, I do agree that the new naming makes more sense. :+1:

PR #463, opened 2015-02-06
Fix error on string concatenation, fixes #462

### LINKED ISSUE #462: error in timber-posts-collection.php
Hi there, i am getting an error on line 26 of timber-posts-collection.php:

``` php
TimberHelper::error_log( $post_type . ' of ' . $rid . ' not found in ' . print_r( $post_class, true ) );
```

I think this is a bug? `$rid` is not a string, it is a `WP_Post`, so concatenation in this way will fail. Changing it to `$rid->ID` fixed this for me.

quinn:
fixes #462

### DIFF
--- a/functions/timber-posts-collection.php
+++ b/functions/timber-posts-collection.php
@@ -23,7 +23,7 @@ class TimberPostsCollection extends ArrayObject {
 
                 } else {
                     if ( is_array( $post_class ) ) {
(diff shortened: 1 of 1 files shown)


jarednova:
Thanks @quinn -- I think the root problem was those variable names weren't any help in documenting what was going on. In addition to your change, I clarified the vars

PR #466, opened 2015-02-11
Show website name if wp_title is empty

JeyKeu:


jarednova:
Thanks @JeyKeu! I flipped the logic to make the `if` test for truthiness -- otherwise, looks good!

JeyKeu:
Awesome. Thanks.

PR #469, opened 2015-02-12
added getter for comment_status

simonmilz:
Currently there is no method to get the comment_status. I could be useful to create a frontend that show all comments:

```html
<tr class="{{ comment.status }}">
    <td>{{ comment.date | date('d.m.y - h:i:s') }} Uhr</td>
    <td>{{ comment.author.name }}</td>
    <td>{{ comment.content }}</td>
</tr> 
```


jarednova:
Looks good, thanks @simonmilz !

PR #476, opened 2015-02-17
Php router alpha

jarednova:
Cleaned-up usage of PHP router which should help other projects that include Timber via composer. This will also resolve #409


austinpray:
:sunglasses:

PR #477, opened 2015-02-17
added static get_loader method

simonmilz:
I removed the duplicated code in the "compile" and "get_sidebar_from_php" methods and put it in the "get_loader"-method Also, you are no able to get the loader in your theme to check if a template exists:

```php
    $loader = Timber::get_loader();
    if ($loader->template_exists('my-file.twig')) {
        // ...
    }
```

I think the $caller_file is unused, or am i wrong?

### DIFF
--- a/lib/timber-user.php
+++ b/lib/timber-user.php
@@ -182,6 +182,14 @@ class TimberUser extends TimberCore implements TimberCoreInterface {
     function slug() {
         return $this->user_nicename;
     }
+    
+    
+    /**
+     * @return string
+     */
+    function email() {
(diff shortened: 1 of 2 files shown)


simonmilz:
Now i have to do sth like this:

```php
    $loader = new TimberLoader();
    if ($loader->template_exists('views/my-file.twig')) {
        // ...
    }
```

this works in my case, but not if i use a child theme or more directories with views.

simonmilz:
I also added a getter for user_email. This is useful if you are working with multiple user for example in a forum or something.

Issue #259, opened 2014-06-12
Version 1.0 Epic


jarednova:
After about a year of active development, it's appropriate to give Timber a 1.0 release, why?

- The API has stabilized. A minimum number of breaking changes have been introduced in the last few months, and of those the vast majority have been minor ones.
- Test coverage exists over 50% of the code
- Timber is in use by [hundreds of sites](http://upstatement.com/timber#showcase)
- [Other projects](http://wpmu.io/blog/rethinking-shoestrap/) are now using Timber as the underlying layer for template
- The GitHub and WP.org channels have solidified as the dev/distribution channels

However, before 1.0 there are several things that need to be completed

- [x] Update Starter Theme with best practices (@ricwa230)
	- _Consider moving starter theme's official home into its own repo, with linkage back to main Timber_
- [x] Request the [Timber page](http://wordpress.org/plugins/timber/) from the WP.org admins
- [x] Launch new project homepage with updated logo
- [x] Ensure that all classes have at least "medium" coverage (50%) (#264)
- [x] Consider separating Routes into its own repo, with support in Timber (#273)
- [x] Audit objects for universal use of magic methods, meta, id, name, title, slug (etc.) (#272)
- [x] Create a 1.0 upgrade guide noting deprecated functions/properties
- [x] Remove deprecated functions (#263)
- [x] Name Spacing of Timber classes _**Contributor wanted**_
- [x] Reduce clutter of main Timber class, implement TimberPostGetter for getting stuff (#262)
- [ ] Resolve Partial Inheritance _**Contributor wanted**_
> Re:namespacing & autoloading: yes! Yes! Yes! :) This is also a good occasion to think about the problem of partial inheritance: make it easier to extend Timber objects, without having to extend the lot. Eg. if you want to extend TimberPost, but not TimberImage, you still have to extend TimberImage and set TimberImage::$post_class to 'MyTheme\Post', because otherwise you'll end up with an ordinary TimberPost instead of your own postclass when you call TimberImage::get_parent. [@mgmartel on #161](https://github.com/jarednova/timber/issues/161#issuecomment-35796577)

- [x] Implementation of WordPress coding standards (http://make.wordpress.org/core/handbook/coding-standards/php/) where appropriate.

- [x] Autoloading classes _**Contributor wanted**_
[...]

PR #485, opened 2015-03-01
Autoloading

mgmartel:
Completely unrelated to the template autoloader, here is an old remnant from #50 (Implement Composer) that was still on the todo list for the big 1.0 (#259). This PR implements very simple class autoloading through a classmap, courtesy of Composer.

By adding a classmap autoloader to `composer.json`, Composer will scan the `lib` dir whenever you run `composer install / update / dump-autoload` and generate a classmap based on all the classes it finds. 

This means that the file containing the class will be automatically included when you use the class for the first time. This means that class files are only included when they are actually needed, improving performance (yay).

This also means that initialization code can't run in the class files (as they aren't included by default any more), so where appropriate, I added a static `init()` method that runs the initialization code for the class. These are all called in `Timber::init()`, which in turn is called in `Timber::__construct()`. As a bonus it's now much clearer what code is actually run during Timber's init.

Lastly, I also got rid of the global vars besides `$timber` because they were never really needed. `$timber_admin` was only used in the unit test and could not be usefully used outside of the plugin, so I rewrote the test and removed it. `$timberRoutes` was only used to call the route object from `Timber`. Instead I installed the `TimberRoutes` object in `Timber::$routes`.

Ideally, Timber moves to namespaces and proper autoloading (PSR-4/PSR-0) mechanisms. But I think this will be quite a big job and should be paired with a larger refactor of the class structure. For now, this gets rid of all those `require`s!


jarednova:
Ahhh! So nice to nice to get rid of that stack of dependency loading! I'm reviewing now.....

jarednova:
Added a few bits of docs at the top of the new class files. Thanks much! The loading all feels so much cleaner

Issue #478, opened 2015-02-17
Potential loop in TimberCore - rethink purpose?


mgmartel:
Hi @jarednova,

There seems to be a potential nasty loop in `TimberCore`: the `__get` method checks `$this->meta( $field )`. If the `meta` method does not exist, the magic method `__call` is called, which in turn calls `__get`. You see where this is headed?

Now, the `TimberCoreInterface` _does_ require a `meta` method, so all objects implementing `TimberCoreInterface` are good. The ones in trouble are the ones that 
- extend `TimberCore`
- don't implement `TimberCoreInterface`

That is:
- `TimberArchives`
- `TimberMenu`
- `TimberTheme`

`TimberSite` does implement `TimberCoreInterface`, but defines the `meta` method to simply call `__get` as well. So it's part of the list.

I think that the problem stems from the fact that TimberCore has 2 purposes:
1. Be a generic object wrapper for all-things-Timber, because it exposes handy `__get` and `__call` magic methods, as well as `import`, etc.
2. Be a wrapper for WordPress objects, where the `__get` and `__call` also interact with object meta.

My guess is that it's best to split the two purposes up into two objects. Something like:

```
-TimberCore (or TimberObject?)
--TimberWPObject (or something..)
```

Where the main addition of WP objects is that they all have object meta. Both classes should implement their own interface, so that objects that do not extend either of them can still 'plug in' my implementing the same interface.

What do you think?

palmiak:
@mgmartel @connorjburton what is the status of this one? I see it has-patch, wasn't merged, but i has over 2 years.

jarednova:
@palmiak I think this remains a live issue. The contradictions and confusion could be (relatively) easily sorted out by re-applying #486 to the current codebase. That said, I think we would want to target it at 2.x

gchtr:
There’s not only a problem with non-existing methods, but with all inaccessible methods. In this part:

https://github.com/timber/timber/blob/3947bf995e30ff836e87354a6a0119bc47fd0afd/lib/Core.php#L38-L48

When `$field` is a the name of **a private method** of an object, then `if ( method_exists($this, $field) ) {` will create an infinite loop, because like mentioned, `__call()` will call `__get()` which will again call `__call()`.

[...]

PR #486, opened 2015-03-01
Add TimberWPObject

mgmartel:
Hi Jared! This PR implements my suggestion for #478. It splits up `TimberCore` and `TimberCoreInterface` into:
- _`TimberCore` & `-Interface`_ For objects that can be used in Twig templates (implementing the magic `__get` and `__call` methods for handiness)
- _`TimberWPObject` & `-Interface`_ For actual WordPress objects that implement object meta: posts, comments, users and terms.

I'm not sure if TimberWPObject is the correct name. If so, then maybe the correct name for `TimberCore` is `TimberObject`. After all, it's a backbone for classes that can be used as objects in timber templates. What do you think?

Also, do you reckon this is the right way to deal with consistent interfaces for Timber object?


mgmartel:
Forgot to include the extra requires in my commits last night. 
But now that master autoloads class files, it's no longer necessary to require the new class files manually, so I merged this branch with master. It should work now!

Mike

PR #489, opened 2015-03-03
Animated GIF TimberImage override

josephbergdoll:
Little override to allow animated GIFs to still play even when using TimberImage to downsize…

### DIFF
--- a/lib/timber-image.php
+++ b/lib/timber-image.php
@@ -97,7 +97,14 @@ class TimberImage extends TimberPost implements TimberCoreInterface {
         }
 
         if ($size && is_string($size) && isset($this->sizes[$size])) {
-            $image = image_downsize($this->ID, $size);
+            $imageUrl = wp_get_attachment_url($this->ID);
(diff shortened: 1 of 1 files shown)


josephbergdoll:
probably not sensible to include in core but might be good to leave hanging for those in need of the fix?

jarednova:
@josephbergdoll: I'm doing a bunch of PR review right now (including fix-up on how gifs are handled overall). Can you demo how your patch is meant to be handled in twig?

jarednova:
@josephbergdoll I merged in part of your PR here. After spending time with it, I think your fix (exempting GIFs from resizing to prevent an animated gif from losing its animation) makes sense via a filter. If you'd like to resubmit a PR that introduces filter hooks (that would let a theme or plugin developer exempt GIFs from these functions), I'd gladly pull it into core. 

That sort of behavior should allow Timber to interact with plugins like this: https://wordpress.org/plugins/animated-gif-resize/

PR #490, opened 2015-03-04
Changes visibility to protected on TimberPost::check_post_id()

osadi:
This method is usable for classes extending TImberPost. Setting it as protected still hides it from being called on the instance.


jarednova:
Looks good, I agree with making these [privates become protected](https://10up.github.io/Engineering-Best-Practices/php/#visibility) -- thanks for the spelling fix too!

Issue #483, opened 2015-02-24
Composer cannot resolve an installable set of packages.


tmoitie:
```
Problem 1
    - Installation request for jarednova/timber dev-master -> satisfiable by jarednova/timber[dev-master].
    - jarednova/timber dev-master requires jarednova/php-router 1.0.1 -> no matching package found.
```
`jarednova/php-router` is not in packagist, and also https://github.com/jarednova/PHP-Router does not have a tag release for 1.0.1 so I cannot force the issue with a vcs repository.

I've noted there is an explicit repository inside your composer.json but from the outside my install never gets the chance to look through it before it throws the unresolvable error

xemlock:
+1. This issue should be marked as a bug, as it makes timber not installable via composer.

PR #491, opened 2015-03-05
New routes

jarednova:
This solves #483 and separates the routing logic into its [own repo](upstatement/routes). Routing is now based on AltoRouter instead of PHP-Router


SheilaRuth:
Does this mean that upgrading Timber in WordPress will break my routes-based plugin? Is there backwards compatibility, or do all routes need to be converted when upgrading? I looked at the routes repo and there isn't much documentation there.

aristath:
@SheilaRuth Routes is now a separate plugin: https://github.com/jarednova/routes

jarednova:
@SheilaRuth @aristath the repo is located at:

https://github.com/upstatement/routes

... @SheilaRuth: everything is still backwards compatible. However, it's officially depreciated so I recommend migrating to the new plugin for new work.

SheilaRuth:
OK, thanks!

Issue #182, opened 2014-03-21
Comments Pagination


matthewsoares:
Is there any support for comments pagination? Also Nested Comments?

lmartins:
Was also looking for a solution for displaying comments showing their hierarchy.

jarednova:
Closed by #554

PR #554, opened 2015-06-02
Add support to sub comments in TimberComment & TimberPost

VincentLoy:
Hello ! This is my suggestion for #182 

This add an attribute to TimberComment class (an array of his sub comments).


jarednova:
Thanks @VincentLoy -- I'm pulling this down to add some tests before mergin

VincentLoy:
Nice ! :)

PR #569, opened 2015-06-11
Store the menu item level as a param for ease of access

newkind:


newkind:
Closing this pull request and opening a new one because of some other pull request that was made by accident on Jan 19

PR #570, opened 2015-06-12
Store the menu item level as a param, custom Read More labels from <!--more--> tag, support for shortcodes when get_preview uses content

newkind:
This is for ease of access and can help with creating hierarchical structures

### DIFF
--- a/lib/timber-menu.php
+++ b/lib/timber-menu.php
@@ -150,8 +150,10 @@ class TimberMenu extends TimberCore {
 		foreach ($index as $item) {
 			if (isset($item->menu_item_parent) && $item->menu_item_parent && isset($index[$item->menu_item_parent])) {
 				$index[$item->menu_item_parent]->add_child($item);
+				$item->level = $index[$item->menu_item_parent]->level + 1;
 			} else {
 				$menu[] = $item;
+				$item->level = 0;
 			}
 		}
 		return $menu;
--- a/lib/timber-post.php
+++ b/lib/timber-post.php
@@ -214,8 +214,8 @@ class TimberPost extends TimberCore implements TimberCoreInterface {
 				$text = $this->post_excerpt;
 			}
 		}
-		if (!strlen($text) && strpos($this->post_content, '<!--more-->') !== false) {
-			$pieces = explode('<!--more-->', $this->post_content);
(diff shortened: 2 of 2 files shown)


jarednova:
This is great @newkind -- the next thing we need before I can merge is a test to verify the behavior. Are you familiar with (or interested in getting acquainted with) PHPUnit + unit tests?

newkind:
@jarednova sure! i'm always happy to learn something new! i just set up the VVV but I'm having issues with getting PHPUnit to run properly with Timber. I followed the wiki instructions but they seem incomplete. Right now I'm struggling with `PHP Fatal error:  require_once(): Failed opening required '/tmp/wordpress-tests-lib/includes/functions.php'` Do I need to install wordpress tests separately ? I tried also going with `./install-wp-tests.sh wordpress_unit_tests wp wp` but looks like the http://wordpress.org/wordpress-master.tar.gz returns 404 - I might need to do some script edits ;)

jarednova:
Hmmmm, it's been a loooooong time since I've set stuff up. Let me run through the instructions and get those in better shape so you can have at it.

KLVTZ:
@newkind @jarednova to setup testing on your local, I suggest grabbing http://wp-cli.org/
After correct install on a proper environment, check out my quick summary guide via:

https://github.com/KLVTZ/WordPressHackersGuide/blob/master/08-unit-testing-in-wordpress/17.wordpress-testing.md

KLVTZ:
@newkind @jarednova keep in mind that the above guide uses the param argument found in scaffold:
https://github.com/KLVTZ/WordPressGalaxy/blob/master/public/content/plugins/justin/bin/install-wp-tests.sh#L4

newkind:
Hmm I have added a new piece of code to detect custom read more labels in the  tag, but I'm not sure why it got added by this pull request ... Anyways, another pull request is here ;)

jarednova:
Thanks @newkind! I've merged the menu parts in and moved the stuff about shortcodes, etc. into #602 for further review and testing

newkind:
No problem @jarednova :) I'm planning to do some more pull requests as I'm going through some Timber stuff and think that allowing a little bit more of user customization here and there would be a good thing.

PR #572, opened 2015-06-13
Xavierpriour master1

jarednova:
My own changes to #544 in order to pass tests


jarednova:
Thanks for your contribution @xavierpriour! I wanted to make sure I fully understood the implications which is why it took me a few days to ensure everything worked on my variety of testing setups

PR #591, opened 2015-07-03
fixed property visibility in timber-menu

prisis:
Now we can extend timber-menu.


jarednova:
Yup! Thanks for the PR

PR #602, opened 2015-07-09
570 shortcodes and moretags

jarednova:
This is @newkind's commits from #570 relating to `get_preview`, `<!-- more -->` tags and shortcodes. I'm moving it into its own PR for discussion and testing

### DIFF
--- a/lib/timber-post.php
+++ b/lib/timber-post.php
@@ -225,9 +225,11 @@ class TimberPost extends TimberCore implements TimberCoreInterface {
 				$text = TimberHelper::trim_words($text, $len, false);
 				$trimmed = true;
 			}
+			$text = do_shortcode($text);
 		}
 		if (!strlen($text)) {
 			$text = TimberHelper::trim_words($this->get_content(), $len, false);
+			$text = do_shortcode($text);
 			$trimmed = true;
 		}
 		if (!strlen(trim($text))) {
--- a/lib/timber-post.php
+++ b/lib/timber-post.php
@@ -218,32 +218,32 @@ class TimberPost extends TimberCore implements TimberCoreInterface {
 				$text = $this->post_excerpt;
 			}
 		}
-		if (!strlen($text) && preg_match('/<!--\s?more(.*?)?-->/', $this->post_content, $readmore_matches)) {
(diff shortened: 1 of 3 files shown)


jarednova:
@newkind can you explain how the `<!-- more -->` tag implementation you setup here differs from the previous? It'd be very helpful to see an example of how a prior usage would have failed with the existing code so I can evaluate the new one

newkind:
Sure :) 
So WordPress (and Timber) allows you to set the custom label for the read more button ie. Read More. However when using WP Core native the_content you can also define the label on a per post basis by doing `<!--more I'm important! Read Me!-->` in the post content. 
If WordPress (and now Timber) will find such modified tag in the post content, they will display `I'm important! Read Me!` as a button label instead of `Read More`, but only for that one post.

Sample :
- Post 1
In condimentum facilisis porta. Sed nec diam eu diam mattis viverra. Nulla. <a>Read More</a>

- Post 2
In condimentum facilisis porta. Sed nec diam eu diam mattis viverra. Nulla. <a>I'm important! Read Me!</a>

- Post 3
In condimentum facilisis porta. Sed nec diam eu diam mattis viverra. Nulla. <a>Read More</a>

- Post 4
In condimentum facilisis porta. Sed nec diam eu diam mattis viverra. Nulla. <a>Read More</a>

jnweaver:
I was just looking at this too. I think the following would be a test for this feature (which is new to me!).

```php
// content with <!--more --> tag with custom read more defined
$post->post_content = 'this is super dooper<!--moreMy read more--> trooper long words';
$preview = $post->get_preview(2, false);
$this->assertRegExp('/this is super dooper <a href="http:\/\/example.org\/\?p=\d+" class="read-more">My read more<\/a>/', $preview);
```

jnweaver:
I've also been looking at the `do_shortcode` additions on lines 228 and 232. The first one – happening within the test for the  tag – I think will work. The second one however will not I think work because `TimberHelper::trim_words($this->get_content(), $len, false)` will already execute the short code (within *TimberPost::get_content()* because *the_content* filter gets applied). But *trim_words* also strips all tags but *p a span b i br* so most shortcode output (e.g. *caption*, *gallery* etc) will get stripped out.

PR #613, opened 2015-07-16
Passing ID thru to 'timber_image_src' filter

aaronhipple:
We're using [WP Offload S3](https://wordpress.org/plugins/amazon-s3-and-cloudfront/) to serve images through S3/CloudFront, and it works most of its magic through hooks in `wp_get_attachment_url()`. Since Timber never calls this function, we're currently using the timber_image_src hook to rewrite the URL string, but that's not super robust.

Passing the ID through as an arg to this filter makes compatibility super simple. Does this seem like a sensible addition? Alternately, any thoughts on a better way to do this?

### DIFF
--- a/lib/timber-image.php
+++ b/lib/timber-image.php
@@ -113,8 +113,8 @@ class TimberImage extends TimberPost implements TimberCoreInterface {
 		$base = ($dir["baseurl"]);
 
 		$src = trailingslashit($this->_maybe_secure_url($base)) . $this->file;
-		$src = apply_filters('timber/image/src', $src);
-		return apply_filters('timber_image_src', $src);
+		$src = apply_filters('timber/image/src', $src, $this->ID);
+		return apply_filters('timber_image_src', $src, $this->ID);
 	}
 
 	private static function _maybe_secure_url($url) {


jarednova:
@aaronhipple this makes sense to me, my only question is whether it makes more sense to pass the entire object through like this....

```php
$src = apply_filters('timber/image/src', $src, $this);
```

... do you know if there's a general WP guideline for whether to use the ID or the entire object in a hook/filter like this?

aaronhipple:
@jarednova I like the idea (though I'm not really seeing a use case at the moment), but I'm not sure on guidelines. Examples from WordPress core tend to pass through the ID only, as I'm looking around. The `wp_get_attachment_url` filter itself, for example, is applied like so:

```php
$url = apply_filters( 'wp_get_attachment_url', $url, $post->ID );
```
[(source)](https://core.trac.wordpress.org/browser/tags/4.2.2/src/wp-includes/post.php#L5008) But that's not necessarily a good reason to do it one way or another if the aim is to bring better coding paradigms to WordPress.

jarednova:
@aaronhipple thanks for that info. My feeling is to follow WP until there's a good reason to deviate. In this case, a dev would have access to both key data pieces of the src and the ID. Could they need access to something like the `author_id` or the `post_date`? Sure? I guess, but they could also get this by just initializing the object with that ID. 

All of that is a long way to say, let's merge!

aaronhipple:
Yeah. My initial thought was that it could come in handy if somebody were using an extended `TimberImage` object, but if you're doing that it's probably better to write the functionality into your own methods than to fuss with hooks anyway. Thanks!

PR #633, opened 2015-07-31
Fix for Timber ignoring default_comments_page and missing pages in comment pagination

newkind:
Hi,

This is a piece of code from WP core (`comments_template()`) required for Timber to respect properly the default_comments_page setting from the Discussion settings in WP. Without this comments pagination will be broken, missing comments pages can happen (I was getting only 2 pages instead of 3) and first comments would be displayed not accordingly to what's saved in the Discussion settings.

Cheers

P.S Contains my other pull request to add `blockquote` to allowed_tags in `trim_words` and possibility of filtering the `$allowed_tags`

### DIFF
--- a/lib/timber-helper.php
+++ b/lib/timber-helper.php
@@ -172,13 +172,13 @@ class TimberHelper {
 	 * @param string  $allowed_tags
 	 * @return string
 	 */
-	public static function trim_words( $text, $num_words = 55, $more = null, $allowed_tags = 'p a span b i br' ) {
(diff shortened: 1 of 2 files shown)


jarednova:
Thanks @newkind! As a heads-up, I changed the filter name you added to match the project's [hook style](https://github.com/jarednova/timber/wiki/WP-Integration#hooks) of `timber/trim_words/allowed_tags`

newkind:
Awesome! :) Thanks!

Issue #619, opened 2015-07-23
{{ site.url }} does not work when manually defining WP_HOME and WP_SITEURL


herrbischoff:
While hunting for this very annoying oddity, I found out that as long as I am manually defining `WP_HOME` and `WP_SITEURL` within `wp-config.php` as follows, `{{ site.url }}` and other related tags like `{{ theme.parent.link }}`, `{{ theme.link }}`, `{{ theme.parent.path }}` and `{{ theme.path }}` do not work, outputting nothing.

```
define('WP_HOME', 'http://' . $_SERVER['HTTP_HOST']);
define('WP_SITEURL', 'http://' . $_SERVER['HTTP_HOST'] . '/wp');
```

As soon as I declare those within the database, and comment those defines, it starts working as expected again.

I was hoping to circumvent WordPress' rather braindead behavior of being tied to a domain name to make updating between a local dev environment and a staging server more seamless.

jnweaver:
FWIW, I use those exact same definitions in my wp-config.php without any issues in Timber.

herrbischoff:
@jnweaver: Maybe I'm overlooking something incredibly stupid here. Can you think of other reasons that could interfere with those tags/functions?

jnweaver:
Not really.

What does `{{ http_host }}` return for you in one of your twig views when defining those constants in wp-config?

Also, what does `get_bloginfo( 'url' )` return?

ivanjuras:
@jarednova , is this really the prefered behavior?

PauGNU:
Hi,

I confirm this bug.

After setting a custom theme directory using the function:
    register_theme_directory(ABSPATH . '../themes');

The variable {{ theme.path }} doesn't return the correct value anymore. I had to define my own variable in the functions.php file to set it correctly.

xavivars:
I've added a pull request that may partially fix this (at least, @PauGNU's issue).

According to WordPress documentation, path-related constants shouldn't be used by plugin/themes, and it seems the issue was because Timber was assuming the themes folder would always be inside WP_CONTENT_DIR.

By using methods instead of the constants, `{{theme.path}}` returns the right  value

destefanis:
I'm also encountering this bug when developing locally using a VM. Posts and pages won't appear styled as the path appears as ```<link rel="stylesheet" href="/style.css" type="text/css" media="screen">```

herrbischoff:
@jarednova: Any update on this?

xavivars:
[...]

PR #666, opened 2015-08-31
Remove constant usages from TimberTheme

xavivars:
According to WordPress documentation, path-related constants shouldn't be used by plugin/themes, and it seems the issue was because Timber was assuming the themes folder would always be inside WP_CONTENT_DIR.

By using methods instead of the constants, `{{theme.path}}` returns the right value.

This should fix `PauGNU`'s issue in #619

### DIFF
--- a/lib/timber-url-helper.php
+++ b/lib/timber-url-helper.php
@@ -225,14 +225,26 @@ class TimberURLHelper {
 	 * @return boolean if $url points to an external location returns true
 	 */
 	public static function is_external_content( $url ) {
-		// using content_url() instead of site_url or home_url is IMPORTANT
-		// otherwise you run into errors with sites that:
-		// 1. use WPML plugin
-		// 2. or redefine upload directory
-		$is_external = TimberURLHelper::is_absolute( $url ) && !strstr( $url, content_url() );
+		$is_external = TimberURLHelper::is_absolute( $url ) && !TimberURLHelper::is_internal_content($url);
+                
 		return $is_external;
 	}
 
+    private static function is_internal_content($url) {
+        // using content_url() instead of site_url or home_url is IMPORTANT
+		// otherwise you run into errors with sites that:
+		// 1. use WPML plugin
+		// 2. or redefine content directory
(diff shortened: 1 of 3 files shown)


jarednova:
@xavivars thanks for the find! As I keep working on this I'm finding all kids of ways to configure WP setups (and all kinds of ways to Timber to mix-up paths in the process). 

Could you update this PR with tabs instead of spaces for the indentation? From here out I'm following WordPress's [PHP Coding Standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/) to play (mostly) nicely with the rest of the code.

xavivars:
Hi @jarednova, I've updated the pull request. When you have time, take a look at it.

Thanks!

xavivars:
Hi, I've just found another (related) bug: when changing the UPLOADS constant to be outside of WP-CONTENT, `TimberURLHelper` was detecting the image as an external image. With this patch this is no longer happening.

I've tried to run the unit tests, to see if there's anything broken, and also to add tests to cover this case, but I couldn't run it: it seems I'm missing something to set up your tests environment, but I don't see documented how to set it up.

lib/timber-url-helper.php
```diff
@@ -225,14 +225,25 @@ public static function is_absolute( $path ) {
...
-		// 2. or redefine upload directory
-		$is_external = TimberURLHelper::is_absolute( $url ) && !strstr( $url, content_url() );
-		return $is_external;
-	}
+        $is_external = TimberURLHelper::is_absolute( $url ) && ! TimberURLHelper::is_internal_content( $url );
+                
+        return $is_external;
+    }
+
+    private static function is_internal_content($url) {
```
prisis: need more spaces

xavivars: Thanks, just fixed it.

xavivars:
Just added unit tests to it.

jarednova:
@xavivars -- thanks for the PR and the tests! It's now merged and will release tomorrow to wp.org

PR #680, opened 2015-09-11
Adding WP Types integration

xavivars:
Filter based on example provided in https://github.com/jarednova/timber/issues/169


jarednova:
Moving this to #1021

PR #700, opened 2015-10-06
Update timber.php

erik-landvall:
If you allow it to be fetch as such, I wont have to use `ob_start` => `Timber::render` => `ob_get_contents` => `ob_end_clean`

### DIFF
--- a/timber.php
+++ b/timber.php
@@ -310,7 +310,7 @@ class Timber {
 	 * @param string  $cache_mode
 	 * @return bool|string
 	 */
-	public static function render( $filenames, $data = array(), $expires = false, $cache_mode = TimberLoader::CACHE_USE_DEFAULT ) {
+	public static function fetch( $filenames, $data = array(), $expires = false, $cache_mode = TimberLoader::CACHE_USE_DEFAULT ) {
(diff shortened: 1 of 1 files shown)


xavivars:
This feature would be great :)

timber.php
```diff
@@ -319,6 +319,18 @@ public static function render( $filenames, $data = array(), $expires = false, $c
...
+
+	/**
+	 * @param array   $filenames
+	 * @param array   $data
+	 * @param bool    $expires
+	 * @param string  $cache_mode
+	 * @return bool|string
+	 */
+	public static function render( $filenames, $data = array(), $expires = false, $cache_mode = TimberLoader::CACHE_USE_DEFAULT ) {
+		$output = static::fetch( $filenames, $data = array(), $expires = false, $cache_mode = TimberLoader::CACHE_USE_DEFAULT );
```
xavivars: This line is incorrect, as it's rewriting the values of the variables with default values.

It should be 

    $output = static::fetch( $filenames, $data, $expires, $cache_mode);

erik-landvall: Sorry, sloppy of me :) thanks for pointing it out @xavivars

PR #708, opened 2015-10-13
Added Reply Link for threaded comments.

josephbergdoll:
This enables accessing the comment reply link via {{ comment.reply_link }} in your comments loop.

Fixes https://github.com/jarednova/timber/issues/319

### DIFF
--- a/lib/timber-post.php
+++ b/lib/timber-post.php
@@ -695,6 +695,35 @@ class TimberPost extends TimberCore implements TimberCoreInterface {
 		$children = array_values($children);
 		return $children;
 	}
+	
+		/**
+	 * Enqueue the WP threaded comments javascript,
+	 * and fetch the reply link for various comments.
+	 * @internal
+	 * @param int $comment_id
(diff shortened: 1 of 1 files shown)


jarednova:
Thanks @josephbergdoll! I've moved the function you wrote to be a public method within `TimberComment`. This makes it easier to add arguments like:

{{ comment.reply_link( "Respond" ) }}

... so that it's easier to customize. Tests are running now -- once they complete (and pass) I'll merge-in.

josephbergdoll:
@jarednova ah, you beat me to it! was trying to do that yesterday but in timber-post.php but obviously that's using the wrong object, was going to move it over to timber-comment today. :+1:

tehlivi:
<code>TimberComment_reply_link()</code> was never added to the WordPress install although <code> get_comments()</code> was updated with <code>$timber_comment->reply_link = $this->TimberComment_reply_link($comment->comment_ID, $this->ID);</code>

PR #720, opened 2015-10-28
Add WordPress feed urls to TimberSite

rpkoller:
Urls for rdf, rss, rss2 and atom are retrieved


jarednova:
Thanks @rpkoller -- this is merged in #721

Issue #724, opened 2015-10-28
Latest version breaks ACF date field dates


t-wright:
I think pull request 715 https://github.com/jarednova/timber/commit/a6ba9ddb7f49a4231450912efa987058dc5d2093 exposed an issue with ACF date field values. Returned values are always numeric (even if you specify a string or date format in field settings) but they aren't unix timestamps, rather Ymd (20140917). As such date_i18n() doesn't play nice with them.

rpkoller:
Yep same here. problem is dates are saved numeric in the 20151028 format in acf . so with that patch the else if statement matches and intval is run instead of strtotime. so with intval the value remains the same instead of getting converted into the necessary unix timeformat. :/

jarednova:
@t-wright @rpkoller thanks for the heads-up on this. I'll take a look at it first-thing tomorrow

xavivars:
Found a possible fix. I'll commit the change in a pull request

xavivars:
Didn't realize @rpkoller fixed it earlier on #727. Just closed #728.

jarednova:
Closed by #727 I also added a test in dd1c09738f072a0bfe17d6d822d59ce34b5e1b3c to catch the condition in the future

Issue #726, opened 2015-10-29
How to get the timber phpunit tests running in vvv


rpkoller:
I've never used phpunit so far but wanted to let it run through before i place my next pull request. i am running my sites via vvv locally. i've then installed the timber version via git. 

```
git clone git@github.com:jarednova/timber.git
```

then used 

```
composer install
```
to install all the dependencies. in the next step i've just hit `phpunit`in the timber directory (version 4.7.6.). but then i get two warnings:

```
Warning: require_once(/tmp/wordpress-tests-lib/includes/functions.php): failed to open stream: No such file or directory in /Users/danton/Documents/Coding/Vagrant/vvv/www/timbertest/wp-content/plugins/timber/tests/bootstrap.php on line 7

Fatal error: require_once(): Failed opening required '/tmp/wordpress-tests-lib/includes/functions.php' (include_path='.:') in /Users/danton/Documents/Coding/Vagrant/vvv/www/timbertest/wp-content/plugins/timber/tests/bootstrap.php on line 7
```

based on the settings in the bootstrap.php i dont have a `/tmp/wordpress-tests-lib` folder as well as no `/includes/functions.php`  file and there isnt also any `/includes/bootstrap.php` . i am bit confused. :/ a brief headsup where and what to change to get the phpunit tests running would be more than appreciated. :) thanks r.

xavivars:
If I'm not wrong, what I did some time ago was to create that folder (`/tmp/wordpress-tests-lib`) before running the tests. But probably you already tried that...

rpkoller:
thanks for the headsup... hmmmmm ok manually creating those two folders helped at least a little... guess i also have to create the includes folder inside the `wordpress-tests-lib` folder but then i am still missing the two already mentioned files `functions.php` and `bootstrap.php. which do i have to use there? still a bit uncertain and puzzled. ;)

xavivars:
(info coming from https://github.com/jarednova/timber/blob/master/.travis.yml)

Try to do this BEFORE running phpunit

`bash bin/install-wp-tests.sh wordpress_test root '' localhost $WP_VERSION`
`composer install --dev --prefer-source`

with `$WP_VERSION` being the version you want to test.

Then, you'll be able to run your tests.

connorjburton:
@jarednova I assume this is all okay now?

Issue #695, opened 2015-09-30
Execution of unprepared SQL statements


arall:
Some queries seems to not being sanitised before it's execution.

In `wp-content/plugins/timber-library/lib/timber-archives.php` file, there are a bunch of calls to `$wpbd->get_results()` method, with a non sanitised query as parameter. Those are some examples:

```php
$query = "SELECT YEAR(post_date) AS `year`, count(ID) as posts FROM $wpdb->posts $join $where GROUP BY YEAR(post_date) ORDER BY post_date $order $limit";
// [...]
$results = $wpdb->get_results($query);
```

```php
$query = "SELECT YEAR(post_date) AS `year`, MONTH(post_date) AS `month`, count(ID) as posts "
. "FROM $wpdb->posts $join $where GROUP BY YEAR(post_date), MONTH(post_date) "
. "ORDER BY post_date $order $limit";
// [...]
$results = $wpdb->get_results($query);
```

```php
$query = "SELECT YEAR(post_date) AS `year`, MONTH(post_date) AS `month`, DAYOFMONTH(post_date) AS `dayofmonth`, count(ID) as posts FROM $wpdb->posts $join $where GROUP BY YEAR(post_date), MONTH(post_date), DAYOFMONTH(post_date) ORDER BY post_date $order $limit";
// [...]
$results = $wpdb->get_results($query);
```

jarednova:
Hey @arall thanks for the heads-up on this. For the most part this is taken directly from WP Core's [`wp_get_archives` function](https://core.trac.wordpress.org/browser/tags/4.3.1/src/wp-includes/general-template.php#L1483). As you can see, they've got some other stuff going on for caching, etc. too -- but the same non-sanitized queries exist.

Can you show an example (or a PR!) of what these should look like instead?

arall:
@jarednova In the `get_items` method, query arguments seems to be correctly sanitised, but in `get_items_monthly` and `get_items_yearly`, the parameters go directly to the query. However, those methods are only called from `get_items` method, after the sanitisation, so everything looks fine on those.

The `$join` and `$where` variables are generated using the `apply_filters` WP method, which is using `$post_type` and `$args` non-sanitised parameters, that come from outside. 
I'm not sure how safe is this. I will try to reproduce it later.

arall:
Ok, this is actually injecting SQL:

```php
$args = ['post_type' => '" OR 1="1'];
$defaults = [
	'post_type' => 'post',
];
[...]

PR #727, opened 2015-10-29
Fix for the ACF date field issue #724

rpkoller:
Add a second condition to check if the $date variable contains a unix timestamp. If a unix timestamp is present strtotime returns false. if you have a date formated by acf strtotime returns the unix timestamp.

### DIFF
--- a/lib/timber-twig.php
+++ b/lib/timber-twig.php
@@ -266,7 +266,7 @@ class TimberTwig {
 
 		if ( $date instanceof DateTime ) {
 			$timestamp = $date->getTimestamp();
-		} else if (is_numeric( $date ) ) {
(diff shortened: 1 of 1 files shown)


jarednova:
@rpkoller thanks, I like it!

rpkoller:
@xavivars sorry for not posting alongside the pull request also into the issue thread noting about the pull request. but quickly pushed things out cuz i was late for an appointment; i had to leave rather quickly. 

@jarednova np yw. btw could you take a quick look at #726 . i have another patch more or less ready preparing the reported sql statements from issue #695 . but i wanted to run the phpunit tests beforehand to see if there are any grave mistakes in there. but never worked with phpunit yet and somehow the startup already throws warnings at me :/

PR #728, opened 2015-10-29
Allow date formats like '20151029'

xavivars:
Allow numeric values that can be parsed with strtotime (like `20151029') to be used.


xavivars:
Duplicate of #727

PR #733, opened 2015-11-02
Eaton master

jarednova:
This also closes #723


lib/image/timber-image-operation-tojpg.php
```diff
@@ -58,9 +58,14 @@ function image_create( $filename, $ext = 'auto' ) {
 				$ext = $ext['ext'];
 			}
 		}
-		if ( $ext == 'gif' ) {
+		if ( strtolower($ext) == 'gif' ) {
```
bartveneman: Why not do `strtolower` a single time when declaring the variable?

jarednova: good call @bartveneman! that makes things easier. I've added that in f3f6d2c

PR #758, opened 2015-11-20
Added time functions to Timber post and Timber comment

lggorman:


jarednova:
Great stuff @lggorman! I really dig the inline docs + tests! I made a small tweak in #760 which I can explain best in-person next week. Once those test complete, I'll merge in this in can release to WP.org!

Issue #676, opened 2015-09-07
Timber related errors and warning on image delete


rpkoller:
A friend informed me regarding an issue deleting an image from the media library showing timber related notices and warnings. I've tried as well which lead to the following white screen showing the following errors and warnings: 

```
Notice: Undefined index: dirname in /Applications/MAMP/htdocs/mypage/wp-content/plugins/timber-library/lib/timber-image-helper.php on line 171

Notice: Undefined index: extension in /Applications/MAMP/htdocs/mypage/wp-content/plugins/timber-library/lib/timber-image-helper.php on line 172

Warning: Cannot modify header information - headers already sent by (output started at /Applications/MAMP/htdocs/mypage/wp-content/plugins/timber-library/lib/timber-image-helper.php:171) in /Applications/MAMP/htdocs/mypage/wp-includes/pluggable.php on line 1207
```
I am using WordPress 4.3 and Timber 0.21.9 locally. In the wp-config.php i've set the following debug settings and the php error reporting is set to strict: 

```
define( 'WP_DEBUG', true );
define( 'WP_DEBUG_DISPLAY', true );
 ```

Got feedback from my friend. on his local install debug was switched off, there no notices and warnings showed up, while on his remote install debug is switched on and showed the errors. By adding `isset` checks for the indexs in the `delete_generated_files` function the warnings could be eliminated: 

```php
	static function delete_generated_files( $local_file ) {
		if (!isset($info['dirname']) || !isset($info['extension']) ) {
			return false;
		}
		if (TimberURLHelper::is_absolute( $local_file ) ) {
			$local_file = TimberURLHelper::url_to_file_system( $local_file );
		}
		$info = pathinfo( $local_file );
		$dir = $info['dirname'];
		$ext = $info['extension'];
		$filename = $info['filename'];
		self::process_delete_generated_files( $filename, $ext, $dir, '-[0-9999999]*', '-[0-9]*x[0-9]*-c-[a-z]*.' );
		self::process_delete_generated_files( $filename, $ext, $dir, '-lbox-[0-9999999]*', '-lbox-[0-9]*x[0-9]*-[a-zA-Z0-9]*.' );
	}
```

Would have created a pull request but the code changed in the master branch to Timber 0.2.19 and i also dont know against which branch in particular i would have to place the pull request. 

Best regards r.

connorjburton:
Closing this as the fix has been merged.

PR #762, opened 2015-11-27
Change action hook 'delete_attachment'

lgaechter:
The description of the hook `delete_post` in the WordPress Codex:

> `delete_post` is fired before and after a post (or page) is deleted from the database.
> However by this time the post comments and metadata have been already deleted. Use the `before_delete_post` hook to catch post deletion before that.

When I want to delete all associated attachments together with a post, I would use something like this in my functions.php

```php
add_action( 'before_delete_post', function ( $post_id ) {

	global $post_type;

	if ($post_type !== 'mycustomposttype') {
	    return;
	}

	$attachments = get_children( array(
	    'post_parent' => $post_id,
	    'post_type'   => 'attachment'
	) );

	if ( empty($attachments) ) {
	    return;
	}

	foreach ( $attachments as $attachment ) {
	    wp_delete_attachment($attachment->ID);
	}
});
```

If I would use the `delete_post` hook for that, it wouldn’t work, because the relation between the post and the attachments would already have been deleted, `$attachments` would be empty and no attachments would be deleted, although they would still exist.

The action `delete_attachment` triggers before post meta data is deleted and would also catch cases like the one described above.

The check for the attachment post type (`if ( $post->post_type == 'attachment' )`) would also become obsolete, because `delete_attachment` triggers only on attachment post types.

**Other changes**
- Add check for existing `file_loc` before trying to delete files for that attachment (fix for #676)
- Fix indentation for action function body

### DIFF
--- a/lib/timber-image-helper.php
+++ b/lib/timber-image-helper.php
@@ -159,14 +159,16 @@ class TimberImageHelper {
 	 * Deletes all resized versions of an image when the source is deleted
 	 */
 	protected static function add_actions() {
-		add_action( 'delete_post', function ( $post_id ) {
-				$post = get_post( $post_id );
-				$image_types = array( 'image/jpeg', 'image/png', 'image/gif', 'image/jpg' );
-				if ( $post->post_type == 'attachment' && in_array( $post->post_mime_type, $image_types ) ) {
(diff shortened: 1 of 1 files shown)


lggorman:
Merged in with #765.  Thanks

PR #763, opened 2015-11-30
Added list filter and corresponding test

lggorman:
No Oxford comma, a la AP style.  But could pass a param if you think it'd be worth it!


jarednova:
@lggorman this is great. As part of our convo today, we should make sure to document all of these custom Twig filters in a central and logical location. I'll leave that to you as an issue

Issue #711, opened 2015-10-20
No 'Awaiting Moderation' message after posting comments


dinolatoga:
Normally, when a commenter leaves a message and comment moderation is on, the commenter should be able to read a *Your comment is awaiting moderation* message. This doesn't work for me in Timber. Anybody got a clue?

dinolatoga:
My code looks kinda looks like this:

    {% for comment in post.comments %}
        {% if comment.approved %}
            <li>{{comment.author.name}}, says {{comment.content}}</li>
        {% else %}
            <li>{{comment.author.name}}, says {{comment.content}}<br>Your comment is awaiting moderation.</li>
        {% endif %}    
    {% endfor %}`
`

connorjburton:
This seems to have been fixed in #770 so i'm going to close this. Let me know if you are still having issues @dinolatoga

PR #770, opened 2015-12-03
Added logic to return unmoderated comments of current user

lggorman:
'get_comments' will return a user's unmoderated comments, if that user is logged in or still has their email address stored by a browser cookie.  This will allow developers to show a message indicating that a user's comment is awaiting moderationg by checking for `comment.approved`, as described in #711.


jarednova:
Notes from 12/7: @lggorman to investigate applying test based on hooking into WP filter

PR #775, opened 2015-12-08
Fix a typo in `TimberHelper::get_comment_form()` documentation

Plou:


jarednova:
Thanks @Plou — those are actually automatically-generated pages based on the PHPDoc comments. @lggorman — can you update the source to reflect the typo that @plou found?

Plou:
Oh, 😅 I'll do that

jarednova:
Sounds good, thanks @Plou — we'll leave it to you

Plou:
I wish I could do it, but the source seems to be okay, the change of quote happens during the doc building process. I don't know if there is a easy way to fix it.

jarednova:
@lggorman: looks like the parser in https://github.com/jarednova/PHP-Markdown-Documentation-Generator is responsible here. This is the output:

http://jarednova.github.io/timber/#get_comment_form

... we can talk later today about how this fits into prirorities

jarednova:
The problem exists at the Slate level, I've filed an issue with them:

https://github.com/tripit/slate/issues/466

jarednova:
@Plou @lggorman after much head-banging the issue lies with an error in redcarpet (a dependency of slate); I fixed the issue by not italicizing the "deprecated" messages

Plou:
:+1:

PR #797, opened 2015-12-16
Fix for "Method TimberFunctionWrapper::__toString() must not throw an…

newkind:
Fix for "Method TimberFunctionWrapper::__toString() must not throw an exception"

From time to time we get stopped by this error while working with Gantry and we need to manually change the code to see where the real issue lies - that's why I created this PR ;)

https://monosnap.com/file/FAxian75O4yVcfbR8BG84CxhhvE8Py.png

### DIFF
--- a/lib/timber-function-wrapper.php
+++ b/lib/timber-function-wrapper.php
@@ -7,7 +7,11 @@ class TimberFunctionWrapper {
 	private $_use_ob;
 
 	public function __toString() {
-		return (string)$this->call();
+		try {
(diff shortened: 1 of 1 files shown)


jarednova:
@newkind before I merge I want to write an appropriate test — can you give an example of data or a function that would cause this error to be thrown?

newkind:
@jarednova basically any function that will throw an exception ;)

jarednova:
@lggorman: can you review, write a test and merge?

PR #840, opened 2016-02-01
Add id as a duplicate of ID in timber-image.php

jonginn:
In a TimberPost, you can access the id [through `post.id` or `post.ID`](https://github.com/jarednova/timber/blob/master/lib/timber-post.php#L628). However with a `post.thumbnail`, you can only use `post.thumbnail.ID`, which doesn't seem intuitive and leads to errors.

This adds `id` to `post.thumbnail` as a duplicate of `ID`.


jarednova:
Thanks @jonginn — merged!

PR #844, opened 2016-02-04
Allow post object as parameter of TimberImage()

connorjburton:
Caveat: I have only just started using Timber so this may already be solvable easily.

**Issue**

Using ACF, if you have a repeater field that contains a relationship field, there is no easy way to get the featured image of the post that has been selected in the relationship field. You can init a new ```TimberPost``` however that is adding more logic to the views whilst it can be easily solved within the ```TimberImage``` class

**Solution**

If you pass a post object to ```TimberImage```, it will grab a new instance of ```TimberPost```, get the ```_thumbnail_id``` and re-init ```TimberImage``` with the correct id.

**Impact**

This will be slightly slower than just passing the ID, but in the instance I described above the only way to do this in the first place is this method, it's just moved the logic away from the view. I haven't benchmarked this change.

**Usage**

``````

**Considerations**

We may not _want_ TimberImage to be so flexible to reduce chances of bugs, however we already allow other formats than ID's to be passed to ```TimberImage```, so it seems to me a natural progression to allow post objects.

### DIFF
--- a/docs/wiki/cookbook-images.md
+++ b/docs/wiki/cookbook-images.md
@@ -85,7 +85,7 @@ This can be used in conjunction with other filters, so for example:
 #### Using images in custom fields:
 Let's say you're using a custom field plugin (like the amazing [Advanced Custom Fields](http://www.advancedcustomfields.com/)). You can use the resulting images in your Twig templates very easily.
 
-When setting up your custom fields you'll want to save the `image_id` to the field. The image object, url, etc. _will_ work but it's not as fool-proof.
+When setting up your custom fields you'll want to save the `image_id` to the field. The image object, post object, url, etc. _will_ work but it's not as fool-proof.
 
 ##### The quick way (for most situations)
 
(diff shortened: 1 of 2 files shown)


jarednova:
Awesome, thanks @connorjburton! I think I follow the logic you have here. I'm gonna pull this down and write a test to make sure I get it 100%

connorjburton:
A couple of things after using Timber more. This only seems to be a problem in instances where a regular WP Post object is returned, not a TimberPost object. Maybe the root cause is that we are not transforming WP Post object in acf repeater relationship fields to TimberPost objects in the first place? We would have .thumbnail available if we did that.

However I think catching more things for TimberImage is better. I don't think my changes will support if you pass a TimberPost, only WP Post. I will check tomorrow and add further commits to support Timber Post if needed.

connorjburton:
Okay so i've added support for ```TimberPost```, this shouldn't really ever be used as you would have ```.thumbnail``` available. However in the name of consistency if we are saying ```TimberImage``` supports post objects it should support ```TimberPost``` too.

We are getting into lots of ```elseif```s in ```TimberImage->init```, may be worth refactoring into a different function if we add more cases. Also not sure where this change would go in the docs.

jarednova:
Merged!

PR #852, opened 2016-02-13
Added timber_transients_slug filter

ruscon:
If you are using i18n, you can use this filter


jarednova:
Thanks @ruscon! I changed this to `timber/transient/slug` to match the filter syntax we have going on elsewhere.

ruscon:
thx

PR #853, opened 2016-02-16
Adding request.get/post to TimberSite

connorjburton:
**Issue**

There is no easy way to access GET/POST variables in the Twig templates

**Solution**

Add ```site.request``` array that contains ```site.request.get``` and ```site.request.post```

Eg.

/vacancies/?salary=30000

```
Requested salary: £{{site.request.get.salary}}
```

```
Requested Salary: £30000
```

### DIFF
--- a/lib/timber-request.php
+++ b/lib/timber-request.php
@@ -1,0 +1,26 @@
+<?php
+
+/**
+ * TimberRequest exposes $_GET and $_POST to the context
+ */
+class TimberRequest extends TimberCore implements TimberCoreInterface {
+
+	$post = array();
+	$get = array();
+	
+	/**
+	 * Constructs a TimberRequest object
+	 * @example
+	 */
+	function __construct() {
+		$this->init();
+	}
+
+	/**
+	 * @internal
+	 */
+	protected function init() {
+		$this->post = $_POST;
+		$this->get = $_GET;
+	}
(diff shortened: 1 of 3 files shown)


jarednova:
Thanks @connorjburton! I love this. My only question is whether it makes intuitive sense to make this property a child of Site. Does this merit being its own object? IE:

```twig
Requested salary: ${{request.get.salary}}
```

(also correcting that strange dollar symbol you used in yr example) :-)

connorjburton:
Yeah ```request.get.salary``` is much better, was my first thought.

What is this $ you speak of!

connorjburton:
Re-factored to it's own class, will test fully tomorrow.

jarednova:
Awesome, thanks @connorjburton! The `"$"` are legal tender we use with dead kings on 'em

connorjburton:
I've just added the required interface functions, other than that it works. One thing I noticed, the composer autoloader needs to be updated with the new class. How is that handled on Plugin update? As of now, if I just drop these files in it throws a fatal error.

jarednova:
@connorjburton — I was able to get everything working. Are you talking about when deploying the plugin to wp.org?

connorjburton:
Oh I guess its just pre-generated before deploying to wp.org

PR #858, opened 2016-02-18
Fixing get_preview inconsistencies

### LINKED ISSUE #624: excerpt filter and get_preview: spacing inconsisitencies
When using the excerpt filter the output is the content and then "..." with no space between. 

    Lorem ipsum dolor sit amet...

When using .get_preview() it outputs the content, then a space, then '...' another space and then the 'read more' link. 

    Lorem ipsum dolor sit amet ... Read More

It would be nice to be able to replace the ellipsis with a string as well.

connorjburton:
Also updated tests.

Fixes #624


jarednova:
Thanks @connorjburton — I'll take a look at this one!

jarednova:
Looks good! I added a test for the new custom $end parameter

Issue #843, opened 2016-02-03
TimberHelper::function_wrapper doesn't support array($class, 'func')


connorjburton:
Would be super useful for ```TimberHelper::function_wrapper``` to support class methods, means we could create a class called TemplateHelpers and use reflection to go through all the helpers and add them to function wrappers automatically. Eg: (Not tested, just threw it together)

```
class TemplateHelpers {
  public function run() {
     $reflection = new ReflectionClass(get_class($this));
     $methods = $reflection->getMethods();
     unset($methods[0]); //remove this function

     foreach($methods as $method) {
          $params = array();
          foreach($method->getParameters() as $param) {
                 $params[] = ($param->getDefaultValue()) ?: '';
          }
          TimberHelper::function_wrapper(array($this, $method->name), $params);
     }
  }
 
  private function imAHelper($message = 'hey') {
      return $message;
  }
}
```

PR #866, opened 2016-02-24
Adding support for classes in Function Wrapper

### LINKED ISSUE #854: TimberFunctionWrapper has problems with Namespaces
If I just use the `TimberFunctionWrapper` in an unsupported way: Sorry and please tell me. :)

I wanted to make a **static function** available in the context by doing this:

```PHP
$context["helper"] = new \TimberFunctionWrapper(
      "Name\\Space\\Class::staticHelperFunction",
      ["should_be_set_from_twig", "default", "default2"]
);
```
and expexted that I can use it in Twig like this:
```twig 
{{ helper }}
```
(Works)
```twig
{{ helper("first_parameter") }}
```
(Does not work because Twig does not have a `helper` function (only `Name\Space\Class::staticHelperFunction`))

I helped myself by **extending** `TimberFunctionWrapper` and supporting a **fourth parameter** `twig_function_name` that is passed as a first parameter into the `Twig_SimpleFunction` constructor (instead of `$this->_function`) in `add_to_twig`. See: [timber-function-wrapper.php](https://github.com/jarednova/timber/blob/master/lib/timber-function-wrapper.php#L41)

connorjburton:
Adds official (you can currently do `TimberFunctionWrapper('MyClass::staticFunction')`) support for classes in `TimberFunctionWrapper`. Fixes #854 and #843

**Issue**

It's currently not documented that you can do `TimberFunctionWrapper('MyClass::staticFunction')` and also that does not solve the issue with non static functions.

**Solution**

Allows the first parameter of `TimberFunctionWrapper` to be an array, consisting of an `object` (class) OR a `string` that references a class. It seems you need to pass a string for name spaced classes `TimberFunctionWrapper(array('MyNamespaces\Container', 'some_function'))`, however you can also pass a class `object` as the first parameter of the array.

**Impact**

Negligibly slower.

**Usage**

`TimberFunctionWrapper(array($classObject, 'function_name'))`
`TimberFunctionWrapper(array('MyNamespace\Container', 'function_name'))`

**Considerations**

I have only added 1 extra test (haven't ran the test either), we may need to create a dummy namespaced class for testing the namespace version of this? Input on how to handle this best would be great.

Also I am wondering if we can get away from `call_user_func_array` as it's super slow.

### DIFF
--- a/docs/wiki/wp-integration.md
+++ b/docs/wiki/wp-integration.md
@@ -110,7 +110,7 @@ For a less quick-and-dirty way, you can use the TimberFunctionWrapper. This clas
 TimberHelper::function_wrapper( $function_name, $defaults = array(), $return_output_buffer = false );
 ```
 
-Classes (including namespaced) are also support for `function_wrapper`:
+Classes (including namespaced) are also supported for `function_wrapper`:
 
 ```php
 # Namespaced has to be a string
--- a/docs/wiki/wp-integration.md
+++ b/docs/wiki/wp-integration.md
@@ -102,13 +102,25 @@ For a less quick-and-dirty way, you can use the TimberFunctionWrapper. This clas
 
 ```php
 /**
- * @param string $function_name
+ * @param mixed $function_name or array( $class( string|object ), $function_name )
  * @param array (optional) $defaults
  * @param bool (optional) $return_output_buffer Return function output instead of return value (default: false)
  * @return \TimberFunctionWrapper
  */
 TimberHelper::function_wrapper( $function_name, $defaults = array(), $return_output_buffer = false );
 ```
+
+Classes (including namespaced) are also support for `function_wrapper`:
+
+```php
+# Namespaced has to be a string
+TimberHelper::function_wrapper('Example\Class', 'function_in_class');
+
+# Otherwise, you can pass the object of the class
+TimberHelper::function_wrapper($this, 'function_in_class');
+```
+
+You can then call the function like so `{{function_in_class}}`
 
 So if you want to add `edit_post_link` to your context, you can do something like this:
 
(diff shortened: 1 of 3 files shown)


connorjburton:
I may be wrong that it requires a string just for namespaced, you probably need a string for any class that you don't have access to the object. Maybe I need to update the docs to reflect that.

jarednova:
:boom:

connorjburton:
Cool! I'm gonna assign this to myself so I remember to update docs/tests when my setup is a bit better

jarednova:
I'm all for killing `call_user_func_array`, but I don't know if the alternatives are any better. This [thread](http://stackoverflow.com/questions/15057573/alternative-to-call-user-func-array) on StackOverflow notes that `ReflectionFunction` is even _slower_ than `call_user_func_array`. Other [alternatives](http://stackoverflow.com/a/15057864/528654) listed seem almost dangerously inelegant.

jarednova:
@connorjburton — lemme know if you're hitting a wall with test setup. I put together this wiki page:

https://github.com/jarednova/timber/wiki/Unit-testing

... on setting things up, but VVV and vagrant are known to throw curveballs

lmartins:
Is there any plans on when this will be released?

connorjburton:
Not quite sure, that'd be a question for @jarednova

lmartins:
@jarednova could you share if this is expected to get into the library?

connorjburton:
Hi @lmartins - this has been in the library since `0.22.6`

https://github.com/timber/timber/blob/0.22.6/lib/timber-function-wrapper.php

lmartins:
Thanks @connorjburton, It didn't worked for me and I just assumed that wasn't in yet. Will look into it, probably an error on my part. Thanks for the update.

connorjburton:
If it still doesn't work please post an example on here and I'll take a look

PR #873, opened 2016-03-01
Updating http reference in composer.json to https

connorjburton:
Sorry original PR was branched from a branch.

The composer file references http://wpackagist.org however that causes an error when doing `composer install` due to https://getcomposer.org/doc/06-config.md#secure-http.

Fix just changes the reference to `https`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,7 +45,7 @@
   "repositories": [
     {
       "type": "composer",
-      "url": "http://wpackagist.org"
+      "url": "https://wpackagist.org"
     },
     {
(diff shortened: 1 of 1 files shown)


connorjburton:
One of the tests failed but all the others went through okay, think it's just an anomaly?

jarednova:
Yeah, there are sometimes random failures due to timeouts with packagist and such. I'm re-running it now

connorjburton:
All went through okay now so merging. Thanks

Issue #867, opened 2016-02-28
Issue with the datefilter


Langstra:
When using the datefilter the timezone is completely omitted. For instance, I inserted the DateTime object with the timezone "Europe/Amsterdam", the time was 21.30. Now it converts this with gettimestamp() to an integer. Here is omits the timezone. Afterwards the time is printed as 20:30. 

A solution could be the following code:
`if ( $date instanceof DateTime ) {
   $timestamp = $date->getTimestamp() + $date->getOffset();
} else if (is_numeric( $date ) && strtotime( $date ) === false ) {
   $timestamp = intval( $date );
} else {
   $timestamp = strtotime( $date );
}`

connorjburton:
Hi @Langstra, thanks for this. Could you please post full error reproducible code so I can look into this? Thanks

Langstra:
PHP-code
`
$date = new DateTime("now", new DateTimeZone("europe/amsterdam"));
echo $date->format("d F \\a\\t G:i");
return Timber::compile('partials/signup_closed.twig', array('date' => $date));
`
Timber template
`
<br>
{{ date|date("d F \\a\\t G:i") }}
`
The timezone of $signup_open is ignored and the dates have one hour difference.

connorjburton:
Okay have tested this and implemented a fix, just need to write a test then will submit a PR.

connorjburton:
Merged

PR #874, opened 2016-03-01
Ref #867 - Fixes timezone when using date filter

connorjburton:
This solves #867 - also added test to confirm functionality.

#### Issue
```php
$context['amsterdam'] = new DateTime('now', new DateTimeZone('europe/amsterdam'));
```

```twig
{{ amsterdam|date('g') }}
```

Will return the wrong time.

The `date` filter does not honour the `DateTimeZone`

#### Solution
We add the offset of the timezone onto the date (offset can be negative, so `04:00 + -1 = 03:00`)

#### Impact
None

#### Usage
No change

#### Considerations
There may be more instances of this, we need to be aware of honouring timezone offsets.

### DIFF
--- a/tests/test-timber-dates.php
+++ b/tests/test-timber-dates.php
@@ -75,7 +75,7 @@
 			$date = new DateTime('2015-09-28 05:00:00', new DateTimeZone('europe/amsterdam'));
 			$twig = "{{'" . $date->format('g:i') . "'|date('g:i')}}";
 			$str = Timber::compile_string($twig);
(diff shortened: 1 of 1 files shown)


connorjburton:
Just need your eyes on this @jarednova (honestly, it won't take up more than 10 seconds of your eye time) then I can merge.

jarednova:
Easier than #876! Looks good to me; please merge when you're ready

connorjburton:
Thanks! Merged

Issue #836, opened 2016-01-29
Preview Changes and TimberPost


rwstone:
Thanks for Timber!

I was having problems previewing changes in the admin. The preview showed the published version rather than the changes. 

I replaced
`$post = new TimberPost();`
with
`$post = Timber::query_post();`
in all my php files except for the blog home. That seems to fix the issue. 

However, I don't want to introduce new problems to my theme, changing things I don't fully understand, so I thought I'd ask about this behavior here.

Is this a TimberPost bug, a feature, or a subtlety I haven't grasped yet?

Thanks!

(On a related note, it seems like you need to use `post.get_field('field')` rather than `post.field` on _all_ field types to properly preview ACF changes. That makes sense, but it wasn't clear to me from the ACF Cookbook.)

connorjburton:
This is solved and merged.

PR #876, opened 2016-03-02
[Needs Review] Ref #836 - Allow to preview when using TimberPost

connorjburton:
This solves #836. This needs review as it touches a few important bits of code. @jarednova 

Also still needs tests that I will write once functionality is complete.

#### Issue
If you use `TimberPost` in your PHP files and try to preview the page it won't show any previews.

#### Solution
If `$_GET['preview']` is set and the `nonce` verifies we then go onto to test if the current user _should_ be seeing this post/page, if they are we get the latest revision and return the ID of that post.

The ID will also fall back to the original post if a revision isn't found or the user doesn't have permissions to see the post preview.

#### Impact
Possibly bad ones if we miss something.

#### Usage
No change

#### Considerations
Okay a few things:

1. I'm using `$_GET` - is there a wrapper around this? I don't see why `$_GET` wouldn't be fine in this situation, just wondering for consistency
2. At the moment I just check for `edit_posts/pages` and `edit_others_posts/pages`, however does this cover custom post types? I don't think so. What about if a user should be able to see the post but not necessarily be able to edit the page/post, is there another capability type for that?
3. Is there an easier way to do this? The original issue said he just changed it to `Timber::query_post` and it worked, can we hook into that?
4. Do we need to check for passworded posts?

### DIFF
--- a/lib/timber-post.php
+++ b/lib/timber-post.php
@@ -181,7 +181,11 @@ class TimberPost extends TimberCore implements TimberCoreInterface {
 			&& is_object($wp_query->queried_object)
 			&& get_class($wp_query->queried_object) == 'WP_Post'
 			) {
-			$pid = $wp_query->queried_object_id;
+				if( isset( $_GET['preview'] ) && isset( $_GET['preview_nonce'] ) && wp_verify_nonce( $_GET['preview_nonce'], 'post_preview_' . $wp_query->queried_object_id ) ) {
+					$pid = $this->get_post_preview_id( $wp_query );
+				} else if ( !$pid ) {
+					$pid = $wp_query->queried_object_id;
+				}
 		} else if ( $pid === null && $wp_query->is_home && isset($wp_query->queried_object_id) && $wp_query->queried_object_id )  {
 			//hack for static page as home page
 			$pid = $wp_query->queried_object_id;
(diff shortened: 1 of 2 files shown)


jarednova:
@connorjburton thanks as always for the thorough work. I did a review on my end and tried to break stuff — luckily, unsuccessfully. Some specifics...

**_I'm using $_GET - is there a wrapper around this? I don't see why $_GET wouldn't be fine in this situation, just wondering for consistency_**
I did a scan of the WP core source and found plenty of `$_GET`s including some specifically for `$_GET['preview_nonce']`. Obviously, there's plenty in WP core to not copy, but I don't see an issue here.

**_At the moment I just check for edit_posts/pages and edit_others_posts/pages, however does this cover custom post types? I don't think so. What about if a user should be able to see the post but not necessarily be able to edit the page/post, is there another capability type for that?_**
I did a search through [Roles & Capabilities](https://codex.wordpress.org/Roles_and_Capabilities) and didn't see anything around previewing. I was able to successfully preview custom post types in y tests

**_Is there an easier way to do this? The original issue said he just changed it to Timber::query_post and it worked, can we hook into that?_**
I'm not sure how to hook into Timber::query_post here to solve the issue. I think there's an opportunity to refactor `determine_id` into a cleaner set of methods (perhaps even its own class), but I'd call that a battle for another day. I see this fix as solving the issue without the introduction of significant technical debt.

**_Do we need to check for passworded posts?_**
I did a test for PW-protected posts and it works using the convention from the [Timber Starter Theme](https://github.com/timber/starter-theme). There are good arguments about whether the PW protection should come at a higher level, but let's leave that for another day.

I'm good to merge this from here, unless you feel strongly otherwise

scottfatbeehive:
@jarednova There seems to be a partial wrapper for the `$_GET`s [here](https://codex.wordpress.org/Function_Reference/is_preview) with the `is_preview()` function. Perhaps that helps?

connorjburton:
@scottfatbeehive - I did look at that, but `is_preview` seems to assume that `$wp_query` is already the previewed post, however the issue here is that `$wp_query` isn't yet the previewed post until we do the checks in the PR. Unless I am wrong here?

@jarednova - okay thanks for checking, leave it with me and I will try and write a test for this use case. If I can figure it out. I'll let you know when it's all okay to merge.

scottfatbeehive:
@connorjburton No you're right, it requires $wp_query already being set. Good point.

connorjburton:
Spent way too much time getting this test working, but it works! Had to refactor the edit link test a bit to get them to place nicely together.

Also tested this fix with custom post types having `capabilities` registered as described in the codex and all worked okay.

Ready for you to merge @jarednova

PR #877, opened 2016-03-02
Adding second test for function wrapper

connorjburton:
Very simple, just adds second test for `TimberHelper::function_wrapper` to make sure `array('ClassName', 'FunctionName')` works as the first parameter.


connorjburton:
Ready to be merged if you are okay with this @jarednova

jarednova:
Woohoo! Tests! Tests! Tests! Tests! :microscope:

PR #879, opened 2016-03-04
876 preview tests

jarednova:
This picks up #876:

@connorjburton I made two changes I want to call your attention to.

First on `TimberPost::get_post_preview_id` I switched the assumption on true/false. For security reasons (admittedly, hypothetical) it's better practice to start with a no-access assumption. That way if something screwy happens with permissions, the code makes the assumption that viewing the post preview should _not_ be allowed.

On the test itself (`testPreviewContent`), PHPUnit assertEquals compares:

```
$this->assertEquals( Expected Thing, Thing to Test );
```

The test will still pass either way, but the formatting of errors/failures will read correctly. 

All this is just tweaks on your earlier work — give it a look and if you like what you see this should be good to (finally) merge!

### DIFF
--- a/lib/timber-post.php
+++ b/lib/timber-post.php
@@ -228,25 +228,27 @@ class TimberPost extends TimberCore implements TimberCoreInterface {
 	 		$can[] = 'edit_others_' . $query->queried_object->post_type . 's';
 	 	}
 
-	 	$continue = true;
+	 	$can_preview = false;
 
 	 	foreach( $can as $type ) {
-	 		if( !current_user_can( $type ) ) {
-	 			$continue = false;
+	 		if( current_user_can( $type ) ) {
+	 			$can_preview = true;
 	 			break;
 	 		}
 	 	}
-		
-		if( $continue ) {
-			$revisions = wp_get_post_revisions( $query->queried_object_id );
 
-			if( !empty( $revisions ) ) {
-				$last = end($revisions);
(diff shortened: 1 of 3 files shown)


connorjburton:
Makes sense about the test.

For the refactor of `get_post_preview_id`, I can see a couple of issues:

`$can` is an array that contains _all_ the permissions that a user needs to have to preview the post, as there is a`break` on the first instance of `current_user_can`:

1. We are assuming if the user can do any of `$can` they should be able to preview the post, which isn't true, it needs to pass all of the array `$can`.
2. If we removed the break we still have an issue where `$can[0]` may be false (keeping `$can_preview` at `false`), but the `$can[1]` may be true, and because we no longer have the break `$can_preview` is now  `true`. However that isn't right because the user shouldn't be able to preview.

Phew, think that makes sense.

I propose to change the `foreach` to (not the prettiest, if you can think of a better way I'm all for it):

```php
                $can_preview = array();

	 	foreach( $can as $type ) {
	 		if( current_user_can( $type ) ) {
	 			$can_preview[] = true;
	 		}
	 	}

	 	if ( count( $can_preview ) !== count( $can ) ) {
	 		return;
	 	}
```

Hopefully that all makes sense? Can clarify if not.

jarednova:
Yep! Makes sense — thanks for the fix on that. I merged that in with #887

PR #880, opened 2016-03-04
Adding trim around get_widgets

### LINKED ISSUE #511: Timber::get_widgets('sidebar'); always returns true
In my single.php I have the following code

$context['sidebar'] = Timber::get_widgets('sidebar-right');

And in my base.twig I have:

                    {% if sidebar %}
                        <div class="col-xs-12 col-sm-4">
                            <aside class="layout-sidebar">
                                {{ sidebar }}
                            </aside>
                        </div>
                    {% endif %}

It always returns true even if there is no widgets in the sidebar-right.

This seems to function quite differently to dynamic_sidebar() which returns false if there is no content in the sidebar.

Am I doing something wrong or is this a feature not a bug?

connorjburton:
Fixes #511 

#### Issue
If you are using a plugin such as Display Widgets of Dynamic Widgets, the sidebar may be active but not actually containing content, as you can choose which widgets in a sidebar are shown on what page, however `is_active_sidebar` will only check if a sidebar has _any_ widgets.

#### Solution
We just `trim` the result so that we don't have any empty content.

Normally I wouldn't put a fix in for plugin compatibility (massive rabbit-hole) however this is literally just a `trim`.

#### Impact
None

#### Usage
No change

#### Considerations
I can't see how this would effect any existing code, if you do a trim in `twig` as some people were suggesting in #511 it will still work.

### DIFF
--- a/lib/timber.php
+++ b/lib/timber.php
@@ -348,7 +348,7 @@ class Timber {
 	 * @return TimberFunctionWrapper
 	 */
 	public static function get_widgets( $widget_id ) {
-		return TimberHelper::function_wrapper( 'dynamic_sidebar', array( $widget_id ), true );
+		return trim( TimberHelper::function_wrapper( 'dynamic_sidebar', array( $widget_id ), true ) );
 	}
 
 


jarednova:
This one's a lot easier than #847!

Issue #804, opened 2016-01-04
Resize based on height parameter ignoring width


ashrayer:
Hi,

Happy New Year! and thank you for the Timber -- it made me happy with WP development again :)

One thing I wanted to ask -- is there any way to resize image to certain height (instead of width)? sometimes I need this feature. 
I know it is not difficult to implement, what method/function should I modify?

jarednova:
You bet! Just send a `0` or `false` to the width parameter, and your desired height in pixels

ashrayer:
Thanks!
P.S. Updated the docs https://github.com/jarednova/timber/wiki/TimberImage#arguments

jarednova:
@lggorman — can you make sure this is reflected in the docblock/new docs?

lggorman:
@jarednova Sure no problem!

connorjburton:
This just needs docblock/new docs updated before we can close this

PR #881, opened 2016-03-04
Updating docblock for TimberImage(url)|resize(0, 50)

connorjburton:
Finished off #804 - just adds comments in docblock


jarednova:
:tada:

PR #882, opened 2016-03-04
Timber::get_posts('custom-post-type') ensures class is child

connorjburton:
Hard to fix description into title! This is just an extension of #847, for calls like `Timber::get_posts('cpt')`.


connorjburton:
Closing as have a better refactor

PR #883, opened 2016-03-04
post.parent.children now will only return published posts

### LINKED ISSUE #677: Post parent includes draft pages
When I use 

```
{% for livingPage in post.parent.children %}
```

in the parent childrens posts are included posts with wordpress status draft

connorjburton:
Fixes #677 

#### Issue
`post.parent.children` will return all children of that post, regardless of status, as described in the codex here. https://codex.wordpress.org/Function_Reference/get_children

#### Solution
Explicity requests `publish`ed children only.

#### Impact
Maybe someone somewhere is using this and doesn't realise their children are not published.. maybe but I doubt it.

#### Usage
No change

#### Considerations
None

### DIFF
--- a/lib/timber-post.php
+++ b/lib/timber-post.php
@@ -674,7 +674,7 @@ class TimberPost extends TimberCore implements TimberCoreInterface {
 		if ( $post_type == 'parent' ) {
 			$post_type = $this->post_type;
 		}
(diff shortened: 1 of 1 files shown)


jarednova:
That is bizarre — why would anyone want draft things to appear by default? I wonder how many sites I have out there that are showing draft posts without realizing it. Oh WordPress

connorjburton:
Haha, yeah, it's odd. I can only assume `get_children` is for admin purposes, maybe we are supposed to use `get_posts(array('post_parent' => ID))`

PR #884, opened 2016-03-04
Allowing allowable_tags customization

### LINKED ISSUE #130: get_preview() should allow to explicitly specify html tags when calling trim_words()
Since trim_words() defaults tags are 'p a span b i br', it is impossible show embedded videos in the preview unless I specify a custom post_excerpt which has been filtered with "the_content" because $last_p_tag will = false if there are no <p> tags in the post. Which is the case when a client wants to have a post with only an embedded video in it.

connorjburton:
Fixes #130 

#### Issue
You can't customize what tags `get_preview` returns

#### Solution
Allows the `$strip` param to accept a string of tags to strip whilst keeping original `true`/`false` functionality and not effecting the `$strip` tag itself so when it's used later on in the function everything else works as before.

#### Impact
None

#### Usage
`get_preview(length, force, readmore, tags, end);`

#### Considerations
None


jarednova:
Looks good! (better docs, too)

connorjburton:
Not that it changes anything, but for someone looking at this PR, the tags string is actually tags to keep, not remove. My misunderstanding sorry!

PR #885, opened 2016-03-04
[Needs Review] Auto register classes based on post type

### LINKED ISSUE #799: Timber::get_posts - instantiating class based on post type
I'm extending the TimberPost class for each of my custom post types - SiteArticle, SiteVideo, SiteAuthor, etc.

Is it possible to have Timber::get_posts instantiate the correct class for a given post type automatically?  I know that I can specify a class to use as the second parameter, but sometimes I have collections of posts that are of multiple post types.  For instance, on the home page I'd like to get a mixed collection recent videos and articles.

Thanks.

connorjburton:
Sorry this is a long one, but I think it's important to get this code right as I think it's a bit confused at the moment, and I may be mis-understanding it's purpose.

Fixes #799 and replaces #847

#### Issue
After investigating #799 I realise the function that checks for auto registering could have some nasty side effects and in some instances won't work as expected. Take the example below:

```php
class cpt extends TimberPost {}

# 1.
Timber::get_posts(array(
    'post_type' => 'cpt'
));
# doesn't work

# 2.
Timber::get_posts('cpt');
# works

# 3.
Timber::get_posts(array(
    'name' => 'cpt',
    'post_type' => 'page'
));
# if i did a simple fix, would return the cpt class, however we are querying a page, not a cpt
```

So in the code there were a couple of issues, it was checking if any of the array properties you passed to `Timber::get_posts` matched any classes that extended `TimberPost`, which can lead to some confusing bugs (3). However, I also noted the function just returned `true`, but when it came to the `if` statement, it'd replace the `$PostClass` with the first parameter you passed into `Timber::get_posts`, which would be an array (1), which doesn't seem to be what `$PostClass` should be, but maybe im jumping to conclusions here? And finally, it's undocumented but if you just pass the `post_type` name as the first parameter it will use the class (1).

#### Solution
Refactored the function to:

1. Just have 1 check for if the class exists
2. If the first parameter is an array, only check if `post_type` has a matching class, not every property
3. Always assign a string to `$PostClass`

#### Impact
This is obviously an important function, so goes without saying it needs eyes on it.

#### Usage
```php
TimberPost::get_posts(array(
    'post_type' => 'cpt'
));
# Returns the cpt class
```

#### Considerations
Auto assigning the class by `post_type` or the first parameter being a string is *not* documented and I think right now it's fairly confusing if this is even supposed to happen or not. I think the refactor I have done makes more sense, it's more explicit and what I would personally expect in terms of "Magic".

Currently (and with this fix) if you run `Timber::get_posts(array('post_type' => 'cpt', 'posts_per_page' => -1));` on a `single.php` page, it won't return all the `cpt`s, it will only return the one you are on. That's because when it matches a class it will remove the original query and just assume you want the class. Maybe it should carry over the query too?

### DIFF
--- a/lib/timber-post-getter.php
+++ b/lib/timber-post-getter.php
@@ -90,7 +90,13 @@ class TimberPostGetter {
 	 * @return bool
 	 */
 	static function is_post_class_or_class_map($arg){
-		$type = (is_string($arg)) ? $arg : (is_array($arg) && isset($arg['post_type'])) ? $arg['post_type'] : false;
+		$type = false;
+
+		if(is_string($arg)) {
+			$type = $arg;
+		} else if(is_array($arg) && isset($arg['post_type'])) {
+			$type = $arg['post_type'];
+		}
 
 		if(!$check) return false;
 
--- a/lib/timber-post-getter.php
+++ b/lib/timber-post-getter.php
@@ -32,8 +32,8 @@ class TimberPostGetter {
 	 * @return array|bool|null
 	 */
 	static function query_posts($query = false, $PostClass = 'TimberPost' ) {
-		if (self::is_post_class_or_class_map($query)) {
-			$PostClass = $query;
+		if ($type = self::is_post_class_or_class_map($query)) {
+			$PostClass = $type;
 			$query = false;
 		}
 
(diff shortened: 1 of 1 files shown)


jarednova:
@connorjburton — thanks for your work on this. I think your notion at top: _I think it's a bit confused at the moment, and I may be mis-understanding it's purpose._ is a good one. Basically (as you call-out in your note) there may be a bit too much "magic" going on at the detriment of clear "I do this, and get this back" 

So I want to fix what is broken in #847 with this fix (my changes are located in #891 — is there some trick to pushing to a PR?). That said, I think the kitten has outgrown the litterbox here. Once the bleeding has stopped, it would seem to me that registering the relationship between a post type and a class should be defined somewhere in a configuration. 

I'm not sure of the exact setup on this (should this be defined in the TimberSite? the context? a `.timber` config file?) but I want to make note of my official objection (to how I wrote this ~2 years ago) as over-ambitious and foolish.

connorjburton:
Okay I agree, after doing this PR I was thinking this seems odd, less magic is better. I will continue discussion on #891 

A `.timber` config file would be nice.

Objection noted :pencil2: :notebook: 

Regarding pushing to PR, easier ways would be either:

1. Give you access to my fork so you can push directly to the branch
2. Create branches on this repo so you can push directly to them

Anywho, closing this one now.

PR #886, opened 2016-03-04
Fixing extra slash on images not in year/month folders

### LINKED ISSUE #498: TimberImage resize URL extra forward slash issue
Timber Version 0.21.1
```
<img src="{{post.thumbnail.src|resize(720, 540)}}" alt="" />
```
The resize filter produces this URL: http://localhost/wp-content/uploads\/1024x400-720x540-c-default.jpg

There is an extra forward slash before the image file name, it prevents the image to be loaded correctly.

connorjburton:
Fixes #498

#### Issue
If your settings are to *not* upload in year/month folders, when you get file info the `subdir` will be `/` which causes a double `/` before the image filename, e.g

`http://site.com/wp-content/uploads//1024x1024-400x400x-c-default.jpg`

#### Solution
Don't register the `/` if it's on its own.

#### Impact
None

#### Usage
No change

#### Considerations
I think this will fix #498, but they were reporting backwards slash, but I am thinking that is due to system environment.

### DIFF
--- a/lib/timber-image-helper.php
+++ b/lib/timber-image-helper.php
@@ -351,7 +351,7 @@ class TimberImageHelper {
 			}
 		}
 		$parts = pathinfo($tmp);
-		$result['subdir'] = $parts['dirname'];
+		$result['subdir'] = ($parts['dirname'] === '/') ? '' : $parts['dirname'];
 		$result['filename'] = $parts['filename'];
 		$result['extension'] = $parts['extension'];
 		$result['basename'] = $parts['basename'];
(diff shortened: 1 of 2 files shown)


connorjburton:
Just updated branch so travis would re-run. This could do with a unit test too. Will do a bunch of unit tests soon.

connorjburton:
Fingers crossed the test works :pray:

connorjburton:
Umm, I tried to match `//`, I mean URL's don't start with `http://` right? It's a monday, that's my excuse.

connorjburton:
Ready for you to merge if you are happy with it. @jarednova

PR #888, opened 2016-03-07
docs: Fix broken anchor link

gianpaj:
the link is in the 3rd step here http://jarednova.github.io/timber/#use-the-starter-theme


connorjburton:
Thanks for this @gianpaj - merged.

PR #890, opened 2016-03-07
Dev composer

jarednova:


connorjburton:
Oops I branched off image slash fix! As long as the one is okay this is fine.

PR #891, opened 2016-03-08
Connorjburton cjb auto register class

jarednova:


connorjburton:
Looks good, can you clarify usage? So I can update docs after merge. This is all un-documented AFAIK

jarednova:
@connorjburton I added docs in Timber.php here:

https://github.com/timber/timber/pull/891/commits/f47dada3529cf209eae65ec54c9c0a7f641e55d1#diff-85bc15c4c650971e1fe6e17dc0602106R92

... it should appear on http://timber.github.io/timber/ once I re-run the doc generator. Curious if you think it's a feature even worth promoting (as I mentioned on a diff't thread, it feels like the sort of relationships you'd want to define at the Site level). The basics are that instead of a `$PostClass` (string) you can send an array with:

```php
$class_map = array('post' => 'MyPostClass', 'portfolio' => 'MySpecialClass');
$posts = Timber::get_posts($query, $class_map);
```

This will make it so any posts of post type "post" will be returned as `MyPostClass` objects and post_type=portfolio will be returned as `MySpecialClass` objects

connorjburton:
I agree, however if it's something that is available, it should be documented, at least in a docblock.

I think registering what post type goes to what class would be good, although I think it should be defined earlier than `Timber::get_posts`, how we do that however, I'm not sure? At the moment we don't really have a global config space.

jarednova:
Sounds good, I'm all on-board with documentation (but maybe not promotion). 

Way back when I was thinking about a "View" object that would be a more localized place that things like contexts could be defined (thus a site could share several diff't contexts and not conflict if a plugin were also using Timber). That to me feels like the place that you'd want to configure something like this — bigger than one place, but not truly "global". Something like this might look like....

```php
$view = new MyTimberView(); //MyTimberView would contain built-in context items 
//you need for your templates and not interact with the main $data or $context array
$data = array();
$data['posts'] = Timber::get_posts($query, $view->get_class_map());
$view->render('single.twig', $data);
```

Totally hypothetical at this point, but this would be ripe space to also clean-up some of the overlapping responsibilities of `Timber`, `TimberLoader`

PR #893, opened 2016-03-08
Refactoring Twig_Loader_String existing Twig environment

### LINKED ISSUE #863: Twig_Loader_String should not be used
Timber is using `Twig_Loader_String` to compile strings in: 

https://github.com/jarednova/timber/blob/master/lib/timber.php#L233

This class should never be used; its for internal use only and there is a lot better way to do the same:

Here is the idea how to do it properly:

    $template = $twig->createTemplate($string);
    return $template->render($data);

You can even use existing twig environment allowing user to load and extend existing twig template files.

connorjburton:
Fixes #863 

#### Issue
Tests (and possibly existing code in themes) using `compile_string` currently use `Twig_Loader_String` which is not recommended and deprecated since `1.18.1`

#### Solution
Hooks into the existing Twig environment and loads the template via that.

#### Impact
None

#### Usage
No change

#### Considerations
None

### DIFF
--- a/lib/timber.php
+++ b/lib/timber.php
@@ -232,12 +232,9 @@ class Timber {
 	 */
 	public static function compile_string( $string, $data = array() ) {
 		$dummy_loader = new TimberLoader();
-		$dummy_loader->get_twig();
-		$loader = new Twig_Loader_String();
-		$twig = new Twig_Environment( $loader );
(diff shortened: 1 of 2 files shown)


jarednova:
Indeed seems to be an error from when `Timber_Loader_String` wasn't escaping that

PR #895, opened 2016-03-08
Composer update

connorjburton:
This is the correct PR you will want to merge, I changed all the names in `composer.json` to `timber/timber`

All ready to go if you are okay with it, will need to update docs afterwards.


connorjburton:
Hmm, the tests are failing, I guess we need to require `vendor/autoload.php` in `tests/bootstrap.php` - is there a way to get travis to run `composer install` before it runs the tests? It'll then create a `classmap` we can load.

connorjburton:
Tried a couple of quick changes but they didn't work, will fix this properly tonight if you don't get to it before me

PR #904, opened 2016-03-11
Fix formatting of final example in Getting Started: Themeing

cameronbaney:
Matches the previous examples to be in a syntax highlighter block


connorjburton:
Thanks for this @cameronbaney!

Issue #387, opened 2014-11-07
[Enhancement] Add current user to the global Twig context


ste93cry:
Hi,
I realized that there is no easy way to check if the user is loggedin inside Twig without using `function('is_user_logged_in')`. It would be nice to always have a `user` variable defined in the global context, with maybe a methods that returns a `boolean` indicating if the user is anonymous or loggedin. What do you think about it?

PR #909, opened 2016-03-15
Adding user to context

connorjburton:
**Ticket**: #387 
**Reviewer**: @jarednova 

#### Issue
There isn't an easy way to grab the current user in Twig

#### Solution
Created a new `TimberUser->get_context` function that either returns `false` if there is no ID, or `get_userdata` if there is.

#### Impact
I don't know how expensive `get_userdata` is

#### Usage
`{{user}}`

#### Considerations
I know that we already do a `$this->import` in the `init` function, however that doesn't assign the properties in an easily accessible way. E.g we can't just do `$this->user`

#### Testing
I need to write tests for this, should be quite simple, does the function return `false` when no user is logged in and does it return an array if a user is logged in (ideally check the id too of the array, just to make sure it's correct).


jarednova:
Hi @connorjburton, the reasoning makes sense to me to provide `{{ user }}`. I'm wondering why not just provide the `TimberUser` object instead of the result of `get_userdata`?

```php
$data['user'] = new TimberUser();
```

This would provide the logged-in user or a "false" anonymous user when not logged in (I think....)

connorjburton:
Good point! Will review this in next couple of days

connorjburton:
I haven't managed to test the tests, will do that tonight.

jarednova:
In!

PR #912, opened 2016-03-17
Fix function_wrapper docs

PhilippBaschke:
#### Issue

The Wiki documentation of `TimberHelper::function_wrapper` contains an error regarding the new possibility to pass in Classes or Objects. They have to be passed in as an Array. Also the DocBlock of `TimberHelper::function_wrapper` does not reflect that change.

#### Solution

Updated the DocBlock (with the updated DocBlock that is shown in the wiki documentation) and added the `array()` part in the wiki documentation.


connorjburton:
Thanks you! Forgot about those.

Issue #412, opened 2014-12-14
[Enhancement] Expose WordPress translations functions to Twig context


ste93cry:
Currently, Timber provides only the standard `__` translation function, but more are needed in many other cases, for example when we have to support singular/plural translations. I would like to see Timber exposing to Twig all the common WordPress translations functions. Here a list of them:

* `translate`
* `__`
* `_e`
* `_n`
* `_x`
* `_ex`
* `_nx`
* `_n_noop`
* `_nx_noop`
* `translate_nooped_plural`

For sure not all will be useful, I just copied here all the translation functions I found on Codex. At least, Timber should support the ones giving singular/plural forms of a text.

skynet:
:+1:

connorjburton:
Will sort this soon.

PR #918, opened 2016-03-21
Exposing all translation functions

connorjburton:
**Ticket**: #412 
**Reviewer**: @jarednova 

#### Issue

Full range of translation functions aren't exposed to context

#### Solution

Expose rull range

#### Impact

Some users may already have added `translate` or other functions to their context, I assume this will override those functions?

#### Usage

All translation functions are exposed to context, e.g`{{translate()}}`

#### Testing

I don't think they are necessary, we would just be testing Wordpress' functions.

### DIFF
--- a/lib/timber-twig.php
+++ b/lib/timber-twig.php
@@ -185,6 +185,33 @@ class TimberTwig {
 		$twig->addFunction( '__', new Twig_SimpleFunction( '__', function ( $text, $domain = 'default' ) {
 					return __( $text, $domain );
 				} ) );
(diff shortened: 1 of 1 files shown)


jarednova:
Looks good, thanks!

yeahsmaggy:
Does the translation function as in {{__('text to translate', 'text-domain')}} work as specified in docs?

connorjburton:
Yep should do

yeahsmaggy:
Ah great. Yeah, it does work. The problem I am having is that the Loco Translate plugin doesn't parse Twig files.

Issue #911, opened 2016-03-15
Update WP.org Script


jarednova:
Update the WP.org script so that the plugin works with the new file setup

jarednova:
Closed by #919

xavivars:
I'm using timber as a "plugin dependency", but from wpackagist (so I'm really pulling `timber-library` in my composer). The new setup won't have an impact in my setup, as I'm almost downloading it from WP.org, right?

PR #919, opened 2016-03-21
Updated WP.org build script with timber.php file

jarednova:
**Ticket**: #911 
**Reviewer**: @connorjburton 
#### Issue
With the new Composer setup, usage as a WordPress plugin is no longer native.

#### Solution
This re-introduces the `timber.php` file (in the `bin` directory) to be used in the build script. When deploying to WP.org, this file is copied into the main directory so that it's read by the WordPress plugin page.

#### Impact
Not really. This will not interfere with the theme-based Composer installation.

#### Considerations
Is it worth adding docs for the edge-edge case of someone installing through non-wp.org but who still want it to use Timber as a plugin versus a theme dependency?

#### Testing
To test, you'll need to manually move `/bin/timber.php` to `/timber.php`


connorjburton:
Looks good!

Issue #944, opened 2016-04-19
ACF Post Object does not return as TimberPost


connorjburton:
### Expected behavior

For posts than are selected via ACF Post Object to return as a `TimberPost`

### Actual behavior

Returns as `WP_Post`

### Steps to reproduce behavior

1. Create an ACF field that is of type `Post Object`
2. Get that field in the template `{% set foo = post.get_field('foobar') %}`
3. `foo` = `WP_Post`

### What version of WordPress, PHP and Timber are you using?

All latest

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)

Composer

stefthoen:
Has this change been reversed? I'm still getting `WP_Post` when I assign a ACF `Post Object` to `foo`.

PR #945, opened 2016-04-19
ACF posts now return as TimberPosts

connorjburton:
**Ticket**: #944 
**Reviewer**: @jarednova 

#### Issue

ACF fields that container a post would return as `WP_Post`, not `Timber\Post`

#### Solution

If `get_field` returns an array, we run through each array element to see if it's either an array(1) or is an instance of `WP_Post`(2)

1. We run `$this->convert` again recursively till `$ele` is no longer an array
2. We pass the `WP_Post` into `Timber\Post`

#### Impact

None that I know of, if someone already has done this manually, it will just end up passing `Timber\Post` into `Timber\Post`

#### Usage

There is less usage as you won't have to cast `WP_Post` to `TimberPost` anymore

#### Considerations

Performance I think should be good, I don't think `instanceof` or `gettype` are expensive?

#### Testing

No tests added, may add some after namespacing is merged.

### DIFF
--- a/lib/Post.php
+++ b/lib/Post.php
@@ -1026,7 +1026,7 @@ class Post extends Core implements CoreInterface {
 			}
 		}
 		$value = apply_filters('timber_post_get_meta_field', $value, $this->ID, $field_name, $this);
-		$value = $this->convert( $value );
+		$value = $this->convert( $value, __CLASS__ );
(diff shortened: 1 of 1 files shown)


jarednova:
looks good, let's add it to the namespace PR!

PR #947, opened 2016-04-20
Use default wp rewrite base for pagination

matsrietdijk:
#### Issue
`Timber::get_pagination` ignores WP's default pagination base in `$wp_rewrite` and just uses `'page'`.

#### Solution
Replace `'page'` in `Timber::get_pagination`'s format with `$wp_rewrite->pagination_base`.

#### Impact
The codebase would only improve because it now uses WP's rewrite settings.

### DIFF
--- a/lib/timber.php
+++ b/lib/timber.php
@@ -412,7 +412,7 @@ class Timber {
 				parse_str( $url[1], $query );
 				$args['add_args'] = $query;
 			}
-			$args['format'] = 'page/%#%';
+			$args['format'] = $wp_rewrite->pagination_base . '/%#%';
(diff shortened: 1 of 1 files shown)


connorjburton:
I like this, thanks. Will merge after `1.0.0` (all the files are changing so no point merging no as will cause conflicts).

matsrietdijk:
That is fine, I can work around this issue for now by overwrite the format in the function call.

```
global $wp_rewrite;
$context['pagination'] = Timber::get_pagination(array('format' => $wp_rewrite->pagination_base . '/%#%'));
```

connorjburton:
#968 takes over from here

Issue #957, opened 2016-04-26
Compatibility with WPBakery Visual Composer's front end editor


felthy:
### Expected behavior
When I use the `timber-starter-theme` in conjunction with [WPBakery Visual Composer](https://vc.wpbakery.com/), and use the Front End Editor on a page, I expect to see the inline editing anchors when I hover my mouse over editable parts of the content, like in this screenshot:
![front-end-editor](https://cloud.githubusercontent.com/assets/1940215/14808674/b1759332-0bca-11e6-9a23-adade53a7112.jpg)

### Actual behavior
No inline editing anchors pop up.

### Steps to reproduce behavior
- Start with a fresh WP install
- Install Timber
- Install and activate a copy of the `timber-starter-theme`
- Install _WPBakery Visual Composer_
- In the _Pages_ section of the WP admin interface, select _Edit With Visual Composer_ for **Sample Page**.

### What version of WordPress, PHP and Timber are you using?
WordPress 4.6-alpha-37307, PHP 5.4.45, Timber 0.22.5

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
- via WP.org

PR #958, opened 2016-04-26
Call "the_post" action to fix Visual Composer editor

felthy:
**Ticket**: #957

#### Issue
Gets WPBakery Visual Composer front end editor working with Timber. See #957 for more detail.

#### Solution
Visual Composer uses the `'the_post`' action hook to inject its inline editing anchors into the content; because the timber theme doesn't use `the_post()` to render content, this action is not called. I added a call to the action to Timber.

#### Impact
Minimal.

#### Usage
No change.

#### Considerations
This could also have been achieved by calling `setup_postdata($post);` at the same line instead of `do_action` but I felt that calling `do_action` would have fewer potential side effects.

#### Testing
None - not sure how testing could be usefully automated for this issue. Also VC is not free.


connorjburton:
Taken over by #969

PR #930, opened 2016-04-07
Allow filtering of new image url and path

danielbitzer:

PR #959, opened 2016-04-26
Danielbitzer master

jarednova:
This picks up on #930


jarednova:
Thanks @Danielbitzer!

Issue #943, opened 2016-04-19
Author Avatar?


alzearafat:
Hi,

Thanks for excelent tool! It's really help me a lot,

Anyway, how I can get author avatar URL? Something like this : 

```
{{ post.author.avatar }}
```

The documentation only explain comment avatar.

Thank you in advance for any help you can provide,
Regards

connorjburton:
This actual doesn't seem to be possible at the moment, it shouldn't be too difficult of a fix. Will look at this post `1.0.0` (soon)

alzearafat:
Hi @connorjburton 

Wow, cool! :D 

I will follow the progress ;)

Thank you

connorjburton:
Merged

PR #971, opened 2016-05-03
Adding avatar to user

connorjburton:
**Ticket**: #943 
**Reviewer**: @jarednova 

#### Issue

There was no built in way to retrieve the post author's URL.

#### Solution

Use `get_avatar_url` to return the URL to the class property

#### Impact

Need to consider how this may effect comments, will this conflict? We should probably deprecate `comment.avatar` and make it an alias of `comment.author.avatar`

#### Usage

`user.avatar`

#### Testing

No tests included.

### DIFF
--- a/lib/User.php
+++ b/lib/User.php
@@ -40,6 +40,12 @@ class User extends Core implements CoreInterface {
 
 	/**
 	 * @api
+	 * @var string The URL of the author's avatar
+	 */
+	public $avatar;
+
+	/**
+	 * @api
 	 * @var  string The first name of the user
 	 */
 	public $first_name;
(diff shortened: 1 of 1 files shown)


jarednova:
Lookin' good @connorjburton! I think as a next step we should consider whether to make an object for `Timber/Avatar`; but for now this will do (and for backwards compatibility the `__toString` of a future object would return the url). I'm just gonna wait for the tests to finish and then I'll merge

connorjburton:
Good shout on that, will give it some thought

connorjburton:
Tested this on starter theme and works well

Issue #979, opened 2016-05-05
Using Timber library with Bedrock


chrisgherbert:
I've been using Timber with [Bedrock](https://roots.io/bedrock/).

Until RC3, there were no major issues.  Now Timber calls `is_admin()` and uses WordPress hooks on being loaded.  Because Bedrock loads Composer dependencies before WordPress this will cause a fatal error.

Is the only solution to switch to the plugin version of Timber?

connorjburton:
Hi, do you have stack traces please? Our preference is Timber is to be used with composer, so if we can solve this we will.

chrisgherbert:
```
[05-May-2016 18:31:42 Europe/Berlin] PHP Fatal error:  Call to undefined function Timber\is_admin() in /Users/cherbert/web/bedrock-timber-test/vendor/timber/timber/lib/Timber.php on line 61
[05-May-2016 18:31:42 Europe/Berlin] PHP Stack trace:
[05-May-2016 18:31:42 Europe/Berlin] PHP   1. {main}() /Users/cherbert/web/bedrock-timber-test/web/index.php:0
[05-May-2016 18:31:42 Europe/Berlin] PHP   2. require() /Users/cherbert/web/bedrock-timber-test/web/index.php:5
[05-May-2016 18:31:42 Europe/Berlin] PHP   3. require_once() /Users/cherbert/web/bedrock-timber-test/web/wp/wp-blog-header.php:13
[05-May-2016 18:31:42 Europe/Berlin] PHP   4. require_once() /Users/cherbert/web/bedrock-timber-test/web/wp/wp-load.php:42
[05-May-2016 18:31:42 Europe/Berlin] PHP   5. require_once() /Users/cherbert/web/bedrock-timber-test/web/wp-config.php:7
[05-May-2016 18:31:42 Europe/Berlin] PHP   6. ComposerAutoloaderInit658bcd7a4d9b35310726249c356c62aa::getLoader() /Users/cherbert/web/bedrock-timber-test/vendor/autoload.php:7
[05-May-2016 18:31:42 Europe/Berlin] PHP   7. composerRequire658bcd7a4d9b35310726249c356c62aa() /Users/cherbert/web/bedrock-timber-test/vendor/composer/autoload_real.php:45
[05-May-2016 18:31:42 Europe/Berlin] PHP   8. require() /Users/cherbert/web/bedrock-timber-test/vendor/composer/autoload_real.php:55
[...]

PR #981, opened 2016-05-06
ref #979 -- is_admin() lives in \is_admin();

jarednova:
**Ticket**: #979
**Reviewer**: @connorjburton 

#### Issue
User @chrisgherbert is getting an error of `[05-May-2016 18:31:42 Europe/Berlin] PHP Fatal error:  Call to undefined function Timber\is_admin() in /Users/cherbert/web/bedrock-timber-test/vendor/timber/timber/lib/Timber.php on line 61`

#### Solution
`is_admin` is a WP function in the global namespace so it should be referred to as `\is_admin()` so PHP doesn't look for `Timber\is_admin()`

#### Considerations
This doesn't quite make sense. From what I read here:

http://php.net/manual/en/language.namespaces.fallback.php

PHP should always look in the global namespace when a function is not found in the current namespace; thus this PR shouldn't be necessary. Could there be a setting that this user has on in their PHP or WP settings?

#### Testing
All tests pass locally. I tested both `is_admin` and `\is_admin()` on a test site and they both work.

### DIFF
--- a/lib/Timber.php
+++ b/lib/Timber.php
@@ -58,10 +58,10 @@ class Timber {
 	 * @return
 	 */
 	protected function test_compatibility() {
-		if ( is_admin() || $_SERVER['PHP_SELF'] == '/wp-login.php' ) {
+		if ( \is_admin() || $_SERVER['PHP_SELF'] == '/wp-login.php' ) {
 			return;
 		}
-		if ( version_compare(phpversion(), '5.3.0', '<') && !is_admin() ) {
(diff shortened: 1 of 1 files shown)


chrisgherbert:
The issue I reported isn't related to namespaces, despite the slightly confusing PHP error message.  The issue is that the global function `is_admin()` doesn't exist when it's being called because Bedrock loads all Composer dependencies (including Timber) *before* loading WordPress.

connorjburton:
Yeah it looks like we are going to have to have a specific Bedrock integration, that waits for an event to tell us Wordpress is setup.

Issue #984, opened 2016-05-09
PHP error with the "use" statement in Term.php


Jmayhak:
In lib/Term.php there is a line to use TImber\Term on line 9 which is referencing itself. should be removed

### What version of WordPress, PHP and Timber are you using?
Timber 1.0.1

PR #985, opened 2016-05-09
remove "use" statement for Term in the Term file

Jmayhak:
Simple line removal to fix bug with "use" statement

#984


connorjburton:
Thank you!

PR #986, opened 2016-05-09
You can now re-initialize a Timber Image with a Timber Image

jarednova:
**Reviewer**: @connorjburton 

#### Issue
If the `TimberImage` or `Image` function in twig was passed a `\Timber\Image` object it would cause a compile error.

#### Solution
If passed an `Image` object, the `\Timber\Image::init` function will now grab the ID and re-initialize

#### Impact
It would be bad practice to do this, but it's worse to have a silent error. 

#### Usage
No.

#### Considerations
Rather than re-initialize, the `Image` object should be able to grab the public assignments of properties so that no new queries need to be run.

#### Testing
New Unit tests included demonstrating failing behavior and passing behavior

### DIFF
--- a/lib/Image.php
+++ b/lib/Image.php
@@ -155,6 +155,17 @@ class Image extends Post implements CoreInterface {
 	}
 
 	/**
+	 * @return array
+	 */
+	protected function get_post_custom($iid) {
+		$pc = get_post_custom($iid);
+		if ( is_bool($pc) ) {
+			return array();
+		}
+		return $pc;
+	}
+
+	/**
 	 * @internal
 	 * @param  int $iid the id number of the image in the WP database
 	 */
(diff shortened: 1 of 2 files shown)


lib/Image.php
```diff
@@ -212,7 +223,9 @@ public static function wp_upload_dir() {
 	 */
 	function init( $iid = false ) {
 		if ( !$iid ) { Helper::error_log('Initalized TimberImage without providing first parameter.'); return; }
-		
+		if ( $iid instanceof Image ) {
```
connorjburton: Would `$iid instanceof __CLASS__` work here? In case we ever change the class name.

connorjburton:
Just that one note then I'll merge

lib/Image.php
jarednova: You'd think that, right? but when I change it to...

```php
if ( $iid instanceof __CLASS__ )
```

I get: `Parse error: syntax error, unexpected '__CLASS__' (T_CLASS_C) in /srv/www/timber/lib/Image.php on line 226`

jarednova: ... though this works:

```php
if ( $iid instanceof self ) {
```

jarednova:
@connorjburton check it out now!

connorjburton:
Beautiful. Merged.

Thanks!

Issue #977, opened 2016-05-04
Call to Timber::get_widgets  kills wordpress


Jakkkub:
### Expected behavior
This code should load data for three widget areas:

```
add_filter('timber_context', 'add_to_context');
function add_to_context($data){ 
    //footer widgets
    for($i = 1; $i<4; $i++){ 
        $data['footer_widgets_'.$i] = Timber::get_widgets('footer_widgets_'.$i);
    }
}
```

Code is places in timber_context filter

### Actual behavior
Nothing is loaded in browser.
After some time I see just Chrome`s ERR_CONNECTION_RESET page.
If I comment out line with Timber::get_widgets then it works.

It is fresh WP installation so there are no custom widgets - just WP`s default ones, so there should be no problem inside widgets.

I have used this snippet of code on many themes before, always worked. Maybe 1.0 upgrade issue?

### What version of WordPress, PHP and Timber are you using?
WordPress 4.5.1, PHP 7.0.1, Timber 1.0 -->

### How did you install Timber? 
From wp.org

connorjburton:
Hi @Jakkkub - did this happen pre `1.0.0`?

connorjburton:
Replicated this without the `add_filter` part and it works okay.

elivz:
I'm seeing something similar. With `Timber::get_widgets` inside the `timber_context` action callback, it appears to be creating an infinite recursion. I see the first widget area repeated many times on the page (and no other content), and then PHP gives up with `Fatal error: Maximum function nesting level of '256' reached, aborting!`. Let me know if there is anything I can do to help track this down - I can post the stack trace if that would be helpful.

connorjburton:
Hi, yes the stack trace would be useful. Thanks

connorjburton:
Hi guys,

I am doing the following and i'm not getting any errors, a full stack trace would be useful along with your guys full setup in terms of what is in your widgets.

functions.php
```php
register_sidebars(2, array('name'=>'foobar%d'));

	add_filter('timber_context', 'add_to_context');
	function add_to_context($data){
		for($i = 1; $i<2;$i++){ 
	        $data['foobar'.$i] = Timber::get_widgets('foobar'.$i);
	    }

		return $data;
	}
```

elivz:
[...]

PR #989, opened 2016-05-10
Adding memoization to get_context

connorjburton:
**Ticket**: #977
**Reviewer**: @jarednova 

#### Issue

Calling `get_context` inside the `timber_context` filter is common, and at the moment causes an infinite loop.

This would also not cache any data so is unnecessarily slow

#### Solution

Adds memoization to the `get_context` function so it will only run the filters once (Fixes infinite loop) and also is faster when using `get_context` multiple times.

#### Impact

Faster, fixes issue with infinite loop.

#### Usage

No change

#### Considerations

This needs some real world testing, will get the guys on the ticket to give it a go.

#### Testing

No tests included.

### DIFF
--- a/lib/Timber.php
+++ b/lib/Timber.php
@@ -194,24 +194,28 @@ class Timber {
 	 * @return array
 	 */
 	public static function get_context() {
-		$data = array();
-		$data['http_host'] = 'http://'.URLHelper::get_host();
-		$data['wp_title'] = Helper::get_wp_title();
-		$data['wp_head'] = Helper::function_wrapper('wp_head');
-		$data['wp_footer'] = Helper::function_wrapper('wp_footer');
-		$data['body_class'] = implode(' ', get_body_class());
+		static $cache = [];
+		
+		if(empty($cache)) {
+			$cache['http_host'] = 'http://'.URLHelper::get_host();
+			$cache['wp_title'] = Helper::get_wp_title();
+			$cache['wp_head'] = Helper::function_wrapper('wp_head');
+			$cache['wp_footer'] = Helper::function_wrapper('wp_footer');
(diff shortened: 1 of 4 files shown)


jarednova:
@connorjburton — I'd like to bounce this back to you for final review of my changes. My only concern is calling `Timber::get_context` more than once. As you can see in my tests I had to clear the cache each time to allow other things to be added to the context. 

Are there situations where this could happen in the real world? I can only think of total edge cases (a single.php that loads a sidebar.php where there's a filter) but I want to make sure I'm not missing something else obvious.

connorjburton:
Yeah I thought something like that would crop up. I'll take a look, I just needed a proof of concept.

Thanks for the tests also!

I am wondering if a better solution would be to cache data that won't change and have a static property `Timber::context `. Just throwing ideas out there

connorjburton:
At least if makes more sense that you aren't generating the context or running any filters. You are just getting a property. Then we add the filters back in on every request to `Timber::get_context`. Although the naming is weird now, as `Timber::get_context ` is actually more like `Timber::set_context`

jarednova:
@connorjburton where I want to get to is something that removes the idea of a universal context all together where it's scoped based on the view you're working in (which would enable plugins and parent/child themes to work much more flexibly), maybe something like...

```php
$context = new Context('my-theme');
Timber::render('home.twig, $context);
```

```php
add_filter('timber/context/my-theme, function($context) {
    $context['foo'] = 'bar';
    return $context;
});
```

For now, I'm good with this fix if you are

connorjburton:
Yeah I agree, context is an important part and should be more fleshed out than it currently is. Shall I make an issue for it and assign it for `1.1.0` release?

connorjburton:
Merged

PR #990, opened 2016-05-11
Comment form redux1

jarednova:
**Reviewer**: @connorjburton 

#### Issue
Many people (especially on the WordPress.org forums) have reported problems with the deprecated `TimberHelper::get_comment_form` function. Well of course they have, it was removed!

https://wordpress.org/support/topic/fatal-error-class-not-found-in-101?replies=2
https://wordpress.org/support/topic/get_comment_form-returns-empty?replies=4

... that said, the problem they're running into isn't entirely of their making. The `single.php` file in the Starter Theme had a call to `TimberHelper::get_comment_form` that I'm sure many developers left in w/o realizing what it was doing. I'd like to reduce 1.0 pain as much as possible

#### Solution
Restore the `Helper::get_comment_form` method (for now). But introduce a better method through `{{ post.comment_form }}` for folks to use.

#### Impact
A little extra code

#### Usage
We should update elements of the Starter Theme to use `{{ post.comment_form }}`

#### Considerations
I don't like the idea of a Timber object method returning HTML. But I also sympathize that hand-building a comment form's HTML is no one's idea of a good time.

#### Testing
Unit tests included!

### DIFF
--- a/lib/Helper.php
+++ b/lib/Helper.php
@@ -444,6 +444,18 @@ class Helper {
 
 	/**
 	 *
+	 * Gets the comment form for use on a single article page
+	 * @deprecated 0.21.8 use `{{ function('comment_form') }}` instead
+	 * @param int     $post_id which post_id should the form be tied to?
+	 * @param array   $args this $args thing is a fucking mess, [fix at some point](http://codex.wordpress.org/Function_Reference/comment_form)
+	 * @return string
+	 */
+	public static function get_comment_form( $post_id = null, $args = array() ) {
+		return self::ob_function( 'comment_form', array( $args, $post_id ) );
+	}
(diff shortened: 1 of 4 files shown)


connorjburton:
Yes I agree, this is probably the biggest issue I have seen with upgrades. Would this be the final version of this, or are you thinking of developing it further (other than removing the deprecated version eventually).

jarednova:
@connorjburton I would call this final "for now". I think the next place to go is actually show how to include a partial in the [starter theme](timber/stater-theme) so that the HTML remains fully accessible. That said, if someone just wants to slap-in the standard comment form (or one that's modified by other plugins) this represents a stable format for that use case

Issue #991, opened 2016-05-11
Timber killing site load when installed as a dependency of the whole site instead of the composer.json of a theme


stewarty:
### Expected behavior
Timber not to kill a site loading

### Actual behavior
Following error on page view

> Fatal error: Uncaught Error: Call to undefined function Timber\is_admin() in /vagrant/vendor/timber/timber/lib/Timber.php:61 Stack trace: #0 /vagrant/vendor/timber/timber/lib/Timber.php(49): Timber\Timber->test_compatibility() #1 /vagrant/vendor/timber/timber/init.php(5): Timber\Timber->__construct() #2 /vagrant/vendor/composer/autoload_real.php(54): require('/vagrant/vendor...') #3 /vagrant/vendor/composer/autoload_real.php(45): composerRequire38245ed637a44f902e4cbd66fc63b950('/vagrant/vendor...') #4 /vagrant/vendor/autoload.php(7): ComposerAutoloaderInit38245ed637a44f902e4cbd66fc63b950::getLoader() #5 /vagrant/public/wp-config.php(99): include('/vagrant/vendor...') #6 /vagrant/public/wordpress/wp-load.php(42): require_once('/vagrant/public...') #7 /vagrant/public/wordpress/wp-admin/admin.php(31): require_once('/vagrant/public...') #8 /vagrant/public/wordpress/wp-admin/index.php(10): require_once('/vagrant/public...') #9 {main} thrown in /vagrant/vendor/timber/timber/lib/Timber.php on line 61

### Steps to reproduce behavior
Install timber at the composer.json in the root of a WordPress project, not in the composer.json of a theme

### What version of WordPress, PHP and Timber are you using?
WordPress 4.5.2, PHP 7, Timber 1.01 -->

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
`composer require timber/timber`

connorjburton:
There has been another report of this, where do you require the autoloader.php?

And what is your setup that you can't require it in the theme?

Thanks!

connorjburton:
Hi, can you please try the fix in #997 (you should be able to do `require timber/timber:dev-cjb-autoload-root`)

jarednova:
@stewarty I merged @connorjburton's changes into master — can you give the newest version a test to confirm that it solves your issue?

PhilippBaschke:
[...]

PR #997, opened 2016-05-13
Trying to fix init at root wp level

connorjburton:
**Ticket**: #991 and #979 
**Reviewer**: @jarednova 

#### Issue

If you include Timber at root level it will throw a fatal error as wordpress is not available at that point and we make use of wordpress functions.

#### Solution

Defer setup until `WP` is available, we defer by calling the `init` function on each `compile` method, if the new constant `TIMBER_LOADED` is not defined.

#### Impact

We will see, early days on this one.

#### Usage

No change

#### Considerations

I tried to make this as small of a change as possible to achieve the desired result and we should keep that in mind if we need to develop it further.

#### Testing

No tests included

### DIFF
--- a/lib/Timber.php
+++ b/lib/Timber.php
@@ -91,12 +91,11 @@ class Timber {
 	 * @codeCoverageIgnore
 	 */
 	protected function init() {
-		if(class_exists('\WP') && !defined('TIMBER_LOADED')) {
+		if( class_exists('\WP') && !defined('TIMBER_LOADED') ) {
 			Twig::init();
 			ImageHelper::init();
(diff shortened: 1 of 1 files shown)


jarednova:
Just fixed the Travis bug in `master` — with that clear I'll take a look at this guy today!

jarednova:
We're looking good! I made some stylistic tweaks, but otherwise I like the approach of "touch as little as possible". As picked-up by the Coveralls warning, there's no test coverage yet. Given that this is an active bug, I'm still going to merge but something we should circle-back on at some point

Issue #992, opened 2016-05-11
Q: Getting post type labels


swthate:
I've been looking all over on how to get post type labels/names into my theme. Originally I did `post.post_type`, but that returns something like "galleries" or "cabinet-meetings". What I want it to return is the *name* label: "Cabinet Meetings". Is it buried away in some post object or something? I did `dump(post)` but all it showed as far as post types was `["post_type"] => string(16) "cabinet-meetings"`.

Thanks!

connorjburton:
This is a very good question, got a feeling this will require a new `Timber\PostType` object that you can pass the slug/id into and get all the details of.

Thoughts @jarednova - should be fairly simple to implement

jarednova:
actually `{{ post.get_post_type }}` will return the object that @swthate is looking for. That said, this might be ripe for revisiting with some clearer syntax soon (I have an idea that I'll PR)

swthate:
Ah-ha. `dump(post.get_post_type)` does return exactly what I'm looking for. Not sure how to go about extracting `["labels"] => { ["name"] => "Calendars" }` though (or is that what you have an idea on? :))

nickmoline:
since in twig objects and arrays are treated the same way (dot notation), it should be `{{ post.get_post_type.labels.name }}`

swthate:
Sweet. I wasn't sure that I was able to make a chain that long!

Thanks, @nickmoline

jarednova:
closed by #1003

christophhapple:
Since `post.get_post_type` has been deprecated in 1.4 I was just wondering if there is any sort of replacement for this?
If not, what is the best practice to get post type labels in a template?

jarednova:
@christophhapple here ya go ...

`{{ post.type.labels.name }}`

if you do a `{{ dump( post.type ) }}` you can see everything inside the object you might want to use

christophhapple:
Thanks @jarednova for getting back to me so quickly.
Shame on me for missing that ...

jarednova:
This is WordPress development — there's no room left for shame

PR #1003, opened 2016-05-17
992 post type labels

jarednova:
**Ticket**: #992 
**Reviewer**: @connorjburton 

#### Issue
Need an API to access post type information that matches other methods on Post

#### Solution
`{{ post.type }}` can store and return this data

#### Impact
If a user has a custom field named "type" they're gonna have a bad time

#### Usage
`{{ post.type }}` and the old method `{{ post.get_post_type }}` still work

#### Considerations
Should `{{ post.post_type }}` return this data too? At first I thought so, but then I changed my mind. Thoughts?

#### Testing
Tests included!

### DIFF
--- a/lib/Comment.php
+++ b/lib/Comment.php
@@ -42,7 +42,7 @@ class Comment extends Core implements CoreInterface {
 	public $post_id;
 	public $comment_author;
 
-	public $children = array();
+	protected $children = array();
 
 	/**
 	 * @param int $cid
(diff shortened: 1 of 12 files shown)


jarednova:
I just caught a failing test, please hold on reviewing for now...

fewerflatlands:
as you say, **If a user has a custom field named "type" they're gonna have a bad time** - in fact this appears to be a breaking change in 1.04... I've just had to downgrade to 1.03 as a result. any suggestions for a workaround?

connorjburton:
We should probably change this to `post.post_type` @jarednova ?

connorjburton:
Related #1057

connorjburton:
We could also make this pluggable, so only add the `type` property if it doesn't already exist.

fewerflatlands:
either of those (`post_type` rather than `type` or the conditional add) seem better than bulldozing any custom field called type... it's got to be a pretty common/likely field name, no?

jarednova:
@connorjburton @fewerflatlands thanks for the heads-up on that. I agree and will work on a patch now

Issue #1005, opened 2016-05-17
Q: Timber plugin version method


ogrosko:
You have made great job with version 1.0.*
Is there some method for getting timber version?

There are namespaces naming changed and I need to keep compatibility for older versions.
I would need something like:

`if (version_compare(\Timber::get_version(), '1.0.0', '<')) {
	# old namspace functionality
}
else {
	# new namspace functionality
}`

connorjburton:
Hi,

Timber 1.0.X should be fully backwards compatible with old versions. Is there something specific that you are doing that isnt working?

ogrosko:
Yes I am using small plugin for cache clearing https://github.com/ogrosko/clear-cache-for-timber/blob/master/clear-cache-for-timber.php

See line 40:    `echo TimberCommand::clear_cache();` results in error due to namespace changed in v 1.0

connorjburton:
Aha, I must have missed `TimberCommand` when transitioning to namespacing. I will fix this in `1.0.3`. I will also introduce/investigate a version method/attribute so you can check the current version.

PR #1008, opened 2016-05-17
Adding version attribute

connorjburton:
**Ticket**: #1005 
**Reviewer**: @jarednova 

#### Issue

There is no way to get the current version of Timber

#### Solution

Add static property with current version number

#### Impact

No impact

#### Usage

`Timber::$version`

#### Considerations

We could use a function, but just makes more sense for it to be an attribute

#### Testing

Not included

### DIFF
--- a/lib/Timber.php
+++ b/lib/Timber.php
@@ -32,7 +32,7 @@ use Timber\Loader;
  */
 class Timber {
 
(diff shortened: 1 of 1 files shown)


jarednova:
lookin' good!

PR #1012, opened 2016-05-20
Fixed in_plugin_update_message bug

lggorman:
#### Issue
The in_plugin_update_message was treating the [$plugin_data](https://developer.wordpress.org/reference/hooks/in_plugin_update_message-file/) parameter like an object, when it's actually an array.  As a result I was seeing a ` Trying to get property of non-object` error.

#### Solution
Changed the arrow syntax to square bracket.

#### Impact
Nah.  But it would be nice if there were some way to propagate this tweak to earlier versions.

### DIFF
--- a/lib/Admin.php
+++ b/lib/Admin.php
@@ -3,7 +3,7 @@
 namespace Timber;
 
 class Admin {
-	
+
 	public static function init() {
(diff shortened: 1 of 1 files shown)


jarednova:
Thanks @lggorman !

Issue #65, opened 2013-09-18
Rethink get_preview API


jarednova:
Looking for feedback on how to best structure this method, currently it's:

{{post.get_preview(60, $force, 'Read More >>', $strip_tags)}}

where.
60 (number of words to show), 
$force (whether to "force" manually set post excerpts to a shorter length; or let them in at full length)
'Read More >>' (read more hyperlink)
$strip_tags (strips any tags found in the text)
#### Some questions...
- Should arguments be re-ordered?
- How to handle ellipses that are added?
- Should some of this be controlled by global vars or filters?

jeffjack:
For those looking for a way to remove the ' `&hellip;` ' string and don't want to edit the plugin files (don't edit the plugin files. bad. no.), you can just set up a simple script to switch that out with something else:

in your twig file:

    {{post.get_preview(25,false,"",true)|fixEllipsis}}


in your functions.php:

    function add_to_twig($twig){
        $twig->addFilter('fixEllipsis', new Twig_Filter_Function('fixEllipsis'));
        return $twig;
    }
    add_filter('get_twig', 'add_to_twig');

    function fixEllipsis($text){
        $text = str_replace(' &hellip; ', '&hellip;', $text);
        return $text;
    }

aeblin:
@jeffjack Digging the concept, but I can't seem to make that work. Any idea if things may have changed since you threw that together?

jeffjack:
@aeblin Hmm it's still working for me after updating to Timber 0.21.7. I'm on WP 4.2.2 as well. I didn't include it in the example above for the sake of brevity, but make sure you also add the StringLoader extension to $twig as well. For example:
```
function add_to_twig($twig){
    $twig->addExtension(new Twig_Extension_StringLoader());
    $twig->addFilter('fixEllipsis', new Twig_Filter_Function('fixEllipsis'));
    return $twig;
}
add_filter('get_twig', 'add_to_twig');
```
Here's another example use in a twig file:
```
{{post.get_preview(50,false,"Read More",true)|fixEllipsis({'text_replacement': '&hellip; '})}}
```

aeblin:
Hmm, even with the StringLoader extension added, I tend to just get back a syntax error.
```
 unexpected 'add_filter' (T_STRING), expecting function (T_FUNCTION)
```

jeffjack:
[...]

PR #1015, opened 2016-05-21
65 post preview

jarednova:
**Ticket**: #65
**Reviewer**: @connorjburton 

#### Issue
`{{ post.get_preview }}` is a hodge-podge of arguments. Too many arguments. And in an order that makes no particular sense.

#### Solution
Create an object to manage post previews!

#### Impact
`{{ post.get_preview}}` is unaffected. The `PostPreview` object should make it easier to add customizations down the line

#### Usage
It's easy and fun: `{{ post.preview.length(20) }}` or `{{ post.preview.length(20).force.read_more("Keep Reading") }}` or `{{ post.preview.length(25).end('...') }}`

#### Considerations
`{{ post.get_preview }}` could be refactored to use the processing in the `PostPreview` object, but now as I think about it, it might just be best to leave it so that that the Preview object can evolve w/o backwards compatibility concerns.

#### Testing
Tests included!

### DIFF
--- a/lib/Post.php
+++ b/lib/Post.php
@@ -126,7 +126,7 @@ class Post extends Core implements CoreInterface {
 	public $post_date;
 
 	/**
-	 * @var string 	$post_exceprt 	the raw text of a manual post exceprt as stored in the database
+	 * @var string 	$post_excerpt 	the raw text of a manual post excerpt as stored in the database
 	 */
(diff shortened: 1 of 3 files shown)


connorjburton:
Will review today

connorjburton:
I like it, very functional programming-esque.

PR #1016, opened 2016-05-22
Passing a name as a first argument to the Twig_Environment::addFunction method is deprecated since version 1.21

simonmilz:
As the headline said, i removed some deprecated function calls.


jarednova:
Thanks @simonmilz!

Issue #169, opened 2014-03-12
Problem with Fieldname-Prefix from Types Custom Fiels Plugin


motionVector:
I’m trying to use Timber with the Custom Fields Plugin from http://wp-types.com/. 
Because the plugin prefixes all fieldnames with “wpcf-”, the dash causes a problem, when I want to output a field with TWIG. I know I could use the attribute function {{ attribute(page, 'wpcf-fieldname') }} , but I’d rather not have to use it all the time, because its so much more complicated. 
Is there a workaround for this problem, maybe with a subclass of TimberPost or adding an apply_filter that removes the prefix, when the meta info is loaded into TimberPost?

xavivars:
Wouldn't be good to add this as an integration, similar to the one with ACF?

connorjburton:
Hey everyone. Am I missing something here, can you not just do `{{ post['my-thing'] }}`?

PR #1021, opened 2016-05-23
Xavivars add wptypes integration

jarednova:
**Ticket**: #169 and #680 
**Reviewer**: @jarednova

Hey @xavivars — this is long overdue. Thanks for submitting the PR in #680. I'm sorry it took me so long to follow-up. I'd love to pull this into Timber, but I want to make sure we have proper test coverage for the additional integration. Could you take a look at writing unit tests that will cover Types?


jarednova:
@xavivars I picked this up hoping I could bring it home, but it needs some more work to get merged. The two things that jump out at me are...

- [ ] Support for types beyond posts: users, taxonomies (any others I'm not seeing?): _to truly provide integration w another plugin, I'd want to provide full integration_
- [ ] Unit tests

I made some starts in tests and bootstrapping here: https://github.com/timber/timber/tree/xavivars-add-wptypes-integration

I'm going to close for now, but please submit a new PR once you (or someone else) can bring this all the way

xavivars:
In fact, my goal was simply to be able to use the same methods than with ACF, and could not do it because wp types adds wpcf as a prefix for all meta fields. But I agree that full coverage would be desirable. 

In my case, we started migrating from WP Types to ACF, so I don't think I'll have the need to work on this on the near term, but if I have the time, I'll give it another try.

Issue #942, opened 2016-04-19
ACF has_field


connorjburton:
### Expected behavior
`post.has_field('foobar')` => boolean depending on if it exists or not on that post

### Actual behavior

`NULL`

### Steps to reproduce behavior

- Create a post with ACF field 'foobar'
- Try `post.has_field('foobar')`
- Will return `NULL`

### What version of WordPress, PHP and Timber are you using?

Latest all

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)

Composer

jarednova:
Closed by #1024

PR #1024, opened 2016-05-24
Adding has field + tests

connorjburton:
**Ticket**: #942
**Reviewer**: @jarednova 

#### Issue

`post.has_field(fieldname)` doesn't exist

#### Solution

Add an alias to the `get_field` method that returns a `boolean` rather than `mixed`

#### Impact

None, new method

#### Usage

`post.has_field('field')` => `boolean`

#### Considerations

There isn't any real benefit to this, it's mostly syntax sugar, but it is more inline with ACF itself.

#### Testing

Tests included, I hope they run...

### DIFF
--- a/lib/Post.php
+++ b/lib/Post.php
@@ -636,6 +636,13 @@ class Post extends Core implements CoreInterface {
 		return get_comments_number($this->ID);
 	}
 
+	/**
+	 * @param string $field_name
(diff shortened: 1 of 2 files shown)


jarednova:
Thanks @connorjburton — just waiting for tests to finish and then this is ready for mergin

Issue #933, opened 2016-04-11
escaping file extension (.twig) while including or extending a file.


kala725:
### Expected behavior

Should take the file. Suppose I have a file called "default.twig" so {extends 'default'} should render the file. Currently it is throwing an load error. The similar functionality is provided in TwigBridge (Laravel)

### Actual behavior

Throwing an error

### Steps to reproduce behavior

just include a file without .twig extension. It will throw Twig_Error_Load

### What version of WordPress, PHP and Timber are you using?

Wordpress 4.3
Timber 0.21.10
Php 5.5

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)

Github through composer dependency

jarednova:
I'm closing out old issues. Please open a fresh issue if this is still a live issue that we should investigate — thank you!

PR #1025, opened 2016-05-24
Adding support for loading views without twig extension

connorjburton:
**Ticket**: #933
**Reviewer**: @jarednova
#### Issue

Can't load views without specifying `.twig` extension
#### Solution

Check if the filename is a string, check if the template doesn't exist (if someone wants to load `myview` without an extension and the file exists, then we should let them), then check if the filename doesn't end in `.twig` (reverse the string as it's quicker). If all that is true, we add `.twig` on the filename.
#### Impact

Little bit slower
#### Usage

`{% include 'views/hello' %}`
#### Considerations

This only works for strings, not arrays of views. I think we should refactor this method to be recursive, I think it would make more sense. Let me know if you want me to do this. 
#### Testing

None included


connorjburton:
Tests are failing for some reason, will come back to this

connorjburton:
Should be okay to merge @jarednova

connorjburton:
Not sure how coverage has gone down here?

jarednova:
I've noticed that Coveralls can throw back weird numbers on occasion — you def. have coverage on those lines. Before merging I want to make sure this works with arrays so that we don't end up with a situation where not adding `.twig` _sometimes_ works.

I could see this being accomplished through a private `maybe_add_twig_extension` method (sorry for being so prescriptive) so the `choose_template` method doesn't get too unwieldy. That said, use your judgement — almost there!

jarednova:
I just want to confirm the expected functionality here...

If I try to load `{% include 'views/hello' %}` and the file `views/hello` exists, the content of `views/hello` will be used.

If I try to load `{% include 'views/hello' %}` and the file `views/hello` DOES NOT exists, the content of `views/hello.twig` will be used (if it exists).

jarednova:
@lggorman and I never had much luck with this guy, going to close for now and pursue a diff't solution to #933

Issue #927, opened 2016-04-01
Make TimberImage support ACF's "image array" type


tremby:
### Expected behavior

Be able to pass an "image array" as set up in ACF to `TimberImage` in twig, and have a usable image.

### Actual behavior

The constructed `TimberImage` object works fine for the purposes of getting its `src`, but doesn't have `width` and `height` and possibly other things. I have to do `TimberImage(image_field.ID)` to make it work, or change the custom fields to return the image ID instead.

This sent me into debugging mode to try to figure out why the `TimberImage` was seemingly fine (`src` works) but didn't have width and height. It's confusing that it partially works and partially doesn't. Bugs would actually be caught much sooner if it didn't work at all when passed an array like this, but obviously it'd be much more useful if it handled the "image array" properly.

### Steps to reproduce behavior

1. Make image type ACF field, set to return "image array".
2. Get this field in Twig and do something like `{{ dump(TimberImage(image_array).width) }}`, and see that it shows null.

### What version of WordPress, PHP and Timber are you using?

WP 4.4.2, Timber 0.22.5.

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)

Composer

connorjburton:
Thanks for this, the docs do recommend to return ID's rather than the array, however I will put this on the list to fix post `1.0.0`

davidmaneuver:
Just went into debugging mode myself to figure this out and came to the same conclusion.

Only difference is that I have a ACF gallery field (instead of 1 image field) and it always returns as an array, there is no option to return as an ID.

But the following works just fine.

`````
{% for image in post.get_field('gallery_images') %}
  {% set image = TimberImage(image.id) %}
  ...
{% endfor %}
`````

**Comment by connorjburton** [comment]
I have created a PR for this, if you guys could give it a test that'd be great. Thanks

**Comment by jarednova** [comment]
@tremby @davidmaneuver #1026 is now merged in, please give it a look and confirm that this solves your issue and we can close this guy up!

**Comment by tremby** [comment]
Works for me. With this change, instead of

```
{% set image = TimberImage(post.get_field('my_image').ID) %}
```

I can now do

```
{% set image = TimberImage(post.get_field('my_image')) %}
```

jarednova:
Great!

PR #1026, opened 2016-05-24
Catching acf image array

connorjburton:
**Ticket**: #927
**Reviewer**: @jarednova 

#### Issue

Passing an ACF image array would not return a full `Timber\Image` object.

#### Solution

Check if an array has been passed and if that array has an ID property, if it does grab the ID and continue.

#### Impact

`Timber\Image` instances initiated with an ACF image array will work properly.

#### Usage

No change

#### Considerations

`init` method is getting ever larger, it seems to do some logic which should probably be moved out.

#### Testing

None included.

### DIFF
--- a/lib/Image.php
+++ b/lib/Image.php
@@ -222,10 +222,19 @@ class Image extends Post implements CoreInterface {
 	 * @param int $iid
 	 */
 	function init( $iid = false ) {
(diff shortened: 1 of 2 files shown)


jarednova:
All good — thanks @connorjburton

Issue #568, opened 2015-06-11
Multisite and {{site}} variable


haroldangenent:
On Multisite, `TimberSite::init_with_multisite()` is called, instead of `TimberSite::init()`. However, some attributes are not set or set to the wrong value. For example:

- `charset` is not set
- `language_attributes` is not set
- `language` is not set
- `url` is set to the wrong value (Site URL instead of Home URL like in single-site installs).

There are probably several other differences.

Is this by design?

jarednova:
Hi @haroldangenent, not by design -- likely by oversight (multisite was added later). Are the properties you're referring to on the main Timber `$context` object? Or ones on the `TimberSite` object?

haroldangenent:
The ones on the `TimberSite` object (so in the `$context['site']` variable). You can see the differences by looking at `TimberSite::init_with_multisite()` and `TimberSite::init()`.

jarednova:
cool, thanks @haroldangenent I'll take a look!

jarednova:
@haroldangenent -- this has a patch: PR #583. Tests are running now in Travis. If you have a chance to evaluate, I'd love your feedback before merging into master

haroldangenent:
@jarednova Seems to work great, thanks!

haroldangenent:
@jarednova While the patch worked and 0.21.4 (https://github.com/timber/timber/blob/0.21.4/lib/timber-site.php) was working as well, this seems to have been reverted in later installs.

Is this done purposefully? If so, I am now having issues again with a different `home` than `site_url` (for example, when installing WP in a separate folder).

connorjburton:
It seems to have been reverted in this commit https://github.com/timber/timber/commit/83b04f6aaf613acfd5a5ed78b96666036d173009 - will investigate

jarednova:
@connorjburton has patched this in #1028. @haroldangenent: are you able to confirm this by trying out the current `master` in your project?

haroldangenent:
@connorjburton @jarednova Thanks! Couldn't test really thorough, because most Multisite projects and starters themes are broken under 1.0 (congratulations on 1.0 by the way!), because of namespacing. 

After some superficial fixing, it does seem to work properly. I would recommend using `home_url()` over `get_bloginfo()`, since `get_bloginfo('url')` just calls `home_url()`. `home_url()` is a bit more descriptive, since it specifically shows the destinction between the home and site url.

connorjburton:
[...]

PR #1028, opened 2016-05-24
Revert a revert for multiste url fix

connorjburton:
**Ticket**: #568
**Reviewer**: @jarednova 

#### Issue

A fix was made for the multsite `site.url` property that was reverted at a point afterwards.

#### Solution

Revert the revert as the issues seems to still exist according to the ticket

#### Impact

Should fix multisite urls

#### Usage

No change

#### Testing

None included

### DIFF
--- a/bin/phpunit-multisite.xml
+++ b/bin/phpunit-multisite.xml
@@ -1,0 +1,23 @@
+<phpunit
+	bootstrap="../tests/bootstrap.php"
+	backupGlobals="false"
+	colors="true"
+	convertErrorsToExceptions="true"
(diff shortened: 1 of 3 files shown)


jarednova:
Woohoo, thanks for all the PRs @connorjburton! I'll start here and make my way down the list in the next few days

connorjburton:
It's nearly everything for 1.0.4, apart from that composer require from root issue.. 

Probably lots of failing tests, if you don't get to them first I'll sort them

jarednova:
Thanks @connorjburton — let's confirm that this takes care of #568 once and for all and we can close that one out too

PR #1029, opened 2016-05-26
get pids? what was I thinking

jarednova:
**Reviewer**: @connorjburton 

#### Issue
`PostGetter::get_pids` — did anyone ever use it? I did waaaaaay at the beginning and then kinda forgot about this method. It was never documented and is a remnant of "Jared's personal WordPress toolbelt" vs. Timber

#### Solution
Kill it

#### Impact
While it's possible that there's one person out there who used it — I doubt it. It was never documented and anyone who actually saw the code would see that it's no more efficient vs. `get_posts`

#### Testing
Not covered by tests either

### DIFF
--- a/lib/PostGetter.php
+++ b/lib/PostGetter.php
@@ -59,17 +59,9 @@ class PostGetter {
 		}
 	}
 
-	static function get_pids( $query ) {
-		$posts = self::get_posts($query);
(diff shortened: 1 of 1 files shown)


connorjburton:
It's deaded!

Issue #1017, opened 2016-05-22
TimberImage resize working locally but not on amazon server


kmturley:
### Expected behavior
I was expecting TimberImage to resize images on my amazon ec2 server as it does on my local machine

### Actual behavior
It is outputting the original image, for example:

    <img src="{{ post.thumbnail.src|resize(640, 360) }}" alt="" />

When outputted locally the result is:

    http://localhost:8888/new/wp-content/uploads/2016/05/abc-640x360-c-default.jpg

When running the same code on amazon ec2 the result is:

    http://mysite.com/new/wp-content/uploads/2016/05/abc.jpg

### Steps to reproduce behavior
Normal resize image code, on an amazon ec2 instance. I've tried enabling errors to ensure it wasn't erroring:

    define('WP_DEBUG', true);
    define('WP_DEBUG_LOG', true);
    define('WP_DEBUG_DISPLAY', true);

and also in my php.ini, but nothing errors out:

    display_errors = on

### What version of WordPress, PHP and Timber are you using?
Local: WordPress 4.4.2, Timber 1.0.3, PHP 5.5.29
Server: WordPress 4.4.2, Timber 1.0.3, PHP 5.3.29

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
via Wordpress plugin manager

Guides used to install wordpress:
http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/install-LAMP.html
http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/hosting-wordpress.html

kmturley:
I added some code to TimberImage.php to add a param to the image url depending on where it is served:

    if ( file_exists($new_server_path) ) {
        if ( $force ) {
            // Force operation - warning: will regenerate the image on every pageload, use for testing purposes only!
            unlink($new_server_path.'?force');
        } else {
            // return existing file (caching)
            return $new_url.'?cache';
        }
    }
    // otherwise generate result file
    if ( $op->run($old_server_path, $new_server_path) ) {
        if ( get_class($op) === 'TimberImageOperationResize' && $external ) {
            $new_url = strtolower($new_url);
        }
        return $new_url.'?success';
    } else {
        // in case of error, we return source file itself
        return $src.'?error';
    }

And the result I am getting is:

    http://mysite.com/new/wp-content/uploads/2016/04/Photo-Mar-10-11-49-23-PM.jpg?error

kmturley:
[...]

PR #1030, opened 2016-05-26
Fixing double image object issue

connorjburton:
**Ticket**: None
**Reviewer**: @jarednova 

#### Issue

I discovered a bug that I haven't seem reported anywhere, I got a fatal error when trying to resize an image.

If you look at the code before/after, you will see that we run `wp_get_image_editor` in both `run_animated_gif` and `run`, the result of that then gets passed into `get_target_sizes` which will then run `wp_get_image_editor` once again, and pass in the result of the previous `wp_get_image_editor`

#### Solution

Firstly I moved the errors further up the method so we catch any errors without throwing fatal errors on production sites.

Then I passed the result of `wp_get_image_editor` into `get_target_sizes` and didn't run that method again once we were in there.

#### Impact

Should fix this bug

#### Usage

No change

#### Considerations

I couldn't find any other mention of this, so maybe it's my setup, but the code didn't look right to me.

#### Testing

None included

### DIFF
--- a/lib/Image/Operation/Resize.php
+++ b/lib/Image/Operation/Resize.php
@@ -205,7 +205,11 @@ class Resize extends ImageOperation {
 			// @codeCoverageIgnoreStart
 			Helper::error_log( 'Error loading ' . $image->error_data['error_loading_image'] );
 		} else {
-			Helper::error_log( $image );
+			if(!extension_loaded('gd')) {
+				Helper::error_log( 'Can not resize image, please installed php-gd' );
+			} else {
+				Helper::error_log( $image );
+			}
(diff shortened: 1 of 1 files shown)


connorjburton:
Couple of issues thrown up by tests, will trace and fix them

jarednova:
cool, I'll bounce it back to you for the test adjustments

connorjburton:
Also added a fix for #1017 in this PR

connorjburton:
Type hinting, I like it. I think I'll go through and add type hinting to everything that I can at some point. (After v1.0.4)

jarednova:
@connorjburton sounds good!

Issue #837, opened 2016-01-29
Timber::query_posts() && count


ruscon:
How to count posts from the TimberQueryIterator ?
May be need to add Countable interface ?

```php
public function count() 
{
    return $this->_query->post_count;
}
```

PR #1032, opened 2016-05-27
Adding count property to queryiterator

connorjburton:
**Ticket**: #837
**Reviewer**: @jarednova 

#### Issue

If you have a `Timber\QueryIterator` object, there is no way to get the total posts returned, for example of a search results page with the default `posts` object in context.

#### Solution

Add a method to return the private `_query.post_count` property

#### Impact

No impact

#### Usage

` {{ QueryIteratorObject.post_count }}`

#### Considerations

There may be other properties that would be useful to expose. Also I am not sure on the reason behind returning a `QueryIterator` rather than a collection of `Timber\Post`s, you could then do `{{ posts | length }}` to get the total posts

#### Testing

None


jarednova:
Thanks @connorjburton — @lggorman, could you write a test to cover the new code?

lggorman:
@jarednova sure thing

Issue #671, opened 2015-09-02
next_post_link with argument


MateuszCisowski:
Hi,

First of all - great job with Timber!

I have small issue with function. What I'm trying to achieve is a next/prev links for the custom post types with my labels.

Example from Codex:
`<?php next_post_link( '%link', 'Next post in category', TRUE ); ?>`

When I use just function (without any arguments) in this way:
`{{ function('previous_post_link') }}`
it works correctly but when I add arguments:
`{{ function('previous_post_link', '%link', 'next project', TRUE) }}`
nothing shows up.

I'll be really grateful for help.
Thanks in advance!

MateuszCisowski:
I have tried with the
`$context['pagination'] = Timber::get_pagination();`
and in my twig file
`<a href="{{ pagination.prev.link }}">next project</a>`

and it doesn't work... "next project" links to the current page.

What is strange is that when I delete `$content['pagination']` it works exactly the same...  when I dump `$context['pagination']` it shows
`array(2) { ["pages"]=> array(0) { } ["prev"]=> string(0) "" }`

Any ideas?

MateuszCisowski:
OK, I have found temporary solution, I paste it below in case that someone will need it:
```
$next_post = get_previous_post();
if ( is_a( $next_post , 'WP_Post' ) ) {
   $context['nextPost'] = '<a href="' . get_permalink( $next_post->ID ) . '">' . 'next project' .  '</a>';
}

$prev_post = get_next_post();
if ( is_a( $prev_post , 'WP_Post' ) ) {
    $context['prevPost'] = '<a href="' . get_permalink( $prev_post->ID ) . '">' . 'previous project' .  '</a>';
}
```

However I'm still curious how to solve this problem using first (function with arguments) and second (Timber::get_pagination) approach. Thanks!

stewarty:
I think I might be running into something similar.

I'm trying to get previous/next posts within the same category but Timber's post object is having none of it  , both the prev() & next() methods are returning false.

`$context['post']->next = $post->next(true);
$context['post']->prev = $post->prev(true);`

connorjburton:
Thanks for the update @stewarty, will look at this next week

henrikwirth:
For me `$context['pagination'] = Timber::get_pagination();` returns only this:
```
array (size=5)
  'current' => int 1
  'total' => float 1
  'pages' => 
    array (size=0)
      empty
  'next' => null
  'prev' => string '' (length=0)
```
Any thoughts?

[...]

PR #1040, opened 2016-06-01
Set next and prev to use paginate_links array 

lggorman:
**Ticket**: 
**Reviewer**: @jarednova

#### Issue
The `Timber::get_pagination` function accepts a preferences array, which can include arguments like a custom base or format for the pagination links.  These args are then passed into the the paginate_links function, which generates the `pages` attribute of the pagination array.  However, the `next` and `prev` attributes of the array are generated separately, using `get_next_posts_page_link` and `previous_posts`, neither of which uses the prefs array.  So basically the issue is that if you try to pass custom args, they will not be incorporated into the `next` and `prev` links.

#### Solution
Instead of using `get_next_posts_page_link` and `previous_posts`, just get the next and prev links from the `pages` array generated by `paginate_links`. 

#### Impact
None

#### Usage
No changes

#### Considerations
--

#### Testing
Tests are included!


jarednova:
Awesome – thanks @lggorman! I'll review this once the tests pass

jarednova:
This is great @lggorman! With this gap patched, I'd still like you to think about the best API to interact with pagination; we can talk next week a little more about how this might take shape

connorjburton:
Does this solve #671 ?

lggorman:
@connorjburton Don't think so unfortunately -- this just sets next and prev to pull from that `pages` array, so if that's empty as well there's still going to be an issue.

PR #1041, opened 2016-06-02
[Proposal] Composer compatible initalization

mrgrain:
**Ticket**: #979 #981 #991 
**Reviewer**: @jarednova @connorjburton @chrisgherbert

#### Issue
Timber includes a `init.php` in the autoloader for initialization. However, as there is no guaranteed point in time (in terms of the WordPress lifecycle) when the autoloader is included. This approach effectivly limits Timber to only use 'itself' during initialization. To overcome this limitation, the current implementation checks for typical WordPress functionality to determine if WordPress has been loaded or not.

Unfortunately it solves the issue only half-way. In a set-up where Timber is included as a real dependency (e.g. [Bedrock ](https://github.com/roots/bedrock) or [Autobahn](https://github.com/mrgrain/autobahn)), the initialization won't happen.

#### Solution
This Pull Requests proposes to remove any automatic initialization functionality. It suggest moving this responsibility to the theme. Therefore the documentation has been updated.
The plugin version of Timber will continue to ship with this auto initialization. The necessary code has been moved to the plugin file.

Proposed changes:
- Removed init.php from autoloader
- Removed init.php
- Added default view dirname
- Extended Readme with initialization instructions
- Added init to WordPress plugin file
- Adjusted `tests/bootstrap.php` to work without `init.php`

#### Impact
Set-ups relying on this initialization are likely to break, but most of them won't work anyway with releases between 1.0.0-rc2 and 1.0.3. 

#### Usage
A theme using timber will need to add the following line (or something similar):
```
$timber = new \Timber\Timber();
```
The documentation has been updated accordingly.

#### Considerations
The Timber object is currently used in a mostly static concepts. Initialization could happen on the first static call on Timber. However, there is not clearly defined entry point for themes (could be `get_context`, `get_post` or `compile`).

The starter theme should be adjusted accordingly, but as of now it's still based on [Timber 0.22](https://github.com/timber/starter-theme/blob/master/composer.json#L27) anyway.

Since this is an API change and possibly BC-breaking, the change could be introduced to version 1.1. The current fix (as proposed for 1.0.4) is not ideal, but gets the job done.

#### Testing
No additional tests required. Test bootstrap has been adjusted to work without `init.php`. Interestingly, the test is using a custom initialization already.


connorjburton:
I agree with this PR completely, it should be up to theme to init Timber. The only remaining question is `1.0.4` or `1.1.0`. @jarednova thoughts?

Thanks very much for this PR @mrgrain

mrgrain:
Go with `1.1.0` but make sure, that `1.0.4` gets released ASAP, maybe only including the check for WordPress.

Also it might be time to have planned releases now. ;)

jarednova:
Thanks @mrgrain! This will go out in version 1.1 — I really appreciate the thoroughness of updating the Readme and WP.org files, etc. That's the stuff I often find I have to do (or forget) as a part of updating the code.

PR #1042, opened 2016-06-02
Fix usage of old classnames

mrgrain:
**Reviewer**: @jarednova 

#### Issue
The code is still relying on the old classnames without namespaces.

#### Solution
Changing oldclassnames to new namespaces names.

#### Impact
None.

#### Usage
No changes.

#### Considerations
I'm quite sure I still missed some.

#### Testing
No changes. Everything is still passing.


jarednova:
Thanks @mrgrain, merged!

Issue #687, opened 2015-09-15
Conflict with installed WP CLI


kevin-coyle:
I've installed wp cli separately but when I enable timber I get the following error:

````
PHP Fatal error:  Cannot redeclare cli\render() (previously declared in phar:///usr/local/bin/wp/vendor/wp-cli/php-cli-tools/lib/cli/cli.php:26) in /app/wp-content/plugins/timber/vendor/wp-cli/php-cli-tools/lib/cli/cli.php on line 28
Fatal error: Cannot redeclare cli\render() (previously declared in phar:///usr/local/bin/wp/vendor/wp-cli/php-cli-tools/lib/cli/cli.php:26) in /app/wp-content/plugins/timber/vendor/wp-cli/php-cli-tools/lib/cli/cli.php on line 28
````

Looks like there's a conflict with the version that composer installed and the version that I would normally have on my system anyway.

**Comment by hurrycaner** [comment]
I don't know if it will bring more problems than solutions, but try to remove ```"wp-cli/wp-cli": "*",``` line from composer.json in timber folder. Since it's just a require-dev package, it shouldn't bother in production environment.

IMO this is a wp-cli problem, at ```php-cli-tools``` package... (https://github.com/wp-cli/php-cli-tools/).

**Comment by danielbachhuber** [comment]
The problem is that WP-CLI has already loaded `php-cli-tools`, and then Timber tries to load it again via Composer autoload.

WP-CLI can't do anything different because it needs `php-cli-tools` available at runtime. I'm not sure why Timber is loading WP-CLI again with Composer.

**Comment by danielbachhuber** [comment]
WP-CLI was added as a dependency in c85b643d7c3e2ae98a68b188fbae316073014110 To me, the commit message doesn't explain why.

**Comment by kevin-coyle** [comment]
Could this just be avoided by doing something like `composer install --no-dev`

**Comment by ffub** [comment]
My solution to this has been to skip the Timber plugin in my wp-cli.yml file:

```yaml
skip-plugins:
  - timber
```

mortona42:
I was able to install timber by removing wp-cli from composer.json, removing composer.lock, and running composer update.

connorjburton:
@jarednova - are you able to shed any light on this? Thanks

andrejcremoznik:
It makes no sense to me to require wp-cli in project. Developers relying on it have it installed system-wide already. Also, wp-cli as the name implies is a command line tool. It has no purpose in an application.

jarednova:
[...]

PR #1047, opened 2016-06-02
Removing wp cli

connorjburton:
**Ticket**: #687 
**Reviewer**: @jarednova 

The ticket explains it much better than I can.


mrgrain:
If we don't need it, than let's just get rid of it.

However, I think it's worth mentioning that this is a common issue with WP CLI. If you require it via composer, you'll have to use the `./vendor/bin/wp` version of it. Any other will fail.

jarednova:
Good riddance

Issue #1043, opened 2016-06-02
Image resizing uppercase file extensions


commercial-hippie:
When resizing files that have uppercase extensions I'm running into an issue.

``` twig
// Original file: post.thumbnail = foo.JPG
{{ post.thumbnail | resize(400, 500) }}
// Resized file: foo-400x500-c-default.jpg
// Resized link: foo-400x500-c-default.JPG
```
As you can see above, the final link goes to the wrong image and turns into a 404.

#### Steps to reproduce behavior

Import a image with a uppercase extension.  You can't just upload a image because Wordpress converts it to a lowercase extension.  (I imported all my posts, images from a ExpressionEngine install).

WordPress 4.5.22, PHP 7.0.4, Timber 1.0.3

commercial-hippie:
Such a quick update! Thanks for looking into it so quickly, really appreciate is! Great plugin :smile:

jarednova:
Thanks @commercial-hippie! @connorjburton submitted a PR that I just merged. Look for a release to WP.org early next week!

PR #1048, opened 2016-06-02
Fixing uppercase extension

connorjburton:
**Ticket**: #1043 
**Reviewer**: @jarednova 

#### Issue

If you tried to resize an image with an uppercase extension it would return the URL with the uppercase extension, which would be fine if we didn't `strtolower` the extension when saving the resized image.

#### Solution

`strtolower` the extension that we get back from `pathinfo`

#### Impact

None

#### Usage

No change


jarednova:
Thanks @connorjburton !

Issue #386, opened 2014-11-07
[Enhancement] Support get_term function of Wordpress in Timber core


ste93cry:
Hi,
Timber's core already supports the `get_terms` wrapper function of the Wordpress original function. It would be nice to have a `get_term` wrapper too, as now the only way I found to retrieve a single term object is to use `array_pop`.

connorjburton:
@jarednova @ste93cry - i was going to go ahead and start on this but need some clarification, should we be searching over the current posts terms, which is what I would assume, and if so should we be giving the options to search by term id, slug, name or term_taxonomy_id. Replicating `get_term_by`?

ste93cry:
I would expect the function to work as the WordPress [one](https://codex.wordpress.org/Function_Reference/get_term), but returning a TimberTerm object instead

jarednova:
Closed in #1049

PR #1049, opened 2016-06-02
Adding get term

connorjburton:
**Ticket**: #386 
**Reviewer**: @jarednova @ste93cry

#### Issue

There was no wrapper for `get_term`

#### Solution

Adds wrapper

#### Impact

None

#### Usage

`Timber::get_term(params)`

#### Considerations

This needs testing, I literally just added a wrapper, `get_terms` seems to do a lot more.

#### Testing

Needs a test

### DIFF
--- a/lib/TermGetter.php
+++ b/lib/TermGetter.php
@@ -7,6 +7,11 @@ use Timber\Helper;
 
 class TermGetter {
 
+	public static function get_term( $term, $taxonomy, $output = OBJECT, $filter = 'raw', $TermClass = 'Term' ) {
+		$term = get_term($term, $taxonomy, $output, $filter);
+		return new $TermClass($term->term_id, $term->taxonomy);
+	}
+
 	/**
 	 * @param string|array $args
 	 * @param array $maybe_args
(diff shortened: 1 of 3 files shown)


jarednova:
Hey @connorjburton: thanks for submitting! Here's my only thought on tweak....

While it makes sense to mirror WP's argument order (which in this case is... `($term, $taxonomy, $output, $filter)`) the last two (`$output` and `$filter`) seem not particularly useful — or at least _less_ useful than `$TermClass`; what are your thoughts on tweaking this to be: 

`$term, $taxonomy, $TermClass, $output, $filter` ?

connorjburton:
Gotcha, will resolve

ste93cry:
Does it makes sense to have `$output` an `$filter`? The first according to the WordPress [documentation](https://codex.wordpress.org/Function_Reference/get_term) should tweak the type of the result (which is not needed in our case because we're going anywhere to wrap it with `TimberTerm`) while the second should filter how the data is filtered (what's the purpose of it in our function?)

connorjburton:
Remove the un-necessary args

jarednova:
Thanks @connorjburton! Just added a quick test – we're good to go!

PR #1050, opened 2016-06-02
Keeping it DRY and switching to home_url()

connorjburton:
**Ticket**: #568
**Reviewer**: @jarednova 

#### Solution

In response to latest comments on #568, switching to `home_url` and keeping it dry by just moving it to the base `init` method


jarednova:
@connorjburton — I definitely like this approach but the problem I keep hitting with this is fucking WordPress. The problem you hit is when I want to get information on a site other than the one currently set as the "active" blog. And unless I'm a total homer, there's no way to get the `home_url` for a specific blog without switching to it first (unless you wanted to directly query the options table, which might actually not be the worst idea in the world).

That said, there might be a more efficient way to NRY based on where/how `switch_to_blog` is called (or doing something dirty like querying the table directly, which the WordPress gods seem to generally not recommend)

haroldangenent:
@jarednova @connorjburton I'm not sure if I should reply to this PR or use the issues. Hope I'm not cluttering this up, but I've got some input that might help.

First of all, I get your frustration about Multisite. I was running into exactly the same issues about a month ago. Multisite is a bit wonky when it comes to almost all functions, switching blogs and using global variables.

**`Timber\Site` object**
When looking at the code in `Timber\Site`, I realized a `Timber\Site` object should be able to get initialized for _any_ blog, by passing the ID. Which is a great effort. This indeed poses the problem of actually getting this information. For example, retrieving the Home URL:

- `get_home_url($blogId)` can be called, but -internally in WP- this just calls `switch_to_blog()` and `restore_current_blog()`
- `get_blog_details()` is already called in `Timber\Site` and contains the home URL, but also just calls `switch_to_blog()` to get it (on a similar note, there also is [`WP_Site::get_instance()`](https://core.trac.wordpress.org/ticket/32450) since 4.5.0 which might be less expensive but contains less information). `get_blog_details()` doesn't pass the URL through the `home_url` filter.
- `get_blog_option()` could retrieve home URL (used for `$site->description` right now) but also just calls `switch_to_blog()`

It might be more efficient to call `switch_to_blog()` once on initialization and then retrieve all information by using regular single-site functions.

**Getting information directly from the database**
This would be less expensive, but retrieving this information without using built-in WordPress functions would not apply any filters. This would lead to less compatibility with existing code. For example, it would be quite hard to create a Multisite network which sites run on separate domains if the Home URL was guessed based on information from `wp_blogs`.

Hope this helps.

connorjburton:
Hi guys, thanks for feedback. Not sure what code changes need to happen here, we do already `switch_to_blog` and you can init Timber\Site manually if you need to get details of a blog that isn't the current one. Am I missing something here? Don't have too much experience with multi-site

haroldangenent:
You can init `Timber\Site` manually, but the properties would not be correct. For example, in case of `$site = new Timber\Site(2)` and `2` being a subdirectory Multisite install (using `cjb-multisite-home-url` branch):

- `$site->url` contains the URL of the main install (blog ID: 1)
- `$site->language` contains the language of the main install
- `$site->language_attributes()` would output the language attributes of the main install

There might be other properties which return the wrong value, but the most critical one would be the URL. 

Since that is what this PR is about, the code changes for the would require `home_url()` to be called inside `Site::init_as_multisite()` between `switch_to_blog($info->blog_id)` and `switch_to_blog($old_id)`. This means it would need to be set in both `Site::init_as_multisite()` and `Site::init_as_singlesite()`.

**DRY solution**
A more dryer solution would be to have an `init()` method that initializes all the properties. On Multisite, it just calls init like this:

```
if (is_multisite()) {
  switch_to_blog($blogId);
  $this->init();
  restore_current_blog();
} else {
  $this->init();
}
```

But that would be an idealistic (and probably too simplistic) approach and changing it would probably have impact on backwards compat.

jarednova:
I'm not so sure @haroldangenent — I implemented your suggestion and it seems to do the trick. I'd love a second set of eyes before I go ahead and merge. Lemme know if this checks out for you!

PR #1051, opened 2016-06-02
836 preview page fix

dknoben:
**Ticket**: #836 

#### Issue
The "Preview Changes" button in a WordPress post on the admin side was grabbing the last or oldest revision, not the newest revision, meaning the preview changes button would not show your latest changes, but the oldest revision of the post or page.  

#### Solution
instead of grabbing the post id from last element of the $revisions array, I am grabbing it from the first element. 

…last revison from a list of revisons sorted in desc order, meaning it was always showing you the first revision ever made, and not the latest change

### DIFF
--- a/lib/Post.php
+++ b/lib/Post.php
@@ -256,8 +256,8 @@ class Post extends Core implements CoreInterface {
 		$revisions = wp_get_post_revisions($query->queried_object_id);
 
 		if ( !empty($revisions) ) {
-			$last = reset($revisions);
-			return $last->ID;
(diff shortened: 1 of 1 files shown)


jarednova:
Thanks @dknoben! Great catch on this guy. I added a test in #1052 where we can double-confirm this and merge it in!

jarednova:
Thanks @dknoben! Great catch on this guy. I added a test in #1052 where we can double-confirm this and merge it in!

connorjburton:
Thanks @dknoben !

PR #1055, opened 2016-06-08
Changing static call to regular call

connorjburton:
Can you please merge and release @jarednova - just a small bug with `1.0.4`

Merci!

### DIFF
--- a/lib/Timber.php
+++ b/lib/Timber.php
@@ -53,7 +53,7 @@ class Timber {
 			$this->test_compatibility();
 			$this->backwards_compatibility();
 			$this->init_constants();
-			$this->init();
+			$this::init();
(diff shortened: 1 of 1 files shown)


jarednova:
@connorjburton thanks for the fix! The failing tests are just for a problem with the gravatar server, we can ignore those. However, since `Timber::compile` is a static function, wouldn't the call to `init` also need to be static — `$this` wouldn't be available, right? Or am I missing something here?

connorjburton:
Oh yeah gotcha, will change that now

connorjburton:
How's that?

jarednova:
That'll do it! I'm just re-running the tests with the latest master and will merge!

PR #1060, opened 2016-06-09
Avoid test to break if no image processing extensions are installed

FlyingDR:
#### Issue
Currently tests run can cause fatal errors in a case if they're running on PHP with no image processing extensions (GD, Imagick).

#### Solution
This pull request marks affected tests as skipped in a case if no such extensions are loaded.

#### Impact
Raise of chance on success tests run in more environments.


connorjburton:
Thank you very much @FlyingDR!

PR #1061, opened 2016-06-09
Loader locations normalization

FlyingDR:
#### Issue
Currently `Timber\Loader` class collects location information from several methods and their checks and normalization is inconsistent. It results into duplicates into resulted set of paths in some cases, e.g. when using paths containing symlinks. Moreover prepared paths are re-normalized upon use ([twice](https://github.com/timber/timber/blob/master/lib/Loader.php#L215)) while it can be safely omitted if locations storage itself will be not accessible directly from outer code.

#### Solution
Provided patch unifies paths checks and normalization across information gathering methods and removes methods mean for internal use from being publicly accessed.

#### Impact
Minor performance gain and slightly better stability

### DIFF
--- a/lib/Loader.php
+++ b/lib/Loader.php
@@ -115,23 +115,18 @@ class Loader {
 	 */
 	private function get_locations_theme() {
 		$theme_locs = array();
-		$child_loc  = realpath( get_stylesheet_directory() );
-		$parent_loc = realpath( get_template_directory() );
-		if ( is_dir( $child_loc ) ) {
-			$theme_locs[] = $child_loc;
-			$child_loc    = trailingslashit( $child_loc );
-			foreach ( $this->get_locations_theme_dir() as $dirname ) {
-				$tloc = realpath( $child_loc . $dirname );
-				if ( is_dir( $tloc ) ) {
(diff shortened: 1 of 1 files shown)


FlyingDR:
Moreover, I'm not sure if [this](https://github.com/timber/timber/blob/master/lib/Loader.php#L223) piece of code is necessary. On Windows path `'/'` is obviously wrong, on Linux it is [removed by Twig itself](https://github.com/twigphp/Twig/blob/1.x/lib/Twig/Loader/Filesystem.php#L97). But I've kept it for now because original author may have some own reasons for it, I'm not aware of.

connorjburton:
Thanks for this, will review over the weekend

jarednova:
Yeah thanks @FlyingDR! this looks like a big boost in simplicity without any loss of functionality. I'll leave it to @connorjburton's review. But this looks (at a glance) awesome!

jarednova:
Thanks as always @FlyingDR! Made just a few tweaks for private vs. protected based on the [visibility rec from 10up](https://10up.github.io/Engineering-Best-Practices/php/#code-style)

PR #1068, opened 2016-06-16
ref #1067 -- added test for private properties

jarednova:
**Ticket**: #1067 

This should cause a failing test in PHP 7

PR #1069, opened 2016-06-16
Flying dr import error php7

jarednova:
**Ticket**: #1068 

This merges the issue and the test


jarednova:
This can close #1068 and #1067 — thanks again @FlyingDR !

Issue #1070, opened 2016-06-16
Legacy {{ post.type }} can cause unexpected behavior


jarednova:
### Expected behavior
If you had a custom field with `{{post.type}}` it should work as it did in 1.0.3

### Actual behavior
You get the data from the `Timber\PostType` object

As suggested in #1003 — this is a breaking change that should be rectified in the `1.0.*` version and resolved in `1.*`

jarednova:
Closed by #1071

Issue #1057, opened 2016-06-08
Cannot access private property Timber\Post::$_type 


jerome-quere:
Wordpress: 4.5.2
PHP: 5.6
Timber: 1.0.4

![image](https://cloud.githubusercontent.com/assets/526046/15896234/af6b59dc-2d8f-11e6-9166-a5d7d7da9d5a.png)

When Core::import is called with $info containing a field `_type` PHP throw an error because `Post::_type` is marked private since 0d218854642d243b8121843e78e51aba4dd645f4

jarednova:
Closed by #1071

PR #1071, opened 2016-06-16
ref #1070 -- ref #1057 -- resolve errors with {{ post.type }} fallout

jarednova:
**Ticket**: #1070 and #1057 
**Reviewer**: @connorjburton 

#### Issue
Some errors were caused by #1003. This should resolve them:
- If a post had a custom field named "type", you would get the `PostType` object instead of the value
- If a post had a custom field named "_type" all hell broke loose

#### Solution
Return a custom field value if set for `post.type`

#### Impact
Version `1.0.5` can restore expected functionality, we can plan for how to do things best in `1.1.x`

#### Usage
None

#### Considerations
We should consider if/how to disallow the names of certain custom fields

#### Testing
You betcha

### DIFF
--- a/lib/Post.php
+++ b/lib/Post.php
@@ -163,7 +163,7 @@ class Post extends Core implements CoreInterface {
 	/**
 	 * @var PostType $_type stores the PostType object for the Post
 	 */
-	private $_type;
+	protected $__type;
 
 	/**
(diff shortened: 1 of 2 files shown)


connorjburton:
Thanks!

PR #1075, opened 2016-06-20
Fixed issue with paginating over 10 pages of posts

lggorman:
**Reviewer**: @jarednova 

#### Issue
As pointed out by @davidallansson, the changes to pagination I made [here] (https://github.com/timber/timber/commit/9b9d11a1fd6b954e63badbdfc0d76110f07705e7) caused issues when there were a large number of posts and paginate_links was generating dots instead page links.

#### Solution
When the total number of pages exceeds the page links generated by paginate_links, loop through pages array which entry includes the current page, and calculate next and prev from there (thanks again to @davidallansson for the suggested solution).

#### Impact
N/A

#### Usage
N/A

#### Considerations
This fix should resolve the issue for now, but this is something we may want to revisit/refactor is we make broader changes to pagination.

#### Testing
Added a test for when there are enough posts for paginate links condense the pages array with dots


jarednova:
Merged, thanks @lggorman !

PR #1077, opened 2016-06-21
Add extra contextual parameters to render filters

johnbillion:
This change adds extra parameters to the `timber/loader/render_data` and `timber/output` filters so more context is available when using the filters.

### DIFF
--- a/lib/Loader.php
+++ b/lib/Loader.php
@@ -67,7 +67,7 @@ class Loader {
 				do_action('timber_loader_render_file', $result);
 			}
 			$data = apply_filters('timber_loader_render_data', $data);
-			$data = apply_filters('timber/loader/render_data', $data);
+			$data = apply_filters( 'timber/loader/render_data', $data, $file );
 			$output = $twig->render($file, $data);
 		}
 
@@ -75,7 +75,7 @@ class Loader {
 			$this->set_cache($key, $output, self::CACHEGROUP, $expires, $cache_mode);
 		}
 		$output = apply_filters('timber_output', $output);
-		return apply_filters('timber/output', $output);
+		return apply_filters( 'timber/output', $output, $file, $data );
 	}
 
 	/**


jarednova:
Thanks @johnbillion! I'll review this week

lib/Loader.php
```diff
@@ -67,15 +67,15 @@ public function render( $file, $data = null, $expires = false, $cache_mode = sel
 				do_action('timber_loader_render_file', $result);
 			}
 			$data = apply_filters('timber_loader_render_data', $data);
-			$data = apply_filters('timber/loader/render_data', $data);
+			$data = apply_filters( 'timber/loader/render_data', $data, $file );
```
jarednova: So here the arg order is `$data, $file`...

lib/Loader.php
```diff
@@ -67,15 +67,15 @@ public function render( $file, $data = null, $expires = false, $cache_mode = sel
...
 			$output = $twig->render($file, $data);
 		}
 
 		if ( false !== $output && false !== $expires && null !== $key ) {
 			$this->set_cache($key, $output, self::CACHEGROUP, $expires, $cache_mode);
 		}
 		$output = apply_filters('timber_output', $output);
-		return apply_filters('timber/output', $output);
+		return apply_filters( 'timber/output', $output, $file, $data );
```
jarednova: ... but here it's `$file, $data` — is there a reason for the diff?

jarednova:
@johnbillion I added a quick question in [line notes](https://github.com/timber/timber/pull/1077/files), lemme know what you think!

jarednova:
@johnbillion — thanks again for submitting this, I added #1091 with some tests and tweaks (re-ordering `$file` and `$data` so the order is consistent). Lemme know if you have any thoughts on this, otherwise I'll plan to merge in sometime tomorrow

Issue #1073, opened 2016-06-20
User password hash from DB exposed in [user] object


streetyingyang:
### Expected behavior
Wasn't expecting to see my hashed password in an object available publicly.

### Actual behavior
In  [user] => Timber\User Object, the user_pass for my user is available.

### Steps to reproduce behavior
Login, display user_pass in twig template. Available for entire session.

{{ user.user_pass }}

### What version of WordPress, PHP and Timber are you using?
PHP 5.5.35 - MySQL 5.5.5 - WP 4.5.2 - Timber 1.0.3

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Composer

jarednova:
Good point @streetyingyang — I see no reason why the user_pass should be returned by default. Any dissenting opinions?

jarednova:
@streetyingyang — specifically are you concerned about someone in a template being able to retrieve via `{{ user.user_pass }}`? or is this more about someone doing a `print_r` on that object (for example, the [Timber Debug Bar](https://wordpress.org/plugins/debug-bar-timber/) would make that info available right now)

streetyingyang:
Tough question. It's obviously not something somebody would want to use in a view. In some rare circumstances of someone being ignorant or having nefarious intentions, a list of users in a twig view could possibly expose user_pass.

I created a new user object via $context['user_test'] = new TimberUser(2); and got the same output my user has during my session (user_pass being printed). It's a stretch, but it might be attractive to crack a list of WP user hashes for particular sites through just an edited template.

I guess I'm not so concerned about WP Debug Bar showing my hash as an intern or human turd listing user hashes (sorted by username and email addresses as well) for fun.

jarednova:
@streetyingyang thanks for highlighting this and your info. @lggorman made a PR that will be a part of the next release (#1084)

PR #1084, opened 2016-06-24
Set user password to be hidden by default

lggorman:
**Ticket**: #1073 
**Reviewer**: @jarednova 

#### Issue
As described in #1073, the user's hashed password exposed is in the TimberUser object.

#### Solution
Added an `include_pass` flag to the constructor and init functions, which is set to false by default.  Unless the flag is set to true, the user password will not be exposed in the user object.   

#### Impact
Would only affect users that are using that hashed password somewhere, which seems like an edge case.

#### Usage
The TimberUser constructor now accepts an optional second param, the `include_pass` flag.  

#### Considerations
-

#### Testing
Tests are included!

### DIFF
--- a/lib/User.php
+++ b/lib/User.php
@@ -67,9 +67,10 @@ class User extends Core implements CoreInterface {
 
 	/**
 	 * @param int|bool $uid
+	 * @param bool $include_pass Whether to include the user's hashed password
 	 */
-	public function __construct( $uid = false ) {
-		$this->init($uid);
+	public function __construct( $uid = false, $include_pass = false ) {
+		$this->init($uid, $include_pass);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


lggorman:
@jarednova Should I revise this? Maybe it would be better to just make it protected instead of unsetting it entirely.

jarednova:
@lggorman — can you look at some of the comments in #1073? My inclination is to unset with no ability to access the user's PW. The two biggest potential vulnerabilities I see are...

1) A `print_r` either in a template or using the Timber Debug Bar
2) A user having access to a template editor which would allow them to do something like `{{ TimberUser(4).user_pass }}` and have access to any user's (hashed) PW

I can't think of a use case where you'd ever want to display or access this data. I see a far greater risk in both exposing this and giving the appearance that Timber is less than secure. 

Your revision could be as simple as removing the `$include_pass` flag functionality

lggorman:
@jarednova Makes sense to me, got rid of the password flag

jarednova:
Awesome, thanks @lggorman !

Issue #1085, opened 2016-06-24
Dates rendering incorrectly


thedamon:
### Expected behavior
`{{ '1457395200' | date('Y-m-d') }} => 2016-03-08`
`{{ 1457395200 | date('Y-m-d') }} => 2016-03-08`
`{{ function('date_i18n', 'Y-m-d', 1457395200) }} => 2016-03-08`
`{{ function('date', 'Y-m-d', 1457395200) }} => 2016-03-08`

### Actual behavior
`{{ '1457395200' | date('Y-m-d') }} => 5200-06-24`
`{{ 1457395200 | date('Y-m-d') }} => 5200-06-24`
`{{ function('date_i18n', 'Y-m-d', 1457395200) }} => 2016-03-08`
`{{ function('date', 'Y-m-d', 1457395200) }} => 2016-03-08`

### Steps to reproduce behavior
Simply using the function. Most dates work fine, certain ones are rendering unexpectedly (with the last 4 digits coming out as the year and the day/month appearing to be today's month and day)
If I add ` | number_format(0, '.', '')` before `| date`, different dates will be messed up.

### What version of WordPress, PHP and Timber are you using?
WordPress 4.5.1, PHP 5.6, Timber 1.0.5 -->

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Downloaded from WP plugins page

jarednova:
Closed by #1087

PR #1086, opened 2016-06-26
remove strtotime check from dates

thedamon:
**Ticket**: https://github.com/timber/timber/issues/1085

#### Issue
Bug in the date filter where some timestamps were being dubiously evaluated with `strtotime()` instead of their actual value 

#### Solution
Remove the check for `strtotime($date) === false` when deciding to use intval()

#### Impact
I believe the impact is only positive. I'm trying to think what purpose that check had in the first place... couldn't find anything in the commit history. I'm worried perhaps this was covering a different edge case, but I'm confident the current behaviour is not idea

#### Testing
Testing the date filter with '1457395200' could snag this.

some timestamps can be interpreted as differing dates by strtotime, such as strtotime('1457395200') => 101944277859. 
So strings like this should be treated with `intval()`.
I cannot think of an example where a date that would satisfy `is_numeric()` should be translated with `strtotime()`

### DIFF
--- a/lib/Twig.php
+++ b/lib/Twig.php
@@ -271,7 +271,7 @@ class Twig {
 
 		if ( $date instanceof \DateTime ) {
 			$timestamp = $date->getTimestamp() + $date->getOffset();
-		} else if ( is_numeric($date) && strtotime($date) === false ) {
+		} else if ( is_numeric($date) ) {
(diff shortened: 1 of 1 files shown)


thedamon:
Going to do this better, with a test

PR #1087, opened 2016-06-26
Fix issue where timestamps could be erroneously converted by strtotime

thedamon:
**Ticket**: #1085 
Reviewer: @jarednova

#### Issue
Bug in the date filter where some timestamps were being dubiously evaluated with `strtotime()` instead of their actual value because some timestamps can be interpreted as differing dates by strtotime, such as strtotime('1457395200') => 101944277859.

#### Solution
Only use `strtotime` on numeric values that are 8 characters in length (assuming they will be in format YYYYMMDD). 

#### Impact
Timestamps that are 8 chars in length will suffer, but this is preferable to previous behaviour. Most dates in WordPress will probably not be timestamps

#### Testing
Added a test with a timestamp that `strtotime` will interpret as a valid, different date.

### DIFF
--- a/lib/Twig.php
+++ b/lib/Twig.php
@@ -271,7 +271,7 @@ class Twig {
 
 		if ( $date instanceof \DateTime ) {
 			$timestamp = $date->getTimestamp() + $date->getOffset();
-		} else if ( is_numeric($date) && strtotime($date) === false ) {
+		} else if ( is_numeric($date) && (strtotime($date) === false || strlen($date) !== 8) ) {
 			$timestamp = intval($date);
 		} else {
 			$timestamp = strtotime($date);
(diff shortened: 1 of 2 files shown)


jarednova:
@thedamon as I understand the risk of this is that if someone had an 8-digit timestamp from the early 70s (It seems between 4/26/1970 - 3/3/1973) it will result in an error. Yes? Or anything else I'm missing?

thedamon:
@jarednova yes.. though it's not a new problem.

Not sure any way to know for sure whether 19970402 should be treated as YYYYMMDD or a timestamp without further information. the current code would treat it as YYYYMMDD as well.

jarednova:
Sounds good — I'll review over the next few days and merge in!

jarednova:
Thanks @thedamon — this will go out in the next releasse!

Issue #405, opened 2014-12-11
Images URLs issues with WPML


RealBigB:
WPML overloads the standard wp `home_url` function to add the language either as a path element (ie: `http://www.mydomain.tld/` becomes `http://www.mydomain.tld/fr/`) or as a query string (`http://www.mydomain.tld/` becomes `http://www.mydomain.tld/?lang=fr`). This breaks a whole lot of things in timber-image-helper.php, and potentially some more in at least timber-image.php and timber-url-helper.php. One of the breakage might not be obvious: it makes the `resize` filter re-sideload and re-resize images that are local and already resized. 

I replaced every call to `home_url()` by `site_url()` in my local instal (in the three aforementioned files) and it seems to solve the problem but I can't garantee it doesn't break something else...

jarednova:
Closed by #502

jedrzejchalubek:
It seems that there is still something wrong. After adding thumbnail to the translated version of post it generates new image file with md5 as name (instead of picking already added).

`twig custompost.thumbnail.src|resize(x,y)` returns relative path `/wp-content/uploads/2015/03/7ab1cdd0648fd5faefb0aa4813d7a583.jpg`

When viewing post in main language, everything works perfectly.

RealBigB:
You have to replace _all_ calls to `home_url()` with calls to `site_url()` to solve the issue, cf my first post.

jedrzejchalubek:
This is not a good way to go. For now, I'll try to get around without timber (get_the_post_thumbnail or something).

laurentvd:
I'm also experiencing this issue when using the latest WPML and latest Timber. Additionally, I'm running WordPress from a subdir (`/wp`). The default language works just fine, but the alternative language fails. It generates url's like `http://domain.ltdhttp://domain.ltd/sv/path/to/image`. 

So `home_url()` returns with the suffix of the locale and `site_url()` returns with the `/wp` suffix. Not really sure how this can be resolved for everyone, but I had to solve it using a hard-coded update to `timber-image-helper.php`.

jarednova:
@laurentvd can you post a snippet of the twig code that you're using? I was able to setup PHPUnit with multisite, so I'd like to give it a shot with what you've got

laurentvd:
Sure! 
```
<img class="photo" src="{{ photo.src | resize(428, 428) }}" 
    width="214" height="214" alt="{{ photo.alt }}"/>
```
[...]

PR #1088, opened 2016-06-27
405 wpml image url fix

TuureKaunisto:
#### Issue
WPML language identifier breaks the ImageHelper
https://github.com/timber/timber/issues/405

#### Solution
Use site_url() instead of home_url()

#### Impact
This might break some things that the current tests don't cover.

#### Testing
The test mocks the filter added by WPML since WPML is not a free plugin one could easily install using composer. Without the home_url() -> site_url() change the test fails.

### DIFF
--- a/lib/ImageHelper.php
+++ b/lib/ImageHelper.php
@@ -187,7 +187,7 @@ class ImageHelper {
 	 */
 	protected static function add_constants() {
 		if ( !defined('WP_CONTENT_SUBDIR') ) {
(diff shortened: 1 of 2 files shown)


jarednova:
Thanks @TuureKaunisto! Getting good with WPML and Multisite always feels like a tough road. This gets us closer!

jarednova:
Merged! Thanks again @TuureKaunisto —this will go out in the next release

Issue #1104, opened 2016-07-06
Undefined variable: current


nikola3244:
### Expected behavior

No notices.
### Actual behavior

Couple notices of undefined variables.
```
Notice: Undefined variable: current in /home/nikola/public_html/wphelium/wp-content/plugins/timber-library/lib/Timber.php on line 464
Notice: Undefined variable: current in /home/nikola/public_html/wphelium/wp-content/plugins/timber-library/lib/Timber.php on line 467
Notice: Undefined variable: current in /home/nikola/public_html/wphelium/wp-content/plugins/timber-library/lib/Timber.php on line 464
Notice: Undefined variable: current in /home/nikola/public_html/wphelium/wp-content/plugins/timber-library/lib/Timber.php on line 467
```
### Steps to reproduce behavior

I got this error while trying out "Helium" template from RocketTheme.
[Can be downloaded here](http://gantry.org/downloads#ci-builds).

### What version of WordPress, PHP and Timber are you using?

Wordpress 4.5.3, Apache/2.4.20 (Unix) OpenSSL/1.0.2h PHP/5.6.23, Timber 1.1.0 

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)

Fresh install from WordPress, "Add new plugin".

jarednova:
Closed by #1105

PR #1105, opened 2016-07-06
Fix for #1104

nikola3244:
**Ticket**: # 
**Reviewer**: @ 

#### Issue

#1104 

#### Solution

If `$current` is still not set on that line, exit from function.
Makes sure that PHP Notice is not shown.

#### Impact

Impact: none/small.
Backwards compatibility: 100%
Performance impact: none.

#### Usage

No usage changes.

#### Considerations

/

#### Testing

Scenario when current code fails is in #1104.


jarednova:
@nicolascava — thanks for the PR! I haven't run into this issue in my projects so my Q is what is the ideal behavior to see when `$current` is not available. Is it...

A) No pagination controls are returned (what I'm pretty sure your PR does)
B) A stripped-down set of pagination controls are returned (number nav?)

I would think we'd want to give users _something_ if possible. So my question: is it possible to return the data necessary for B without `$current` being available?

nikola3244:
Something like this?

I tested it for my problem, and it still works for me.
No notices and this is the actual output from the function after second fix:
```
array(4) { 
    ["current"]=> int(1)
    ["total"]=> float(1)
    ["pages"]=> array(0) {}
    ["prev"]=> string(0) ""
 }
```

With the first fix output was just null.

jarednova:
Lookin' good @nikola3244, merged!

PR #1114, opened 2016-07-12
Added functionality to get direct children of a term excluding grandc…

justono1:
#### Issue

Trying to loop through each level of a taxonomy but with existing functionality of get_children, after the first level the function would return all children (get_term_children) even the grandchildren. I needed to apply different styles and functionality to each level of the taxonomy.
#### Solution

Added a function named get_direct_children to return only terms directly related to the term parent. Essentially return terms one level deep.
#### Testing

Only tested within current project.

…hildren, useful for working with mutli leveled taxonomies.

### DIFF
--- a/lib/Term.php
+++ b/lib/Term.php
@@ -49,6 +49,7 @@ class Term extends Core implements CoreInterface {
 	public static $representation = 'term';
 
 	public $_children;
+	public $_directChildren;
 	/**
 	 * @api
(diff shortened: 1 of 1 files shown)


jarednova:
Thanks @justono1! I just have two Qs:

1) is `direct_children

justono1:
@jarednova I'll look into it and do a little more research on accepted terminology but honestly I think having it be an argument on the `get_children` method would be the way to go. Also I'll try and get a unit test written for it.

PR #1121, opened 2016-07-15
add escaping for esc_url and wp_kses_post

matgargano:
**Reviewer**: @jarednova

#### Issue
There was no escaping for `esc_url` or `wp_kses_post` natively in Twig (or Timber)

#### Solution
I added escapers as described in the [Twig Docs](http://twig.sensiolabs.org/doc/filters/escape.html#custom-escapers)

I'm not sure if `wp_kses_post` should be an escaper or if it should be it's own filter a la `wpautop` what are your thoughts?

I added the code in `Timber\Twig` and it acts as an analog to the filters, seems like the logical place for this to live.

I also added an extra filter `timber/twig/escapers` which is pretty much an analog to `timber/twig/filters`

#### Impact
Should be minimal impact.

#### Usage
```twig
<a href="{{ post.get_field('custom_link')|e('esc_url') }}">My Link</a>
```

```twig
<p>{{ post.get_field('wysiwyg_field')|e('wp_kses_post') }}</p>
```

#### Considerations
This is pretty straightforward, we may want to move wp_kses_post into a filter - that's all.

#### Testing
included unit tests -- can make them more robust if needed

### DIFF
--- a/lib/Loader.php
+++ b/lib/Loader.php
@@ -246,6 +246,7 @@ class Loader {
 
 		$twig = apply_filters('twig_apply_filters', $twig);
 		$twig = apply_filters('timber/twig/filters', $twig);
+		$twig = apply_filters('timber/twig/escapers', $twig);
 		$twig = apply_filters('timber/loader/twig', $twig);
 		return $twig;
 	}
--- a/lib/Twig.php
+++ b/lib/Twig.php
@@ -26,6 +26,7 @@ class Twig {
 	 */
 	public function __construct() {
 		add_action('timber/twig/filters', array($this, 'add_timber_filters'));
+		add_action('timber/twig/escapers', array($this, 'add_timber_escapers'));
 	}
 
 	/**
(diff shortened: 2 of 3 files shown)


jarednova:
@matgargano — in regards to making `wp_kses_post` an escape option vs. its own filter, my thinking is to _reduce_ the number of filters (or keep them as small as possible). I'll be honest, I don't really use this escaping method (maybe I should be!!!) — is it "cleaner" to have...

```twig
<p>{{ post.get_field('wysiwyg_field')|e('kses') }}</p>
```

... or does this just add confusion?

matgargano:
@jarednova I agree that the escaping is the better route to go, I use wp_kses_post on any piece of content that the user will/may enter in HTML, it strips out "evil scripts" ... 9 times out of 10 end users are fine with what it strips out.

i would use ``wp_kses_post`` in favor of `|e('kses')` because that strips out (well, allows, rather) specific elements/attributes that WP natively strips out for posts. KSES is more a generic term and `wp_kses_post` is a specific implementation of KSES.

jarednova:
Awesome, I didn't know there was a diff between `kses` and WP's implementation. Thanks again for the PR and tests, let's merge!

Issue #819, opened 2016-01-13
Timber::get_post() behaviour differs from WP get_post()


ondratra:
Very unfortunate inconsistency in get_post function while working with custom post types. 

assuming we have two posts:
- one post with ID 1234 of regular post type 'post'
- one post with ID 6789 and post_type 'myCustomPostType'
then 

get_post(1234); // returns WP post object
Timber::get_post(1234, 'MyPostClass'); // returns MyPostClass instance
get_post('6789'); // returns WP post object
Timber::get_post(6789); // returns NULL!!!

It's caused by Timber::get_post() internally calling Timber::get_posts() that looks for 'post_type' query parameter; WP get_post() has it's own logic separate from get_posts()

The ugly solution to this situation is to rewrite clause:
Timber::get_post($postId, 'MyCustomClass');
into
$tmpPost = get_post($postId); // get actual post
$post = $tmpPost ? Timber::get_post($tmpPost, 'MyCustomClass') : null; // create desired class instance

Is there a way to fix this easily? (if there is not we should probably update documentation and mention this problem)

chrisgherbert:
I'm not getting the same behavior.  If I run:

```
Timber::get_post(7023)
```

I get the correct TimberPost object, even if the post is a custom post type.

connorjburton:
Hi @ondratra is this still an issue? `Timber::get_post` has always worked for me for custom post types, but maybe it's to do with the fact you want `MyCustomClass`?

connorjburton:
I'm going to close this as it seems inactive, let me know if it's still an issue and I will re-open.

kmturley:
I'm also getting this issue. When I log out the following in page.php:

    print_r(get_posts());

    Array
    (
        [0] => WP_Post Object
            (
                [ID] => 20
                [post_content] => sadbasjdkhsa jhdaskjh 
                [post_title] => This is a post
                [post_type] => post
            )

        [1] => WP_Post Object
            (
                [ID] => 1
                [post_content] => Welcome to WordPress. This is your first post. Edit or delete it, then start writing!
                [post_title] => Hello world!
                [post_type] => post
            )
    )

But when I use the Timber version it's outputting the current page instead:

    print_r(Timber::get_posts());

    Array
    (
        [0] => TimberPost Object
            (
                [ID] => 26
                [post_content] => hello ya
                [post_title] => This is my homepage yup
                [post_type] => page
            )
    )

jscontrust:
[...]

PR #1123, opened 2016-07-19
Edited get_post method to get posts directly by id

lggorman:
**Ticket**: #819
**Reviewer**: @jarednova 

#### Issue
As described in the ticket, Timber::get_post using an ID may not return anything when a post with a custom type is requested, in contrast to the WP `get_post` function.

#### Solution
Instead of using the `get_posts`  method, which will only get post types with `exclude from search` set to true, just get posts directly using `new $PostClass()`.  The new post constructor uses Wordpress' `get_post`, which is pretty much a direct database hit to get a post based on ID and circumvents a lot of the logic applied to `get_posts`.  There's also a conditional in there to handle post previews.

#### Impact
- 

#### Usage
nope!

#### Considerations
This use of the `maybe_get_preview` function here is a little funky -- it's intended to accept and return an array, which makes sense in the context of a `PostsCollection` but isn't super clean here.  We could potentially set the method to accept either an object or an array, and in that instance I think it might make sense to move it to the `Helper` or `Timber` class.  Let me know what you think.

#### Testing
Tests included!

### DIFF
--- a/lib/PostGetter.php
+++ b/lib/PostGetter.php
@@ -13,6 +13,16 @@ class PostGetter {
 	 * @return array|bool|null
 	 */
 	public static function get_post( $query = false, $PostClass = '\Timber\Post' ) {
+		// if a post id is passed, grab the post directly
+		if ( is_numeric($query) ) {
+			$post = new $PostClass($query);
+			// get the latest revision if we're dealing with a preview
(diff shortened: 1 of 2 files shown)


jarednova:
Merged! Thanks @lggorman

PR #1135, opened 2016-08-02
add additional escapers

matgargano:
This is just adding more escapers that are native to WP, the `|e` escaper is not the same as WordPress' `esc_html` as I had thought and while I was adding that I added in the native WordPress `esc_js` as well to the escapers.

Added tests as well


jarednova:
Thanks @matgargano! Really appreciate your focus on this escaping stuff so that we can get that 100%. Could you also help out with some docs so that we can fully explain what these offer? It can be as simple as a markdown doc here: https://github.com/timber/timber/tree/master/docs/wiki. The files in that directory automatically compile to the docs here: http://timber.github.io/timber/

matgargano:
@jarednova https://github.com/timber/timber/pull/1137 :)

PR #1137, opened 2016-08-03
add escapers documentation

matgargano:
this is the documentation for my escapers


jarednova:
Score! Thanks — I'll deploy this into the docs right now

PR #1142, opened 2016-08-08
Possibility to change "Read More..." link using filters

LiljebergXYZ:
#### Issue
Unable to edit the "Read more" link in full, e.g. to add extra elements (div, li, etc)

#### Solution
Using the wordpress filters it's possible to create a filter that we can then use in our functions.php to customize the link in full

#### Usage
```
function read_more_link( $link ){
	$post = get_post();
	$link = '<div><a class="read-more" href="'. get_permalink($post->ID) . '">Read more...</a></div>';
	return $link;
}
add_filter( 'timber/post/get_preview/read_more_link', array($this, 'read_more_link') );
```

### DIFF
--- a/lib/Post.php
+++ b/lib/Post.php
@@ -389,6 +389,7 @@ class Post extends Core implements CoreInterface {
 	 */
 	public function get_preview( $len = 50, $force = false, $readmore = 'Read More', $strip = true, $end = '&hellip;' ) {
 		$text = '';
+		$link = '';
 		$trimmed = false;
 		$last_p_tag = null;
 		if ( isset($this->post_excerpt) && strlen($this->post_excerpt) ) {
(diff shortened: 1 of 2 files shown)


jarednova:
@LiljebergXYZ I'll still review this for inclusion, but I recommend this approach instead using the new Preview object...

```twig
{{ post.preview.read_more("Keep Reading....") }}
```

LiljebergXYZ:
Does the new Preview object accept different type of HTML formatting as well, so it's possible to fully customize?

jarednova:
@LiljebergXYZ — you are correct; the actual markup is not accessible. I'll review this today and try to get it merged in time for 1.1.3

jarednova:
Just needed one little tweak, thanks @LiljebergXYZ !

LiljebergXYZ:
Ah yes, the test was wrote in quite a hurry.
Glad to hear it was accepted, finding good use for it myself. Glad I could contribute!

Issue #1149, opened 2016-08-12
Timber 1.1.2: get_post cannot handle post classmap array


vilpersson:
In Timber 1.1.2 you can't use classmap array when retrieving single post:
```
	Timber\Timber::get_post(1, array(
		'portfolio' => 'MyPortfolioClass',
		'alert'     => 'MyAlertClass'
	));
```
This gives you an error:
```
Fatal error: Class name must be a valid object or a string in /srv/www/xxxx.com/current/web/app/plugins/timber-library/lib/PostGetter.php on line 18
```

Mapping post types to classes worked previously also with `get_post` method. New behavior was introduced in this pull request: https://github.com/timber/timber/pull/1123

Classmapping is a superb feature, I can create appropriate classes for every post type. Now I have to know the post type of a post before querying it.

By the way, it would be very cool if I could map all classes with filters instead of overwriting class variables and using method parameters.

jarednova:
closed by #1155

PR #1155, opened 2016-08-16
Allow usage of classmap in get_post

vilpersson:
**Ticket**: #1149

#### Issue
In Timber 1.1.2 you can't use classmap array when retrieving single post.

#### Solution
Get post_type and figure out the class to be used before instantiating it. This is handled by static method `get_post_class` which is also used in `PostCollection`. Added possibility to set classmap by filter `Timber\PostClassMap`.

#### Impact
Backward compatible.

#### Usage
`get_post` usage is now the same than prior 1.1.2 version.

#### Considerations
There is need to investigate if `get_post_class` method could be used for example in method `get_posts()` of `\Timber\Term\`

#### Testing
Tests included and passed

### DIFF
--- a/lib/PostGetter.php
+++ b/lib/PostGetter.php
@@ -9,12 +9,14 @@ class PostGetter {
 
 	/**
 	 * @param mixed $query
-	 * @param string $PostClass
+	 * @param string|array $PostClass
 	 * @return array|bool|null
 	 */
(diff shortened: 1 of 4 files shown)


jarednova:
@vilpersson — thank you so much for the issue and fix! (and the tests! oh the tests!). This will be released later today in 1.1.3

PR #1166, opened 2016-08-22
Add Countable interface to QueryIterator

aaemnnosttv:
**Ticket**: # 
**Reviewer**: @ 

#### Issue

The special Twig `loop` variable [requires the `Countable` interface to be implemented](http://twig.sensiolabs.org/doc/tags/for.html#the-loop-variable) when iterating over an object for some loop properties to work as expected (read: exist).

`Timber\QueryIterator` does not implement nor inherit this interface, so when iterating over it in Twig, some properties of the `loop` variable are not there.

#### Solution

The solution is to simply implement the interface on the `QueryIterator` class.  It requires a single `count` method.  The existing `post_count` method remains and is used by the new `count` method.

#### Impact

I have no reason to believe that this would be a backwards-incompatible change.  If anything, it improves the experience of using Timber by filling the gap in expected behavior.

#### Usage

Nope.

#### Considerations

n/a

#### Testing

Batteries included.

### DIFF
--- a/lib/QueryIterator.php
+++ b/lib/QueryIterator.php
@@ -10,7 +10,7 @@ if ( !defined('ABSPATH') ) {
 	exit;
 }
 
-class QueryIterator implements \Iterator {
+class QueryIterator implements \Iterator, \Countable {
 
 	/**
 	 *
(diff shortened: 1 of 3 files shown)


jarednova:
@aaemnnosttv thanks for your work on this. I ran into a related issue to this the other day and had no idea that `Countable` existed to solve this. Bravo on the code and tests!

Issue #1160, opened 2016-08-17
Shiny Updates Visual Bug


Daronspence:
### Expected behavior

The info should be in one line. I expect this is due to how 4.6 parses the info for shiny updates

### Actual behavior

![screen shot 2016-08-17 at 3 23 24 pm](https://cloud.githubusercontent.com/assets/4062087/17751890/dacf66fa-648e-11e6-8463-d15f49c5f690.png)

### Steps to reproduce behavior

Upgrade to WP 4.6, then rollback the timber version number manually to force an update to show.

### What version of WordPress, PHP and Timber are you using?

WordPress 4.6

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)

Plugin

jarednova:
Closed by #1167

PR #1167, opened 2016-08-23
Replace paragraph tags with line breaks in plugin update row

Daronspence:
**Ticket**: #1160

#### Issue
See #1160

#### Solution
Replacing the paragraphs with line breaks bring the update information in line with 4.6 changes to shiny updates

#### Impact
The impact should be minimal. The double line break functions similarly as a paragraph. It's not as "semantic" as before but that's debatable since now the entire update message is wrapped in a single `<p>` tag automatically as of 4.6. Wrapping the different sections in a `<div>` could be ideal since they are block level, but `<div>`s are hoisted out of `<p>` tags in most browsers.

Here is what the proposed change looks like visually in 4.6

screen shot 2016-08-22 at 11 12 41 pm

Here is what is looks like with the proposed changes on 4.5.3 (and below presumably)

screen shot 2016-08-22 at 11 25 09 pm

### DIFF
--- a/lib/Admin.php
+++ b/lib/Admin.php
@@ -48,18 +48,18 @@ class Admin {
 
 		if ( version_compare("1.0.0", $plugin_data['new_version']) <= 0 ) {
 			//a version of 1.0.0 or greater is availalbe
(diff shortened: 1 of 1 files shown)


jarednova:
Thanks so much for the fix @Daronspence!

PR #1169, opened 2016-08-23
Added get_scheme URL helper

natenolting:
#### Issue
The url helper was missing the ability to tell what HTTP scheme it was in. 

#### Solution
This fixes that by adding URLHelper::get_scheme() to check if HTTPS is on or off.

#### Impact
This removes the hardcoded HTTP scheme in Timber::get_context()

#### Usage
No special usage

#### Considerations
There's the possibility of adding a check to URLHelper::get_scheme() to see what port is being used, but it's not guaranteed that port 443 will be encrypted.

#### Testing
Test for this method are included.

### DIFF
--- a/lib/Timber.php
+++ b/lib/Timber.php
@@ -225,7 +225,7 @@ class Timber {
 	 */
 	public static function get_context() {
 		if ( empty(self::$context_cache) ) {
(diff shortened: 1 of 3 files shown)


jarednova:
Clean code! Tests! I love it all! Merged!

Issue #1118, opened 2016-07-15
Create new Pagination API


jarednova:
Relates to #816 

Create new pagination object that directly relates to a `PostsCollection` so that pagination more naturally relates to the collection of posts the user is handling

RadoslavGeorgiev:
@jarednova why are you associating the pagination with a `PostsCollection`?

I think it would make more sense to associate it with `QueryIterator` and to use the native WordPress pagination.

This way, if you also include a `$query` element in `Timber::get_context()`, which wraps the main `$wp_query`, even in the default templates, like archives and posts page, you would be able to simply use `query.pagination` without any additional setup.

lggorman:
@RadoslavGeorgiev I like pagination associated with the PostsCollection largely because it seems intuitive from an API perspective, e.g. `{% for page in posts.pagination.pages %}`.  But I can also see the logic in using the `QueryIterator`.  I'm not quite clear -- what would that look like on the php side, in terms of setting up the context?

PR #1173, opened 2016-08-26
1118 pagination api

lggorman:
**Ticket**: #1118 
**Reviewer**: @jarednova 
#### Issue

Pagination is generated based on the default query, requiring an extra `query_posts` to make pagination accurate for a custom query.
#### Solution

Set pagination as a method on the PostsCollection object .  The pagination object associated with the collection will use the same query params used to generate the collection, negating the need to tweak the default query (in order to make this work I've set the PostsCollection to keep a private reference to the wp_query object used to create it).  I've also added a `Timber::get_posts_collection` method to allow for a cleaner way to return a PostsCollection object instead of an array.  
#### Impact

Should have no effect on existing methods.
#### Usage

``` php
 $context['posts'] = new Timber\PostCollection('post_type=recipe');
```

``` twig
{% for post in posts %}
    <h1>{{ post.title }}</h1>
    <p>{{ post.preview }}</p>
{% endfor %}
<ul class="pages">
    {% for page in posts.pagination.pages %}
        <li>
            {% if page.link %}
                <a href="{{page.link}}" class="{{page.class}}">{{page.title}}</a>
            {% else %}
                <span class="{{page.class}}">{{page.title}}</span>
            {% endif %}
        </li>
    {% endfor %}
</ul>
```
#### Considerations

I set the new pagination up to use object attributes rather than an associative array, which seemed to make sense given that we are using a class.   To support the old `get_pagination` method I'm basically casting the object into an associative array using `get_object_vars`.  This approach seems to be passing tests fine, but let me know if you see any risks.
#### Testing

Tests included!


lggorman:
@jarednova Made the following changes:
- Turned former `PostsCollection` class into `PostsArray` and created new PostsCollection class that can be instantiated with a query, ie. `new PostCollection(array('post_type' => 'book'))`
- The new `PostsCollection` class contains a reference to the `QueryIterator`, used to get posts and generate pagination.

jarednova:
@lggorman we're so so close. Scrutinizer picked-up this potential bug that I want to take a look a w you on Monday....

https://scrutinizer-ci.com/g/timber/timber/inspections/eb00d952-9e07-4eb9-8b17-a7588a3ea97e/issues/files/lib/PostCollection.php?status=new&selectedLabels%5B0%5D=9&orderField=path&order=asc&fileId=lib%2FPostCollection.php&honorSelectedPaths=0

jarednova:
Hi @Timber folks! @lggorman and I would love any feedback on this PR. It's a pretty major (improvement!) that will link pagination to posts that resolve a lot of the confusion with re-running `query_posts`, etc. that occur now. Here's the basic usage....

#### Usage
```php
 $context['posts'] = new Timber\PostCollection('post_type=recipe');
```
```twig
{% for post in posts %}
    <h1>{{ post.title }}</h1>
    <p>{{ post.preview }}</p>
{% endfor %}
<ul class="pages">
	{% for page in posts.pagination.pages %}
		<li>
			{% if page.link %}
				<a href="{{page.link}}" class="{{page.class}}">{{page.title}}</a>
			{% else %}
				<span class="{{page.class}}">{{page.title}}</span>
			{% endif %}
		</li>
	{% endfor %}
</ul>
```

As seen in the PHP sample above, the new preferred way to get posts (outside the normal query) would be to call a `new PostCollection` object. The object is a PHP `ArrayObject` that (as the same suggests) contains the benefits of both data types. This allows both the easy iteration of an array, and methods to retrieve the pagination for that Collection (and future things).

### Considerations
By default, the `Timber::get_context()` array's `post` attribute will return a `Timber\PostCollection` (it presently returns a `Timber\QueryIterator`). Despite this change, all unit tests pass

### Questions
#### The name
The "collection" nomenclature was inspired by Laravel and Backbone. Does it make sense in this context? Another one I've considered is `Timber\PostQuery`

#### Default?
Does it make sense to replace the `posts` attribute from the Timber::get_context (as noted above)? The 👿  on my shoulder says "do it, it's better!" but the 👼  says "don't make it the default quite yet until it's thoroughly tested by the community!" I think I know which I should listen to but I need your help!

#### Anything else?
This has been a big lift and I'm really proud of the work @lggorman has done to bring it this far. I'm interested in opinions on whether there are opportunities we've missed with this approach or tweaks we could make to help make things both as flexible as possible, and as backwards compatible as possible

Thanks! 🌲

PR #1174, opened 2016-08-26
Add possibility to insert custom Loaders to the environment

tnottu:
#### Issue
I want to add custom Loaders to the Twig environment. I am trying to use [Fractal](https://github.com/frctl/fractal) to make Living Style Guide for WordPress sites and to take full advantage, the `include` and `extends`-tags should use an id handle instead of a path to the template. For example `{% extends '@base' %}` instead of `{% extends 'views/base.twig' %}`

#### Solution
Originally the `get_loader` function returns a Filesystem Loader. I changed it to return a Loader Chain (which is basically an array of Loaders) which includes the Filesystem Loader. I then added filter which allows theme developers to add custom Loaders to the environment. User @LostKobrakai was helpful to share his [Fractal Twig Loader](https://github.com/LostKobrakai/frctl-twig/blob/master/src/Loader.php) made for this exact purpose (though for another CMS). The custom Loaders and the original Filesystem Loader are merged into an array, which is used to create the Loader Chain. More about loaders [in Twig documentation](http://twig.sensiolabs.org/doc/api.html#built-in-loaders).

Example code how to actually take Fractal in to use in functions.php:

```php
require_once(__DIR__ . '/fractal-twig-loader.php'); // Custom Loader by LostKobrakai
function fractal_twig_loader() {
	// PHP array of template handles and paths to their corresponding twig files.
	// LostKobrakai has a NodeJS script for this purpose in one of his repos:
	// https://github.com/LostKobrakai/processwire-fractal/blob/master/fractal.js#L44
	$alias_handles = require(__DIR__ . '/build/fractal-handles.php');
	return array(new Alias_Loader($alias_handles));
}
add_filter('timber/loader/custom', 'fractal_twig_loader');
```

#### Impact
I am not a very experienced PHP developer and I'm not very familiar with Timber codebase, but my estimation is that this is a simple change and will have no ill effects on the current codebase, performance or backwards compatibility.

#### Usage
No usage changes that I know of. The Loader Chain includes the original Filesystem Loader and if you don't do anything with this feature, you won't be affected.

#### Considerations
It would be good to add some way to control if the Custom Loaders would be loaded before or after the original Filesystem Loader.

#### Testing
I didn't write any tests, because I'm not familiar with PHP testing.


jarednova:
Thanks so much for this PR @tnottu. Even without tests, the thoroughness of your description here gives me enough to work from. I'm going to start cracking into this so we can release it with Timber 1.1.4!

jarednova:
All merged! Thanks again @tnottu

Issue #1186, opened 2016-09-07
Non-static method Timber\Timber::init() should not be called statically in timber/lib/Timber.php on line 262


montchr:
### Expected behavior

Timber should just work when calling `Timber::render()`.

### Actual behavior

Any time I load a page that uses `Timber::render()`, I get the following error with a similar call stack each time:

```
Deprecated:  Non-static method Timber\Timber::init() should not be called statically in /var/www/genericwebsite.dev/htdocs/vendor/timber/timber/lib/Timber.php on line 262
Stack trace:
  1. {main}() /var/www/genericwebsite.dev/htdocs/index.php:0
  2. require() /var/www/genericwebsite.dev/htdocs/index.php:17
  3. require_once() /var/www/genericwebsite.dev/htdocs/wp-blog-header.php:19
  4. include() /var/www/genericwebsite.dev/htdocs/wp-includes/template-loader.php:75
  5. Timber\Timber::get_context() /var/www/genericwebsite.dev/htdocs/wp-content/themes/pedestal/single-pedestal_factcheck.php:7
  6. apply_filters() /var/www/genericwebsite.dev/htdocs/vendor/timber/timber/lib/Timber.php:242
  7. Pedestal\Frontend->filter_timber_context() /var/www/genericwebsite.dev/htdocs/wp-includes/plugin.php:235
  8. Timber\Timber::render() /var/www/genericwebsite.dev/htdocs/wp-content/themes/pedestal/inc/class-frontend.php:417
  9. Timber\Timber::fetch() /var/www/genericwebsite.dev/htdocs/vendor/timber/timber/lib/Timber.php:332
 10. Timber\Timber::compile() /var/www/genericwebsite.dev/htdocs/vendor/timber/timber/lib/Timber.php:317
```

### Steps to reproduce behavior

This happens on every page using Timber so I cannot isolate the issue.

### What version of WordPress, PHP and Timber are you using?

WordPress 4.6 // PHP 7.0.10 // Timber 1.1.3

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)

Composer

kingkool68:
Found this https://github.com/timber/timber/issues/979#issuecomment-224545836 don't know if it helps or not...

jarednova:
@montchr can you try upgrading to Timber 1.1.5 and see if this still occurs?

montchr:
@jarednova I upgraded and it still occurs.

montchr:
[...]

PR #1189, opened 2016-09-09
ref #1186 -- remove instance init calls; set init to be static

jarednova:
**Ticket**: #1186


mahagr:
Looks good and is backwards compatible, too.

PR #1193, opened 2016-09-15
Add excerpt chars filter

CROSP:
#### Issue

Add separate `Filter` class for commonly used filters. In this case add useful filter for trimming text
string to a certain amount of characters.
Decouple `Twig` class. Move all filter methods like `add_list_separators` , `time_ago`. (Not included in this pull request).

#### Solution

Create new class `Filter`. Add `trim_characters` method. Modify `Twig` class, add new `Twig_SimpleFilter`

#### Impact

Uses `mb_strimwidth` (PHP 4 >= 4.0.6, PHP 5, PHP 7) , wordpress functions `__()` for localization and
`wp_strip_all_tags` for striping all tags from input text.

#### Usage

By default string is trimmed to 60 characters. 

```
   <p class="post-content__text">
                    {{ post.post_content|excerpt_chars }}
    </p>
```
You can specify desired characters amount 

```
   <p class="post-content__text">
                    {{ post.post_content|excerpt_chars(130) }}
    </p>
```

It is possible to specify custom excerpt (more text). By default `&hellip;` is used

```
   <p class="post-content__text">
                    {{ post.post_content|excerpt_chars(130,'[...]') }}
    </p>
```
#### Considerations

I would suggest to decouple `Twig` class and move all filters to the `Filter` class.

### DIFF
--- a/lib/Filter.php
+++ b/lib/Filter.php
@@ -1,0 +1,30 @@
+<?php
+
+namespace Timber;
+
+/**
+ * Class provides different commonly used in Wordpress development
+ */
+class Filter
+{
+    /**
+     * Trims text to a certain number of characters.
+     * This function can be useful for excerpt of the post
+     * As opposed to wp_trim_words trims characters that makes text to
+     * take the same amount of space in each post for example
+     *
+     * @param string $text      Text to trim.
+     * @param int    $num_chars Number of characters. Default is 60.
+     * @param string|null $more      Optional. What to append if $text needs to be trimmed. Default '&hellip;'.
+     * @return string trimmed text.
+     */
+    public static function trim_characters($text, $num_chars = 60, $more = null)
+    {
+        if ($more === null) {
+            $more = __('&hellip;');
+        }
+        $text = wp_strip_all_tags($text);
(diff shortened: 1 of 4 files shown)


CROSP:
What do you think about moving all filters to separate class ? 
Also I found really useful excerpt character function, because it makes text to take the same amount of space.

jarednova:
Thanks @CROSP! I agree, it might be time to simplify the organization of the Filter code. I'll take a look at this guy in the next few days

CROSP:
@jarednova Thanks for answer. Also please consider to add characters excerpt function/filter everywhere where you have now word excerpt. The problem with word excerpt is that sometimes words are really long and each block will take different amount of space and this behaviour is not desirable sometimes. 
Thanks

jarednova:
I'm back from vacation and this is all-merged! Thanks @CROSP! I re-named your `Filter` class to `TextHelper` (to mirror the existing `Helper` and `URLHelper` classes), and pulled some related methods in. I also added character control to the new `PostPreview` object....

```twig
<div class="preview">{{ post.preview.chars(25) }}</div>
```

to also use your new method. Please let me know of any other locations I can provide filters to (or keep submitting more awesome PRs like this one)

PR #1206, opened 2016-10-06
Regenerate image if modified date of source image newer than existing…

dudewithamood:
**Ticket**: # 
**Reviewer**: @ 

#### Issue
When the original file is replaced with a new version all generated files remain unchanged reflecting the original version.

#### Solution
By comparing file modified dates between the source file and generated files the generated files can be regenerated if the source file is newer (has been changed more recently, indicating it has been changed/updated)

#### Impact
It is backward compatible.
The only performance implication is the regeneration of images.

#### Usage
No.

#### Considerations
Some way to clear generated images? One of the issues while developing is that images are resized to multiple sizes until the correct sizing is established and this generates image files that are not needed.

#### Testing

They're not included - please check the code and test. I understand if it's not correct and needs to be adjusted as I'm not necessarily and expert.

Test cases would be...

Without this code: original image is replaced (with same file name) and resized generated images reflect the image prior to being replaced

With the code:  original image is replaced (with same file name) and resized images are regenerated based on the updated source image


jarednova:
Thanks for your work @dudewithamood! I've merged this and #1207 — they'll be a part of the next WordPress.org release!

dudewithamood:
Sure, no problem @jarednova, and thank you @gchtr for you encouragement and guidance.

Sorry for disappearing - feels like I didn't complete properly - been out of contact for a few weeks and only getting back into stuff now with lots to catch up on. No doubt you'll see more from me as this is integral to the work I'm doing and am very grateful for what's been done.

PR #1207, opened 2016-10-07
Delete generate files for image when metadata generated

dudewithamood:
**Ticket**: # 
**Reviewer**: @ 

#### Issue
Generates files not affected when Wordpress regenerates thumbnails

#### Solution
Delete generated files for an image/attachment when Wordpress is called to regenerate thumbnails so that images generated by Timber can be regenerated when next called

#### Impact
All generated images for an attachment would be deleted and would be recreated when first accessed.
Backward compatible.

#### Usage
Just for folks to know that if they regenerate Wordpress thumbnails that any resized images created by Timber will be deleted and will only be regenerated when first accessed.

#### Testing
No tests.
Without the code, generated images remain.
With the code, generated images are deleted for an image/attachment.

### DIFF
--- a/lib/ImageHelper.php
+++ b/lib/ImageHelper.php
@@ -166,7 +166,7 @@ class ImageHelper {
 	}
 
 	/**
-	 * Deletes all resized versions of an image when the source is deleted
+	 * Deletes all resized versions of an image when the source is deleted or its meta data is regenerated
 	 */
 	protected static function add_actions() {
 		add_action('delete_attachment', function( $post_id ) {
@@ -178,6 +178,17 @@ class ImageHelper {
 					ImageHelper::delete_generated_files($attachment->file_loc);
 				}
 			}
+		} );
+		add_filter('wp_generate_attachment_metadata', function( $metadata, $post_id ) {
+			$post = get_post($post_id);
+			$image_types = array('image/jpeg', 'image/png', 'image/gif', 'image/jpg');
+			if ( in_array($post->post_mime_type, $image_types) ) {
+				$attachment = new Image($post_id);
+				if ( $attachment->file_loc ) {
+					ImageHelper::delete_generated_files($attachment->file_loc);
+				}
+			}
+			return $metadata;
 		} );
 	}
 
--- a/lib/ImageHelper.php
+++ b/lib/ImageHelper.php
@@ -170,24 +170,10 @@ class ImageHelper {
 	 */
 	protected static function add_actions() {
 		add_action('delete_attachment', function( $post_id ) {
-			$post = get_post($post_id);
-			$image_types = array('image/jpeg', 'image/png', 'image/gif', 'image/jpg');
-			if ( in_array($post->post_mime_type, $image_types) ) {
-				$attachment = new Image($post_id);
-				if ( $attachment->file_loc ) {
-					ImageHelper::delete_generated_files($attachment->file_loc);
-				}
-			}
+			ImageHelper::delete_generated_if_image($post_id);
 		} );
 		add_filter('wp_generate_attachment_metadata', function( $metadata, $post_id ) {
-			$post = get_post($post_id);
-			$image_types = array('image/jpeg', 'image/png', 'image/gif', 'image/jpg');
-			if ( in_array($post->post_mime_type, $image_types) ) {
-				$attachment = new Image($post_id);
-				if ( $attachment->file_loc ) {
-					ImageHelper::delete_generated_files($attachment->file_loc);
-				}
-			}
+			ImageHelper::delete_generated_if_image($post_id);
 			return $metadata;
 		} );
 	}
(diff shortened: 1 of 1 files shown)


gchtr:
I think this is a great addition.

This will make Timber make compatible with plugins like [Regenerate Thumbnails](https://wordpress.org/plugins/regenerate-thumbnails/) or [Replace Image](https://wordpress.org/plugins/replace-image/).

Regenerate Thumbnails would just delete all generated image sizes. The only implication I see is that the first visitor of a page would have longer load times, because the image sizes would have to be regenerated first. You could also manually visit all pages with images yourself to trigger regeneration, which is also a hassle. But it’s still a better option than not to be compatible with other plugins and to make if difficult to delete generated image sizes.

lib/ImageHelper.php
```diff
@@ -179,6 +179,17 @@ protected static function add_actions() {
 				}
 			}
 		} );
+		add_filter('wp_generate_attachment_metadata', function( $metadata, $post_id ) {
+			$post = get_post($post_id);
+			$image_types = array('image/jpeg', 'image/png', 'image/gif', 'image/jpg');
+			if ( in_array($post->post_mime_type, $image_types) ) {
```
gchtr: The 3 lines

```
$post = get_post($post_id);
$image_types = array('image/jpeg', 'image/png', 'image/gif', 'image/jpg');
if ( in_array($post->post_mime_type, $image_types) ) {
```

could be simplified to

```
if ( wp_attachment_is_image( $post_id ) ) {
```

lib/ImageHelper.php
```diff
@@ -179,6 +179,17 @@ protected static function add_actions() {
...
+				$attachment = new Image($post_id);
+				if ( $attachment->file_loc ) {
+					ImageHelper::delete_generated_files($attachment->file_loc);
+				}
+			}
+			return $metadata;
+		} );
```
gchtr: The filter uses practically the same code as the `delete_attachment` action above. Why not extract the logic inside the action and the filter as a new method of the `ImageHelper` class?

dudewithamood:
@gchtr Thanks for the review and feedback. I've not worked as part of a team like this so it's a new and appreciated experience.

lib/ImageHelper.php
```diff
@@ -170,24 +170,10 @@ public static function img_to_jpg( $src, $bghex = '#FFFFFF', $force = false ) {
...
-			$image_types = array('image/jpeg', 'image/png', 'image/gif', 'image/jpg');
-			if ( in_array($post->post_mime_type, $image_types) ) {
-				$attachment = new Image($post_id);
-				if ( $attachment->file_loc ) {
-					ImageHelper::delete_generated_files($attachment->file_loc);
-				}
-			}
+			ImageHelper::delete_generated_if_image($post_id);
 			return $metadata;
 		} );
```
gchtr: When using `add_filter` with a function that uses more than one parameter, you need to state the amount of parameters that you want to pass to the function. In this case, it’s `2` parameter with the default priority of `10`.

```php
add_filter('wp_generate_attachment_metadata', function($metadata, $post_id) {
	// Your function content
}, 10, 2);
```

lib/ImageHelper.php
```diff
@@ -216,6 +202,24 @@ static function add_filters() {
...
 	//-- end of public methods --//
+
+	
+	/**
+	 * Checks if attachment is an image before deleting generated files
+	 *
+	 * @param  int  $post_id   an attachment post id
+	 *
+	 */
+	static function delete_generated_if_image( $post_id ) {
```
gchtr: Scrutinizer reports the following issue:

> It is generally recommended to explicitly declare the visibility for methods.

Since this function could be useful to use for others, I guess it’s ok introduce it as a `public` function.

```php
public static function delete_generated_if_image( $post_id ) {
```

And in that case, you’d better move it above the comment `//-- end of public methods --//`.

lib/ImageHelper.php
```diff
@@ -216,6 +202,24 @@ static function add_filters() {
...
+		if ( wp_attachment_is_image( $post_id ) ) {
+			$attachment = new Image($post_id);
+			if ( $attachment->file_loc ) {
+				ImageHelper::delete_generated_files($attachment->file_loc);
+			}
+		}
```
gchtr: This function now is just a wrapper for `delete_generated_files` method with additional checks. Maybe `delete_generated_if_image` could be incorporated into `delete_generated_files`, so we do not need two function. Maybe like this:

```php
/**
 * Deletes the auto-generated files for resize and letterboxing created by Timber.

 * @param string|int  $local_file   File URL or Attachment ID
 * 									ex: /var/www/wp-content/uploads/2015/my-pic.jpg
 *	                            	or: http://example.org/wp-content/uploads/2015/my-pic.jpg
 */
public static function delete_generated_files( $local_file ) {
	// Check if file is an attachment
	if (wp_attachment_is_image($local_file)) {
		$attachment = new Image($local_file);
		$local_file = $attachment->file_loc;
	}

	if (URLHelper::is_absolute($local_file)) {
		$local_file = URLHelper::url_to_file_system($local_file);
	}
```

And because this function misses a visibility statement as well, we could also add `public` to it.

What do you think?

jarednova:
Thanks @dudewithamood and @gchtr! I'm going to work today on releasing a new version (with the current master) and then turn my attention to brining this into the next release. Stay tuned!

PR #1218, opened 2016-10-21
First pass at object factory

joshlevinson:
#### Issue

Begins to solve the problem of a lack of control over which classes Timber uses to instantiate WordPress objects.
#### Solution

This introduces a factory for creating Timber posts and terms.
Abstracts the class mapping functionality previously available only to posts
#### Impact

This change doesn't itself introduce back compat issues; future changes/usage might.
#### Usage

Changes how Timber WP objects are instantiated. These should be created with the appropriate Factory, rather than directly instantiating the Timber class objects.
#### Considerations

This solution doesn't introduce a factory for instantiating every Timber object. That may or may not be desirable. I'm up for discussion 🙂
#### Testing

Yet to be done!


joshlevinson:
Note that the travis-run unit test appear to have failed due to an http/svn-related issue. The test passed on all but one environment. I'm running PHP7 locally, but didn't make use of any features that are specific to > 5.4.

jarednova:
@joshlevinson I'm about to get to bed, but I'm super excited about what I see. This seems to really clean things up (I didn't realize how much of the Post and Term class could really be delegated to the Factory to handle the BS like `check_post_id`). I hate to fully communicate my ignorance, but I want to make sure I fully educate myself on the use of factories as a programming concept before I get to deep. Is there any recommended 📖 that you have to help get me better up-to-speed?

joshlevinson:
I don't have any good books handy to recommend; most of my reading has been online.
You can check out these links though:

* [Factory Method](http://designpatternsphp.readthedocs.io/en/latest/Creational/FactoryMethod/README.html) _personal fav_
* [Simple Factory Pattern](http://designpatternsphp.readthedocs.io/en/latest/Creational/SimpleFactory/README.html)
* [Design Patterns in _PHP The Right Way_](http://www.phptherightway.com/pages/Design-Patterns.html) 
* [The factory design pattern in PHP explained](http://phpenthusiast.com/blog/the-factory-design-pattern-in-php)

acobster:
@jarednova @gchtr @joshlevinson I would be interested in picking this up again if there's any chance of getting a feature like this into 2.x. Especially with regard to being able to (somehow) specify a factory to use when calling `get_all` and friends.

Example use-case:

* say I have custom `Page`, `Post`, and `CustomPost` classes that all extend `Timber\Post`
* I want end-users to be able to search all of them, across post types
* I want to display their search results polymorphically in a Twig template that calls `{{ post.get_search_excerpt }}`

jarednova:

Issue #1221, opened 2016-10-26
Post->get_terms() not compatible with Post->terms()


haroldangenent:
### Expected behavior

`Post->get_terms()` (deprecated since 1.0) being backwards compatible.
### Actual behavior

Not compatible, `Post->get_terms()` doesn't receive the `$TermClass` parameter anymore.
### Steps to reproduce behavior
1. Use `Post->get_terms()` with another `$TermClass` before 1.0
2. Upgrade to 1.0+
### What version of WordPress, PHP and Timber are you using?

WordPress 4.6.1, latest Timber, PHP 7.
### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)

Composer. Or any other.

haroldangenent:
By the way, the issue seems to be created in https://github.com/timber/timber/commit/6b2f438d6ae6f2f889aea56e6cdd479f0d135d7b.

jarednova:
Closed by #1222

PR #1222, opened 2016-10-26
Post->get_terms() return class

haroldangenent:
**Ticket**: #1221
**Reviewer**: @jarednova / @connorjburton?
#### Issue

See #1221.
#### Solution
- Test for `Post->terms()` return class
- Test for `Post->get_terms()` return class
- Receive parameter for `Post->get_terms()` and pass it
#### Impact

More backwards compatibility with pre-1.0 versions.
#### Usage

No changes.
#### Considerations

The `$TermClass` parameter could be better documented (this is missing right now), but that's probably better as a separate issue.
#### Testing

Tests are included.


haroldangenent:
The failing test seems to have [a different origin](https://travis-ci.org/timber/timber/jobs/168654477) and also fails on `master`.

jarednova:
This looks good, thanks @haroldangenent !

PR #1227, opened 2016-11-02
Fix images when wordpress is not installed in default location

njbarrett:
**Ticket**: # 
**Reviewer**: @ 

#### Issue
When Wordpress is in a different directory other than the default, (for example when using [Bedrock](https://github.com/roots/bedrock), images are not found by the ImageHelper class.

#### Solution
Replace uses of site_url() with get_home_url()

#### Impact
The change is designed to have no impact.

#### Considerations
Will need to make sure this has no unintended side effects on multisite installs.

#### Testing
Tests are already in place.


jarednova:
Awesome, thanks @njbarrett !

PR #1232, opened 2016-11-07
Joshlevinson feature/interfaced object factories

jarednova:
Jared's tweaks to #1220. Not necessarily for merging


jarednova:

PR #1229, opened 2016-11-03
1227 image test fix

jarednova:
**Ticket**: #1227 

We might have gotten a little ambitions with replacing `site_url` with `get_home_url()`. I found some errors in tests relating to WPML and external images. This splits the tests to better measure and resolves errors

PR #1234, opened 2016-11-09
fix resize for urls when upload directory is not in default location

njbarrett:
Unfortunately the reverting in #1229 has caused issues with resizes when uploads is not in a standard location. This attempts to fix it again, and not break WPML.

### DIFF
--- a/lib/ImageHelper.php
+++ b/lib/ImageHelper.php
@@ -367,9 +367,11 @@ class ImageHelper {
 			}
 		} else {
 			if ( !$result['absolute'] ) {
-				$pos = strpos(content_url(), CONTENT_DIR);
+				$pos = strpos($upload_dir['baseurl'], $upload_dir['relative']);
 				if ($pos !== false) {
-					$tmp = substr_replace(content_url(), '', $pos, strlen(CONTENT_DIR)) . $tmp;
+					$tmp = substr_replace($upload_dir['baseurl'], '', $pos, strlen($upload_dir['relative'])) . $tmp;
+				} else {
+					$tmp = site_url().$tmp;
 				}
 			}
 			if ( 0 === strpos($tmp, $upload_dir['baseurl']) ) {
--- a/lib/ImageHelper.php
+++ b/lib/ImageHelper.php
@@ -367,7 +367,10 @@ class ImageHelper {
 			}
 		} else {
 			if ( !$result['absolute'] ) {
-				$tmp = site_url().$tmp;
+				$pos = strpos(content_url(), CONTENT_DIR);
+				if ($pos !== false) {
+					$tmp = substr_replace(content_url(), '', $pos, strlen(CONTENT_DIR)) . $tmp;
+				}
 			}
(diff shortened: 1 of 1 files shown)


lib/ImageHelper.php
```diff
@@ -367,7 +367,10 @@ private static function analyze_url( $url ) {
 			}
 		} else {
 			if ( !$result['absolute'] ) {
-				$tmp = site_url().$tmp;
+				$pos = strpos(content_url(), CONTENT_DIR);
```
jarednova: did you mean `WP_CONTENT_DIR`?

lib/ImageHelper.php
```diff
@@ -367,7 +367,10 @@ private static function analyze_url( $url ) {
...
+				if ($pos !== false) {
+					$tmp = substr_replace(content_url(), '', $pos, strlen(CONTENT_DIR)) . $tmp;
```
jarednova: did you mean `WP_CONTENT_DIR`?

lib/ImageHelper.php
```diff
@@ -367,7 +367,10 @@ private static function analyze_url( $url ) {
 			}
 		} else {
 			if ( !$result['absolute'] ) {
```
jarednova: Right now `$tmp` only gets a value when `$post !== false`, should this line be brought back?

jarednova:
Hey @njbarrett — thanks for the PR! I'm seeing a few errors in Travis. I left some notes. Please let me know if there's a test you have so I can write/apply something to test against — thanks!

lib/ImageHelper.php
njbarrett: My new commit fixes this

lib/ImageHelper.php
njbarrett: Not in this case, CONTENT_DIR is a constant defined by the bedrock code which I didn't realize isn't core wordpress. I've changed it now.

lib/ImageHelper.php
njbarrett: Not in this case, CONTENT_DIR is a constant defined by the bedrock code which I didn't realize isn't core wordpress. I've changed it now.

njbarrett:
Hey @jarednova , only just got back on this today.
I have done some more testing locally and adjusted my fix.

The issue I am having is weird - the first time a resize is done, the correct URL is passed into `Timber\ImageHelper analyze_url`, e.g:
`http://example.org/app/uploads/2016/11/527f4446420f3c34c44033a01bd8bbe9.jpg`

then subsequent urls come into the function as:

`/app/uploads/2016/11/527f4446420f3c34c44033a01bd8bbe9.jpg` (not absolute urls)

then in the current version, `analyze_url` changes this to

`http://example.org/wp/app/uploads/2016/11/527f4446420f3c34c44033a01bd8bbe9.jpg` (note the additional `/wp`)

njbarrett:
Upon writing the above comment and going through the debugger again, I found a much cleaner fix.

jarednova:
I love it, thanks @njbarrett !

PR #1237, opened 2016-11-11
fix get menu by slug

### LINKED ISSUE #1230: Multiple Menu Issues
### Expected behavior
I'm trying to implement multiple menus. One for the header and one for the footer. I have made both menus in WordPress. I have also registered both menus and issued their own specific locations. 

### Actual behavior
The Primary nav display correctly using the TimberMenu class. I make another call for the footer menu and the result is null.

**functions.php**
```
function add_to_context( $context ) {
  $context['menu'] = new TimberMenu('primary-menu');
  $content['footer'] = new TimberMenu('footer-menu'); 
  $context['site'] = $this;
  return $context;
}
```

**footer.twig**
```	
<div class="sub-footer-menu">
 {% if footer %}
      <ul>
        {% for item in footer.get_items %}
         <li class="{{item.classes | join(' ')}}">
          <a href="{{item.get_link}}">{{item.title}}</a>
        </li>
       {% endfor %}
     </ul>
  {% endif %}
</div>
```

**base.twig**
```
{% block footer %}
   <footer id="footer">
     {% include "footer.twig" with {"footer": footer.get_items} %}
   </footer>
   {{ function('wp_footer') }}
{% endblock %}
```

I can include the menu fine if I use the regular `wp_nav_menu()` function but I'd rather not go that route. 

+ WordPress 4.6.1
+ PHP Version 7.0.10
+ Timber 1.1.8
+ Environment = MAMP Local

Upgraded to newest version via plugin updater in WordPress dashboard

motia:
fix #1230 
TimberMenu($slug) did not work correctly for menu slug input, because ```get_menu_id_from_terms``` was not reachable in case wordpress locations exist and ```get_menu_id_from_locations``` fails in finding the menu id.

### DIFF
--- a/lib/Menu.php
+++ b/lib/Menu.php
@@ -80,6 +80,7 @@ class Menu extends Core {
 	 * @param integer|string $slug
 	 */
 	public function __construct( $slug = 0 ) {
+		$menu_id = false;
 		$locations = get_nav_menu_locations();
 		if ( $slug != 0 && is_numeric($slug) ) {
 			$menu_id = $slug;
--- a/lib/Menu.php
+++ b/lib/Menu.php
@@ -87,7 +87,8 @@ class Menu extends Core {
 			$menu_id = $this->get_menu_id_from_locations($slug, $locations);
 		} else if ( $slug === false ) {
 			$menu_id = false;
-		} else {
+		}
+		if ( !$menu_id ) {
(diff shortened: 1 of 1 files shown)


jarednova:
Hey @motia, thanks so much for the PR! I made one tweak and we're good!

PR #1239, opened 2016-11-12
fixed guest authors loading

motia:
#### Issue
First, I had like to thank you for the great plugin!
I found a bug in the co-authors-plus integration. For instance, the code below displays only co-authors using a Wordpress account but it doesn't show guest-authors. 
``` 
{% for author in post.authors %}
    {{ author.name }}
{% endfor %}
```
It seems that the User class fails to create init its fields when created from a guest-author.

#### Solution
Create a CoAuthorsPlusUser class which extends Timber\User, and make CoAuthorsPlus decide which User class to represent each author.

#### Considerations
As guest-authors may have a linked account and can have use a featured image as an avatar, I added options to rule co-authors-plus integration. 
```
// guest-author information is loaded from user if he exists
Timber\Integrations\CoAuthorsPlus::prefer_gravatar = false;
 //edit: Use instead
 //global $coathors_plus
 // $coathors_plus->force_guest_authors = true;

// guest-author gravatar takes precedence to its featured image
Timber\Integrations\CoAuthorsPlus::prefer_gravatar = true;
```

### Testing
I manually tested to dump post.authors in single.twig for those situations:

1.  **normal author**: dumps an User object
2.  **guest author not having linked account**: dumps a Timber\Integrations\CoAuthorsPlusUser
3.  **guest author having linked account**: 
            - ```prefer_guur == false```: dumps a Timber\User
            - ```prefer_user == true```: dumps a Timber\Integrations\CoAuthorsPlusUser
4.  **guest author having featured image**: 
            - ```prefer_gravatar == false && have_featured_image```:  author.avatar displays the image image
            - ```prefer_gravatar == true  && have_featured_image```: author.avatar displays a gravatar image
            - ```prefer_gravatar == true  && !have_featured_image```: author.avatar displays a gravatar image


jarednova:
Thanks @motia! I noticed you were still updating some tests. Will you have some time in the next few days to fix the `TestTimberIntegrationsCoAuthors::testGuestAuthorAvatar` test? Thanks again!

motia:
@jarednova , I finished fixing the tests. Please, let me know if I need to make any change.

jarednova:
Thanks so much @motia — I added another test (and mucked with the spacing a bit), but this is 🎉 fantastic 🎉 . Thanks so much for carefully considering the usage/implications and providing tests. Anything else you're thinking of improving/fixing?

Issue #1129, opened 2016-07-26
Incompatibility with Yoast SEO on multisite / network


jordant1089:
There is an incompatibility issue between Timber and [Yoast SEO](https://wordpress.org/plugins/wordpress-seo/). I've made a comment on a [similar issue in their repository](https://github.com/Yoast/wordpress-seo/issues/4989). 

The issue occurs on multisite installs where both plugins are active. It occurs because when saving a post, their metabox checks if the site is a multisite and if the site has switched in the process of saving, and if it has, the saving process bails. When instantiating a Site object in Timber on a multisite, the process switches sites even if the Site object being created is for the current site. 

It seems like the issue was introduced from changes for issue #568. 
### Expected behavior

Changes made in the Yoast SEO metabox would be saved.
### Actual behavior

Changes made in the Yoast SEO metabox are not being save.
### Steps to reproduce behavior

Have Timber plugin, a timber theme, and Yoast SEO plugin active and edit a post. In the Yoast SEO metabox, make changes to the title or description. Save the post and the changes to the title or description will not be saved. 
#### Reproduction From a clean install

I was able to reproduce the issue on a clean installation of everything, in case that is any help.
1. Setup new multisite
2. Add and activate Yoast SEO and Timber plugins
3. Copy the 'timber-starter-theme' directory from the Timber plugin directory to the themes directory
4. Activate the 'My Timber Starter Theme' theme on a site
5. Create or edit a post and make changes in the SEO metabox
6. Save the post
### What version of WordPress, PHP and Timber are you using?

WordPress 4.5.3
Timber 1.1.1
Yoast SEO 3.4
### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)

wordpress.org download and copy to plugins directory, however in my reproduction steps, the plugin was installed via the admin plugin search.

leonmanifesto:
Any update with this?

Seems like a pretty major issue.

kmmathis:
I can confirm this bug. Has anyone found a way to fix this?

kmmathis:
As a temporary fix, I literally removed lines 113 through 118 of Site.php. Not sure what the repercussions of this are, as I don't really know what this code is doing.

```php
if ( is_multisite() ) {
  $blog_ids = self::switch_to_blog($site_name_or_id);
  $this->init();
  $this->init_as_multisite($blog_ids['new']);
  return switch_to_blog($blog_ids['old']);
} 
```

alexandernanberg:
[...]

PR #1244, opened 2016-11-16
Fix for Yoast SEO metabox on multisite #1129

alexandernanberg:
Use Wordpress internal function to restore current blog instead of manually doing it. Solves issue with Yoast SEO metabox on multisites. 

I'm not sure if my changes will break other functionality. But I've done some testing on our site and not encountered any problems at all. We use ACF, WPML and many more plugins and it seems like it hasn't affected them.

### DIFF
--- a/lib/Site.php
+++ b/lib/Site.php
@@ -124,11 +124,15 @@ class Site extends Core implements CoreInterface {
 	/**
 	 * Switches to the blog requested in the request
 	 * @param string|integer|null $site_name_or_id
-	 * @return array with the ID of the old and new blogs
+	 * @return integer with the ID of the new blog
 	 */
 	protected static function switch_to_blog( $site_name_or_id ) {
 		if ( $site_name_or_id === null ) {
-			$site_name_or_id = get_current_blog_id();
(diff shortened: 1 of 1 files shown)


jarednova:
Thanks @alexandernanberg! Unfortunately the change causes a test to fail; can you take a look at resolving that? (looks to be pretty straightforward) — then we can merge this in!

jarednova:
Here's a link to the Travis job: https://travis-ci.org/timber/timber/jobs/176334361

alexandernanberg:
@jarednova My intended fix apparently failed more tests. I haven't used Travis before but is there some simple way to test everything before you push your commit? Googled a bit but didn't find anything concrete.

jarednova:
@alexandernanberg that did the trick! For next time, here's a guide to getting tests setup locally:

https://github.com/timber/timber/wiki/Unit-testing

alexandernanberg:
@jarednova Ahh I completely missed the wiki for unit testing, I'll keep that in mind for future pull requests! Thanks for merging this btw

Issue #1249, opened 2016-11-23
Pagination issue when using query string parameters


xavivars:
### Expected behavior
Given a page like this:

https://www.softcatala.org/noticies/?cerca=gimp

where `cerca` is defined as a query_var

```
function add_query_vars_filter( $vars ){
    $vars[] = "cerca";

    return $vars;
}
add_filter( 'query_vars', 'add_query_vars_filter' );
```
pages should get paginated as 

https://www.softcatala.org/noticies/page/2/?cerca=gimp

### Actual behavior
Pagination links are created like

https://www.softcatala.org/noticies/page/2/?cerca=gimp/

### Steps to reproduce behavior
The problem is in this line:
https://github.com/timber/timber/blob/master/lib/Pagination.php#L172

```
	public static function is_search_query( $url ) {
		if ( strpos($url, 's=') !== false ) {
			return true;
		}
		return false;
	}
```

where `s` is the only query var that is checked against.

### What version of WordPress, PHP and Timber are you using?
WordPress 4.6, PHP 5.6, Timber 1.1.10

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
WP.org

jarednova:
Closed by #1250

PR #1250, opened 2016-11-23
Add support for public query_vars

xavivars:
**Ticket**: https://github.com/timber/timber/issues/1249
**Reviewer**: @ 

#### Issue
Problem with pagination and usage of query vars

#### Solution
Take query_vars into account when deciding if pages need to be removed

#### Impact
Pagination

#### Considerations
I've also removed tests I added in the past when working on some other pagination issues, as one of them is in fact doing the wrong thing.

#### Testing
Unit tests are included.

### DIFF
--- a/lib/Pagination.php
+++ b/lib/Pagination.php
@@ -167,17 +167,20 @@ class Pagination {
 				if ( $args['show_all'] || ($n <= $args['end_size'] || ($args['current'] && $n >= $args['current'] - $args['mid_size'] && $n <= $args['current'] + $args['mid_size']) || $n > $args['total'] - $args['end_size']) ) {
 					$link = str_replace('%_%', 1 == $n ? '' : $args['format'], $args['base']);
 					$link = str_replace('%#%', $n, $link);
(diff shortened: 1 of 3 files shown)


xavivars:
I've faced some problems locally with some tests: `testWPMLurlRemote` (in test-timber-image-resize.php) is failing for me, and I can't see how that test could have ever passed. However, it seems the tests is not being executed in Travis, as GD-dependant tests are ignored.

xavivars:
I've identified a better way to handle this. I had to rewrite most of `paginate_links`, as when looking at it the current logic didn't have much sense: fragment (`#`) being added before the query strings, trailing/untrailingslashit after adding both fragment and query strings....

I'm making sure all tests pass now, and I'll commit the new code.

xavivars:
I've done another round of refactor, now changing more stuff (even some tests that, IMHO, were wrongly setup). But I'd like to leave it for a second round of review in case this one gets merged.

jarednova:
So much cleaner, thanks @xavivars — and tests too!!!!

Issue #1253, opened 2016-11-28
Unable to add additional Twig Filesystem Loaders


adamtomat:
### Expected behavior
We need to change the way twig handles file look-ups as we are using a pattern library with WordPress. We therefore would like to register our own Loader (that implements `Twig_LoaderInterface`, see example below) with Timber.

For example: Our file structure looks like this:

- <theme>/views/templates/
    - home/
        - template.twig

Given this command:

`Timber::render('templates/home', $context);`

We want it to find `views/templates/home/template.twig` in our theme.

I don't expect this exact use-case to be fixed. Instead, having a hook when Timber initialises the `Twig_Loader_Filesystem` class would enable us to extend it with our own loader.

### Actual behavior

The current behaviour, when trying to find `templates/home` gives this error:

```
Fatal error: Uncaught exception 'Twig_Error_Loader' with message 'Unable to find template "templates/home" (looked into: <some_dir>/themes/<theme_name>, <some_dir>/themes/<theme_name>/views, ).' in <some_dir>/vendor/twig/twig/lib/Twig/Loader/Filesystem.php:247
```

### Steps to reproduce behavior

1. Create a twig file in `<theme>/views/templates/home/template.twig`

2. Register load paths for Timber (in `functions.php`)

```php
new Timber();

Timber::$dirname = [
    'views/templates',
];
```

3. In a controller, e.g. `page.php` add the following:

`Timber::render('templates/home', $context);`

### Suggested Fix

Add a hook in `lib/Loader.php` in the `getLoader()` method, after initialising the `Twig_Loader_Filesystem`.

*Current method (as of 1.1.0):*

```php
/**
 * @return \Twig_Loader_Filesystem
 */
public function get_loader() {
    $open_basedir = ini_get('open_basedir');
    $paths = array_merge($this->locations, array( $open_basedir ? ABSPATH : '/'));
    $paths = apply_filters('timber/loader/paths', $paths);

    $rootPath = '/';
    if ( $open_basedir ) {
        $rootPath = null;
    }
    $fs = new \Twig_Loader_Filesystem($paths, $rootPath);
    return $fs;
}
```

*Suggested fix (will submit a PR for this too):*

```php
/**
 * @return \Twig_Loader_Filesystem
 */
public function get_loader() {
    $open_basedir = ini_get('open_basedir');
    $paths = array_merge($this->locations, array( $open_basedir ? ABSPATH : '/'));
    $paths = apply_filters('timber/loader/paths', $paths);

[...]

PR #1254, opened 2016-11-28
Unable to add custom twig loaders

adamtomat:
**Ticket**: https://github.com/timber/timber/issues/1253

#### Issue
Unable to add custom twig loaders.

#### Solution
Add a new filter (hook), that exposes the `Twig_Loader_filesystem` that TImber creates. This allows theme developers to modify this or add their own loader and create a new [loader chain](http://twig.sensiolabs.org/doc/api.html#twig-loader-chain).

#### Impact
This fix is additive, and therefore should be backwards compatible. It's also possible that developers could not return the correct thing in the filter which could cause unexpected behaviour. I'm not sure there's much value in trying to defend against this though.

#### Usage
An example use case:

```php
namespace App;

// Hook into when Timber initialises the Twig_Loader_Filesystem
add_filter('timber/loader/loader', function ($loader) {
    $customLoader = new \App\CustomLoader(get_stylesheet_directory() . '/views');
    // Add our loader to the end of the chain (i.e. Timber does its lookup, then runs ours)
    $chainLoader = new \Twig_Loader_Chain([$loader, $customLoader]);

    return $chainLoader;
});

class CustomLoader implements \Twig_LoaderInterface
{
    // Base path for patterns lib.
    protected $patternsPath;

    public function __construct($patternsPath)
    {
        $this->patternsPath = $patternsPath;
    }

    /**
     * Gets the source code of a template, given its name.
     *
     * @param  string $name string The name of the template to load
     *
     * @return string The template source code
     */
    public function getSource($name)
    {

        // Remove the extension at this point as we add it ourselves below
        $id = preg_replace('/\.twig$/', '', $name);

        // Add template.twig into path
        $path = $this->patternsPath . '/' . $id . '/template.twig';

        return file_get_contents($path);
    }

    /**
     * Gets the cache key to use for the cache for a given template name.
     *
     * @param  string $name string The name of the template to load
     *
     * @return string The cache key
     */
    public function getCacheKey($name)
    {
        return md5($name);
    }

    /**
     * Returns true if the template is still fresh.
     *
     * @param string    $name The template name
     * @param timestamp $time The last modification time of the cached template
     */
    public function isFresh($name, $time)
    {
        return true;
    }
}
```

#### Considerations
This solution simply exposes the loader, which allows people to do a lot more. Another approach may be to let Timber handle the `Twig_Loader_Chain` and just expose an `addLoader()` method or something similar. However, that feels less flexible and scaleable.

#### Testing
Sorry, no tests written. An example test could be something like:

```php
function testTwigLoaderFilter () {
    $php_unit = $this;
    add_filter('timber/loader/loader', function ($loader) use ($php_unit) {
        $php_unit->assertInstanceOf('Twig_Loader_Filesystem', $loader);
    });
    $str = Timber::compile('assets/single.twig', array());
}
```


mahagr:
`timber/loader/filesystem` sounds better than the `timber/loader/loader`..?

adamtomat:
I would agree.

My original thinking was that the filesystem was only one type of loader. You could in theory have a database loader. So I tried to keep the name generic - but actually `timber/loader/filesystem` better describes what is actually going on, i.e. the filesystem loader is being exposed.

mahagr:
For database loader I would have event `timber/loader/database` just because of people start assuming that loader is always of the same type and when it gets changed -- plugins will stop working.

It would be nice to have chain loader in the top, though :)

adamtomat:
Been thinking about the name change this morning & I think I disagree with `timber/loaders/filesystem` now.  Timber has a method called `get_loader()` which currently returns a `Twig_Loader_Filesystem`. But Timber, as far as I know, doesn't care that this loader is the filesystem; the only thing it cares about is that it implements `Twig_LoaderInterface`.  

The fix in this PR allows developers to:
- overwrite the filesystem loader with their own
- create a `Twig_Loader_Chain` with Timber's filesystem loader and any additional loaders

Maybe the method needs better naming:

```php
// ...
$loader = new \Twig_Loader_Filesystem($paths, $rootPath);
$loader = apply_filters('timber/loader/loader', $loader);
return $loader;
```

jarednova:
@adamtomat Thanks for the PR! I'm sympathetic to the idea that `timber/loader/filesystem` is much prettier, but given your reasoning and @mahagr's, it makes sense that _accuracy_ is the most crucial guiding factor at this "deep" level. Merged!

jarednova:
also, extra special points for the little test you put in your notes. I added it to PHPUnit so we can make sure `loader/loader` works well into the future

mahagr:
Forgot to reply earlier: I'm fine with loader/loader.

PR #1255, opened 2016-11-28
load children inheriting published status

motia:
#### Issue
Child posts with inherit status are not loaded with post.children

#### Solution
Add ```post_status=inherit``` to get_children query.

#### Impact
None

#### Usage
Attachments with post.number are now correctly loaded

#### Considerations
```post.children``` returns also the thumbnail making ```post.children``` cumbersome it in twig templates. So, what do you think about excluding it from the method it self?

#### Testing
Done

### DIFF
--- a/lib/Post.php
+++ b/lib/Post.php
@@ -849,7 +849,10 @@ class Post extends Core implements CoreInterface {
 		if ( is_array($post_type) ) {
 			$post_type = implode('&post_type[]=', $post_type);
 		}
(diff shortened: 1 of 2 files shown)


jarednova:
Thanks @motia! I'm not quite sure of which thumbnail issue with `post.children` you're referring to though? If the concern is about thumbnails loading, those are only called once the actual `{{ post.thumbnail }}` is called from the templates (it's not a value loaded at init, or at least it _shouldn't_ be)

jarednova:
Also, bonus points for the TESTS!!!! woohooo!!

motia:
@jarednova , I meant to say that ```post.children``` contains the post thumbnail (not what we want in our templates), I think it should be removed, is not it?

PR #1258, opened 2016-12-02
Resolve issue where file_loc isn't set in some situations

jarednova:


jarednova:
Never mind, this is a rat's nest for another day 🐀

PR #1260, opened 2016-12-02
Refactor of Pagination

xavivars:
**Ticket**: # 
**Reviewer**: @ 

#### Issue
Bigger refactor of the Pagination class. 

The main problem of the current implementation is that it looks for a public_query_var (before my previous pull request, for a hard-coded English search var) and, if it exists, it always removes the trailing slash. The problem here is that, if a query string exists, the trailing slash should have never  been added after the query string, but before. 

Ideally, we should just use `paginate_links` from core, but right now it's impossible to use it, as it doesn't receive the links as parameters and uses global variables or other WordPress methods instead. 

The ultimate goal is to have https://core.trac.wordpress.org/ticket/37056 implemented, but in the mean time, this new implementation tries to take care of add, adding the trailing slash where it belongs, and following a logical build steps for URLs

#### Impact
There are some breaking changes in cases where pretty permalinks are disabled (see unit tests changed) but IMHO they were wrong: URLs should never be generated as XML encoded strings, but only rendered like that when they are placed into XML tag's attributes (which may not be what the user will do) 

#### Considerations
The pull request is probably missing some code formatting (spaces, etc) that you introduced with a previous commit. If you can't merge it like this, I 'll try to rebase the changes on top of the current master.


xavivars:
@jarednova: do you think this is a good improvement to the existing codebase? I do think so, but I can close the pull request if you don't ;)

jarednova:
Hi @xavivars, thanks for submitting! I've been out for break so just getting to this now. I'll review and hit you back with notes/etc.

lib/URLHelper.php
```diff
@@ -292,6 +292,42 @@ public static function remove_trailing_slash( $link ) {
 	}
 
 	/**
+	 * Pass links through user_trailingslashit handling query strings properly
```
jarednova: @xavivars — if we were to use WP's `user_trailingslashit` w/o this it wouldn't handle query strings properly, do I have that right?

xavivars: Yes, `user_trailingslashit` acts only on top of "paths"

jarednova: cool, that's what I thought

lib/URLHelper.php
```diff
@@ -292,6 +292,41 @@ public static function remove_trailing_slash( $link ) {
...
+			// absolute url
+			if( isset($link_parts['path']) && $link_parts['path'] != '/' ) {
+				$new_path = user_trailingslashit( $link_parts['path'] );
+				
+				if ( $new_path != $link_parts['path'] )	{
+					$link = str_replace($link_parts['path'], $new_path, $link);
+				}
+			}
+		} else {
+			// relative url
```
jarednova: Sounds good! My thought here is garbage-in / garbage-out

jarednova:
Thanks @xavivars! I added a few lasts tests and merged your commits under #1284; thanks again for this and your other PRs!

PR #1265, opened 2016-12-08
Deprecated Twig getExtension argument

codesman:
**Ticket**: # 
**Reviewer**: @ 

#### Issue

Passing the string `'core'` to the `$twig->getExtension` method was deprecated in v1.26 and will be removed in 2.0. The API now requires that the fully qualified class name be used instead.

#### Solution

Changing the string `'core'` to `'Twig_Extension_Core'`

#### Impact

Shouldn't have any effect, very minor change.

#### Usage

N/A

#### Considerations

Deprecation notice is gone and the code works without error, can't get more ideal than that!

#### Testing

There were no tests for this string previously and I didn't add any. Based on the deprecation notice, an app will fail without this change if `2.0 >= $twig_version `. I did not test this.

### DIFF
--- a/lib/Twig.php
+++ b/lib/Twig.php
@@ -220,18 +220,18 @@ class Twig {
 	 */
 	public function add_timber_escapers( $twig ) {
 
-		$twig->getExtension('core')->setEscaper('esc_url', function( \Twig_Environment $env, $string ) {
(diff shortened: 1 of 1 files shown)


jarednova:
Looks good, thanks @codesman!

PR #1268, opened 2016-12-09
Adds Menu Item Thumbnail

edolyne:
**Ticket**: # 
**Reviewer**: @ 

#### Issue
When looping through menu items, you are not able to obtain the referenced post thumbnail image.

#### Solution
This adds a function to expose the post thumbnail url.

#### Impact
This change provides additional functionality.

#### Usage
When looping through a menu you can now use `{{item.thumbnail}}` to obtain the post thumbnail.

### DIFF
--- a/lib/MenuItem.php
+++ b/lib/MenuItem.php
@@ -311,4 +311,19 @@ class MenuItem extends Core implements CoreInterface {
 			return $this->__title;
 		}
 	}
+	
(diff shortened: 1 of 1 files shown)


jarednova:
@edolyne thanks! I did some clean-up to re-use the existing property for `$master_object` that already stores that data and added a test

PR #1283, opened 2017-01-05
fix php notice: "WP_User->id was called with an argument that is depr…

drumba:
…ecated since version 2.1.0! Use WP_User->ID instead."

**Ticket**: # 
**Reviewer**: @ 

#### Issue
The usage of CoAuthorsPlus is generating a php notice.

#### Solution
Change id object property to uppercase.


jarednova:
Awesome, thanks for the catch @drumba !

PR #1285, opened 2017-01-09
Fix a misspelled function in Timber/Admin

alexandernanberg:


jarednova:
Whoops! Thanks @alexandernanberg !

PR #1310, opened 2017-01-27
Use 'pre_get_avatar_data' filter for avatars

xavivars:
#### Issue
Avatar method in the comment class is ignoring filters set up previously (like 'pre_get_avatar_data).

#### Solution
Add filter to avatar method.

#### Impact
Very low, improving consistency with WordPress core.

#### Usage
Transparent for users.


jarednova:
Thanks @xavivars — will take a look tonight!

jarednova:
Thanks as always @xavivars !

Issue #1131, opened 2016-07-28
TimberImage not hooking into all the right functions


sftsk:
### Expected behavior

We are using OffloadS3 on our Amazon server to keep images on S3. When accessing their url like this `TimberImage(icon).src` the path returned is not the right path. It gives the local path and not the s3 path (and the local file does not exist anymore as it is removed upon upload to s3).
`http://new.bpgen.com/app/uploads/2016/07/plate-with-53.jpg`

If I use a image size like here `TimberImage(icon).src('box-link-image')` it gives the right path: `http://s3-eu-west-1.amazonaws.com/bpg2-test/app/uploads/2016/07/20123314/plate-with-53-315x300.jpg`.

The problem is that this does not work with `svg` images as they won't be resized.

You can see this here: http://new.bpgen.com/how-to-order/

Could you please have a look at this and tell me if there is any workaround? It seems like you are not hooking into all the right function for getting the file url at all times. Also using `|resize` does not work for pretty much the same reason.

Thank you for your help,
Lukas

sftsk:
I just saw that this has been discussed on #429 and that there have been some solutions.

Please look into this and fix this within Timber.
For the moment I have "fixed" it by applying the following filter:

```
// fix .src to use WordPress functions for attachment url (trigger s3 plugin hook)
function fix_timber_s3_src( $src, $ID ) {
	$size = '';
 	return wp_get_attachment_image_src($ID, $size)[0];
}
add_filter( 'timber_image_src', 'fix_timber_s3_src', 10, 2 );
```

would be great if you could pass size into the filter than this would almost be a working solution but in any case please fix this within TimberImage.

jarednova:
Thanks @sftsk! I'll take a look at this, but feel free to submit a PR that takes care of the fix you're looking for

jarednova:
@sftsk I believe this is now solved by #1320 — can you test against master to confirm? thanks!

sftsk:
Hey @jarednova , I am sorry but I do not have control over that project anymore so I can not test it. Looking at the pull request it seems like everything is fine :)

Thanks a lot & great work!

Issue #429, opened 2014-12-26
TimberImage: Incorrect URL returned when image is stored off-server


TomSeldon:
When calling `get_src` (either directly or through typecasting a `TimberImage` instance to a string), the method doesn't take into account any of the standard WP methods for retrieving the image's URL (e.g. `wp_get_attachment_url` or `wp_get_attachement_image_src`) and instead goes off and does it's own thing for deriving the URL.

This works fine in what I would expect to be the majority of use-cases, however where other plugins are used to modify the location of files (in my case, moving uploaded files to an S3 bucket and then removing them from the server), this no longer works.

I suppose this is partially an issue with [the other plugin](https://wordpress.org/plugins/amazon-s3-and-cloudfront/), however they correctly apply a filter to the result of the above core method calls that would otherwise allow it to play nicely with other plugins. i.e. Calling `wp_get_attachment_url($id)` shows the Amazon S3 URL, not the (incorrect) location on the server.

Is there a reason why you went with a custom implementation over using a core method?

As a workaround for my case (and would work for anyone else affected by this issue), I just sub-classed `TimberImage` and overrode the `get_src` method:

``` php
<?php

namespace MyProject\PostType;


class Image extends \TimberImage {
    /**
     * @var int
     */
    private $iid;

    /**
     * @param int $iid
     */
    public function __construct($iid) {
        $this->iid = $iid;

        parent::__construct($iid);
    }

    /**
     * @param string|array $size - See http://codex.wordpress.org/Function_Reference/wp_get_attachment_image_src
     * @return array|bool|string
     */
    public function get_src($size = 'full') {
        return wp_get_attachment_image_src($this->iid, $size)[0];
    }
} 

```

I'm happy to put a PR up to correct this, but as I don't fully understand the reasoning behind the current implementation I'm reluctant to do so without hearing from you first.

Thanks (and thanks for the library, it's awesome!) :)

tmoitie:
Even with the change suggested here, the image resize filters aren't compatible with the S3 plugin either. It would be better if TimberImage was refactored to use the more traditional WordPress APIs for managing files.

bryanknauber:
I'm having this same issue (with, say, the URL returned by `post.thumbnail` not being the version on S3) – but I'm a bit too new to get subclassing `TimberImage` with @TomSeldon's code right. Most of my structure is still the starter theme. Any tips?

j:
[...]

PR #1320, opened 2017-02-10
ref #429 -- fixed server location to build-in filter fix

jarednova:
**Ticket**: #429

#### Issue
#429 and #1131 describe issues where images stored on Amazon S3 do not load from S3 when Timber is activated

#### Solution
Apply a widely-used patch directly to `Timber\Image`: 

```php
// Add this to functions.php
function fix_timber_s3_src( $src, $ID ) {
    $size = '';
    $attachment_src = wp_get_attachment_image_src($ID, $size)[0];
    $fixed_src = $attachment_src ? $attachment_src : $src;
    return $fixed_src;
}
add_filter( 'timber_image_src', 'fix_timber_s3_src', 10, 2 );
```

#### Impact
Makes is cleaner! But also could have fallout based on what's being over-written

#### Usage
None

#### Considerations
This does NOT solve the resizing issues described in #429 

#### Testing
Tests pass locally, no new tests though

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,7 @@
     "asm89/twig-cache-extension": "~1.0"
   },
   "require-dev": {
-    "phpunit/phpunit": "5.*",
+    "phpunit/phpunit": "4.*",
     "satooshi/php-coveralls": "1.0.*",
     "wpackagist-plugin/advanced-custom-fields": "4.*",
     "phpdocumentor/phpdocumentor": "2.*",
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,7 @@
     "asm89/twig-cache-extension": "~1.0"
   },
   "require-dev": {
-    "phpunit/phpunit": "4.*",
+    "phpunit/phpunit": "5.*",
     "satooshi/php-coveralls": "1.0.*",
     "wpackagist-plugin/advanced-custom-fields": "4.*",
     "phpdocumentor/phpdocumentor": "2.*",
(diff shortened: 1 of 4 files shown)


jarednova:
@tvanantwerp can you review and see if this solves the issues for you in #429?

tvanantwerp:
Output of `post.thumbnail.src` and sized variants exactly matches the expected S3 URLs with these changes.

I have no idea how the line deletions in the `src` function will affect any other Timber\Image or Timber\ImageHelper functions, if at all--but I pretty much only ever used `post.thumbnail.src` and wouldn't notice any potential side effects.

---

Not directly related to #429 but when I tested `post.thumbnail.width` or `post.thumbnail.height`, it returns `NULL`. Not something I really need or is within the scope of that issue and the bounty, but thought it worth mentioning. I've used `{{function('getimagesize', post.thumbnail.src)[0]}}` as a workaround in rare cases when I want that info. I know that `wp_get_attachment_image_src` also returns the image dimensions, so I imagine `width` and `height` could be fixed similarly if it's not just me experiencing the problem.

jarednova:
Great, thanks @tvanantwerp! I figure while we're here I'd love to take care of the width/height stuff too, so I'll create a fresh issue + PR to work from on that front

FlickerBean:
This merge results in a php >=5.4 requirement that did not previously exist in the codebase.

<php5.4 now throws

` PHP Parse error:  syntax error, unexpected '[' in /home/www.postalmuseum.org/the-postal-museum-2016/wordpress/wp-content/plugins/timber-library/lib/Image.php on line 444`

Due to 

`$src = wp_get_attachment_image_src($this->ID, $size)[0];`

jarednova:
@davidmaneuver thanks for the heads-up on that; @gchtr submitted a fix in #1330 that I'll be releasing later today

PR #1337, opened 2017-02-27
Fixes incorrect strpos comparison

anugupta:
Should be tested against `false`, not 0

**Ticket**: # 
**Reviewer**: @

### DIFF
--- a/lib/ImageHelper.php
+++ b/lib/ImageHelper.php
@@ -354,7 +354,7 @@ class ImageHelper {
 		);
 		$upload_dir = wp_upload_dir();
 		$tmp = $url;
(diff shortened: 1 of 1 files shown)


jarednova:
Hi @anugupta I think the reason that you're running into problems here is that I'm actually testing to see if the comparison string starts with those various strings (thus being equal to "0"). Is there a better way to write this that would help clarify?

anugupta:
Hey @jarednova - yeah, i'm not sure what I was thinking - the actual issue was that the image filters (eg `letterbox` use a file location, not a URL (has that changed in Timber 1.x from previously). 

Anyway, this PR can be discarded :)

PR #1343, opened 2017-03-03
[Fix] Wrong image path breaking timber resize when using external wp-content-dir (WP_CONTENT_DIR)

gillesgoetsch:
It took me a while to get this one sorted out. This is (probably) happening since the big Timber-Update (from 0.22).

#### Issue
If a wp-setup with an external WP_CONTENT_DIR (ie. site.dev/content instead of site.dev/wp-content) in combination with an external WP_SITEURL (ie. site.dev/wp instead of site.dev/) is used, the analyze_url function creates a faulty path (/site.dev/wp/wp/content/uploads/..) instead of /site.dev/content/uploads..). This breaks the resize function, which therefore will only return the original file src.

#### Solution
In analyze_url we first check, if the site_url is contained in baseurl. If not, the site_url can't be used to build the path and therefore we create the path based on the home_url and the images relative path. 

#### Considerations
Could break the whole image-resize processing if this approach does not fit all cases.

#### Testing
Did test it successfully in the mentioned use case. Needs to be tested on default-directory environments etc. to assure a clean fix.

### DIFF
--- a/lib/ImageHelper.php
+++ b/lib/ImageHelper.php
@@ -366,7 +366,12 @@ class ImageHelper {
 				$tmp = str_replace(WP_CONTENT_DIR, '', $tmp);
 			}
 		} else {
-			if ( !$result['absolute'] ) {
+			// if upload dir does not contain site_url, the content-directory seems to be outside of the site_url
+			// therefore using site_url() would lead to a wrong content/ path
+			if (false === strpos($upload_dir['baseurl'], site_url())) {
+				// use HOME_URL and relative image path
+				$tmp = get_home_url().$tmp;
+			} else if ( !$result['absolute'] ) {
 				$tmp = site_url().$tmp;
 			}
 			if ( 0 === strpos($tmp, $upload_dir['baseurl']) ) {


gillesgoetsch:
Use these constants in your wp-config to reproduce:

```
define('WP_CONTENT_URL', 'http://' . $_SERVER['HTTP_HOST'] . '/content');
define('WP_CONTENT_DIR', $_SERVER['DOCUMENT_ROOT'] . '/content');

define('WP_SITEURL', 'http://site.dev/wp/');
define('WP_HOME', 'http://site.dev/');
```

jarednova:
@gillesgoetsch thanks so much for this PR. Good news! It seems to keep everything working with "default"-style setups. The only last thing I need is a test to cover line [373](https://coveralls.io/builds/10423315/source?filename=lib%2FImageHelper.php#L373) that you added. This way when things change in the future, we know that your scenario won't break. Is that something you'd be able to write?

jarednova:
Thanks @gillesgoetsch — I managed to write a test to cover, thanks for the PR!

gillesgoetsch:
Thank you @jarednova , shout out to Timber 👍 !

oxyc:
On bedrock sites `site_url()` returns `http://domain/wp` which now causes this to always evaluate to false. https://github.com/roots/bedrock/blob/master/.env.example#L11

lib/ImageHelper.php
```diff
@@ -365,7 +365,12 @@ public static function analyze_url( $url ) {
 				$tmp = str_replace(WP_CONTENT_DIR, '', $tmp);
 			}
 		} else {
-			if ( !$result['absolute'] ) {
+			// if upload dir does not contain site_url, the content-directory seems to be outside of the site_url
+			// therefore using site_url() would lead to a wrong content/ path
+			if (false === strpos($upload_dir['baseurl'], site_url())) {
+				// use HOME_URL and relative image path
+				$tmp = get_home_url().$tmp;
```
oxyc: This doesn't actually use a relative path. It just concatenates causing `http://domain.comhttp://domain.com/app/uploads/`

jarednova:
Thanks @oxyc — does the fix in #1404 resolve the issue you see here?

oxyc:
@jarednova yes, that fixed it. I apologize that I didn't write up a proper issue report myself.

fatk:
Had the same issue, working with Themosis framework, this fix resolved it for me too.

Thanks!

Issue #1346, opened 2017-03-08
HTTP error. - Uploading Media Files Windows


DamirPecnik:
Trying to upload image to media library but getting HTTP error.

Whoops \ Exception \ ErrorException (E_WARNING)
preg_match(): Compilation failed: unknown property name after \P or \p at offset 4

See screenshot of the error.
![2017-03-08 09 55 10](https://cloud.githubusercontent.com/assets/10489911/23698132/ab24f888-03ea-11e7-9f5d-3b2fc42a9ac2.jpg)

### What version of WordPress, PHP and Timber are you using?
WordPress 4.7.3, PHP 7.0.13, Timber 1.2.4 - Windows 10 - Mamp Pro

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Part of the Gantry 5 Plugin

@mahagr will have some additional info

mahagr:
Looks like Timber is not properly escaping `$filename`, `$regexdir` and `$ext`, causing regexp to have `F:\P...` in it.

Please use `preg_quote($var, '/')` for all untrusted variables, which are not meant to be regular expressions.

mahagr:
Here is the fix:

```php
	protected static function process_delete_generated_files( $filename, $ext, $dir, $search_pattern, $match_pattern = null ) {
		$searcher = '/'.$filename.$search_pattern;
		foreach ( glob($dir.$searcher) as $found_file ) {
			$pattern = '/'.preg_quote($dir, '/').'\/'.preg_quote($filename, '/').$match_pattern.preg_quote($ext, '/').'/';
			$match = preg_match($pattern, $found_file);
			if ( !$match_pattern || $match ) {
				unlink($found_file);
			}
		}
	}
```

Note that I didn't test it. :)

jarednova:
@mahagr could you submit this as a PR?

mahagr:
There you are. Looks like I did the change against my master, but yeah, I can re-branch that later. :)

jarednova:
Closed by #1348

PR #1348, opened 2017-03-09
Fixed HTTP error when uploading media files in Windows (#1346)

mahagr:
**Ticket**: #1346 
**Reviewer**: @jarednova 

#### Issue
Bad response when uploading media files in Windows

#### Solution
Use `preg_quote()` to escape path to the file.

#### Impact
No impact.

#### Usage
No usage changes.

#### Considerations
Make sure that `preg_quote()` is being used elsewhere in the codebase as well.

#### Testing
The code is hard to unit test in different platforms.


jarednova:
Awesome, thanks @mahagr!

Issue #1362, opened 2017-03-17
TimberImage src function will return NULL if the attachment is PDF


hoandang:
The behaviour was correct in version 1.2.1. However, in the version 1.2.4, it looks like the src function now handles only image attachment and returns false for pdf attachment.

    $attachment = new TimberImage(<image_id>);
    $attachment->src() // return correct path

    $attachment = new TimberImage(<pdf_id>);    
    $attachment->src() // return NULL

jarednova:
PDFs!!!! what will they think of next???

jarednova:
but yes, I agree it should return the correct src — @hoandang, is this something you'd be able to look at and submit a PR?

jarednova:
Closed by #1365

PR #1365, opened 2017-03-19
Fix non-image attachment always returns NULL

hoandang:
**Ticket**: #1362 1362
**Reviewer**: @jarednova 

#### Issue
Fix TimberImage src function return NULL if the attachment is a non-image.


jarednova:
Thanks @hoandang! The code looked good — the one change I made was on the test where...

```php
$this->assertEquals('http://example.org/wp-content/uploads/'.date('Y/m').'/dummy-pdf.pdf', $result);
```

In PHPUnit the first argument should be the expected output and the second arg the value we're testing against

jarednova:
@hoandang thanks for finding the issue and the quick PR — this will be in the next release

PR #1372, opened 2017-03-23
fix doc block to avoid IDE warnings

azeemhassni:
#### Issue
On PHPStorm the `Timber\Timber::render('views/view.twig')` issues a warning `Expected array got string`

#### Solution
Updated doc block to accept string as well


jarednova:
Looks good, thanks @azeemhassni !

PR #1374, opened 2017-03-29
Update Internationalization section in docs

gchtr:
With the new version 2 of Poedit, it’s now easier to internationalize a Timber theme. I updated the documentation accordingly:

- Differentiated between using Poedit 1 and Poedit 2
- Added the HTML attribute caveat to Poedit 1 section (http://stackoverflow.com/a/38834931/1059980)
- Listed internationalization functions that are wrapped by default in [Twig.php](https://github.com/timber/timber/blob/master/lib/Twig.php)
- Applied coding standards to code examples

### DIFF
--- a/docs/wiki/_cookbook-text.md
+++ b/docs/wiki/_cookbook-text.md
@@ -8,8 +8,8 @@ There's tons of stuff you can do with Twig and Timber filters to make complex tr
 
 ```twig
 <p class="byline">
-	<span class="name">By {{post.author.name}}</span>
-	<span class="date">{{post.post_date|date('F j, Y')}}</span>
+	<span class="name">By {{ post.author.name }}</span>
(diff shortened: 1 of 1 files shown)


jarednova:
Thanks so much @gchtr! As an 🇺🇸  I'm always a little behind on Internationalization — great to see these updates!

gchtr:
@jarednova I think it would even make sense to have Internationalization as a separate guide instead of having it inside the Text Cookbook. Does the documentation automatically grab all `.md` files in the docs/wiki folder in alphabetical order or are they referenced somewhere?

I could create the new file and do another pull request then.

jarednova:
@gchtr Yep! It's all auto-magic. I totally agree, please make into it's own doc and send a new PR

PR #1375, opened 2017-03-29
Extract Internationalization into separate guide

gchtr:
As discussed in https://github.com/timber/timber/pull/1374

Use own guide for Internationalization. It makes more sense to cover this topic separately than in the Text Cookbook.

- Added general introduction
- Refactored headings
- Added hint about Twig i18n extension
- Added hint about using sprintf in Twig
- Added general information about generating localization files

A lot of the added changes catch questions I saw on Stackoverflow or special cases I ran into myself.


jarednova:
Awesome, thanks @gchtr! I'll re-deploy the docs with this

PR #1376, opened 2017-03-29
Add post_count to TimberArchives output

tnottu:
**Ticket**: # 
**Reviewer**: @ 

#### Issue
TimberArchive is missing the post count value. It is available with the WordPress native archive function `wp_get_archives` and it is quite useful thing to show in archives.

#### Solution
Added a new `post_count` key to the year/month/week/day (and also postbypost) output of `TimberArchives`. That information was basically already available from the `$wpdb` calls, it just wasn't used in the output.

#### Impact
Shouldn't be a problem with backwards compatibility, since this simply introduces a new key and doesn't change output otherwise. Only place where there could be a performance hit is probably line 172 where the monthly post counts are summed up for the total post count of  the year `array_sum(array_column($months, 'post_count'))` but I doubt that's an expensive thing to do.

#### Usage
`{{ archive_item.post_count }}` and that's about it.

#### Considerations
I set the post count to 1 for the `postbypost` and `alpha` types. But I'm not sure if that makes any sense. Should the key exist at all for those types?

#### Testing
No tests unfortunately. I can't really think of a scenario where this would fail if everything else works right.


jarednova:
Thanks @tnottu! are you able to write a unit test that can verify this behavior? I can do it as well, but pretty backed up right now and might take me a bit of time to get to

tnottu:
I added some unit tests now. I've never written php tests before, so if there's any problems or gotchas, I'd be happy to hear.

I also removed the `post_count` from the `postbypost` and `alpha` types because it didn't make much sense to have it there. No point having a `post_count` for single post items.

jarednova:
Awesome, thanks @tnottu !

Issue #1370, opened 2017-03-22
Twig_Autoloader removed in favor of Composer


curtisghanson:
[Timber/Timber.php](https://github.com/timber/timber/blob/master/lib/Timber.php#L75) checks for the Twig_AutoLoader class. As of Twig 2.0.0, the class has been removed, see [Twig/CHANGELOG](https://github.com/twigphp/Twig/blob/563f6d0de423c9b798d8177a7c74a4ff76939bd9/CHANGELOG#L29). This affects all installations of Timber, see [Timber/composer.json](https://github.com/timber/timber/blob/master/composer.json#L32).

yantei:
quite serious bug, breaks site after composer update command

Twansparant:
Jup same here. Getting fatal error:
```
Fatal error: You have not run "composer install" to download required dependencies for Timber, you can read more on https://github.com/timber/timber#installation in /srv/www/example.dev/current/vendor/timber/timber/lib/Timber.php on line 76
```

exophunk:
Confirming that Twig 2.2.0, Timber 1.3.0 and WP 4.7.3 do not work anymore. Same error as @Twansparant.

Any way to fix this while staying on 1.3.0 ?

jarednova:
Thanks so much for this report, I'll take a look at that ASAP. @curtisghanson — if you comment out lines 75-77, does everything work as expected?

curtisghanson:
Yes. That's how I got around it. I'd submit a pull request but I'm new to this project so I don't want to try and make changes before I know what I'm doing.

jarednova:
@curtisghanson @yantei @exophunk @Twansparant: thanks so much for your flagging and reporting on this issue. I was very confident of the Twig 2.0 and PHP 7 compatibility based on earlier Travis tests passing — but I failed to read the logs which showed that PHP 7 was actually using Twig 1.x despite the composer spec.

Regardless, I've created #1379 that now shows everything working. I'm going to merge this in and release as `1.3.0-beta2` — I'd love any feedback you're able to give about whether it resolves the issues you saw (or creates any new ones). 

Thanks again!

jarednova:
@curtisghanson @yantei @exophunk @Twansparant: I've merged #1379 which resolves the issue here. Give it a whirl and let me know if it works, as highlighted above my next step is to update a beta release and see if anything else shakes out before making 1.3.0 official

PR #1379, opened 2017-03-31
1370/twig class

jarednova:
**Ticket**: #1370

#### Issue
Changes in Twig 2.x broke Timber!

#### Solution
- Test for `Twig_Token` class instead of `Twig_Autoloader` (which was removed in 2.x) and don't add functions to Twig through function wrapper.

#### Impact
- This introduces a risk that users who had done something like....

###### PHP
```php
new TimberFunctionWrapper('my_cool_function');
Timber::render('my-template.twig);
```

###### Twig
```twig
{{ my_cool_function }}
```

... will no longer work. That said, this was never _supposed_ to work. It was totally an un-documented piece of functionality that I didn't realize was in there (and is easily achieved in other ways)

#### Testing
All tests pass! I've also added new tests to ensure key functionality from `FunctionWrapper` still works


gchtr:
@jarednova Do I understand it right that the Timber\FunctionWrapper class will no longer work for exposing functions to Twig?

I ran into `Fatal error: Uncaught Twig_Error_Syntax: Unknown "my_custom_function" function.`, because I use 

```php
\Timber\Helper::function_wrapper( 'my_custom_function' );
```

Apparently the way to go is to use the `timber/twig` filter:

```php
    $twig->addFunction( new \Twig_SimpleFunction( 'my_custom_function', 'my_custom_function' ) );
} );
```

- Shouldn’t it be described as a breaking change in the Release notes?
- Shouldn’t it be updated in the Documentation (writing out how exactly it can be *easily achieved in other ways*, using the `timber/twig` filter) in https://github.com/timber/timber/blob/master/docs/wiki/_wp-integration.md#functions? I guess it wasn’t totally undocumented, because I learned to use it that way.
- Can we provide a deprecation error message through `Helper::warn` for developers?
- Should it be [removed from context](https://github.com/timber/timber/blob/master/lib/Timber.php#L233-L234)? The [proposed way to do it in the starter theme](https://github.com/timber/starter-theme/blob/master/templates/html-header.twig#L21) is with `{{ function('wp_head') }}` anyways.

I can help with the changes and do some pull requests, but it would be cool if you could confirm that I’m on the right track.

And finally, I don’t really get the purpose of the FunctionWrapper class when I can’t expose functions with it. Does is still have it’s purpose. Could we document it better?

jarednova:
Hey @gchtr these are all good points (and thanks for testing out the release candidate — it's great to catch these issues before publishing to WP.org). The big rock I ran into (and why this stuff is removed) is b/c of changes with how Twig 2.0 initializes conflict with how `FunctionWrapper` was registering these functions. That said to answer yr questions...

**Shouldn’t it be described as a breaking change in the Release notes?**
**Shouldn’t it be updated in the Documentation**
**Can we provide a deprecation error message through Helper::warn for developers?**
**Should it be removed from context? The proposed way to do it in the starter theme is with {{ function('wp_head') }} anyways.**
Yes. 100%. Absolutely. Totally. These are all the right things to do and in my overloaded state neglected to do full diligence to thoroughly address these.

If you want to PR any those things, I'd really appreciate it. 

As for the last question: **why does it exist then?**
Originally (like 3 years ago) the only way I knew to store the results of functions that would otherwise output was to use PHP `ob_start`, `ob_flush`, etc. FunctionWrapper was added to handle that in a tidy place. I believe it was later that we introduced adding `{{ function('my_cool_function') }}` directly in a TWIG file. 

It's very possible that I never took a second to ask: does `FunctionWrapper` even need to exist then? I'm not so sure it does (don't worry, I'm not going to suggest we remove it right here and now). But it makes sense to investigate if there are any use cases where a desired outcome could not be achieved by including a function call within Twig.

jarednova:
@gchtr I've opened up #1387 to follow-up on how to resolve this. I really appreciate you flagging this in such a respectful and thorough manner

Issue #1192, opened 2016-09-13
Timber throw exception when trying to convert TIFF to JPEG


chrisgherbert:
### Expected behavior

Timber is able to handle the same image formats that WordPress can, including TIFF. 
### Actual behavior

Timber throws an exception when a trying to convert a TIFF to JPG using the |tojpg filter.  I can understand not being able to do the conversion (not sure if PHP-GD can even do this), but an throwing exception seems like an extreme reaction.
### Steps to reproduce behavior

User uploads a TIFF file (terrible idea, but clients do it), and the site throws an exception when it attempts to convert the image to jpeg in a template.
### What version of WordPress, PHP and Timber are you using?

WordPress 4.6.1, PHP 7, Timber 1.1.5
### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)

Loaded from Composer

jarednova:
Tiffs? Those bastards. Is this something you could look into submitting a PR for? I agree that if WP does it, Timber should too. Hopefully it could follow the same conventions as the PNG tools

chrisgherbert:
I'll check out the offending code and see if I'm comfortable enough with it to submit a PR.

codesman:
@chrisgherbert , I did a little digging into this and I found that `gd` is not capable of converting tiff->jpg(or anything else). ` ImageMagick` is required to do this in PHP.

According to [the codex](https://codex.wordpress.org/Uploading_Files) `.tiff` is not a supported upload format, even though by default they are allowed to be uploaded. A common solution to the problem of users uploading .tiffs is to [remove .tiff from the allowed mimetypes](http://wordpress.stackexchange.com/questions/4142/how-can-i-prevent-uploading-bmp-image) and force the user to convert to a supported format before uploading. 

AFAIK, WordPress core does not do any format conversion (are there official WP themes that convert tiff->jpg in the view?).  Unless Timber adds `ImageMagick` as a dependency & implements it in the `tojpg` filter, it seems that an exception should be thrown to let us know that if we are going to let our users upload unsupported filetypes, that we should handle it and make sure we're not trying to convert an image that cannot be converted.

Perhaps you could check the filetype in your theme and only pass it to `tojpg` if it's a supported format?

I suppose another option would be for Timber to check and see if the filetype is tiff and if it is, do no conversion and return the original tiff, but that seems like a wrong behavior for a `tojpg` filter.

I'm not seeing a clean way to handle this. What do you think?

[...]

PR #1383, opened 2017-04-06
Process ToJPG only if the filetype is supported

mgussekloo:
// First, check if the filetype is a valid one. Processing an invalid filetype
// results in an exception, which is not really useful.
// See issue: "Timber throw exception when trying to convert TIFF to JPEG #1192"

**Ticket**: #  #1192 
**Reviewer**: @ 

#### Issue

Timber throw exception when trying to convert TIFF to JPEG #1192

#### Solution

 First, check if the filetype is a valid one


jarednova:
Thanks @mgussekloo! I just updated the test to handle the result appropriately. Letting those run (and hopefully pass) before merging in...

jarednova:
Merged! Thanks again @mgussekloo (and congrats on your first contrib!)

PR #1393, opened 2017-04-20
1387/Do not use function_wrapper where we don’t have to

gchtr:
Instead of wrapping `language_attributes`, which echoes what `get_language_attributes` returns, we can directly use `get_language_attributes`.


jarednova:
awesome, that helps!

Issue #1387, opened 2017-04-14
Resolve FunctionWrapper regressions


jarednova:
From #1379, I think there are two possible routes here

### Make it work with Twig 2.0
- [ ] Adjust code to make it compatible with Twig 2.0 and maintain old functionality

### It won't work with Twig 2.0 so...
- [x] Describe outcome of #1379 as a breaking change in the Release notes
- [ ] Update documentation to reflect how `FunctionWrapper` should be used or how to include functions
- [x] Provide a deprecation error message through `Helper::warn` for developers
- [x] Remove from default context

gchtr:
@jarednova Thanks for the explanation in https://github.com/timber/timber/pull/1379. I guess the discussion continues here now.

Hopefully I’ll find some time this week to think about a good solution and prepare some pull requests. With a quick search here on Github I found that there are a number of themes that use `function_wrapper` as it is suggested in the documentation.

jarednova:
Closed by #1391

PR #1394, opened 2017-04-21
1387/documentation

gchtr:
**Ticket**: https://github.com/timber/timber/issues/1387
**Reviewer**: @jarednova 

#### Issue

Update documentation to reflect how FunctionWrapper should be used or how to include functions.

#### Solution

- Remove deprecated `wp_head` and `wp_footer` entries in _cheatsheet.md and other references in _getting-started-themeing.md
- Create new guide "Functions" to describe:
    - How functions can be used in Twig, without and with arguments.
    - How functions can be exposed to Twig
    - A small note about the function_wrapper deprecation
    - A section about functions that echo output and how to handle these.
- Remove "Scripts + Styles" section, because `wp_head` and `wp_footer` are handled through `{{ function() }}`. Instead, use `wp_head` and `wp_footer` as the primary example in the Functions guide.
- Harmonize heading levels
- Apply WordPress coding standards to code examples

Please review this carefully, as I may have made some assumptions that might not be exactly correct.

I thought about the use case of `FunctionWrapper` and also saw `Helper::ob_function`, and I asked myself: What’s the difference? Am I right with the following assumptions, that would be inserted into the documentation with this pull request?

- `Helper::ob_function` is used to bypass the output and instead save it as a string for functions that don’t depend on stuff like globals.
- If you have a function that needs to be called exactly where you use it in your template (e.g. because it depends on certain global values) `FunctionWrapper` is a better solution.

If `FunctionWrapper` has it’s own special use case, does it maybe need to be added to https://github.com/timber/timber/blob/master/bin/generate-docs.sh#L27?

### DIFF
--- a/docs/wiki/_cheatsheet.md
+++ b/docs/wiki/_cheatsheet.md
@@ -1,11 +1,11 @@
 # Cheatsheet
 
-Here are some helpful conversions for functions you're probably well familiar with in WordPress and their Timber equivalents. These assume a PHP file with the `Timber::get_context();` function at the top. For example:
+Here are some helpful conversions for functions you’re probably well familiar with in WordPress and their Timber equivalents. These assume a PHP file with the `Timber::get_context();` function at the top. For example:
 
 ```php
 $context = Timber::get_context();
 $context['post'] = new TimberPost();
-Timber::render('single.twig', $context);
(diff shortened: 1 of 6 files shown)


jarednova:
@gchtr thank you thank you thank you for updating all of this — mergin!

PR #1402, opened 2017-04-30
[PERF Optimization] Add optimize-autoloader to composer install

jakub-klapka:
Hello,
I suggest this little change, which has noticable performance improvement without any downsides. With that, I was able to save around 130 file_exists calls on quite small project. This is especialy helpful on Windows platforms.

More on composer optimization: https://getcomposer.org/doc/articles/autoloader-optimization.md

Thanks for great library! Jakub


jarednova:
Awesome! I had no idea about that flag, pulling in now! (Also really glad this helps Windows users)

Issue #1056, opened 2016-06-08
WP_CONTENT_SUBDIR is wrong (causing `get_rel_path` to return an URL)


haroldangenent:
### Expected behavior

`WP_CONTENT_SUBDIR` should be `/app`.
### Actual behavior

`WP_CONTENT_SUBDIR` was defined as an URL: `http://www.example.com/app` on subdirectory installs (Multisite).
### Steps to reproduce behavior
1. Place `wp-content/` contents in another folder (for example, when [Giving WordPress Its Own Directory](https://codex.wordpress.org/Giving_WordPress_Its_Own_Directory))
2. Create a subdirectory website
3. Check `WP_CONTENT_SUBDIR`, defined in `TimberImageHelper::add_constants` (Timber 0.22.5) for the subdirectory install

In my case, this causes `TimberImageHelpers::sideload_image` to return an URL instead of a path, **prefixed with a slash** (because of `TimberURLHelpers::preslashit()`).
### What version of WordPress, PHP and Timber are you using?

WordPress 4.5.2 **Multisite**, PHP 7, Timber 0.22.5. Using a [bedrock](https://github.com/roots/bedrock)-type setup.

Problem persists in Timber `dev-master`.
### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)

Composer, via WPackagist.

connorjburton:
Thanks for this, will investigate.

haroldangenent:
This might be fixed when using `untrailingslashit(network_home_url())` instead of `home_url()` on Multisite, since uploads are served from the main website on subdirectory installs. However, `sideload_image` returns a relative path when the file exists. So the output is different the first time a file is sideloaded (you'll get a full URL). Returning a relative path works poorly on Multisite, since subdirectory websites wouldn't work (`www.example.com/sub/wp-content/uploads/...`).

_Notice `untrailingslashit()` on `network_home_url()`. While the [documentation](https://codex.wordpress.org/Function_Reference/network_home_url) claims a lack of trailing slash, it seems to return an URL **with** a trailing slash._

palmiak:
@haroldangenent is this fixed? I see that some related PR was pushed in 2017 so it may fixed it.

palmiak:
Thanks. I'll try to take a look soon. On the other hand I see you have made some research - maybe you will try to make a PR?

haroldangenent:
[...]

PR #1404, opened 2017-05-03
Custom WordPress directory image resize

domtra:
#### Issue

Unfortunately pr #1355 broke the image resizing functionality for us. This is already mentioned by some other people here.
Our setup uses bedrock with its custom wordpress directory structure. WordPress itself is located in a subdirectory `/wp`, the content folder is in `/content`. 
The problem is, that in the previous fix, not only relative, but also absolute urls were prefixed with the `home_url`. This resulted in a url something like `http://example.orghttp://example.org/some-image.ext`.

#### Solution

Revert the previous fix and use `network_home_url` instead of site url. This url needs to be untrailinsgslashed for some reason. Maybe this can help #1056 as well.

#### Considerations
Unfortunately all different scenarios are quite hard to test and would make the test suite even more complex. This solution seems to be working for all the previous test cases and definitely works with bedrock single site install and multi site install with subdirectories. I haven't tested multi site with subdomains yest.

#### Testing
I removed the runkit dependecy that was introduced by #1355. The same behaviour can be achieved by using filters. Then I added another test for relative urls because I believe that caused an issue before.

I tried to make fine grained commits so that the approach is easier comprehensible. First break the test, add a new one and then fix it.

The new test from #1355 could not be run in isolation locally for me. It failed, or rather only succeeded if some other test ran before it. Unfortunately there were some other issue with this interdependency. E.g. files, db entries, etc created in other test. That is why a added a different image size for the relative url test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,10 +18,6 @@ matrix:
           env: WP_VERSION=4.6 WP_MULTISITE=0
 
 before_script:
-    ## Runkit
-    - pecl install runkit
-    - if [ "$TRAVIS_PHP_VERSION" == 7.1 ]; then git clone --depth=1 git://github.com/runkit7/runkit7.git runkit; cd runkit; phpize; ./configure; make; make install; cd ../; phpenv config-add tests/travis.php.ini; fi
-    ## Other setup
     - if [ "$TRAVIS_PHP_VERSION" == 5.5 ] || [ "$TRAVIS_PHP_VERSION" == 5.4 ] || [ "$TRAVIS_PHP_VERSION" == 7.1 ]; then autodetect | pecl install imagick; fi
     - if [ "$TRAVIS_PHP_VERSION" == 5.3 ]; then autodetect | pecl install imagick-3.0.1; fi
     - bash bin/install-wp-tests.sh wordpress_test root '' localhost $WP_VERSION
--- a/lib/ImageHelper.php
+++ b/lib/ImageHelper.php
@@ -365,13 +365,8 @@ class ImageHelper {
 				$tmp = str_replace(WP_CONTENT_DIR, '', $tmp);
 			}
 		} else {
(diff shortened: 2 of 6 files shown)


jarednova:
@domtra thanks for such a thorough PR — I'll take a look this week and do some tests of my own...

jarednova:
Again, thanks so much @domtra. I opened #1408 to accomodate one small tweak/question I had. I noticed the old tests were using `$_SERVER['DOCUMENT_ROOT']` as the value for `option_upload_path`. In your tests it looks like you changed it to `ABSPATH` — what's the methodology for using one over the other?

domtra:
Hi @jarednova. Thanks for looking into this. You are right. `$_SERVER['DOCUMENT_ROOT']` does fit better in that case. I tried to avoid it, because it is not in your control usually. I have run into issues where this value did not resolve to the actual docroot. This was mostly in shared hosting environments. In general I would not recommend using something like this, but always hardcode the exact path, via a environment variable for example.
Since the test are run in an artificial environment and you control it, it is probably the correct way to go. In the tests it was just a coincident that both of these match, because wordpress is actually located in the root folder, where the `wp-config.php` lives and `ABSPATH` is defined.
So from my side all is good with your change.
Hope to see this in a release soon since we have not been able to update to `1.3.0` yet to use the latest features/bugfixes.

jarednova:
Awesome, thanks @domtra — I'll merge in #1408 with these commits

ruscon:
@jarednova 
The same problem.
Can you tag it ?

jarednova:
@ruscon tagged! 1.3.3

ruscon:
@jarednova  Thank!
P.S. and wp plugin too ? :)

jarednova:
@ruscon I'm going to push that out Wednesday (7/5) — (I like to do those during the week to make sure I'm around to support any hot fixes that need to result it going to ~20k WP sites)

PR #1414, opened 2017-05-10
Return correct menu_id when using WPML and 'theme_location' arg

mikeyb31:


jarednova:
Excellent, thanks @mikeyb31! you're going to be a hero to a lot of users. Before I merge this in I wanted to see if you could look into writing a unit test. That way we can ensure that any future modifications to the code won't break what you've added. Rather than including all of the WPML plugin, we can potentially mock-in the response this function gives to test that functionality

jarednova:
Thanks again @mikeyb31. I wrote a quick test and merged this in with #1437

mikeyb31:
You're welcome @jarednova - apologies for not replying, I've been deep in other project work and wasn't too sure how you wanted me to approach the mock test.

jarednova:
@mikeyb31 no prob!

PR #1420, opened 2017-05-17
Added check to see if source exists

mgussekloo:
Prevents too many errors in the log for every lost file.

**Ticket**: # 
**Reviewer**: @


jarednova:
Looks good, thanks @mgussekloo (and sorry for the slow response, I just got back from a long GitHub-free vacation)

Issue #1412, opened 2017-05-05
Timber method get_widgets - FATAL ERROR


dawidryba:
I have problem with new Timber with Twig 2.0. When i try get widgets with code:
```
$this->context['widget_footer'] = Timber::get_widgets('widget-footer');
$this->context['widget_phone'] = Timber::get_widgets('widget-phone');
```

i got error:
```
 Fatal error: Uncaught LogicException: Function "dynamic_sidebar" is already registered. in /home/dawidryba/projekty/praca/sowdent.dev/wp-content/themes/krframe/krFrame/vendor/twig/twig/lib/Twig/Extension/Staging.php:30 Stack trace: #0 /home/dawidryba/projekty/praca/sowdent.dev/wp-content/themes/krframe/krFrame/vendor/twig/twig/lib/Twig/ExtensionSet.php(162): Twig_Extension_Staging->addFunction(Object(Twig_SimpleFunction)) #1 /home/dawidryba/projekty/praca/sowdent.dev/wp-content/themes/krframe/krFrame/vendor/twig/twig/lib/Twig/Environment.php(807): Twig_ExtensionSet->addFunction(Object(Twig_SimpleFunction)) #2 /home/dawidryba/projekty/praca/sowdent.dev/wp-content/themes/krframe/krFrame/vendor/timber/timber/lib/FunctionWrapper.php(70): Twig_Environment->addFunction(Object(Twig_SimpleFunction)) #3 /home/dawidryba/projekty/praca/sowdent.dev/wp-includes/class-wp-hook.php(298): Timber\FunctionWrapper->add_to_twig(Object(Twig_Environment)) #4 /home/dawidryba/projekty/praca/sowdent.dev/wp-includes/plugin.php(203): WP_Hook->apply_ in /home/dawidryba/projekty/praca/sowdent.dev/wp-content/themes/krframe/krFrame/vendor/twig/twig/lib/Twig/Extension/Staging.php on line 30
```

If i comment one get_widgets then work fine.

I check method get_widgets before update and difference is:

OLD:
```
public static function get_widgets( $widget_id ) {
      return trim(Helper::function_wrapper('dynamic_sidebar', array($widget_id), true));
}
```
NEW:
```
public static function get_widgets( $widget_id ) {
		$output = new FunctionWrapper( 'dynamic_sidebar', array( $widget_id ), true );
		return trim( $output );
}
```

danimbrogno:
[...]

PR #1422, opened 2017-05-17
1412/fix-get-widgets

gchtr:
**Ticket**: #1412
**Reviewer**: @jarednova 

#### Issue
When `Timber::get_widgets()` is called more than once, a fatal error occurs:

```
Fatal error: Uncaught LogicException: Function "dynamic_sidebar" is already registered.
```

As @danimbrogno mentioned, if the function is being registered twice in Twig, it causes a `LogicException` to trigger. Because that exception is not catched, a fatal error is thrown.

#### Solution

By wrapping `$twig->addFunction` into a try/catch block, we can ignore this exception. Although there might be multiple registered `FunctionWrapper` instances with the same function name, it doesn’t get added to Twig.

But! We don’t have to use FunctionWrapper for `Timber::get_widgets()`.

I changed `get_widgets()` method to use `Helper::ob_function` instead of `FunctionWrapper`. Because `get_widgets()` is normally called when we build up our context (http://timber.github.io/timber/#widgets), we can use `ob_function`. As I see it, there would be no easy way to call `get_widgets()` from inside a Twig template anyways, so I guess this is a safe way to go?

#### Considerations

I’m still thinking about a use case for `FunctionWrapper`, and I can’t find one yet. Especially if support for using FunctionWrapper to make a function available to Twig will be removed in version 1.4. What will be the difference to just using `Helper::ob_function` then? Maybe I still don’t get the concept behind it.

#### Testing

Manual test with two registered sidebars:

```
$context['sidebar1'] = Timber::get_widgets('sidebar-1');
$context['sidebar2'] = Timber::get_widgets('sidebar-2');
```

### DIFF
--- a/lib/FunctionWrapper.php
+++ b/lib/FunctionWrapper.php
@@ -38,7 +38,7 @@ class FunctionWrapper {
 			if ( (is_string($function[0]) && class_exists($function[0])) || gettype($function[0]) === 'object' ) {
 				$this->_class = $function[0];
 			}
-			
+
 			if ( is_string($function[1]) ) {
 				$this->_function = $function[1];
 			}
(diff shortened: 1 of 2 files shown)


jarednova:
Thanks @gchtr! I added a test and merged this in under #1438

Issue #1429, opened 2017-05-23
Errors when clearing twig or timber cache using WP_CLI command (wp timber cache_clear_*)


vyarmolenko:
WP_CLI commands for clearing only twig or only timber cache fail to write output to console:

![image](https://cloud.githubusercontent.com/assets/3118280/26351291/bbfb3760-3fbf-11e7-8cdb-c81531afc37c.png)

### Steps to reproduce behavior
Running **wp timber cache_clear_twig** or **wp timber cache_clear_timber** produces error.

### What version of WordPress, PHP and Timber are you using?
WordPress 4.7, PHP 5.6.30, WP_CLI 1.2.0-alpha, Timber 1.2.4

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Timber and WP_CLI were installed using Composer.

palmiak:
@vyarmolenko does this bug still exists? Because I see something merged.

vyarmolenko:
@palmiak it's resolved, closing.

columbian-chris:
Looking at this almost six years later, try `wp timber clear-cache`.

PR #1430, opened 2017-05-23
Update Timber_WP_CLI_Command.php

vyarmolenko:
This change addresses https://github.com/timber/timber/issues/1429.

**Ticket**: # 
**Reviewer**: @ 

#### Issue
https://github.com/timber/timber/issues/1429

#### Solution
Fixed WP_CLI call.
Unified calls and produced result in methods for cache clearing.

#### Impact
The proposed change may impact users/software performing WP_CLI cache clearing commands.
After this update `wp timber cache_clear_twig` and `wp timber cache_clear_timber` should work as expected.
Before this update `wp timber cache_clear` didn't produce any WP_CLI output.
After this update it will report success only when both twig and timber content are cleared, and warning otherwise.

#### Usage
No usage changes required.

### DIFF
--- a/lib/Integrations/Timber_WP_CLI_Command.php
+++ b/lib/Integrations/Timber_WP_CLI_Command.php
@@ -19,7 +19,13 @@ class Timber_WP_CLI_Command extends \WP_CLI_Command {
 	 *
 	 */
 	public function clear_cache( $mode = 'all' ) {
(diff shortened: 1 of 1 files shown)


jarednova:
@vyarmolenko thank you thank you thank you. Fixing the WP-CLI stuff has long been in my backlog, this is a great fix!

PR #1439, opened 2017-05-31
Enable Post::convert() to convert also single WP_Posts

luism-s:
#### Issue
When a developer tries to use get_field on a relationship or repeater field, Timber returns an array with instances of TimberPost. However, that didn't happen for Post Object fields, and a WP_Post was returned instead.

#### Solution
In the convert function used by get_field, a if condition was added to check if the argument given is an instance of WP_Post (and not an array of course), and returns an instance of TimberPost.

#### Impact
None that I'm aware of.

#### Usage
Ability to use get_field() on a Post Object field and get a TimberPost

#### Considerations
None

#### Testing
````
{% set post_obj = post.get_field('single_post') %}
{{ post_obj.thumbnail.src.('medium) }}
````
The last line would fail because `post_obj` wouldn't be a TimberPost, but a WP_Post

### DIFF
--- a/lib/Post.php
+++ b/lib/Post.php
@@ -1186,11 +1186,13 @@ class Post extends Core implements CoreInterface {
 
 	/**
 	 * Finds any WP_Post objects and converts them to Timber\Posts
-	 * @param array $data
+	 * @param array|WP_Post $data
 	 * @param string $class
 	 */
 	public function convert( $data, $class ) {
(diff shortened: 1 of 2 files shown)


jarednova:
Awesome, thanks @luism-s! And with a test! I'm just re-running tests based on the newest master then I'll merge this in!

PR #1441, opened 2017-06-06
Improve documentation for render and compile functions

gchtr:
- Add descriptions for parameters and return values.
- Apply [WordPress Inline Documentation standards](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/).
- Add examples.
- Improve type hinting for IDEs.

#### Considerations

I don’t know what the benefit of the `fetch` function is, if the only difference to the `compile` function is that it applies an additional filter `timber_compile_results`. Is this for backwards compatibility?

### DIFF
--- a/lib/Loader.php
+++ b/lib/Loader.php
@@ -87,8 +87,11 @@ class Loader {
 	}
 
 	/**
-	 * @param array $filenames
-	 * @return bool
+	 * Get first existing template file.
+	 *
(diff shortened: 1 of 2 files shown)


jarednova:
Excellent, thanks @gchtr! Good question on fetch: If I recall, it's a left-over from when it used to do more things which have now been moved to compile. I think I was loathe to totally delete it b/c it was officially marked "public" and has that last filter on it. 

Give that it's doing a whole lot of nothing, I'd be good deprecating it for a future v

PR #1445, opened 2017-06-09
fix urls of images when using WPML with language in directory setting

njbarrett:
**Ticket**: # 
**Reviewer**: @ 

#### Issue
When using WPML with a custom wp content directory (bedrock), images URLs are being returned like `https://example.org/us/app/uploads/sites/2/2017/06/599a117bad71440a5b4f98e3807afc70.jpg`

Note the `/us` in the URL. The correct location is not to have the `/us` in the URL, like this:

`https://example.org/app/uploads/sites/2/2017/06/599a117bad71440a5b4f98e3807afc70.jpg`

#### Solution
I attempt to filter our the language code from the URL if it set in the function `file_system_to_url`

#### Impact
Should all tests pass, there should be no impact.

#### Usage
Usage should be transparent and backwards compatible.

#### Testing
I noticed there is a test already for `testWPMLurlRemote`, and `testWPMLurlLocal` - perhaps these don't work? Or don't work for installs where wordpress is installed differently.

### DIFF
--- a/lib/URLHelper.php
+++ b/lib/URLHelper.php
@@ -166,8 +166,13 @@ class URLHelper {
 	 */
 	public static function file_system_to_url( $fs ) {
 		$relative_path = self::get_rel_path($fs);
-		$home = home_url('/'.$relative_path);
-		return $home;
+		$home_url = home_url();
+
+		if (defined('ICL_LANGUAGE_CODE')) {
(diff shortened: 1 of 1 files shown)


jarednova:
@nickbarrettjb — thanks so much! I took your inspiration and worked through a few more scenarios and found a more robust for that place to live. The results are in #1447

PR #1447, opened 2017-06-12
1445/wpml urls

jarednova:
**Ticket**: #1445


njbarrett:
Thanks @jarednova , your solution looks much more robust.. which can only come from having good knowledge of the library I guess ;)

PR #1453, opened 2017-06-14
Add known arguments to avatar filter

xavivars:
**Ticket**: # 
**Reviewer**: @ 

#### Issue
Known arguments for avatar (size, default) are not passed to the filter, so they are missed by the functions hooked into the filter (for example, returning an image of the wrong size).

#### Solution
Add known args  to filter method.

#### Impact
Small, and bringing behaviour up to expectation.


jarednova:
Lookin' good, thanks @xavivars !

PR #1457, opened 2017-06-19
Remove dead code from test-timber-loader

matteodem:
#### Issue

Our hoster notified us about possible security issues with the `open_basedir` override (`Atomicorp.PHP.Bypass.OpenBaseDir`). It's commented out so removing the code doesn't change anything. There's also no reason that the code should be there in the first place.

#### Solution

Removing dead / commented out code.

#### Impact

No notifications of virus scanning softwares about vulnerabilities in the test-timber-loader file.


jarednova:
Thanks @matteodem. You're right, that def. doesn't need to be there — good riddance!

Issue #1417, opened 2017-05-15
WPML issue - Fatal Error: Argument 1 passed to Twig_Environment::addFunction() must be an instance of Twig_Function, instance of Twig_SimpleFunction given


luism-s:
*Here we go again*

There seems to be a issue that is revealed when WPML Translation Management is installed while using Timber 1.3. A fatal error is shown saying that instead of `Twig_SimpleFunction`, `Twig_Function` should be used in several lines of code. 

I then tried replacing the code in FunctionWrapper: 68 and all occurrences in Twig.php, then the site was working again.

Can you check this and try to use Twig_Function in those files?

### Expected behavior
Expected to open the website I'm developing without any errors

### Actual behavior
```
Fatal error: Uncaught TypeError: Argument 1 passed to Twig_Environment::addFunction() must be an instance of Twig_Function, instance of Twig_SimpleFunction given, called in /var/www/project/wp-content/themes/project/vendor/timber/timber/lib/FunctionWrapper.php on line 70 and defined in /var/www/project/wp-content/themes/project/vendor/twig/twig/lib/Twig/Environment.php:805 
Stack trace: 
#0 /var/www/project/wp-content/themes/project/vendor/timber/timber/lib/FunctionWrapper.php(70): Twig_Environment->addFunction(Object(Twig_SimpleFunction)) 
#1 /var/www/project/wp-includes/class-wp-hook.php(298): Timber\FunctionWrapper->add_to_twig(Object(Twig_Environment)) 
#2 /var/www/project/wp-includes/plugin.php(203): WP_Hook->apply_filters(Object(Twig_Environment), Array) 
#3 /var/www/project/wp-content/themes/project/vendor/timber/timber/lib/Twig.php(219): apply_filters('timber/twig', Object(Twig_Environment)) 
#4 /var/www/project/wp-includes/class-wp-hook.php(298): Timber\Twig->add_timber_filters(Object(Twig_Environment)) 
#5 /var/www/ver in /var/www/project/wp-content/themes/project/vendor/twig/twig/lib/Twig/Environment.php on line 805
```

### Steps to reproduce behavior
* Install Timber v1.3.2
* Install Sitepress-Multilingual-CMS plugin (WPML) v3.7.0
* Install WPML Translation Management v2.3.0

### What version of WordPress, PHP and Timber are you using?
[...]

PR #1464, opened 2017-06-25
Replace Twig_SimpleFunction and Twig_Function with Timber_Function by extending both classes depending on Twig version

luism-s:
#### Issue

#1417

tl;dr: A fatal error was thrown by activating a WPML Translation Management plugin stating that `Twig_Function` was expected instead of `Twig_SimpleFunction`.

This only happens when timber is installed via composer in the theme folder. So the cause could be that the file that was extending `Twig_Function` into `Twig_SimpleFunction` (in Twig code) was loaded after all the plugins.

Another point is that Twig_Simple_Function [is getting deprecated in Twig 2.x](https://twig.sensiolabs.org/doc/1.x/deprecated.html), which Timber uses.

#### Solution
Search & Replace for `Twig_SimpleFunction` to `Twig_Function`.

#### Impact
Users should now be using `Twig_Function`, instead of `Twig_SimpleFunction`
Nothing that I'm aware of. The class named `Twig_SimpleFunction`that was extending `Twig_Function` was empty, so I guess there are no problems.

#### Usage
Use  `Twig_Function` instead

#### Considerations
To protect the ones still using `Twig_SimpleFunction`, we could create a file with the sole purpose of creating a class named  `Twig_SimpleFunction` and extending it from `Twig_Function`

# Update
Due to the discussion on #1478 I had the idea to instead of using `Twig_Function` or `Twig_SimpleFunction`, make a new one called `Timber_Function`, which extends from each one depending on the Twig version the site is running.


luism-s:
oops, Travis failed, but only on PHP5, don't know why

jarednova:
@luism-s thanks! I'm rerunning those PHP 5 tests, it said there was a MySQL error...

luism-s:
@jarednova I made it. Only codecov is failing due to the large number of lines I changed on that file. I covered some lines of code with unit tests, but there's still some to go.

What do you think of this solution?

jarednova:
@luism-s thanks for updating this! Glad to see it works (my confusion on #1478). My only question is whether the naming makes sense. Right now its full path is...

`Timber\Timber_Function`

... because the functionality is really about exposing functions to Twig, I'm wondering if `Timber\Twig_Function` might make more sense?

luism-s:
Hm, OK, I really don't mind your name idea. I can do it later ;)

luism-s:
@jarednova Renaming the function to Twig_Function made the tests in PHP5 fail. Maybe it's a conflict with Twig's Twig_Function? I don'know what I'm doing wrong, honestly

luism-s:
Ok I figured out the problem. Good news, all tests checked 👌 ready to merge

jarednova:
🎉  Hooray 🎉 ! Thanks so much @luism-s — wouldn't have thought that one would be so tough, but it really helps smooth out the Twig 2.0 integration. Thanks for all your work!!!

luism-s:
All I wanted was really to be able to update Timber to the latest version, now I can :) When are you releasing the next release (pleonasm unintended)?

jarednova:
@luism-s I like to do those on Tuesdays (my normal night for working late) — so look out for it tomorrow!

PR #1468, opened 2017-06-28
Update type to wordpress-plugin

sinnwell:
Changed type to wordpress-plugin according to 
https://getcomposer.org/doc/04-schema.md#type
so when using composer "installer-paths" it will installed in plugins.
https://roots.io/bedrock/docs/composer/

**Ticket**: # 
**Reviewer**: @

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,6 @@
 {
   "name": "timber/timber",
-  "type": "library",
+  "type": "wordpress-plugin",
   "description": "Plugin to write WordPress themes w Object-Oriented Code and the Twig Template Engine",
   "keywords": [
     "timber",


chrisgherbert:
Timber isn't really a WordPress "plugin" and will work fine if loaded before WordPress and outside the plugins directory.

xavivars:
@sinnwell: I'd recommend you to use timber-library from wpackagist if you're interested on using Timber as a WordPress plugin. I've been doing that for over a year, and it works like a charm.

https://wpackagist.org/search?q=timber-library&type=any&search=

Having it as a library allows you to include directly as a dependency of your code (i.e. a theme), so you don't need a plugin being active in order for your theme to work.

jarednova:
@sinnwell Thanks very much for the PR, but as @chrisgherbert the GH/Composer method for inclusion is as a standard `vendor/` library. @xavivars solution for WPackagist is perfect though, it's what I use as well when needing to include as a plugin for whatever reason

PR #1472, opened 2017-07-02
Documention - fixed minor misspelling

ahinze1337:
fixed minor misspelling in the documentation part **Adding to Twig**


jarednova:
Thanks for the catch @ahinze1337 !

Issue #1475, opened 2017-07-06
Testing for template existence should be delegated to twig.


heino:
### Expected behavior

The filter 'timber/loader/loader' allows customization to Timber’s default Twig loader. This way it is possible to make alterations of the template paths or even replace the default loader, but such modifications are ignored by Timber during testing for template existence, since Timber does its own testing.

Testing for template existence should be delegated to Twig via Twig_LoaderInterface::exists().

```
--- Loader.orig.php	2017-06-19 16:24:56.000000000 +0200
+++ Loader.php	2017-07-06 22:19:15.000000000 +0200
@@ -111,13 +111,7 @@
 	 * @return bool
 	 */
 	protected function template_exists( $file ) {
-		foreach ( $this->locations as $dir ) {
-			$look_for = $dir.$file;
-			if ( file_exists($look_for) ) {
-				return true;
-			}
-		}
-		return false;
+		return $this->get_loader()->exists($file);
 	}
```
 
### What version of WordPress, PHP and Timber are you using?
Timber 1.3.2

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Not relevant

jarednova:
Thanks @heino — could you submit a PR for this? I'd much prefer to use Twig's magic than the (current) roll-your-own

heino:
I would rather first address a related issue, and suggest an alternative.

The patch would generate some overhead due to multiple calls to get_loader(), which generates a new loader object for each call. This is also the case in get_twig(). Shouldn't the Loader use a singleton pattern?

Also, if I am not mistake template_exists() is called from only choose_template(), which could instead delegate to twig directly, and thus only call get_loader() once.

While at it, the phpdoc related to choose_template() is partially incorrect. According to the doc the method will return the first file that exists. But if none exists, the first file is returned. Wouldn't it be better if the method returned false, if no template file was found? This way Timber::compile() could take relevant action in failure to select template file way earlier.

heino:
An example of an altered compile() based on such boolean return from choose_template():
```
public static function compile( ... ) {
	if ( !defined('TIMBER_LOADED') ) {
		self::init();
	}
	$caller = LocationManager::get_calling_script_dir(1);
	$loader = new Loader($caller);
	$file = $loader->choose_template($filenames);

	$caller_file = LocationManager::get_calling_script_file(1);
	apply_filters('timber/calling_php_file', $caller_file);

	// This is moved up here to allow filtering $file == false
[...]

PR #1476, opened 2017-07-07
Delegate testing template existence to Twig

heino:
**Ticket**: #1475

#### Issue
The filter 'timber/loader/loader' allows customization to Timber’s default Twig loader. This way it is possible to make alterations of the template paths or even replace the default loader, but such modifications are ignored by Timber during testing for template existence, since Timber does its own testing.

#### Solution
Testing for template existence should be delegated to Twig via Twig_LoaderInterface::exists().

#### Impact
- Testing template existence is now handled upstream by Twig, and usage of custom loaders is now possible.
- The protected method template_exists() is marked deprecated (from v1.3.4) in the phpdoc, since it it no longer used internally.

#### Usage
- Choose_template() returns false on failure - instead of the first non-existing template.
- Non-existing templates is no longer passed all the way to Twig's render(), which currently generates an exception (in which only the first template name is mentioned, since Twig knows nothing about the possible array of templates).

#### Considerations
- With the current implementation of get_loader() an extra Twig loader object is created. However this is due to the class not implementing the singleton pattern (get_twig() currently also recreates new Twig loaders on each call).

### DIFF
--- a/lib/Loader.php
+++ b/lib/Loader.php
@@ -25,6 +25,8 @@ class Loader {
 
 	protected $cache_mode = self::CACHE_TRANSIENT;
 
+	private $loader;
+
 	protected $locations;
 
 	/**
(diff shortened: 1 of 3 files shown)


heino:
**The 3 additional commits were not intended to be part of this pull request, but rather a separate on related to the the constant recreation of Twig loader objects...**

heino:
By the way, instead of relying on compile() returning false on failure, compile() could throw an internally created exception that implements \Twig_Error_Loader as a replacement for the current exception thrown by Twig.

This way there would be no minor with backward compatibility (the though thrown exception is not documented in phpdoc, whereas returning boolean is).

Such an exception could be named something like TemplateNotFound, and could include a list of _all_ the given templates, and thus give a more descriptive explanation. Currently only the first template is mentioned by Twig...

jarednova:
Thanks @heino! I merged this in with #1486 w/o those three extra commits. Could you re-sumbit those on a new PR so we can figure out a better solution with the re-creation of Twig loader objects?

PR #1478, opened 2017-07-07
1464/luism s develop

jarednova:
**Ticket**: #1464 

Investigation of failing tests so we can fix 'em!

### DIFF
--- a/docs/wiki/_functions.md
+++ b/docs/wiki/_functions.md
@@ -50,7 +50,7 @@ Nice! Any gotchas? Unfortunately yes. While the above example will totally work 
 
 ## Make functions available in Twig
 
-If you have functions that you use a lot and want to improve readability of your code, you can make a function available in Twig by using `Twig_SimpleFunction` inside the `timber/twig` hook.
+If you have functions that you use a lot and want to improve readability of your code, you can make a function available in Twig by using `Twig_Function` inside the `timber/twig` hook.
 
 ```php
 /**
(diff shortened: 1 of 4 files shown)


luism-s:
`PHP Fatal error:  Cannot instantiate abstract class Twig_Function in /home/travis/build/timber/timber/lib/Twig.php on line 38`

hm... But `Twig_Function` is not an abstract class

jarednova:
@luism-s Ahhhh! I thought it was the `mysqli_connect()` errors that were causing the failure, I missed that line. While `Twig_Function` is not an abstract class in `2.x`, it is in `1.x`...

https://github.com/twigphp/Twig/blob/1.x/lib/Twig/Function.php

That is annoying. How can we make this work w both versions then?

jarednova:
@luism-s the thing that occurs to me is to do something very ugly and hacky. It's so ugly, I'm not gonna propose it and hope you have a better idea instead.

luism-s:
How come you have two versions of Twig in your `composer.json`? Are you expecting devs to already have a Twig installed in their them's before installing Timber, and therefore not forcing the installation of version 2?

luism-s:
I just updated my PR to use only one Twig version (being 2.*).

jarednova:
@luism-s the reason is that Twig 2.x is only compatible w PHP 7+. While this is a wonderful future, that's only about [12% of WP users](https://wordpress.org/about/stats/). So I want to be future-friendly while supporting the other ~88% of users who are on 5.x.

Is there an other way I should be approaching this?

luism-s:
Maybe we could do something like this:

````
/**
 * Some real hacky shit
 */
class Timber_Function extends Twig_Function {

}
````

And then use `Timber_Function` everywhere

**Comment by luism-s** [comment]
shit, it's working

**Comment by jarednova** [comment]
The solution I'm thinking is similar, something like...
```
protected $Function_Class;

function init() {
    $this->Function_Class = '\Twig_SimpleFunction';
    if ( version_compare(\Twig_Environment::VERSION, '2.0.0', '>=') ) {
        $this->Function_Class = '\Twig_Function';
    }
}
```

And then in usage...

```
$twig->addFunction(new $this->FunctionClass('_n', function( $single, $plural, $number, $domain = 'default' ) {
```

luism-s:
Travis is not failing in my PR. This PR isn't updated with the solution

PR #1487, opened 2017-07-22
Fix headings

gchtr:
#### Issue

With the upcoming new documentation that switches from a onepager to single pages, not all headings are on the same page. This means that headings need to have a different hierarchy.

#### Solution

This pull request optimizes the heading structure:

- Fix hierarchy levels.
- Remove headings where not needed, e.g. to announce language before code examples or where they were used for styling purposes.
- Remove links from headings because they break links in the table of contents. Instead, add the links to the text.
- Remove colons from then end of headings.
- Add some more headings to provide a better structure for the document.
- Add empty lines before and after headings.
- Remove manual table of contents from beginning of documents.
- Remove unnecessary whitespace where spotted.

---

This pull request only looks at the heading structure. There are other smaller fixes to the documentation that will follow in separate pull requests.

### DIFF
--- a/docs/getting-started/setup.md
+++ b/docs/getting-started/setup.md
@@ -8,10 +8,10 @@ menu:
 
 ## Installation
 
-#### Via WordPress.org (easy)
+### Via WordPress.org (easy)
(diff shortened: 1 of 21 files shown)


jarednova:
Awesome, thanks for cleaning that up!

PR #1489, opened 2017-07-25
Avoid re-creation of Twig loader objects

heino:
#### Issue
A new Twig loader instance is currently created on every call to get_loader().
Due to changes in #1476 calling choose_template(), which now calls get_loader(), results in creation of a new Twig loader instance.

#### Solution
By moving the creating of the loader instance to the constructor and reducing get_loader() to a simple getter, the same instance is reused.

Since the protected property $locations is now only used internally in the constructor, the property can be removed.

#### Impact
None

#### Usage
None

#### Considerations
Since the 'timber/loader/loader' could replace the Twig loader instance, there could be implemented a sanity check to insure the final loader implements \Twig_LoaderInterface.


jarednova:
closed by #1492

PR #1490, opened 2017-07-27
fix gethost can be empty and generate a warning

ahallais:
**Ticket**: # 
**Reviewer**: @ 

#### Issue
PHP Warning:  strstr(): Empty needle in /web/app/plugins/timber-library/lib/URLHelper.php on line 130
In wp-cli context  self::get_host() can be empty therefore strstr($url, self::get_host() generate a warning

#### Solution
checking that the needle is not empty


jarednova:
Thanks for this catch @ahallais — merging!

PR #1491, opened 2017-07-27
1489/loader tweaks

jarednova:
**Ticket**: #1489 

Some organizational adjustments and changing variable visibility from `private` ==> `protected`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,9 @@
 <div style="text-align:center">
 <a href="http://timber.github.io/timber"><img src="http://i.imgur.com/PbEwvZ9.png" style="display:block; margin:auto; width:100%; max-width:100%"/></a>
-<div>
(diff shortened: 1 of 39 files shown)


heino:
What would be the purpose of exposing the $loader property? The Twig Loader is accessible via get_loader(), and exposing the $loader property makes the class vulnerable to alterations after initialization. Even via get_loader() it is possible to modify the loader, but not replacing it altogether...

heino:
If you rename the new init() method to something like createTwigLoader(), and let it return a newly created Twig Loader object, the controller could set the the $loader property once and for all. This seem to be in line with the purpose of both the class and the new init() method, and would insure both internal integrity and overloading.

jarednova:
Closed by #1492

PR #1492, opened 2017-07-27
Rename init() to create_twig_loader(), and move cache filters to controller.

heino:
The Twig Loader is accessible via get_loader(), and exposing the $loader property makes the class vulnerable to alterations after initialization. Even via get_loader() it is possible to modify the loader, but not replacing the loader object, which could be done via the WordPress hooks already.

Renaming the new init() method to create_twig_loader(), and modifying it to return a newly created Twig Loader object, which the controller stores in the private $loader property once and for all, internal integrity is insured and overloading made possible…

### DIFF
--- a/lib/Loader.php
+++ b/lib/Loader.php
@@ -51,7 +51,7 @@ class Loader {
 			$rootPath = null;
 		}
 		$loader = new \Twig_Loader_Filesystem($paths, $rootPath);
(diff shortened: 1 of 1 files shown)


jarednova:
Thanks @heino! I agree — I was making that variable `protected` out of habit/practice more than any strong methodology — let's merge it in!

PR #1493, opened 2017-07-28
Move Twig functions and filters to an actual Twig extension.

heino:
#### Issue
Timber currently extends Twig via Twig_Environment::addFunction() and ::addFilter(), which is called from WordPress hooks/actions. This way a function or filter is forcefully injected into Twig once and for all, and it cannot be replaced, or even removed (since calling these methods directly on Twig_Environment takes precedence over any registered extension).

#### Solution
By moving the individual functions and filters into an actual Twig extension, overloading becomes possible. Also, it would be possible to extend the extension class, and filter out or replace individual functions/filters. Twig uses extensions this way internally, which enables users to overload Twig’s internal elements, although this is not recommended.

#### Impact
Currently users of Timber are unable to overload functions and filters used by Timber, so there should be no issues with backward compatibility.

Users will be able to override Timber’s functions and filters, just like Twig. According to the manual overloading the built-in elements is not recommended as it might be confusing.

#### Usage
The proposed extension class allows filtering the function and filters via Wordpress hooks. It is also possible to make further manipulations when extension of the class.

#### Considerations
Order matters when overloading via extensions, and last loaded takes precedence.
The currently proposed modification loads the extension in Loader::get_twig(), right after the cache extension is loaded, and would therefore be easily overloaded. Loading the extension via an action hook (like ‘timber/twig/extensions’) should be considered…

#### Testing
Currently no testing has been done beyond verifying that Timber’s functions and filters is available.


jarednova:
Awesome, thanks @heino! I'm really glad you're digging deeper into the Twig side of things. Looks like the current build has errors, but keep at it!

jarednova:
Hi @heino — thanks so much for your work on this and #1500, I'm really excited to dig in. Let me know when you're ready for me to start closely reviewing this. It looks like you're still pushing commits so I don't want to start that process before you're ready

PR #1495, opened 2017-07-28
clean(resize_image): remove unused variables

ahallais:
**Ticket**: # 
**Reviewer**: @ 

#### Issue
unused variables in run_animated_gif method of Resize

#### Solution
remove unused variables in run_animated_gif method of Resize


jarednova:
Thanks @ahallais. You're right, nice of us to get those values but not do anything with them. Merging!

PR #1500, opened 2017-07-30
Major rewrite: New caching mechanism (supports PSR-6/16), and multiple performance related modifications to the Timber, Twig and Loader classes

heino:
## Initial pull request:
- Move code from the Timber::__construct() to more appropriate location in ::init().
- Make the Timber::init_*() methods private and static, since they should not be overloaded, and is called by also static init(), and thus do not have access to an instance.
- Move creation of Twig environment from Loader::get_twig() to new ::create_twig_environment() to prevent repetitive creation of environments.
- Avoid constant recreation of new (Timber) Loader’s in Timber::compile() and ::compile_string() by using internal static variables.
- Move integrity test (related to the 'timber/loader/loader' filter) to constructor to prevent overloading.
- Change the filter ‘timber/twig’ into an action, since replacement of the Twig environment instance is not intended. Manipulation to the object is still possible, as always intended.
- Move call to ‘timber/twig’ filter from Twig::add_timber_filters() to Loader::__construct(), since it is a primary filter. (It would not be called, if the filter ‘add_timber_filters’ were to be removed by a user from Wordpress’ hook chain).
- Calls to the filters 'twig_apply_filters', 'timber/twig/filters', 'timber/twig/functions', 'timber/twig/escapers' and 'timber/loader/twig' is moved from Loader::get_twig() to actions (outside the class), and no called via the ‘timber/twig’ action. This way manipulation of the Twig environment is done via the ‘timber/twig’ action, and the mentioned filters could/should be considered legacy/deprecated. A separate pull request (#1493) suggest moving Timber’s extentions to Timber into a separate extension, which ends Timber’s internal use of the mentioned filters.
- The caller path is no longer mixed into the Twig filesystem loader. This is done by adding introducing a Twig loader chain as then main loader, and adding a separate Twig filesystem loader for the caller path. The paths within this loader is consciously updated, but 
- All references to LocationManager is moved from the Loader class to the Timber class, to make the loader instance unaware of Timber’s internal handling of template locations. This way the Loader class could be used separately.
- The (Timber’s) Loader’s constructor is also changed to receive the primary template location array, instead of the caller directory, which is no handled internally in the Timber class.

## Further changes, since initial pull request
- The filter/action ‘timber/twig’ is now the primary way to extend the Twig environment. Timber uses the filter internally to call various legacy filters/actions.
- New $options array in Timbers creator (can only be used on first object, to imitate singleton pattern, even though multiple instances can be created).
- Reloading of the Twig environment made optional (via experimental option).
- Change of the Twig file loader mechanism made possible (via experimental option).
- Then Twig filesystem loader is optimized to avoid constant recreation of paths array (via experimental option). Full backward compatibility is intended, however the optimizations made backward compatible breaks related to filtering (All paths pass through filters, but the 'timber/loader/paths' filter now receives $dirname paths, and the ‘timber_locations’ filter receives $locations paths. Additions of paths though either filter will be added, but the order will not necessarily be the same at before).
- A supplementary legacy filesystem loader provides full compatibility with current path order and filtering. (This loader is used as default).
- Initial preparation for optional change of default Twig environment and Loader classes, through change of static properties. _(Setter methods are currently unimplemented)._
- New Cache class _(containing caching relevant code from Loader class)._
- The Timber class now handles rendering and creation of Twig environments by itself, instead of relying partially on the Loader class.
- The Loader class is degraded to a legacy forwarder, to preserve backward compatibility. _(features has primarily been split between the Timber class and the new Cache class)._

# Rewrite caching and introduce support for PSR-16

Further updates since original #1500 pull request:
-	**Major rewrite of them caching implementation**, which allow generic **support for any PSR-16 and PSR-6 compliant caching adapters** - through **modular design** with **autoloading of registered adapters**.
-	**New PSR-16 compliant adapters/loaders for each currently supported WordPress caching mechanism**. _(These are not intended to be part of Timber, but instead published separately, since might be of common interest)._
-	**New PSR-16 compliant adapter to the Asm89**. _(Also not intended to be part of Timber, and is proposed to upstream/Asm89)._ Removal of now unused WPObjectCacheAdapter class.
-	Cache implantation specific code moved out of Cache class.
-	Add Composer requirement for ‘psr/simple-cache’, ‘psr/cache’, ‘symfony/cache’
-	Fix internally unimportant error in legacy Loader class.

## Everything should be fully backward compatible.
- **All tests except one pass** (without modifications).
- The failing test relates the change of visibility of the Timber::init_*() methods, and as been removed.


heino:
Reuse of loader objects (incl. Twig environments) is made optional for now (default is disabled).

heino:
Splitting up the loader into a loader chain with multiple loaders (which basically takes care of the static properties $locations and $dirname separately), the data passed through the filters 'timber/loader/paths' and 'timber/locations' no longer contain all paths. I stead 'timber/locations' is feed only the content of $locations, and 'timber/loader/paths' is feed what is generated via $dirname. In other words backward compatibility is no fully intact. However, additions to the arrays, which would be the normal scenario, is unaffected, since additions to the arrays be functional no matter which filter is used.

One test, "TestTimberLoader::testTwigPathFilter" fails, and will continue to fail due to this change...

acobster:
@jarednova can this be merged? Looks like a lot of good stuff in here which could also be helpful for #1793

acobster:
@jarednova I can take a look. It might be too divergent from 2.x to be tractable at this point, but it's worth a shot. I'll spend some time seeing if I can get the tests to pass.

jarednova:
@acobster sounds good! I changed this PR to compare against 2.x to help w that

acobster:
Unfortunately there's just too much I don't understand about the reasoning behind these changes to effectively resolve these merge conflicts, and I don't really have the bandwidth to get into the nitty-gritty. There seem to be some cool ideas here for design improvements, but at the moment this is just too much to reason about. :( We can reopen this if there's interest.

PR #1502, opened 2017-07-31
Documentation: Update composer.json

gchtr:
**Reviewer**: @jarednova 

#### Issue

There are still entries related to the old docs in `composer.json`:

- Link to wiki in `support`. The wiki will be deprecated, so this can be removed. Instead, we should use a link to the new docs.
- Package suggestions in `suggest` that are not needed anymore.
- Entry in `repository` that is not needed anymore.


jarednova:
Awesome, thanks for the cleanup!

PR #1506, opened 2017-08-03
Documentation: clarify limitations of php/python gettext hacks

vslavik:
#### Issue
The workaround to make Poedit extract strings by pretending that Timber files are PHP or Python ones has some limitations: in particular, pretty much any use of quotes prevents extraction from within them. Plenty of people are understandably surprised by this behavior (exhibit [A](https://stackoverflow.com/questions/34243278/poedit-doesnt-extract-string-in-html-tags), [B](https://stackoverflow.com/questions/33910482/xgettext-does-not-extract-string-in-html-attribute), [C](https://stackoverflow.com/questions/38665666/how-to-parse-timber-twig-templates-with-poedit-and-detect-quoted-strings-to-tr) to show a few public ones on SO).

#### Solution
This PR updates documentation on extracting `__` etc. strings to warn about the inability to extract from quoted strings when relying on the workaround hack with pretending that the template is a PHP 
or Python source file.

#### Impact
None.


jarednova:
@gchtr could you take a look at this one? I think you've got the better command on the docs and i18n worlds...

gchtr:
Thanks @vslavik! 🍻

I restructured the section about generating l10n files with Poedit 1.x a little so that the pitfalls appear on top of the list of methods. I think this can’t be stressed enough!

Issue #1461, opened 2017-06-23
`relative` key addition in `upload_dir` filter is slow with Polylang plugin


oxyc:
### What version of WordPress, PHP and Timber are you using?

Wordpress 4.8
PHP 7.1
Timber 1.3.2

### Issue

Calling `get_home_url()` repeatedly in the `upload_dir` filter causes some slow performance with the polylang plugin. I'm not sure why but Polylang is [looping over the stacktrace](https://gist.github.com/oxyc/0afe5215a89eb57ae486edae0d64d1de), checking for whitelisted and blacklisted functions repeatedly on every call to `get_home_url()`.

I guess this could be filed for the polylang plugin also but when I was looking into I didn't actually see the `relative` key being used in Timber? It was added here https://github.com/timber/timber/commit/a55ad978c1a721043e9dbc1dc7ead3a71126e8e6 three years ago. Is it still necessary?

If Timber core isn't using the key, but you do not want to break backwards compatibility in case someone is using it, I propose either caching the value, or moving the closure to a function that can be removed with `remove_filter`.

Before:

<img width="1250" alt="screen shot 2017-06-23 at 00 10 23" src="https://user-images.githubusercontent.com/302736/27467494-738d0688-57a8-11e7-8234-242cbcfb1197.png">

After:

<img width="1245" alt="screen shot 2017-06-23 at 00 10 40" src="https://user-images.githubusercontent.com/302736/27467496-772e0558-57a8-11e7-955c-8c127229edc4.png">

oxyc:
This was fixed.

PR #1507, opened 2017-08-05
Fix polylang performance issue by caching calls to get_home_url()

oxyc:
**Ticket**: #1461

#### Issue

Repeated calls to `get_home_url()` noticeably slows down performance with the polylang plugin. 

#### Solution

Cache the value of `get_home_url()` and also avoid closures as they are difficult to unhook downstream.

#### Impact

On a site I was building earlier with quite a few images page load went from 4.3s to 2.6s. See #1461 for XHProf comparisons.

#### Usage

No.

#### Documentation Requirements:
Check one:
- [x] No usage changes. I've updated the `readme.txt` with a quick summary of the change and my GitHub handle
- [ ] Yes, there's a usage change. I've updated the `readme.txt` with ...
  - [ ] **Change for Theme Developers:** label
  - [ ] A quick summary of the change
  - [ ] New usage instructions

#### Considerations

None except if the `relative` key is actually needed at all for `wp_upload_dir()` calls.

#### Testing

None added.

### DIFF
--- a/lib/ImageHelper.php
+++ b/lib/ImageHelper.php
@@ -29,8 +29,9 @@ class ImageHelper {
 	const BASE_CONTENT = 2;
 
 	public static function init() {
-		self::add_actions();
-		self::add_filters();
+		add_action('delete_attachment', array(__CLASS__, 'delete_attachment'));
+		add_filter('wp_generate_attachment_metadata', array(__CLASS__, 'generate_attachment_metadata'), 10, 2);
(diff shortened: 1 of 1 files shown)


jarednova:
Thanks @oxyc! This is next-up in my queue for review

jarednova:
Thanks @oxyc! I merged this in with some tweaks as #1542, we'll release next week

Issue #1458, opened 2017-06-19
init_with_url() doesn't support wp core in a subdiretory


oxyc:
### Expected behavior

If you init a TimberImage by url it should set the correct `file_loc` so you can retrieve file dimensions.

>  "/var/www/wordpress/web/app/uploads/2015/10/cropped-512x5121.jpg"

### Actual behavior

As `init_with_url()` prefixes the relative path with `ABSPATH`, this breaks on a Bedrock file structure where WP core is in a subdirectory

>  "/var/www/wordpress/web/wp/app/uploads/2015/10/cropped-512x5121.jpg"

### Steps to reproduce behavior

Setup a Bedrock site.
```
{% set image = TimberImage(post.thumbnail.src|resize(400)) %}
<img src="{{image.src}}" width="{{image.width}}" height="{{image.height}}">
```

### What version of WordPress, PHP and Timber are you using?

WordPress 4.8, PHP 7.1, Timber 1.3.2

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)

Composer

Levdbas:
Should be fixed in #2134

PR #1508, opened 2017-08-05
Fix TimberImage\init_with_url() not working in bedrock environment

oxyc:
**Ticket**: #1458
#### Issue

If you init a TimberImage by url it sets the incorrect `file_loc` property using `ABSPATH` which points to the wp core directory.

#### Solution

Analyze `content_url()` and `site_url()`, if different find where their path differs and remove that part from `ABSPATH`.

```
# Bedrock
ABSPATH: /var/www/wordpress/wp
WP_CONTENT_DIR: /var/www/wordpress/app
-> /var/www/wordpress
-> /var/www/wordpress/app/uploads/

# Regular
ABSPATH: /var/www/wordpress
WP_CONTENT_DIR: /var/www/wordpress
-> /var/www/wordpress
-> /var/www/wordpress/app/uploads/

# Subdir
ABSPATH: /var/www/wordpress/subdir/wp
WP_CONTENT_DIR: /var/www/wordpress/subdir/app
-> /var/www/wordpress/subdir
-> /var/www/wordpress/subdir/app/uploads/
```

Some notes:
- Can't use `dirname($_DOCUMENT['SCRIPT_NAME'])` as this needs to work with phpunit as well. Possibly with WP-CLI at some point.
- Need to use the URL functions rather than `ABSPATH` and `WP_CONTENT_DIR` as there's otherwise no way to test it. Unless we somehow could modify `ABSPATH` in these particular tests.

#### Impact

This might cause a performance hit because of `get_root_dir()` being called multiple times.

#### Usage

None.

#### Documentation Requirements:
Check one:
- [x] No usage changes. I've updated the `readme.txt` with a quick summary of the change and my GitHub handle
- [ ] Yes, there's a usage change. I've updated the `readme.txt` with ...
  - [ ] **Change for Theme Developers:** label
  - [ ] A quick summary of the change
  - [ ] New usage instructions

#### Considerations

Might consider caching the result of `get_root_dir()` so improve performance. Also, if this feels too complex maybe just default to `ABSPATH` but add a filter to `get_root_dir()` that allows users to manually specifiy the webroot.

#### Testing

I added tests for the `file_loc` property and normalized the WPML prefix tests so that `ICL_LANGUAGE_CODE` is set and the new WPML integration fixes the URLs.

I'm still not 100% confident in this solution but thought I'd share it at least. If nothing else you can use 75dce024e93674620b82a43c994a8480f5f16aea as a way to test that a solution fixes the issue.


oxyc:
As tests pass locally I assume in Travis `get_root_dir()` for some reason resolves to `/` consequently removing all slashes. I'll look into it again when I have time.

jarednova:
Thanks @oxyc — this is the next one up in my list!

oxyc:
@jarednova not sure how to fix this to be honest :) Maybe you can figure something out. Currently I'm just using a custom Timber\Image subclass and fix `file_loc` on init

jarednova:
Closing as stale!

PR #1513, opened 2017-08-08
update regex

ChrisManganaro:
**Ticket**: # 

#### Issue
When using WPML and a language code is exists in the domain name, e.g. 
Language code: `en`
Domain: `https://enviroment.com/en`

The language code is being stripped out from the beginning of the domain name where it shouldn't be. So URL begin return is `https:/viroment.com`

#### Solution
Improve the regex that removes the language code from the URL, ensure the match is not proceeded by `:/`.

#### Impact
Should all tests pass, there should be no impact.

#### Usage Changes
Usage should be transparent and backwards compatible.

### DIFF
--- a/lib/Integrations/WPML.php
+++ b/lib/Integrations/WPML.php
@@ -13,7 +13,7 @@ class WPML {
 
 	public function file_system_to_url($url) {
 		if ( defined('ICL_LANGUAGE_CODE') ) {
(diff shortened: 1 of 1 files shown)


jarednova:
Thanks @ChrisManganaro. In #1520 I created a test and updated the release notes. Merged!

Issue #79, opened 2013-10-16
Starter theme for WooCommerce


JosefJezek:
Could you make Starter theme for WooCommerce?

rzeka:
Is someone still trying to do this? I've got 2 shops already using Timber, but each template has its own quirks and is modified just for that single shop.
If I can find some spare time, I'll try to adapt templates so they can work with any shop and any settings. Thanks to @asecondwill for those 3 gists. Without them I'd have failed

lmartins:
Im building something that I still need to abstract to something reusable. Mind you, im taking a conservative approach, where most stuff still loads with the standard actions from woocommerce.

bikeatefoucault:
About to get started on some Woocommerce enabled sites and considering using Timber/Twig for them.

 Seems like there's enough people here to help out with a project, does anybody want to start a repo with what they have so far and we can all get cracking on it?

lmartins:
Hey Alan,

I've been building a few WooCommerce projects using Timber and it's brilliant. Having already got some help from the posts above and @jarednova on a few occasions, I might be able to help. Mind you, the first step is to get some experience with building normal wordpress installs using Timber and for that the Wiki is your best choice.  
Once you get accustomed with the way things are structured, then move on into WooCommerce.

Then, you'll have one big decision to make, either you go "woocommerce's way", that is your templates are essentially loading WooCommerce hooks, or you'll try to build everything from scratch. I would advise for the first option, which is what im using. It allows you integrate WooCommerce add-ons without issues, upgrading Woo without issues, while still allowing for template customisation if that's what you're after.

My code right now is structured for child-themes, but i'll try to make a quick repo with sample code for a normal theme with Timber/WooCommerce integration. The code posted above should get you started anyway.

JosefJezek:
See https://press.codes/downloads/maera-woocommerce/

bikeatefoucault:
@lmartins,

Thanks for the response. Everything posted so far should be a big help moving forward.

rbvea:
Looks like @jarednova has a wiki page up, anybody want to put in a couple of pointers? https://github.com/jarednova/timber/wiki/WooCommerce-Integration

lmartins:
I've added some information of how I approach this, including some tips I've got from @jarednova about products context within the loop.

eleirgulen:
I have tried to follow the instructions here: https://github.com/jarednova/timber/wiki/WooCommerce-Integration

[...]

PR #1515, opened 2017-08-08
Start a WooCommerce integration

gchtr:
## Issue

A lot of questions arise on how to best integrate Timber with WooCommerce. The current documentation already provides a lot of information for what the biggest pain points currently are:

- Updating the `$products` global so that product previews in a loop display the correct properties.
- Working with product images and galleries.
- Sorting out which templates to load.

I currently have a few upcoming projects that include WooCommerce, so I thought this would be a good time to tackle problems that people have.

## Solution

An idea that came up was to have develop a [Starter Theme for WooCommerce projects](https://github.com/timber/timber/issues/79). I think this would be a good idea, but there are already a lot of starter themes out there. I think the best case would be if it would be easy to integrate WooCommerce into *any* Timber theme. Because of that, I started developing a new integration for WooCommerce with the following goals:

- Provide an easy way to integrate WooCommerce into a theme developed with Timber.
- Try to stay close to the WooCommerce methodology. This means working a lot with actions, filters and pluggable functions.

I don’t know yet what needs to be in this integration. Of course it shouldn’t recreate all WooCommerce functionality. That’s why I need feedback.

This pull request shows the current state of the integration, which solves the following problems:

### Automatic Twig template selection

When adding WooCommerce to a theme, you can add a `woocommerce.php` file to the theme root where you can render the required theme file. The contents would be something like this:

```php

if ( is_singular( 'product' ) ) {
    $context['post'] = Timber::get_post();

    Timber::render( 'woocommerce/single-product.twig', $context );
} else {
    if ( is_product_category() ) {
        $queried_object = get_queried_object();
        $term_id = $queried_object->term_id;
        $context['category'] = get_term( $term_id, 'product_cat' );
    }

    Timber::render( 'woocommerce/archive.twig', $context );
}
```

Now this integration could provide a default for that. Instead of adding the code above, you could call the `render_default_template()` function and the integration would handle the PHP part, so that you only need to care about Twig files.

**woocommerce.php** (in theme root)

```
\Timber\Integrations\WooCommerce\WooCommerce::render_default_template();
```

Now you would add a `woocommerce/single-product.twig` and `woocommerce/archive.twig` to the folder where your Twig templates lie.

#### Automatic Twig partial selection

WooCommerce allows you to [override templates](https://docs.woocommerce.com/document/template-structure/) by adding files to the `woocommerce/` folder of your theme. By using a filter in `wc_get_template()` that WooCommerce uses to render these template, we can **first check if a Twig template exists**.

So instead of adding a file `woocommerce/single-product/related.php` to your theme, you can directly add a file `views/woocommerce/single-product/related.twig` without adding a PHP file. The arguments that WooCommerce would pass to the template will be available under the `wc` variable. Here’s an example:

**views/woocommerce/single-product/related.twig**

```twig
{% for post in wc.related_products %}
    {% include 'woocommerce/teaser-product.twig' %}
{% endfor %}
```

### Automatically sets the `$products` global

Updates the `$products` global when products posts are looped through with `Timber\PostsCollection`, using a separate `ProductsIterator` class.

If the product posts are already in an array when used in a template, this won’t work. It needs to be an instance of `PostCollection` (which extends `ArrayObject`).

The integration already tries to convert all arrays in arguments passed to templates in `wc` to `PostCollections`. Maybe a filter in Twig will be helpful, too. We’ll have to see.

### New class `Product`

Provides a new `Product` class that extends `Timber\Posts` that makes it easier to work with WooCommerce posts:

- When in a WooCommerce context (when `is_woocommerce()` is true), all posts will automatically be wrapped with this class.
- It sets a new property `$product` that makes the product object available under `{{ post.product }}`. You don’t have to call `wc_get_product()` manually anymore. Example: If you’d want to display product attributes, you could go with `{{ post.product.get_attribute('attribute_name') }}`
- Provides a function `get_product_attribute` for easier access to product attributes. It returns an array of `Timber\Term` objects when the attribute is a taxonomy, otherwise an array with all options instead of the string that `post.product.get_attribute()` would return. Example: `{{ post.get_product_attribute('attribute_name') }}`.

---

## Impact

This approach for an integration would probably not only work for WooCommerce, but also for integrations for other plugins that work in similar ways.

## Usage Changes

Usage changes will be available in the documentation.

## Considerations

- This pull request includes changes requested in #1514.
- This pull request shouldn’t be merged yet.
- Maybe there’s a nicer way to solve all this. Suggestions and help are very welcome!

### Make integration optional

We should consider to make this integration optional. Instead of adding it in `Integrations::maybe_init_integrations()`, calling the integration in functions.php would maybe be fine as well:

```php
new Timber\Integrations\WooCommerce\WooCommerce();
```

By making it optional, it’s also easier to pass configuration options and make it extendable. E.g., if you don’t want to use the `woocommerce` subfolder, but name it `woo`, you could do it like this:

```php
new Timber\Integrations\WooCommerce\WooCommerce( array(
    'subfolder' => 'woo',
) );
```

### Upgrading a theme

WooCommerce shows you in the backend [when you need to update/check template files](https://docs.woocommerce.com/document/fix-outdated-templates-woocommerce/) after an upgrade. This only works if the relevant PHP files are present in the `woocommerce/` folder of your theme. If there are only Twig files present, this won’t work.

## Testing

Not yet.

## Todo

- [ ] Gather common painpoints when using WooCommerce with Timber from developers.
- [ ] Add useful functionality (Twig filters, functions) to the integration.
- [ ] Update WooCommerce Guide in Documentation.
- [ ] Write tests.
- [ ] Resolve TODOs.


pascalknecht:
I would really like to know what the status of this pull-request is. I am really looking forward to having this. The whole global $product variable handling is really a pain in the current setting. Some time `Timber::get_posts` or similar functions are used this global variable breaks. Even just solving this issue for the moment would be really appreciated.

gchtr:
@pascalknecht I decided to add the functionality as a [separate integration](https://github.com/mindkomm/timber-integration-woocommerce). Thanks for triggering me that I should finally [release it into the wild](https://github.com/mindkomm/timber-integration-woocommerce/releases/tag/0.5.1) 😄. That’s why I’m closing this pull request here.

If you only [initialize the integration](https://github.com/MINDKomm/timber-integration-woocommerce/blob/master/docs/installation.md#initialize-integration), it should already handle the product global for you. From there, you might go on and try out other features of the integration. I’d be very happy to get your feedback and opinion on it.

pascalknecht:
@gchtr I've already found that one last week as I was setting up a new WooCommerce Theme. Thank you very much for your integration. I'll give it a try and let's see what happens :) .

Issue #1474, opened 2017-07-06
Implement Carbon Fields Integration


mindfullsilence:
Is this a support or usage question?
No

### Expected behavior
allow carbon fields as replacement for ACF fields. 

Would be great to see this as an integration. I imagine it would be pretty much equivalent to the ACF integration file.

shaneoliver:
Add you sections to the page or global context 

```php
$context['sections'] = carbon_get_the_post_meta( 'your section name' );
```
Loop and include your section templates

```twig
{% for section in sections %}
	{% include "modules/sections/#{section['_type']}.twig" %}
{% endfor %}
```

synapticism:
Since this issue is languishing here... I've got this in an extended post class and it's working fine:

```php
public function get_field( $field_name ) {
    $value = carbon_get_post_meta($this->id, 'crb_' . $field_name);
    $value = $this->convert($value, __CLASS__);
    return $value;
  }
```

I haven't seen a need to override the rest of the ACF-specific methods but it doesn't seem all that tricky.

jarednova:
Thanks @synapticism — are you interested in submitting a PR for this? I think you could follow the patterns in the `Integrations/ACF` class for reference:

https://github.com/timber/timber/blob/master/lib/Integrations/ACF.php

synapticism:
I'll take a look and see if my skills are up to the task!

carstenwawer:
Hi @synapticism - did you have a look? Don't want to steal your possibility to contribute, but i would love to take the task, if dont find time for it!

Need a CarbonFields-Integration in project in some weeks.

xsynaptic:
@carstenwawer I had a stab at it... not entirely sure if it's really this simple (my experience with both CF and Timber is still pretty limited), but there's a PR to have a look at/build on/etc.!

P.S. also made a small username change. Sorry for the at-reply confusion here :)

mindfullsilence:
I used this PR on my last site build. Did great!

For posts and options pages it worked exactly as it should. I didn't use it on menu items so can't speak to its stability in that arena. But everything else seems on the money.

dimitriBouteille:
Hey @mindfullsilence and @jarednova , 

Just out of curiosity, did that ticket make any progress ?
If the ticket is in neutral, I will be able to propose a solution :)

jarednova:
@dimitriBouteille I don't believe it was ever picked up. If you're interested, check out how some of the other integrations have been handled ...

[...]

PR #1535, opened 2017-08-27
Carbon Fields basic integration

xsynaptic:
**Ticket**: #1474

#### Issue

First steps toward an integration with [Carbon Fields](https://carbonfields.net/). Only supports
post meta, term meta, comment meta, and user meta at present, and no
complex fields.

#### Solution

This copies the existing ACF integration class and links a few Carbon Fields functions to their associated Timber filters.

#### Impact

I'm not entirely certain what impact this might have on custom fields made with ACF or vanilla WP.

#### Usage Changes

Some docs for CF users will be helpful once we figure out what final shape this feature might take.

#### Considerations

Here I will add some notes from scanning both the Timber and CF repos. This is about what's not included:
* Complex fields (similar to ACF repeaters as I understand it); this would require an additional parameter to be passed to the CF functions. Possible solution: have Timber users append `:complex` to a field name *e.g.* `get_field('crb_field_name:complex')` and match the string?
* CF menu item meta (as Timber doesn't seem to support fields for menu items or offer the appropriate filter hook; something that shouldn't be too hard to add if desired).
* CF widgets; something I still haven't properly looked into.
* CF theme options; probably out of scope.
* Updating meta; CF can update/set meta values progrmatically but I'm not entirely sure how this works in Timber or how useful or necessary it would be to integrate this.

#### Testing

No tests yet. Edits/tips/comments welcome. I'm new at this 😁

### DIFF
--- a/lib/Integrations.php
+++ b/lib/Integrations.php
@@ -3,33 +3,37 @@
 namespace Timber;
 
 /**
- * This is for integrating external plugins into timber
- * @package  timber
+ * This is for integrating external plugins into Timber
+ * @package Timber
  */
 class Integrations {
 
-	var $acf;
-	var $coauthors_plus;
+  var $acf;
+  var $carbon_fields;
+  var $coauthors_plus;
 
-	public function __construct() {
-		$this->init();
-	}
-    
-	public function init() {
-		add_action('init', array($this, 'maybe_init_integrations'));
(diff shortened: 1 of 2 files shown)


xsynaptic:
This PR is stale and there's a new one in the pipe for 2.x (#2400).

PR #1540, opened 2017-08-30
Changed minor spelling error

crewneck:
Changed "Themeing" to "Theming"

**Ticket**: #

PR #1541, opened 2017-08-30
Fix link after site was renamed in #1540

gchtr:
#### Issue

After the Themeing guide was renamed to Theming in #1540, there’s a broken link in the Setup guide.

#### Solution

Fix link.

I already pushed that change to the docs. So there’s nothing more to do.


jarednova:
awesome! thanks for the fix

PR #1552, opened 2017-09-14
Twig_Function: Use Twig_SimpleFunction instead.

benrogmans:
Twig_Function is an abstract class and can thus not be instantiated.

**Ticket**: #

### DIFF
--- a/docs/guides/functions.md
+++ b/docs/guides/functions.md
@@ -64,7 +64,7 @@ If you have functions that you use a lot and want to improve readability of your
  * @return $twig
  */
 add_filter( 'timber/twig', function( \Twig_Environment $twig ) {
-	$twig->addFunction( new Twig_SimpleFunction( 'edit_post_link', 'edit_post_link' ) );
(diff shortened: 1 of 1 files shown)


gchtr:
Hey @benrogmans, thanks for taking the time to create a pull request.

This is probably related to https://github.com/timber/timber/pull/1464.

Are you getting the error about the abstract class when using `Twig_Function`?

Based on which Twig version you use, it’s either `Twig_SimpleFunction` or `Timber_Function`. To select the proper version, a new class was introduced in version 1.3.4 of Timber:
https://github.com/timber/timber/blob/master/lib/Twig_Function.php

So I’m guessing it should be new `Timber\Twig_Function()`:

```php
	$twig->addFunction( new Timber\Twig_Function( 'edit_post_link', 'edit_post_link' ) );
} );
```

benrogmans:
@gchtr Yep, got an error about the abstract class.

jarednova:
Awesome, sorry you ran into the issue @benrogmans — but glad to have the docs fixed!

PR #1554, opened 2017-09-19
Letterbox transparent background

nlemoine:
#### Issue
You can't use letterbox with a transparent background. 

#### Solution
Handle transparent background when using letterbox on a png file. 

#### Impact
None that I am aware of. 

#### Usage Changes
There's now no default color when using letterbox. 

#### Testing
I wrote a test to check if the image has transparency. The test passes. No regressions seen on other transformations.


jarednova:
Thanks @nlemoine! this is next up in my queue

jarednova:
That was easy. Thank you so much for writing the test! Just made a tweak so `-trans` is appended to the end of those filenames

nlemoine:
Thanks! 👍

Issue #1213, opened 2016-10-18
Add item.is_active to TimberMenu items


bambamboole:
It would be very nice to get a possibility to check if the menu item is the current active item.

Example:
```
<ul>
  {% for item in menu %}
    <li{{ menu.is_active ? ' class="active"' }}>
      <a href="{{ item.get_link }}">{{ item.title }}</a>
    </li>
  {% endfor %}
</ul>
```

I don't know if it's hard to implement something like this, but I think I'm not the only one who would love to have a method like this.

cheers

luukdv:
If it's just for having an 'active' class, you could do `<li class="{{ item.classes|join(' ') }}`> and use the added `.current-menu-item` class.

bambamboole:
and handle this myriad of trash classes from wordpress? thats what I don't want... I try to build a clean wordpress site without all these unusable classes...

roylodder:
You can check if an item in your loop is current by using:
```
{{ item.current ? 'is-active' }}
```

motia:
You can create a twig filter and use it in your templates:
```
  $twig->addFilter( new Twig_SimpleFilter('active', function($item){
     if(strpos(Timber\URLHelper::get_current_url(), $item->get_link()) === 0)
         return 'active';
     return '';
 }));
```

```
	<li class="{{ item|active }}">
```

bambamboole:
the method @roylodder supplied, is the best in my opinion.

It should find the way to the docs...

Thanks for all the advices!

closed :-)

Splendorr:
In case anyone is still running into this problem, I searched all over and finally solved my version of this problem. I had a vanilla installation of Wordpress with the Timber Starter Theme, and just couldn't get the `current_page_item` class to appear in the menu. This turned out to be because I hadn't created a Wordpress menu; I was just using the built-in menu in the Starter Theme. But apparently `new TimberMenu();` needs to interact with a native Wordpress menu to get info such as `current`.

In the Wordpress admin panel, I went to Appearance > Menus and created a menu called "primary_navigation". (I don't think the name of this menu matters.) Then I added all of my pages to the menu, just for good measure. I'm not directly using this menu anywhere in my template, but all of a sudden my TimberMenu has a lot more info in the array, and adds `current_menu_item` to the item's classes automatically!

I didn't see anywhere in the Timber or Starter Theme documentation explaining that you should create a Wordpress menu if there isn't one. Good luck to anyone who has wrestled with this!

PR #1569, opened 2017-10-05
Optimize documentation for menus

gchtr:
#### Issue

Improve documentation for Menus

#### Solution

- Create a new guide «Menus» for extended information about how to use Timber Menus.
- Add section about current menu item in guide, as requested in #1213.
- Improve DocBlocks for class methods and properties.
- Make the following methods public for the Menu class: `get_items()`, `find_parent_item_in_menu()`.
- Make the following methods public for the MenuItem class: `__toString()`, `type()`, `meta()`, `path()`, `title()`, `thumbnail()`.
- Hide MenuItem’s `__construct` in documentation.
- Add inherited properties `$current`, `$current_item_parent` and `$current_item_ancestor` to MenuItem class to make them available in the documentation.

#### Impact

This commit also deprecates the wiki page about Menus.

#### Usage Changes

None

#### Considerations

> Add inherited properties `$current`, `$current_item_parent` and `$current_item_ancestor` to MenuItem class to make them available in the documentation.

I checked out `Timber\Post`, and there a lot of properties listed, that are inherited from `WP_Post`. So I guess that’s the practice that we want to establish, that commonly used inherited properties are listed in Timber classes as well?

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,7 @@ Timber is great for any WordPress developer who cares about writing good, mainta
 #### Related Projects
 * [**Timber Starter Theme**](https://github.com/timber/starter-theme) The "_s" of Timber to give you an easy start to the most basic theme you can build upon and customize.
 * [**Timber Debug Bar**](https://github.com/timber/debug-bar-timber) Adds a debug bar panel that will show you which template is in-use and the data sent to your twig file.
+* [**Timber Dump Extension**](https://github.com/nlemoine/timber-dump-extension) Debug output with nice formatting.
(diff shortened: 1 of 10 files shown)


jarednova:
Thanks @gchtr! As to your question....

> I checked out Timber\Post, and there a lot of properties listed, that are inherited from WP_Post. So I guess that’s the practice that we want to establish, that commonly used inherited properties are listed in Timber classes as well?

That was the original idea. I was thinking these could explicitly inherit from `WP_Post`, etc., however WP has marked these `final` which prevents it from being extended. Here's the ticket on that issue:

https://core.trac.wordpress.org/ticket/24672

PR #1585, opened 2017-10-24
Timber Menu Depth

Soben:
#### Issue
We are building our Timber Base Theme and heavily utilizing this code. I realized that one of the primary features of `wp_nav_menu` (restricting the depth of the menu) was not available for `Timber\Menu` so I implemented a fix.

#### Solution
Allows an optional array of options for Timber\Menu that includes `depth`... sets the depth to an integer, and if a menu is found, after sorting it to have children, run back through and clear out any children that extends beyond the depth.

#### Impact
Zero impact

#### Usage Changes
None, unless depth is set by the user. Standard handling is normal.

#### Considerations
We could probably set this up to loop through a bit more efficiently (why go through and organize all of the children if we're not even going this deep in the end?)

#### Testing
No unit tests yet. I do not know if they would be needed or not.

### DIFF
--- a/docs/guides/menus.md
+++ b/docs/guides/menus.md
@@ -43,6 +43,17 @@ Or pass nothing. This is good if you have only one menu. In that case Timber wil
 $menu = new TimberMenu();
 ```
 
+## Options
+
+Optionally, you can send additional options to `Timber\Menu`. Current only `depth` is supported
+
+```php
+$args = array(
+	'depth' => 2,
+);
+$menu = new TimberMenu( 'primary', $args);
+```
+
 ## Setting up a menu globally
 
 The first thing to do is to initialize your menu using `Timber\Menu`. This will make the menu available as an object to work with in the context. Because we need the menu on every page, we can add it to the global context through the `timber/context` filter:
--- a/docs/guides/menus.md
+++ b/docs/guides/menus.md
@@ -40,18 +40,18 @@ $menu = new Timber\Menu( 'primary' );
 Or pass nothing. This is good if you have only one menu. In that case Timber will just grab what you got.
 
 ```php
-$menu = new TimberMenu();
+$menu = new Timber\Menu();
 ```
 
 ## Options
 
-Optionally, you can send additional options to `Timber\Menu`. Current only `depth` is supported
+Optionally, you can send additional options to `Timber\Menu`. Current only `depth` is supported (see https://developer.wordpress.org/reference/functions/wp_nav_menu/ for reference)
 
 ```php
 $args = array(
 	'depth' => 2,
 );
-$menu = new TimberMenu( 'primary', $args);
+$menu = new Timber\Menu( 'primary', $args );
 ```
 
 ## Setting up a menu globally
--- a/lib/Menu.php
+++ b/lib/Menu.php
@@ -117,6 +117,8 @@ class Menu extends Core {
 	 */
 	protected function set_options ($options) {
 		if (!is_array($options)) {
+			$this->depth = -1;
+			$this->raw_options = [];
 			return;
 		}
 
@@ -248,12 +250,13 @@ class Menu extends Core {
 	 * @param array $menu
 	 */
 	protected function strip_to_depth_limit ($menu, $current = 1) {
-		if ($this->depth <= 0) {
+		$depth = (int)$this->depth; // Confirms still int.
+		if ($depth <= 0) {
 			return $menu;
 		}
 
 		foreach ($menu as &$currentItem) {
-			if ($current == $this->depth) {
+			if ($current == $depth) {
 				$currentItem->children = false;
 				continue;
 			}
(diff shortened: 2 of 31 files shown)


Soben:
Can you please advise me on these codecov and coverage errors? I'd love to resolve them in order to actually implement these changes.

jarednova:
Hi @Soben — thanks so much for the PR! It looks like the coverage errors are due to many of the new lines not being covered by tests. Here's a link to the Coveralls report:

https://coveralls.io/builds/13862766/source?filename=lib%2FMenu.php#L251

Can you take a whack at writing some tests? Here's our guide to getting things up-and-running with tests:

... you'll want to add new ones to the [`test-timber-menu.php`](https://github.com/timber/timber/blob/master/tests/test-timber-menu.php) file

Soben:
Working on some tests now.

tests/test-timber-menu.php
```diff
@@ -17,7 +20,10 @@ function testBlankMenu() {
 		$this->assertEquals( '/home/', $item->path() );
 	}
 
+<<<<<<< HEAD
+=======
 	
+>>>>>>> 911a9aa41d6fd06942f4f5afff002d6ef184f64f
```
JasonTheAdams: Oops. Need to remove the merge lines.

Soben: Thanks. I'll fix that in a few moments. Not sure how much that'll help the overall checks, though. I really don't get how i missed it either.

Soben:
FINALLY.
meme_fuck_yea_guy__15903 1403094419 500 659

jarednova:
Wooohooo, thanks @Soben! This is next-up in my queue to review and merge

docs/guides/menus.md
```diff
@@ -43,6 +43,17 @@ Or pass nothing. This is good if you have only one menu. In that case Timber wil
...
+
+Optionally, you can send additional options to `Timber\Menu`. Current only `depth` is supported
+
+```php
+$args = array(
+	'depth' => 2,
+);
+$menu = new TimberMenu( 'primary', $args);
+```
+
```
gchtr: Would you mind to add some more information about what the `depth` parameter is doing exactly? I think for many users it might not be clear that `depth` means the maximum depth for a menu.

Soben: Sure. I'll try to be more verbose... Suggestions on direction there? This really does follow the same flow as the WP menu, so those who are using Timber, I would hope already know the typical way for handling WP's menu... but I am willing to try to explain it a little more clear than it is now.

gchtr: It’s always hard to assume how much users know.

I think it doesn’t need a lot more information. If the options replicate the functionality of the options that can be passed to [`wp_nav_menu()`](https://developer.wordpress.org/reference/functions/wp_nav_menu/), I think it would be enough to mention that and provide the link to the WordPress Code Reference, so that people can read about the options there.

lib/Menu.php
```diff
@@ -93,6 +111,16 @@ protected function init( $menu_id ) {
 		}
 	}
 
+	/**
+	 * @internal
+	 * @param mixed $options
```
jarednova: should this be `@param array $options`?

jarednova:
@Soben thanks for your work thus far on this. I was diving-in to a deeper review and while I think the code is solid, I wanted to get your take on something I noticed. It looks like in the list of options that @gchtr linked to, the only property that seems relevant to Timber is the `depth` one — the other stuff is IDs, classes, containers, etc. all of which is negated by tossing the data into Twig (I suppose `fallback_cb` _could_ be implemented, but even that I'm not sure about).

I'm wondering if instead of using the options, something like ....

```twig
{% for item in menu.get_items(3) %}
    {# continue menu output here #}
{% endfor %}
```

or perhaps ...

```twig
{% for item in menu.items.depth(3) %}
    {# continue menu output here #}
{% endfor %}
```

... might make more sense to devs. What say you?

Soben:
Hey... Yea, that does make sense. I am interested in doing so. I'm tempted to say we keep both options, though, so devs/people can choose what makes more sense to them, in their context and integrations.

I have been super busy these days, so apologizes for not responding. I hope to find time to work on it this week.

jarednova:
@Soben thanks so much for this. I did a lot of navel-gazing over how this should relate to Timber 2.0 (and a bunch of other refactoring I wanted to do) before coming to my senses tonight to separate the two. 

Some refactoring forthcoming, but that will be in a new PR

Soben:
Hey Jared!

I'm glad it's being added. I agree, it's a non-stop battle of "Is this where I want this project to be heading"... I encounter that all the time.

I'm sorry for not getting around to working on the other solution. Life has certainly got in the way, but I'm glad we can move forward from this pull request :)

jarednova:
Indeed, life vs. code is an eternal struggle — one I'm constantly losing. Thanks again for sending such a thorough PR. Would love other ideas/PRs in the future too!

PR #1594, opened 2017-11-04
An easier way (filter) to get a specific item of an array

pablo-sg-pacheco:
#### Issue
What if you want to get an specific item of an array?

#### Solution
**wp_list_filter** is a native WordPress function that filters a list of objects, based on a set of key => value arguments. Perfect for this situation!
In WordPress [documentation](https://codex.wordpress.org/Function_Reference/wp_list_filter) there aren't examples of it in action, but [here](url) there is a good one

#### Impact
```twig
{# Getting two specific posts from a wp-query, post-1 and post-2 #}

Post-1: {{ posts | filter('post-1') }}
<br />
Post-2: {{ posts | filter('post-2') }}
```

### DIFF
--- a/lib/Helper.php
+++ b/lib/Helper.php
@@ -430,8 +430,8 @@ class Helper {
 	 *
 	 * @return mixed
 	 */
-	public static function filter_array( $array, $value, $key = 'slug' ) {
-		$result = wp_list_filter( $array, array( $key => $value ) );
+	public static function filter_array( $array, $value, $key = 'slug', $operator = 'AND' ) {
+		$result = wp_list_filter( $array, array( $key => $value ), $operator );
 		if ( is_array( $result ) ) {
 			$first_element = reset( $result );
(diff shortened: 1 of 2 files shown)


jarednova:
Thanks @pablo-sg-pacheco! Sorry this has taken me a few weeks to get to. I'm doing some tests and modifications in #1620. The biggest thing I changed was how the `$key/$value` args are sent. The way I have it now allows for an array which will let someone do a more complicated request like...

```twig
{% for post in posts|filter({post_title: "Cheese", post_content: "foo"}, "OR") %}
```

Tests should be done in a bit then I'll merge!

pablo-sg-pacheco:
Great @jarednova :)

But I'm not sure if it was necessary. If I'm not wrong we could already create complex queries using comma parameters. But I really liked your curly bracket syntax.

Thanks anyway :)
I'll give it a try

PR #1596, opened 2017-11-16
Timber\Site: provide (real) site_url

nlemoine:
#### Issue

`site_url` ≠ `home_url`

Timber doesn't distinguish `site_url` (the path to WordPress) from `home_url` (the path to home). 
In a regular usage, they are often the same but if you're using a specific folder structure such as Bedrock, and more globally when using WordPress in a subdirectory, they will be different. 

#### Solution

Provide both values in the `Timber\Site` class.

#### Impact

None, `url` still refers to `home_url`.

#### Usage Changes

You can now use `{{ site.home_url }}` or `{{ site.site_url }}`. 

For example (the reason of this PR), the [comment form action URL](https://github.com/timber/starter-theme/blob/master/templates/comment-form.twig#L4) is presently wrong. The right URL is actually `{{ site.site_url ~ '/wp-comments-post.php' }}`

#### Testing

None.

### DIFF
--- a/lib/Site.php
+++ b/lib/Site.php
@@ -167,6 +167,8 @@ class Site extends Core implements CoreInterface {
 	 */
 	protected function init() {
 		$this->url = home_url();
+		$this->home_url = $this->url;
+		$this->site_url = site_url();
 		$this->rdf = get_bloginfo('rdf_url');
 		$this->rss = get_bloginfo('rss_url');
 		$this->rss2 = get_bloginfo('rss2_url');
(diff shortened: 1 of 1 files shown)


jarednova:
Thank you @nlemoine! Could you also submit a PR to the Starter Theme repo to fix the comment form action URL to go along with this?

nlemoine:
Thanks!
Here it is: https://github.com/timber/starter-theme/pull/55

jarednova:
Great — all merged!

Issue #1597, opened 2017-11-16
[Improvement] Support ACF's get_field_object by TimberPost


piotrkulpinski:
Hi,

Would be great if Timber would support get_field_object the same way it support get_field function: 
https://www.advancedcustomfields.com/resources/get_field_object/

What I'd love to have would be something like that:
```
{% set field = post.get_field_object('field') %}

{{ field.label }}
{{ field.value }}
```

Let me know what you think.

palmiak:
https://github.com/palmiak/timber/tree/get_field_object

Try this - it should be working just as you want to:
```php
{% set fields = post.get_field_object('field_name') %}
{{ function( 'var_dump', fields ) }}
```

piotrkulpinski:
@palmiak Thanks!

Maybe you could create a pull request to include that in the core?

cc @jarednova

gchtr:
This feature was released in [1.6.0](https://github.com/timber/timber/releases/tag/1.6.0).

PR #1599, opened 2017-11-23
#1597 get_field_object support

palmiak:
**Ticket**: #1597 

#### Issue
No support for `get_field_object`

#### Solution
It just adds the missing support

#### Impact
It adds two methods. 

In Post object there is a `get_field_object` (with new filter) and in ACF Integration there is 'post_get_meta_object' and a new `add_filter`

#### Usage Changes
You use it just like `get_field`:
```php
post.get_field_object('field_name')
```

#### Considerations
Now it's working only with posts, but if you think it will be useful I can write term and user version too.

#### Testing
I think the tests should be simmilar to `get_field`


jarednova:
All merged! Thanks @palmiak. This will be in the new release next week!

PR #1600, opened 2017-11-25
PostQuery() example call missing Timber namespace

nateweller:
**Ticket**: #N/A 

#### Issue
First example calls PostQuery() without Timber namespace, causing an "Uncaught Error: Class 'PostQuery' not found" error.

#### Solution
Include Timber namespace in example PostQuery() call.

#### Impact
No impact to codebase. Improves reliability of documentation.

#### Usage Changes
None.

#### Considerations
None.

#### Testing
None.


jarednova:
Yes, that's a problem — thanks for the catch @nateweller !

Issue #1561, opened 2017-09-27
Code sniffing for Timber files (PHPCS, WPCS)


johnbillion:
Twig files can't be sniffed for coding standards by [WPCS](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards) because the files are not written in PHP.

I've opened https://github.com/twigphp/Twig/issues/2564 to ask the Twig developers about a sniffer. If something exists or gets built, it would be great if Timber implemented some WordPress-specific rules if necessary so the sniffs more closely resemble WordPress' PHP coding standards.

gchtr:
> It would be great if Timber implemented some WordPress-specific rules if necessary so the sniffs more closely resemble WordPress' PHP coding standards.

I recently discussed this with @jarednova, and we came to the conclusion that sniffer rules would be great. We plan to create a custom ruleset that comes as close to the [WordPress Coding Standards](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards) as possible, but deviates in certain areas (like class or hook naming).

Additionally to the custom ruleset for PHP Code Sniffer, we want to document where the coding standards are different from the WordPress standard for developers who don’t use a sniffer.

Having sniffs for Twig files as well would be very nice.

PR #1603, opened 2017-12-05
Define coding standards and introduce custom ruleset to apply and check them

gchtr:
**Ticket**: #1561

## Issue

Coding Standards need to be documented and there needs to be an easier way to check and fix violations, maybe even automatically. WordPress just recently [applied the WordPress Coding Standards to the whole codebase](https://make.wordpress.org/core/2017/11/30/wordpress-php-now-mostly-conforms-to-wordpress-coding-standards/).

## Solution

Define and document the coding standards and provide a way for developers to check for errors in coding style automatically.

- Introduce a new `phpcs.xml` file, which is a custom ruleset for [PHP_CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer) that inherits from [WPCS](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards). The WPCS (WordPress-Coding-Standards) project provides sniff rules for PHP_CodeSniffer to enforce WordPress coding standards in a project.
- Document Coding Standards in `CONTRIBUTING.md`. Define WordPress Coding Standards as the standard for Timber and list the exceptions where the Timber standard deviates from the WordPress Coding Standards. Document how developers can use the custom ruleset to check for coding style errors.

I hope we can all agree that we’d want to use as few exceptions to the WordPress Coding Standards as possible. However, considering how the Timber code base has grown and how it’s used by theme developers, certain exceptions need to be introduced. Currently I found only two exceptions where the Timber Coding Standards would deviate from the WordPress Coding Standards:

- Class and file names are defined in `StudlyCaps`. We follow the [PSR-0 standard](http://www.php-fig.org/psr/psr-1/#namespace-and-class-names), because we use autoloading via Composer.
- We use hook names namespaced by `/` instead of underscores (e.g. `timber/context` instead of `timber_context`).

Probably we’ll have to see how this holds up. This is a first version of the Timber Coding Standards, not a definitive one. But I think it would be best if we can stay as close to the WordPress Coding Standards as possible.

To [automate the process](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Fixing-Errors-Automatically), PHP_CodeSniffer comes with the PHP Code Beautifier and Fixer tool (`phpcbf`) to fix code according to sniff rules. This could maybe help. Still, we need to be extra careful to not introduce errors through this.

Maybe fixing the coding style in the whole code base could be something for Timber version 2.0 as well. 

### How big are the changes?

When the coding standards are applied to the code base, mostly it will just be formatting changes.

By inspecting the code with the new sniff rules, I saw that one of the biggest changes introduced would be a naming convention for class name variables. Consider the following line:

https://github.com/timber/timber/blob/4d7609edaf04b9d1fc5d8f2c402cf7c0e5bda7ac/lib/Post.php#L59

PHP_CodeSniffer would mark an error here:

`WordPress.NamingConventions.ValidVariableName.NotSnakeCase: Variable "PostClass" is not in valid snake_case format.`

The variable would need to be changed from `$PostClass` to `$post_class`, which shouldn’t be a problem, because from what I saw, they are only used internally. Deactivating that rule is not really an option, because then other variable names wouldn’t be checked either.

### Whitelisting code

If it’s not possible to adapt to certain rules, certain code could be whitelisted:

- <https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/wiki/Whitelisting-code-which-flags-errors>
- <https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#ignoring-parts-of-a-file>

I think this should be used sparingly.

### Required inline documentation

A large amount of sniff warnings will be about missing inline documentation. I think this will be a good incentive to improve this. The [Reference](https://timber.github.io/docs/reference/) will benefit from this as well.

### Version history

If applied to the current codebase, the version history will be more difficult to read. However, with Github’s advanced blame feature, it’s easy to go back step by step. WordPress did it [all in one patch](https://make.wordpress.org/core/2017/11/30/wordpress-php-now-mostly-conforms-to-wordpress-coding-standards/), so I guess we could do it in one pull request as well.

## Usage Changes

- None for theme developers.
- Usage instructions for how to use PHP_CodeSniffer to apply standards for pull requests can be found in `CONTRIBUTING.md`.

## Considerations

- WordPress Coding Standards don’t define a maximum length. Should *we* define one?
- Does it need to be [integrated into **Travis**](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards#running-your-code-through-wpcs-automatically-using-ci-tools)?
- Does it need to be integrated into **Scrutinizer**? While I don’t know a lot about how Scrutinizer works, it allows you to [run PHP_CodeSniffer](https://scrutinizer-ci.com/docs/tools/php/code-sniffer/).

## Testing

No. Existing tests need to be run after coding styles are applied to the code base.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,7 +13,52 @@
 
 Pull requests are highly appreciated. More than fifty people have written parts of Timber (so far). Here are some guidelines to help:
 
-1. **Solve a problem** Features are great, but even better is cleaning-up and fixing issues in the code that you discover.
-2. **Write tests** This helps preserve functionality as the codebase grows and demonstrates how your change affects the code.
-3. **Small > big** Better to have a few small pull requests that address specific parts of the code, than one big pull request that jumps all over.
-4. **Comply with standards** We follow the [WordPress coding standards](https://make.wordpress.org/core/handbook/coding-standards/php/) – which sometimes deviate from [PSR-2](http://www.php-fig.org/psr/psr-2/) and others.
+1. **Solve a problem** – Features are great, but even better is cleaning-up and fixing issues in the code that you discover.
+2. **Write tests** – This helps preserve functionality as the codebase grows and demonstrates how your change affects the code.
+3. **Small > big** – Better to have a few small pull requests that address specific parts of the code, than one big pull request that jumps all over.
+4. **Comply with Coding Standards** – See next section.
+
+## Coding Standards
+
+We try to follow the [WordPress Coding Standards](https://make.wordpress.org/core/handbook/coding-standards/php/) as close as we can, with the following exceptions:
+
+- Class and file names are defined in `StudlyCaps`. We follow the [PSR-0 standard](http://www.php-fig.org/psr/psr-1/#namespace-and-class-names), because we use autoloading via Composer.
+- We use hook names namespaced by `/` instead of underscores (e.g. `timber/context` instead of `timber_context`).
+
+### Inline Documentation
+
+We follow the [WordPress Inline Documentation Standards](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/). The [Reference section of the documentation](https://timber.github.io/docs/reference/) is automatically generated from the inline documentation of the Timber code base. That’s why we allow Markdown in the PHPDoc blocks.
+
+### Use PHP_CodeSniffer to detect coding standard violations
+
+To check where the code deviates from the standards, you can use [PHP_CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer). Timber comes with a `phpcs.xml` in the root folder of the repository, so that the coding standards will automatically be applied for the Timber code base.
+
+- Install PHP_CodeSniffer globally by following this guide: <https://github.com/squizlabs/PHP_CodeSniffer#installation>.
+- Install WPCS by following this guide: <https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards#installation>.
+
+#### Command Line Usage
+
+To run PHP_CodeSniffer with the default settings on all relevant Timber files, use the following command from the root folder of the Timber library: 
+
+```bash
+phpcs
+```
+
+You could check a single file like this:
+
+```bash
+phpcs ./lib/Menu.php
+```
+
+Use `phpcs --help` for a list of available settings.
+
+#### Use it in your IDE
+
+Please refer to <https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards#how-to-use> for different ways to use PHP_CodeSniffer directly in your IDE. In some IDEs like PHPStorm, you may have to select the `phpcs.xml` explicitly to apply the proper standards.
+
+#### Whitelisting
+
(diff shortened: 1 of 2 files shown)


jarednova:
@gchtr this is great! I was finally able to a deep-dive today. `phpcbf` is beautiful but terrifying. Where to go from here?

- **Travis**: Given the amount we'd have to whitelist or ignore, I vote on holding-off for Travis for now
- **Scrutinzier**: I _think_ I updated the config to use the `phpcs.xml` file 
- **Maximum Length**: is this maximum line length? file length? I think I need some more info on this one
- **`phpcbf` & Version History** I started to have some fun, but then pulled back recognizing what your note about it messing with blames. Let's hold off on mass running for 2.0

I'm going to merge this in on #1614 with a few updates I made. Let me know your thoughts on this other stuff (esp. max length) to see what else we need to do

gchtr:
@jarednova I can’t really comment much on Travis and Scrutinizer, because I don’t know them well enough.

> Maximum Length: is this maximum line length? file length? I think I need some more info on this one

I meant **maximum line length**. The WordPress Coding Standards don’t say anything about this. When looking at PSR-2, they have the convention of staying below 80 characters, but longer lines are allowed. [This question](https://stackoverflow.com/questions/14962667/what-is-a-line-length-soft-limit-and-how-do-i-interpret-this-in-the-psr-2-conv) goes more into detail.

It also [seems to be possible to set this as a rule for PHP_CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer/issues/1512), but the question remains: is this something to think about, or should we just leave this open?

Maybe for code it’s not as important as for multiline DocBlock comments, where I’d like to have standard to follow. For code itself probably it’s often a case of: «It depends.». But for DocBlocks, I think 80 characters per line would be reasonable for better readability even on smaller screens.

> **phpcbf & Version History** I started to have some fun, but then pulled back recognizing what your note about it messing with blames. Let's hold off on mass running for 2.0

I think version 2.0 would be a great place to do this.

---

I saw that in https://github.com/timber/timber/pull/1614/commits/c7f6bfcf8c566e90f33d9cd101c29bd80e29d03d you added additional rules for function call spacing. What’s the reasoning behind this? Would we deviate from the WordPress Coding Standards here?

jarednova:
Cool! That's what I thought with line length. I think implementing line length for doc blocks sounds great.

As for c7f6bfc: I've always done this as a no space thing (ie `foo($a, $b)`) — but that's just a personal preference. I get the feeling that this might be time for me to drop that personal pref in favor of community standards?

gchtr:
Okay then, I’ll see if we could check for comment length only.

And about your personal preferences: Yes, I guess this would be the perfect time to drop those. I did that around 1.5 years ago. First I was struggling, because everything looked ugly in the WordPress Coding Standards. Fast-forward to now and I think everything that’s not WordPress Standard looks ugly and even less readable. So you can definitely get used to it.

It’s really tempting to change and add sniffing rules, but in my opinion, if there’s not a reason like backwards compatibility we shouldn’t deviate from the WordPress Coding Standards. This makes it easier for everybody.

Issue #1609, opened 2017-12-21
Transient entries created by {% cache 'key' data %}....{% endcache %} never expire


salaros:
### Expected behavior
One of the big projects I worked for had some templates [using {% cache %} tag](https://timber.github.io/docs/guides/performance/#cache-parts-of-the-twig-file-and-data). 
Let's use this snippet as example:

```twig
{% cache 'header_widgets' header_posts %}
    <div id="header-widgets" data-count="{{ header_posts|length }}"></div>
    {% for header_post in header_posts %}
        <p class="just an example">
             {{ header_post.title }}
        </p>
        <!-- some other really nice HTML goes here>
    {% endfor %}
{% endcache %}
```

The snippet above behave as intended: [a transient items](https://codex.wordpress.org/Transients_API) is created with the following slug/key `_transient_timber_header_widgets__GCS__<a long hash goes here>`, containing the compiled HTML enclosed by {% cache %}/{% endcache %} tags.

### Actual behavior
Once the data changes, the hash after `__GCS__` changes, so a new transient entry is created and so on.
However old transient items remain in the DB and, moreover, they have autload set to true, so on each request WordPress pulls them from the DB with the following query:
```sql
SELECT option_name, option_value FROM wp_options WHERE autoload = 'yes';
```
after some time we had 15 000 __GCS__-suffixed items, autoloaded by each requests, fortunately we have DB queries cached and we didn't notice it a lot, but as you might imagine it affected the overall performance anyway.

### Possible solution
By default in Timber {% cache %} tags are backed by a 3rd party class called [GenerationalCacheStrategy](https://github.com/asm89/twig-cache-extension/blob/master/lib/Asm89/Twig/CacheExtension/CacheStrategy/GenerationalCacheStrategy.php), by default it creates items with lifetime=0, which means that these items will never expire, however IMO this default behavior must be overridable if one wants WordPress to take care about __GCS__-suffixed items without creating [WP Cron](https://codex.wordpress.org/Function_Reference/wp_cron) task which clean them from time to time etc.

### Steps to reproduce behavior
Create a twig file containing a {% cache 'key' data %}....{% endcache %} block, look at the transient entry created to cache the HTML enclosed by the block above.

### What version of WordPress, PHP and Timber are you using?
any version

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Composer/Packagist

salaros:
[...]

PR #861, opened 2016-02-21
implementing new crop positions: 'top-center' and 'bottom-center'

salaros:
useful when cropping vertical 9:16 images or creating Material Design cards.
I think that the image used in unit tests should be changed before writing new tests for 'top-center' and 'bottom-center' positions

example

PR #1610, opened 2017-12-21
allow overriding default lifetime of {% cache %}-created transients i…

salaros:
**Ticket**: #1609 

#### Issue
{% cache %}-generated transient items never expire and are auto-loaded by WordPress by default. So both performance and database size are affected.

#### Solution
The small change I made allow theme designers to override the behavior above and make GCS-<hash>-suffixed transient items expirable and not auto-loaded.

#### Impact
If one won't implement add_filter('timber/cache/extension/lifetime', ....); WordPress filter, nothing in default Timber behavior changes, so IMO there is no serious impact to the code base.

#### Usage Changes
Now theme developers can make {% cache %}-generated transient items expirable by implementing `timber/cache/extension/lifetime` filter:
```php
add_filter('timber/cache/extension/lifetime', function() {
        return HOUR_IN_SECONDS * 24;
});
```

#### Considerations
Frankly I don't know if all Timber users use [WP Cron](https://developer.wordpress.org/plugins/cron/scheduling-wp-cron-events/) in order to solve the issue described in #1609 and clear all previously-created and potentially unused transient items. If they don't even know about this problem, we'd consider the possibility to set the default cache strategy lifetime to something other than 0.

#### Testing
Nope. Unfortunately I have a lot of time now. The last time [I contributed to Timber](https://github.com/timber/timber/pull/861) I created tests, but this time I don't think I can do it. However if nobody can help us out with it I will create a unit test.

### DIFF
--- a/lib/Loader.php
+++ b/lib/Loader.php
@@ -93,14 +93,14 @@ class Loader {
 	 * @return string|bool             Name of chosen template, otherwise false.
 	 */
 	public function choose_template( $templates ) {
-		// Change $templates into array, if needed 
+		// Change $templates into array, if needed
 		if ( !is_array($templates) ) {
 			$templates = (array) $templates;
 		}
-		
+
 		// Get Twig loader
 		$loader = $this->get_loader();
-		
+
 		// Run through template array
 		foreach ( $templates as $template ) {
 			// Use the Twig loader to test for existance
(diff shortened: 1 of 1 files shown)


salaros:
As you might have noticed the build failed because of some complely unrelated functionality

jarednova:
Thanks @salaros — re-ran those tests against the new `master` (looks like something changed in WP with how those un-related media tags were built). Merged!

salaros:
Wow, thanks @jarednova I will send a PR with unit tests shortly

jarednova:
Thanks @salaros. Does this close #1609?

salaros:
Yeap

Issue #1580, opened 2017-10-12
Version 2.0 Running List


jarednova:
A bunch of things are popping in my head to do for version 2.0. I'm opening this thread as a running list

x Consider default escaping strategy
x Remove deprecated functions
x Remove routes
x Remove import vs. on-demand meta

### General
- [x] Remove Routes @jarednova
- [ ] remove deprecated functions 
- [x] Streamline filter names for `timber/filter_name` @gchtr 
- [x] Deprecate `old_filter_names` @gchtr 
- [x] Remove context argument when calling {% do action %} (#1528). @gchtr
- [ ] Apply Timber Coding Standards
- [x] Deprecate custom debugging methods @jarednova
- [x] Add new Timber\Attachment class for easier handling of attachments (which can also be PDFs or other non-image file types) with methods for getting the filesize or the path/URL. Timber\Image could just be extending this class, adding all the functionality required for image files. @gchtr 
- [x] Replace instances of Timber::get_posts() with PostQuery and Timber::get_post() with Timber\PostQuery(), 
  - [x] In the code base @gchtr
  - [x] documentation @gchtr
  - [x] and starter theme @jarednova 

### Post
- [x] remove import of custom fields; on-demand meta @jarednova
- [x] remove public `get_` methods @jarednova
- [x] escape data by default @jarednova
- [x] remove deprecated functions  @jarednova
- [x] Streamline filter names for `timber/filter_name` @gchtr 
- [x] Deprecate `old_filter_names` @gchtr 

### Term
- [x] remove import of custom fields; on-demand meta
- [x] remove public `get_` methods
- [x] remove deprecated functions
- [x] Streamline filter names for `timber/filter_name` @gchtr 
- [x] Deprecate `old_filter_names` @gchtr 

### User
- [x] remove import of custom fields; on-demand meta
- [x] remove public `get_` methods
- [x] remove deprecated functions
- [x] Streamline filter names for `timber/filter_name` @gchtr 
- [x] Deprecate `old_filter_names` @gchtr

gchtr:
Some more ideas:

- Streamline filter names used in Timber with `/` as separators instead of `_`. Use [`apply_filters_deprecated()`](https://developer.wordpress.org/reference/functions/apply_filters_deprecated/) for deprecated filters.
- Remove context argument when calling `{% do action %}` (https://github.com/timber/timber/issues/1528).
- Apply Timber Coding Standards, which will be as close to WordPress Coding Standards as possible.

[...]

PR #1618, opened 2017-12-28
1580/custom data

jarednova:
**Ticket**: #1580 

#### Issue
Timber was grabbing all the custom data in `post_meta` and slapping it to the object. This had both pros and cons...
Pros: 
- Ability to treat custom data on the same level as WP-blessed data (ie `{{ book.title }}` AND `{{ book.isbn }}`
Cons: 
- Ability of custom data to overwrite methods
- Grabs whatever is stored in `wp_postmeta` which isn't necessarily the data the user really wants (for example, a data filtered by ACF or another custom field plugin)

#### Solution
Remove the "auto" importing of data, but still allow the user to get to properties at the top-level by implementing the magic method `__get`

#### Usage Changes
Should be very light, in cases where someone was getting data from the table and filtering it on their own, this could cause issues with their templates. However, in many cases (for example reading a `Timber\Image` into another `Timber\Image`) Timber is resilient enough to recognize the intention and thus some things might be double-filtered which _shouldn't_ matter (other than performance)

#### Considerations
I've disabled the test `TestTimberPost::testMetaCustomArrayFilter`. I'm going to do some more research, but it seemed to be covering an *extreme* edge-case of wanting to filter the names of fields to retrieve. Because field data is now only fetched at run time, this would no longer be possible

#### Testing
Yep!

### DIFF
--- a/lib/Image.php
+++ b/lib/Image.php
@@ -240,7 +240,7 @@ class Image extends Post implements CoreInterface {
 				return;
 			}
 
-			$relative = false;
+			$relative  = false;
 			$iid_lower = strtolower($iid);
 			foreach ( $this->file_types as $type ) { if ( strstr($iid_lower, $type) ) { $relative = true; break; } };
 			if ( $relative ) {
@@ -285,13 +285,6 @@ class Image extends Post implements CoreInterface {
 				$this->$key = $value[0];
 			}
 			$this->id = $this->ID;
-		} else {
-			if ( is_array($iid) || is_object($iid) ) {
-				Helper::error_log('Not able to init in TimberImage with iid=');
-				Helper::error_log($iid);
-			} else {
-				Helper::error_log('Not able to init in TimberImage with iid='.$iid);
-			}
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


gchtr:
@jarednova I have two question/remarks here:

So the values are still loaded from the database and added to the `custom` property. When using `__get()`, is Timber (or WordPress) retrieving a cached value? Otherwise, should `Timber\Core::__get()` or the `meta()` method check for a custom property and return that value?

I guess there will a lot of developers naming custom fields the same as already existing methods in `Timber\Post`. I will make sure this is documented clearly. Just to make sure, the suggested workarounds would be the following, right?

- Get the value through `{{ post.custom.my_custom_field }}`
- or through `{{ post.meta('my_custom_field') }}`
- or through `{{ post.get_field('my_custom_field') }}` when using ACF.

jarednova:
@gchtr you are right about the `custom` property. As it turns out, those values are cached so we get them for "free" (at least, that was based on my earlier tests about a year ago so assuming nothing changed).

My thinking for developer guidance was...

### `{{ post.my_field }}`
This is the recommended way which will hit `__get` and be filtered by ACF and other custom fields plugins. This will probably be what people use/want in 99% of siutations

### `{{ post.custom.my_field }}`
This is recommended if the dev wants the values straight from the `postmeta` table

### `{{ post.meta('my_field') }}`
This is recommended when the dev wants values filtered by ACF or other custom fields plugins, and also wants to remove any possibility of conflicts with Timber methods (thus, custom fields with risky names like `title` could still be accessed)

This would be applied across `Post`, `Term` and `User`. `get_field` would be deprecated. What do you think of this as the standards? I'm very open to suggestion on what the (hopefully final) standards on this should be

gchtr:
@jarednova 

I think this is a very reasonable approach.

The deprecation of `get_field()` will trigger a lot of warnings in a lot of themes when upgrading, but I think it is the right way to go. The name `get_field()` is quite specific to ACF, but Timber wants to be compatible with other custom fields plugins as well. So I think `meta()` as a method for retrieving custom fields in general is a good solution.

---

I had another idea to avoid conflicts with existing methods. Can we add a check to `__get` whether a method exists with that name and add a proper warning? Something like:

```php
public function __get( $field ) {
    if ( method_exists( $this, $field ) ) {
        Helper::warn( "You cannot access {{ post.{$field} }} directly. The name '{$field}' is reserved for a method of Timber\Post. Use {{ post.custom.fieldname }} or {{ post.meta('{$field}'') }} instead. Refer to https://timber.github.io/docs/reference/timber-post/ for a complete list of method names that Timber\Post uses." );
    }

    if ( 'class' === $field ) {
        return $this->css_class();
    }

    return parent::__get($field);
}
```

I think this could solve a lot of confusion and prevent some issues I’ve seen in the past that are quite tricky to detect.

PR #1621, opened 2018-01-03
2.x Streamline hook names, add documentation

gchtr:
Not all hook names use slashes to namespace them. This pull request brings all hooks to the new standard.

(I rebased my commits onto the current 2.x, that’s why there are more commits in that pull request. Can this be untangled?)

## General

- I didn’t change the formatting of existing hooks that already match the new hook name standard.
- I changed the order so that a deprecated hook appears after the new hook.
- I added a DocBlock to all hooks so that they get picked up by the upcoming Hook Generator for the documentation.

## Deprecating old hooks

WordPress provides the functions [`apply_filters_deprecated`](https://developer.wordpress.org/reference/functions/apply_filters_deprecated/) and [`do_action_deprecated`](https://developer.wordpress.org/reference/functions/do_action_deprecated/) to handle the deprecation of hooks.

When they are used, an error notice will be shown:

I converted all old filters and actions to use the deprecation functions.

## Switch from filter to action

The following filters were used, but the result wasn’t assigned to any variable:

- `timber_comment_get_meta_pre`
- `timber_post_get_meta_pre`
- `timber/calling_php_file`

Technically, it’s okay to use a filter instead of an action, because an action is a filter in the end, but it’s more user-friendly if they are declared as actions, because one expects a value to change when a filter is used.

Also, when an empty value was passed as the first parameter, I removed that, because it’d be of no use.

So, as an example, this is how the filter looked like before:

```php
apply_filters('timber_post_get_meta_pre', array(), $pid, $this);
```

And this is after:

```php
/**
 * Fires before post meta data is imported into the object.
 *
 * @deprecated 2.0.0, use `timber/post/pre_get_meta`
 */
do_action_deprecated(
    'timber_post_get_meta_pre',
    array( $customs, $pid, $this ),
    '2.0.0',
    'timber/post/pre_get_meta'
);
```

**However**, for the following newly introduced filters, I thought it would make more sense to introduce a short-circuit/skip filter (see next two sections) **instead of using an action**:

- `timber/comment/pre_get_meta_fields`

## Added short-circuits

For the following filters, I added an early bailout when a value is already set. This "short-circuiting" pattern is very common throughout WordPress Core.

- `timber/comment/pre_get_meta_field`

## Added skips

In a similar way to short-circuits, when a filter appears before certain functionality (like fetching values from the database), we can skip that functionality, because we already have values and save a little bit of performance.

- `timber/post/pre_get_meta`
- `timber/comment/pre_get_meta_fields`

## Streamline filter and function names

Consider the following functions, which each have two newly added filters, adapted from their old (underscored) filter names:

### Get all custom fields

**Post::get_post_custom()**

- `timber/post/pre_get_meta`
- `timber/post/get_meta`

**Term::get_term_meta**

- `timber/term/meta`

**User::get_custom()**

- `timber/user/pre_get_meta`
- `timber/user/get_meta`

**Comment::get_meta_fields()**

- `timber/comment/pre_get_meta_fields`
- `timber/comment/get_meta_fields`

### Get value for one custom field

**Post::get_field()**

- `timber/post/pre_get_meta_field`
- `timber/post/get_meta_field`

**Term::meta()**

- `timber/term/meta/field`

**User::meta()**

- `timber/user/pre_get_meta_field`
- `timber/user/get_meta_field`

**Comment::get_meta_field()**

- `timber/comment/pre_get_meta_field`
- `timber/comment/get_meta_field`

### What could be improved

1. The functions practically all do the same, but they are called differently. Could we streamline these? For example, `Post::meta()` currently calls `Post::get_field()`. But couldn’t it be the other way around, that `get_field()` is an alias of `meta()`?
2. The filter names are not streamlined as well. I think they should match the function names. For pref-filtering, WordPress uses uses `pre_` as a prefix and not as a suffix. I suggest we do that, too. If there’s a plan to streamline the function names in 2.0.0, it’d be better to change the filter names as well. Because these are newly introduced filters, we can still savely do that. 

This would also also have implications for filters used in `Timber\Integrations\ACF`.

## List of new filters

Please review the naming and parameters for the following hooks:

### Timber

- `timber/render/file` (former `timber_render_file`)
- `timber/render/data` (former `timber_render_data`)
- `timber/compile/file` (former `timber_compile_file`)
- `timber/compile/data` (former `timber_compile_data`)
- `timber/compile/done` (former `timber_compile_done`)

### Post

- `timber/post/pre_get_meta` (former `timber_post_get_meta_pre`)
- `timber/post/get_meta` (former `timber_post_get_meta`)

### User

- `timber/user/get_meta` (former `timber_user_get_meta`)
- `timber/user/pre_get_meta` (former `timber_user_get_meta_pre`)
- `timber/user/pre_get_meta_field` (former `timber_user_get_meta_field_pre`)
- `timber/user/get_meta_field` (former `timber_user_get_meta_field`)

### Comment

- `timber/comment/pre_get_meta_fields` (former `timber_comment_get_meta_pre`)
- `timber/comment/get_meta_fields` (former `timber_comment_get_meta`)
- `timber/comment/pre_get_meta_field` (former `timber_comment_get_meta_field_pre`)
- `timber/comment/get_meta_field` (former `timber_comment_get_meta`)

### Site

- `timber/site/update_option` (former `timber_site_set_meta`)

### URLHelper

The following filter names has to be changed to match the WordPress naming convention for hooks, which says that all hooks should be all lowercase:

- `timber/url_helper/url_to_file_system/path` (former `timber/URLHelper/url_to_file_system/path`)
- `timber/url_helper/file_system_to_url` (former `timber/URLHelper/file_system_to_url`)
- `timber/url_helper/get_content_subdir/home_url` (former `timber/URLHelper/get_content_subdir/home_url`)

### `timber_term_set_meta` and `timber/term/meta/set`

`Term::update()` is deprecated, without replacement, so we need to deprecate these filters as well.

### `timber_post_getter_get_posts`

As I understand it from looking at the code `PostGetter::get_posts()` is only used by `Timber::get_posts()`, which is a function that will be deprecated. So would it wrong to just deprecated the `timber_post_getter_get_posts` without a replacement?

## Term meta filters

Currently, there’s only one filter each for getting all term meta and term meta for a field. All other classes like `Post`, `User` and `Comment` have two filters each.

Maybe the functionality could be changed to match the functionality from other classes, including skipping and short-circuiting?

## timber/calling_php_file

The Timber Debug Bar still uses the old filter `timber_calling_php_file`. This should eventually be changed to `timber/calling_php_file`.

## How does it look like in the documentation?

Regenerate the docs in the `2.x` branch: https://github.com/timber/docs/tree/2.x.

## Testing

I updated the hooks in the test files. There’s still one error in `TestTimberProperty::testMeta`.

## Todo

Rethink function names and hook names based on the points mentioned above and add changes accordingly.


gchtr:
Damn, should have requested a merge into `2.x` and not `master`. Sorry.

Issue #1318, opened 2017-02-09
TimberPost picks up the wrong post ID in preview mode


bfred-it:
### Steps to reproduce behavior

1. Publish a post in the admin
2. `global $post->ID === 1` `new TimberPost()->ID === 1`
3. Make change in post
4. Click "Preview Changes" button
5. `global $post->ID === 1` `new TimberPost()->ID === 2` <-- mismatch

### Expected behavior

`new TimberPost()->ID` should be the same as `global $post; $post->ID`, which is the original post's ID

### Actual behavior

`new TimberPost()->ID` is the revision/preview ID, not the original post's ID.

This causes `new TimberPost()->thumbnail()` to return nothing if the thumbnail was changed.

This fixes it for me:

```php
global $post;
new TimberPost($post->ID);
```

### What version of WordPress, PHP and Timber are you using?
PHP 5.5.37
WordPress 4.7.1
Timber 1.2.1

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)

Composer

daltonrooney:
I was using `$post = new TimberPost()` to get the post in my page.php template, which caused the wrong template to be returned for previews. I'm using something along the lines of 

`Timber::render( array( 'layouts/' . $post->post_name . '.twig', 'layouts/default.twig' ), $context );`

to return the template. The slug should have been 'home' but TimberPost was returning '1983-autosave-v1', so default.twig was returned instead of home.twig.

Quick and easy fix was pass the ID of the current page to TimberPost like

`$post = new TimberPost(get_the_ID());`

mgussekloo:
Same problem here! Please advise.

gchtr:
Is there a difference if you use

```php
$post = Timber::get_post();
```

instead of `$post = new TimberPost();`?

sarukuku:
@bfred-it why was this issue closed?

bfred-it:
Mistake 😅

luigitec:
HI @bfred-it I think It's not an issue. 

The thing is that on the **preview mode**, it returns the post "revision" ID instead of the actual post ID, to retrieve such ID: you can do it requesting `post_parent`, that is `new TimberPost()->post_parent`.

bfred-it:
No, it should be the post id like Wordpress naturally does. I shouldn’t get its parent.

dreamon11:
Same problem here.

PHP 7.2.5
WordPress 4.9.6
Timber 1.7.1

Any ideas? :/

[...]

PR #1622, opened 2018-01-03
2.x Use new methods for querying posts.

gchtr:
**Ticket**: #1580

Replace instances of `Timber::get_posts()` with `PostQuery` and `Timber::get_post()` with `Timber\Post()`, both in the code base, documentation.

#### Testing

I didn’t change anything yet in test files. Do we need to write new tests or update existing tests?

### DIFF
--- a/docs/getting-started/theming.md
+++ b/docs/getting-started/theming.md
@@ -140,7 +140,7 @@ Let’s crack open **index.php** and see what’s inside:
 ```php
 <?php
 $context = Timber::get_context();
-$context['posts'] = Timber::get_posts();
+$context['posts'] = new Timber\PostQuery();
 
 Timber::render( 'index.twig', $context );
 ```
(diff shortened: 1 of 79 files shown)


jarednova:
Thanks @gchtr. I updated the tests to use `Timber\PostQuery` in addition to `Timber::get_posts`, etc. We can remove the instances of the `::get_posts` tests once those methods go from being deprecated to removed.

I uncovered to things that I'd like to resolve before merging this PR:

## Previews
It turns out that this test `TestTimberMainClass::testGetPostPreview` fails when `new Timber\Post` is used. This is think underlies the confusion found in #1318. I have to think we'd want the behavior of `Timber::get_post` and `new Timber\Post` to be the same here no? Where it should retrieve the preview version of the post when the user has sufficient privileges

## Stupid. Stupid. Stupid.
Back in the early days (we're talking late 2012) I had this dumb notion of writing a WP-Query language where someone could do like `post#my-cool-post` or `event.term` like jQuery.

Fucking dumb.

But I did it and I found a few remnants of it in `TestTimberMainClass::testGetPostsFromSlugWithHash` and `TestTimberMainClass::testGetPostsFromSlugWithHashAndPostType` 

so I'm going to remove those, because it's embaressing

jarednova:
So it turns out that post preview functions fine in actual tests, so there's something about the unit test that needs re-examining. I'm going to make that a separate issue rather than hold-up this PR

PR #1623, opened 2018-01-03
2.x Start an Upgrade Guide

gchtr:
#### Issue

We need an upgrade guide for 2.0.

#### Solution

Gradually add to/remove from this guide when working on 2.x.

PR #1624, opened 2018-01-03
2.x Make PostQuery available in Twig

gchtr:
#### Issue

Maybe there are developers out there who’d like to use `PostQuery` directly in Twig?

I thought about this when working on the WooCommerce integration, because there I have only reduced access to PHP templates.

#### Solution

Make `PostQuery` available as a function in Twig.

#### Impact

PHP files could be very lean:

```php
use Timber\Timber;

Timber::render( 'index.twig', Timber::get_context() );
```

But the whole idea probably only works well for simple use cases.

#### Usage Changes

I already added the following to the Upgrade Guide for 2.0.

> You can also use PostQuery in Twig. You can pass in your arguments either as a query string, or define the parameters in an argument hash (in Twig, [key-value arrays are called hashes](https://mijingo.com/blog/key-value-arrays-in-twig), and you use curly braces `{}` instead of brackets `[]` for them).

```twig
{# Query string #}
{% for post in PostQuery({'post_type=post&post_status=publish&posts_per_page=10'}) %}
	<a href="{{ post.link }}">{{ post.title }}</a>
{% endfor %}

{# Hash notation #}
{% for post in PostQuery({
    post_type: 'post',
    post_status: 'publish',
    posts_per_page: 10
}) %}
	<a href="{{ post.link }}">{{ post.title }}</a>
{% endfor %}
```

If this isn’t going to be merged, it can be removed again from the Upgrade Guide.

#### Testing

Yes.

### DIFF
--- a/lib/Twig.php
+++ b/lib/Twig.php
@@ -32,7 +32,11 @@ class Twig {
 	}
 
 	/**
+	 * Add Timber-specific functions to Twig.
 	 *
+	 * @param \Twig_Environment $twig
+	 *
+	 * @return \Twig_Environment
 	 */
 	public function add_timber_functions( $twig ) {
 		/* actions and filters */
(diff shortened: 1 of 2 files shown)


jarednova:
Awesome, thanks @gchtr! This is something that Craft has had that I've always craved. This is great and the usage details in the Upgrade Guide #1623 are perfect. My job is easy on this one ...

PR #1634, opened 2018-01-09
2.x Optimize deprecations

gchtr:
**Ticket**: #1580 

## Deprecated functions

I improved the handling of deprecated functions:

* Removed all occurences of `since` following after `@deprecated` tags. According to the [Inline Documentation Standards](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#1-2-deprecated-functions), the `@deprecated` tag should be directly followed by a version number. The "since" keyword is added automatically in the Documentation Generator.
* Always used full version number, e.g. `2.0.0` instead of `2.0`.
* Added `@see` tags where possible.
* Added `@api` tags and removed `@internal` for all deprecated functions, so that they appear in the documentation.
* If a deprecated function had DocBlock information that the new function didn’t have, I moved it to the new function.
* I didn’t add a warning when a function was marked to be removed in `2.0`.

## Improved `Helper::warn()`

- Do not warn if `WP_DEBUG` is not set to `true`.
- Do not return anything.

## Newly deprecated functions

I found methods that probably should be deprecated as well. Sometimes they were marked as `@internal`, other times they did the same as the newer methods or even already linked to them. So I went ahead and deprecated the following methods:

- `Term::get_edit_url()` in favor of `Term::edit_link()`
- `Term::get_children()` in favor of `Term::children()`
- `MenuItem::external()` in favor of `MenuItem::is_external()`

## Candidates for removal

Do any of the functions that were deprecated in 0.x or 1.0, maybe need to be removed for 2.x as well, instead of staying deprecated? Candidates are the following:

- `Site::get_link()`
- `Site::get_url()`

## Testing

Updated tests where needed, except for PostPreview tests. Sorry, I somehow don’t get how I can generate an output for PostPreview in PHP.


jarednova:
Thanks @gchtr! I'm back on the scene. I screwed up some of the merges on this branch so I cleaned things up in #1650. I also found a cool WP Testing annotation for `@expectedDeprecated` to handle those `{{ post.get_preview }}` warnings. There's now a new `Helper::deprecated` method to help with future warnings:

```php
Helper::deprecated('{{ term.get_link }}', '{{ term.link }}', '2.0.0');
```

... this hooks into WP's own deprecation errors system so we can ride on-top of their logging

PR #1635, opened 2018-01-10
Add ACF's get_field() format_value parameter.

TartanStag:
Add the ability to use ACF's get_field() format_value parameter to get raw values, e.g. Post ID's on a Page Link field type.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,6 @@
 {
   "name": "timber/timber",
-  "type": "library",
+  "type": "wordpress-plugin",
   "description": "Plugin to write WordPress themes w Object-Oriented Code and the Twig Template Engine",
   "keywords": [
     "timber",
--- a/lib/Integrations/ACF.php
+++ b/lib/Integrations/ACF.php
@@ -14,7 +14,7 @@ class ACF {
 
 	public function __construct() {
 		add_filter('timber_post_get_meta', array( $this, 'post_get_meta' ), 10, 2);
-		add_filter('timber_post_get_meta_field', array( $this, 'post_get_meta_field' ), 10, 3);
+		add_filter('timber_post_get_meta_field', array( $this, 'post_get_meta_field' ), 10, 4);
 		add_filter('timber/post/meta_object_field', array( $this, 'post_meta_object' ), 10, 3);
 		add_filter('timber/term/meta', array( $this, 'term_get_meta' ), 10, 3);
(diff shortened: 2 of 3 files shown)


gchtr:
Hey @TartanStag 

If you want to retrieve raw values, it’s already possible to do this by **not using** the `get_field()` method, but to directly access the property on a post:

```php
// Filtered value
$value = $post->get_field( 'field_name' );

// RAW value
$value = $post->field_name;

// RAW value
$value = $post->custom['field_name'];
```

There will be changes to how custom meta fields are retrieved in the upcoming version 2.x of Timber (see https://github.com/timber/timber/pull/1618 for more details). In the new version, we will even deprecate `get_field()` in `Timber\Post` in favor of the `meta()` method, which should work for most custom field plugins, and not only for Advanced Custom Fields.

I agree that it would be best to support all the parameters used in `get_field()` for better compatibility with ACF, but the direction we’re heading won’t support this. What would be a necessary use case where you’d really need to use the third parameter of `get_field()` instead of accessing a raw value?

## Changing composer type to `wordpress-plugin`

You added a change to change the Composer type to `wordpress-plugin`. If you’d want to propose that, it would be best to do that in a separate pull request.

But we probably can’t change that. I assume you’re using Bedrock or the like, where you have a Composer file on the root level of WordPress. If we change the Composer type like that, everyone that installs Timber on the theme level would suddenly see a different folder structure.

Current:

```
wp-content/themes/theme-name/vendor/timber/timber
```

New:

```
wp-content/themes/theme-name/wp-content/plugins/timber
```

I guess this not what we want. To work around this, everybody would have to add an extra `installer_paths` to their composer file. Or am I missing something?

Timber is not necessarily a plugin, it can be used as a library in a theme only, where it won’t show up in the plugins list. With type `library` it’s still possible to [define where the package should be installed](https://getcomposer.org/doc/faqs/how-do-i-install-a-package-to-a-custom-path-for-my-framework.md) if you want to have it in the plugins folder.

jarednova:
@TartanStag thanks for your PR. We recently pulled this into 2.0 in spirit. Version 2.0 will support format_value via:

```twig
{{ post.meta('my_field', {format_value: false}) }}
```

This is documented in #1861. Thanks again for the PR, though we were unable to use the exact code, it helped trigger a more comprehensive solution in the next version of Timber

Issue #1632, opened 2018-01-09
Twig_Function will be final in Twig 3.0


StuartFeldt:
I am getting this notice whenever calling Timber::render

> Overriding Twig_Function is deprecated since version 2.4.0 and the class will be final in 3.0.

\Timber\Twig_Function extends \Twig_Function, which will be final in 3.0 of Twig.

### What version of WordPress, PHP and Timber are you using?
PHP 7.0.25, Timber 1.6.0, WordPress 4.9.1

nlemoine:
I adressed this issue in the PR mentioned above. You can still grab my fork if you want to fix it ASAP.

StuartFeldt:
Awesome, thanks!

7php:
I get this error too.
`PHP 7.0.x, Timber 1.6.0, WordPress 4.9.3`

Will this merge be in the next release of Timber 1.6.x ?

jarednova:
@7php; yep! prepping a release for this week

7php:
Awesome thanks @jarednova

PR #1641, opened 2018-01-23
Alias Twig_Function from Twig 2.4.0

nlemoine:
**Ticket**: #1632

#### Issue
From 2.4.0, Twig does not allow `Twig_Function` class to be extended. Triggers PHP notices if using Twig ^2.4.0

#### Solution
Alias class and map it to new Twig class `\Twig\TwigFunction`.

#### Impact

Locally tested with Twig ^1.0|2.3.0|^2.4.0, looks fine.

#### Usage Changes

None.

#### Considerations

Require Twig ^2.4.0. PHP 7+ is now available on any decent hosting provider.

#### Testing

Sorry, no. I don't know how to deal with multiple versions of a package with `phpunit`.

### DIFF
--- a/lib/Twig_Function.php
+++ b/lib/Twig_Function.php
@@ -3,15 +3,24 @@
 namespace Timber;
 
 /**
-  * Temporary fix for conflicts between Twig_Function and Twig_SimpleFunction
-  * in different versions of Twig (1.* and 2.*)
-  */
-if ( version_compare(\Twig_Environment::VERSION, '2.0.0', '>=') ) {
+ * Handle TwigFunction among Twig versions
+ *
+ * From Twig 2.4.0, extending Twig_Function is deprecated, will be final in 3.0
+ *
+ * Temporary fix for conflicts between Twig_Function and Twig_SimpleFunction
+ * in different versions of Twig (1.* and 2.*)
+ */
 
-  class Twig_Function extends \Twig_Function { }
+if ( version_compare(\Twig_Environment::VERSION, '2.4.0', '>=') ) {
+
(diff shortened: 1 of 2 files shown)


jarednova:
Thank you @nlemoine — looking into how (if) we can test against diff't versions of dependencies.

jarednova:
@nlemoine I tested and was not able to get any errors/warnings pre-patch. What log or location are you seeing these? It looks like in the code there's a `@final` designation on the class but not in the class declaration itself:

https://github.com/twigphp/Twig/blob/2.x/lib/Twig/Function.php#L15

... can you confirm this? If this is a future-proof sort of thing, I'm still in total support. I just want to make sure I'm evaluating against the correct thing

nlemoine:
@jarednova Error doesn't show up because there's an error control operator `@` before `trigger_error`. 

You can see it when using a custom error handler such as [Query Monitor](https://github.com/johnbillion/query-monitor/blob/master/collectors/php_errors.php#L92-L95) or [Wonolog](https://github.com/inpsyde/Wonolog).

However, Twig will soon forbid the current behavior. The patch will make Timber Twig 3.0 ready.

jarednova:
Perfect, now I see it!

nlemoine:
> looking into how (if) we can test against diff't versions of dependencies.

@jarednova Just stumbled upon this post: https://blog.martinhujer.cz/17-tips-for-using-composer-efficiently/
Look at tip #7, it might be a good way to test Timber against different versions of Twig.
I gave it a try here: https://travis-ci.org/nlemoine/timber/builds/337731007

PR #1646, opened 2018-01-27
Documentation: Use relative links for linking between doc pages

gchtr:
#### Issue

The upcoming documentation will be versioned for v1 and v2 of Timber. The latest documentation can always be found under <https://timber.github.io/docs/>. The documentation for v1 will live under `https://timber.github.io/docs/v1/`. The links used to link between documentation are absolute now, so links under `/v1` would always point to the latest versions of the docs, instead of v1.

#### Solution

This pull request makes the links relative, so that links between documentation works as intended.


jarednova:
Thank you @gchtr — lookin' good to me

Issue #857, opened 2016-02-18
jpeg compression / quality


fujo:
Hi everybody

First of all thank you for the great job.

I didn't find how I can set the image quality (jpeg compression). Of course, [this topic](https://premium.wpmudev.org/blog/fix-jpeg-compression/) explains how to set it with a WP filter. However I don't see any changes (file size) in case of TimberImage. Have I missed anything?

Thanx for helping.
Fj

connorjburton:
Hi Fujo, have you tested this with new images after you have added the function to functions.php?

fujo:
Hello connorjburton

Thank you for the answer. Yes, it works well.

However I thought that the jpeg was compressed on client request, once the file is called, like the resize function generates new files.

I would like to optimize the image weight but I want to keep the highest image quality on the server.

It could be a great feature if it was possible to define a jpeg compression through twig templates. This function will generates a new file with decreased quality. Like this we could easily manage website/images performance.

I mean something like this: {{post.get_thumbnail.get_src|resize(600, 400)|**jpeg_quality(80)**}}

Best,
Fj

connorjburton:
Thanks for getting back to me. The original link you posted is just the WP filter for compression, it's not related to Timber itself, it will only effect new images, not existing images and won't compress at runtime.

However, with that said, it's possible there is scope for a compression filter in Timber, it's the question of is this something Timber should be handling or should be left to more fully fledged plugins (such as EWWW) to handle. Personally I am not sure how useful specifying a quality _per image_ is, why would you do that over just setting a static quality for all images on the site?

My thoughts summed up:

1. Ensuring images adhere to the compression quality set by `jpeg_quality` regardless if they are new images or not is something I'd love to solve, as long as it's not a detriment to performance
2. I am not sure how useful a quality filter would actually be, open to hear thoughts on this

Be good to get @jarednova thoughts on this

jarednova:
@fujo and @connorjburton: I can see the need to set image-level quality settings, but I have to confess that it strikes me as an edge case. The syntax I'd recommend would be something like...

`{{post.thumbnail.src | resize(600, 400) | quality(80)}}`

The thing this opens up is a bunch of other potential image filters like black-and-white, tint, etc. All of which I can see certain valid use-cases for. 

[...]

PR #1657, opened 2018-02-06
add optional extra parameter to resize method

columbian-chris:
add optional extra parameter to `resize` method of the `ImageHelper` class to customize jpeg compression quality of image being generated

**Related Ticket**: # [857](https://github.com/timber/timber/issues/857)

#### Issue
The default setting `set_quality` method of Wordpress' [WP_Image_Editor class](https://codex.wordpress.org/Class_Reference/WP_Image_Editor) is 90%, which is not a very desirable quality for jpeg compression for web use. Also, it would just be a really nice feature to have implemented to be able to customize what compression quality `resize()` puts out.

#### Solution
Add a new optional parameter to resize() call to customize jpeg quality, 0-100. Personal suggestion for a default setting if nothing passed: 60 (as reflected in my code).

#### Impact
Since it's an optional extra parameter added to the end of the current parameters, it should be backwards compatible.

#### Usage Changes
`{{ image.src | resize( 300, 200, 'top', false, 50 ) }}` (would produce 50% jpeg quality)

#### Considerations
- Would also maybe help to throw error if user passes something other than the accepted 0-100 integer.
- Would need to update [Timber docs](https://timber.github.io/docs/reference/timber-imagehelper/#resize).

### DIFF
--- a/lib/Image/Operation/Resize.php
+++ b/lib/Image/Operation/Resize.php
@@ -17,14 +17,14 @@ use Timber\Image\Operation as ImageOperation;
  */
 class Resize extends ImageOperation {
 
-   private $w, $h, $crop;
+   private $w, $h, $crop, $quality;
 
 	/**
 	 * @param int    $w    width of new image
 	 * @param int    $h    height of new image
 	 * @param string $crop cropping method, one of: 'default', 'center', 'top', 'bottom', 'left', 'right', 'top-center', 'bottom-center'.
 	 */
-	public function __construct( $w, $h, $crop ) {
+	public function __construct( $w, $h, $crop, $quality ) {
 		$this->w = $w;
 		$this->h = $h;
 		// Sanitize crop position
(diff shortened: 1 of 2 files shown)


jarednova:
@chrisguitarguy — thank you so much for this PR to resolve #857. I added a test and made a change so the default quality is the same as WP's (82). It's easy to override with WP's `wp_editor_set_quality` filter. Running the tests now and will then merge..

https://make.wordpress.org/core/2016/02/22/proposal-increase-the-default-image-compression-in-wordpress/

jarednova:
I'm going to close for now as a result of #1666 and the discussion in #857. Perhaps more to come here with a more flexible way of sending args to the `resize` filter. 

Thanks again @columbian-chris for your work here. Even though this didn't make it in verbatim, we've made good progress on integrating with WP-native compression settings and have a clear roadmap for allowing arbitrary adjustments of specific images

columbian-chris:
@jarednova Perhaps a good goal for the 2.0 release to be able to pass an object parameter with any custom settings?

``

PR #1658, opened 2018-02-12
add repeater in flexible content field example

stefenphelps:
update docs with an example of a repeater in a flexible content field

#### Issue
It took me a few minutes to figure out the proper way this is handled in twig

#### Solution
Having this specific example will help save time in the future

#### Impact
no impact. just a simple docs update.


gchtr:
Cool stuff @stefenphelps 😎! Thanks for making this pull request! This definitely helps others when they want to use repeater fields inside flexible content fields.

This will make it into the live documentation when we publish version 2 of Timber.

PR #1659, opened 2018-02-14
2.x Prepare structural elements for usage with phpDocumentor

gchtr:
#### Issue

The new documentation parser will use phpDocumentor to generate a Markdown documentation. Classes, properties and methods will only be included if they have an `@api` tag. This will make it easier to use the CLI when generating the documentation.

This pull request adds missing `@api` tags and fixes some other issues. It’s "a quick sweep" to bring v2 of the docs on the same level as v1 in terms of what will be included. A more thorough run through the DocBlocks will be required later to add missing summaries, descriptions and tags.

#### Testing

None, because only inline documentation was changed.

### DIFF
--- a/lib/Admin.php
+++ b/lib/Admin.php
@@ -2,6 +2,9 @@
 
 namespace Timber;
 
+/**
+ * Class Admin
+ */
 class Admin {
 
 	public static function init() {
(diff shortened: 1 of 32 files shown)


jarednova:
Wow, getting cleaner everyday. Let's do it!

Issue #1660, opened 2018-02-15
PostGetter::get_post_class unnecessarily instantiates return PostClass


StuartFeldt:
### Expected behavior
I expect the class wrapper for the returned post class in Timber::get_post to only be instantiated once.

### Actual behavior
The class wrapper is instantiated twice.

### Steps to reproduce behavior
Create a class wrapper which extends TimberPost, the constructor will be called twice.

### What version of WordPress, PHP and Timber are you using?
WordPress 4.9.4, PHP PHP 7.0.27, Timber 1.7.0

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Composer

jarednova:
Closed by #1663 and #1661

PR #1661, opened 2018-02-15
Fix double instantiation of post class in Timber::get_post

StuartFeldt:
Do not instantiate a new instance of $post_class_use

**Ticket**: #1660 

#### Issue
This solves the double-instantiation of the post class wrapper from Timber::get_post

#### Solution
Instead of checking is_a on an instantiated object, I am checking is_subclass_of

#### Impact
Should be backwards compatible, slight performance improvement if the Post Class wrapper's constructor is heavy.

#### Usage Changes
None

#### Considerations
None

#### Testing
Existing test cases should cover

### DIFF
--- a/lib/PostGetter.php
+++ b/lib/PostGetter.php
@@ -142,7 +142,7 @@ class PostGetter {
 			return $post_class_use;
 		}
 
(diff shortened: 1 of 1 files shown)


jarednova:
Thank you @StuartFeldt! I confirmed the behavior for myself with a test in #1663. Merged!

Issue #404, opened 2014-12-09
TimberImage Resize combined with Retina creates upscaled image, incorrect URL


mennolui:
My uploaded source is a 3500px wide image. I want to display this image width:100% at a viewport width of 1200px, on retina (2x) displays.

I could just resize to a width of 2400px...

``` html
<img src="{{ post.thumbnail.src | resize(2400) }}" />
```

..which works fine and the image looks great, but i wanted to use the retina filter as shown in the Timber docs (and proposed in issue #371):

``` html
<img src="{{ post.thumbnail.src | resize(1200) | retina(2) }}" />
```

This however generates an unsharp image. Looks like it was actually resized to 1200px, and then upscaled to 2400px.

Also, when using the retina filter, the URL returned is incorrect: '/wp-content/uploads/xxxx.jpg'. Without retina filter the correct full URL is returned, 'http://localhost/website-dir/wp-content/uploads/xxxx.jpg'.

Menno
(Using Timber 0.20.8, WP 4.0.1)

aristath:
I can confirm this.
It looks like retina calculations are off... in order to get something acceptable in the above example I had to so something like this:
```
<img src="{{ post.thumbnail.src | retina(2) | resize(2400) | retina(2) }}" />
```
applying the retina filter twice on an image double the size.

kylehotchkiss:
I've just seen a few casual references to `retina` - will this be able to detect the DPI of the users screen or is that a stretch?

simonmilz:
I can confirm this to. I use this also like this:

```twig
{{ product.thumbnail | retina(2) | resize(700, 600, 'center') | retina(2) }}
```

Twansparant:
Same here, this should be fixed.

kmbt:
That is because in current implementation if `Retina` is chained after resize it is given the url of an already resized image, not the original one. `Retina->run()` currently does not try to resolve the original image path. Also, the `resize`d image is generated as an intermediate step even if it not needed. 

One quick workaround would be to add a Twig filter like this one:

```
    function retinize( $src, $pixel_ratio, $w, $h = 0, $crop = 'default', $force = false) {
        $ih = new Timber\ImageHelper();
        if ( !is_numeric($w) && is_string($w) ) {
            if ( $sizes = $ih::find_wp_dimensions($w) ) {
                $w = $sizes['w'];
                $h = $sizes['h'];
            } else {
                return $src;
            }
        }

        $new_w = $w * $pixel_ratio;
[...]

PR #1672, opened 2018-02-26
Remove reference to resize | retina chaining

zackphilipps:
**Ticket**: #404

#### Issue
Docs state that you can chain the `resize` and `retina` filters, but there's a bug that makes this currently unusable.

#### Solution
Remove this section of the docs.

#### Impact
Nonezo.

#### Usage Changes
Nonezo.

#### Considerations
Fixing the bug would be a better solution, but not sure what's involved there.

#### Testing
Nonezo.


zackphilipps:
@gchtr thanks, and no worries! cheers.

PR #1677, opened 2018-03-01
Remove sniff that deviates from WordPress Coding Standards

gchtr:
**Ticket**: #1603 

#### Issue

Taking up the discussion about coding style and personal preferences in https://github.com/timber/timber/pull/1603. I propose to remove these rules, my reasoning can be found here: https://github.com/timber/timber/pull/1603#issuecomment-354163527. Sorry for being a little pedantic about this.

#### Solution

Remove sniffs that deviate from the WordPress Coding Standards without a strong reason that would justify that they stay.

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -66,11 +66,4 @@
         </properties>
     </rule>
 
-    <rule ref="PEAR.Functions.FunctionCallSignature">
-        <properties>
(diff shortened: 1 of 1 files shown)


jarednova:
I agree, let's follow the rules of the road

PR #1678, opened 2018-03-01
2.x Deprecate non-namespaced functions in Twig

gchtr:
**Ticket**: https://github.com/timber/timber/issues/1580#issuecomment-360980331

#### Issue

In Twig, the following functions still exist: `TimberPost`, `TimberImage`, `TimberTerm`, `TimberUser`. For PHP, we replaced all those instances of non-namespaced functions. To make it more consistent, let’s deprecate these in Twig as well. The shorter versions that work perfectly fine (even without Timber\ as a prefix) already exist: `Post`, `Image`, `Term`, `User`. We can leave them as they are.

#### Solution

- Deprecated `TimberPost`, `TimberImage`, `TimberTerm`, `TimberUser` in Twig. They don’t return anything.
- Updated tests.
- Updated guides where some of these function names were still used.

#### Testing

Yes.

### DIFF
--- a/docs/guides/acf-cookbook.md
+++ b/docs/guides/acf-cookbook.md
@@ -27,7 +27,7 @@ You can retrieve an image from a custom field, then use it in a Twig template. T
 ### The quick way (for most situations)
 
 ```twig
-<img src="{{TimberImage(post.get_field('hero_image')).src}}" />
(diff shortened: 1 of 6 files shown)


jarednova:
Awesome, feels like a relief to have those gone

PR #1681, opened 2018-03-03
2.x Streamline method and filter names for getting meta data

gchtr:
**Ticket**: #1580, picking up on #1621

## Issue

Streamline filter and function names.

Consider the following functions, which each have two newly added filters, adapted from their old (underscored) filter names:

### Get all custom fields

**Post::get_post_custom()**

- `timber/post/pre_get_meta`
- `timber/post/get_meta`

**Term::get_term_meta**

- `timber/term/meta`

**User::get_custom()**

- `timber/user/pre_get_meta`
- `timber/user/get_meta`

**Comment::get_meta_fields()**

- `timber/comment/pre_get_meta_fields`
- `timber/comment/get_meta_fields`

### Get value for one custom field

**Post::get_field()**

- `timber/post/pre_get_meta_field`
- `timber/post/get_meta_field`

**Term::meta()**

- `timber/term/meta/field`

**User::meta()**

- `timber/user/pre_get_meta_field`
- `timber/user/get_meta_field`

**Comment::get_meta_field()**

- `timber/comment/pre_get_meta_field`
- `timber/comment/get_meta_field`

### What could be improved

1. The functions practically all do the same, but they are called differently. We should streamline these. And, in addition, `Post::meta()` currently calls `Post::get_field()`. But couldn’t it be the other way around, that `get_field()` is an alias of `meta()`?
2. The filter names are not streamlined as well. I think they should match the function names.

### Streamline method names to get all custom field values

- `Post::get_meta_values()`
    - `timber/post/pre_get_meta_values`
    - `timber/post/get_meta_values`
- `Term::get_meta_values()`
    - `timber/term/get_meta_values`
- `User::get_meta_values()`
    - `timber/user/pre_get_meta_values`
    - `timber/user/get_meta_values`
- `Comment::get_meta_values()`
    - `timber/comment/pre_get_meta_values`
    - `timber/comment/get_meta_values`

### Streamline method names to get value for one custom field:

- `Post::meta()`
    - `timber/post/pre_meta`
    - `timber/post/meta`
- `Term::meta()`
    - `timber/term/meta`
- `User::meta()`
    - `timber/user/pre_meta`
    - `timber/user/meta`
- `Comment::meta()`
    - `timber/comment/pre_meta`
    - `timber/comment/meta`

### Deprecate namespaced filters

Where a a namespaced filter was already used (introduced in 0.21.x), I deprecated these as well.

- `timber/term/meta` in `Term::get_meta_values`
- `timber/term/meta/field` in `Term::meta`

The filter `timber/term/meta` in `Term::get_meta_values` was removed instead of being deprecated through `apply_filters_deprecated`, because there’s now a filter `timber/term/meta` in `Timber::meta()`. This might lead to weird behavior where `timber/term/meta` is still used in a theme.

### Improve DocBlocks

- Added summary, description and descriptions for parameter and return values for the `meta()` methods.

### Deprecation warnings

Added calls to `Helper::deprecated()` in the following deprecated methods:

- `Post::get_field()`
- `Term::get_meta_field()`
- `User::get_meta_field()`
- `User::get_meta()`
- `Comment::get_meta_field()`

## Testing

- I updated tests where they were failing.
- There’s an error in `TestTimberPost::testPostMetaMetaException`, but I’m not really sure what the test is supposed to be doing.
- Do we need to write new tests for something that’s changed here?

### DIFF
--- a/docs/guides/acf-cookbook.md
+++ b/docs/guides/acf-cookbook.md
@@ -14,7 +14,7 @@ While data saved by ACF is available via `{{post.my_acf_field}}` you will often 
 ```twig
 <h3>{{post.title}}</h3>
 <div class="intro-text">
-     {{post.get_field('my_wysiwyg_field')}}
+     {{post.meta('my_wysiwyg_field')}}
 </div>
 ```
 
@@ -27,7 +27,7 @@ You can retrieve an image from a custom field, then use it in a Twig template. T
 ### The quick way (for most situations)
 
 ```twig
-<img src="{{TimberImage(post.get_field('hero_image')).src}}" />
+<img src="{{TimberImage(post.meta('hero_image')).src}}" />
 ```
 
 ### The long way (for some special situations)
@@ -59,7 +59,7 @@ You can now use all the above functions to transform your custom images in the s
 ## Gallery field
 
 ```twig
-{% for image in post.get_field('gallery') %}
+{% for image in post.meta('gallery') %}
     <img src="{{ TimberImage(image) }}" />
 {% endfor %}
 ```
--- a/lib/Comment.php
+++ b/lib/Comment.php
@@ -124,7 +124,7 @@ class Comment extends Core implements CoreInterface {
 		$this->import($comment_data);
 		$this->ID = $this->comment_ID;
 		$this->id = $this->comment_ID;
-		$comment_meta_data = $this->get_meta_fields($this->ID);
+		$comment_meta_data = $this->get_meta_values($this->ID);
 		$this->import($comment_meta_data);
 	}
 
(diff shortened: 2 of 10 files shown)


jarednova:
@gchtr thanks for the fixes and write-up. The reason for that one weird failing test is for when someone has done something silly like name a custom field (the key that gets stored in the `wp_postmeta` table) as simply "meta" — how to best make it "fail gracefully"?

This extra catch/test provides the expected behavior ...

```
Foo {{ post.meta }} // Foo
update_post_meta($post_id, 'meta', 'bar');
Foo {{ post.meta }} // Foo bar
```

... but I can also imagine how a "loud failure" would be more helpful here (and also avoid unforeseen consequences and extra cruft in the codebase). 

I'm honestly torn on how to best handle. Do you have an impulse one way or the other (and do I make sense here)

gchtr:
@jarednova Isn’t this the same topic that we started to discuss in https://github.com/timber/timber/pull/1618#issuecomment-356441760?

jarednova:
@gchtr not _quite_ (but also, very close — funny how the same stuff keeps coming). The advice that you left there that I'll apply here is using `Helper::warn` to alert that they've made a potentially poor choice of field names which could cause problems. That seems reasonable and helpful

PR #1682, opened 2018-03-04
2.x Remove backwards compatibility for non-namespaced class names

gchtr:
**Ticket**: https://github.com/timber/timber/issues/1580#issuecomment-360980331

#### Issue

Since namespaced classes were introduced in 1.x, we’re trying to promote the use of the namespaced classes. Until now, there was backwards compatibility through class aliases. Version 2.x would be a good moment to ditch this.

However, one alias would make sense for ease-of-use: `Timber` as an alias for `Timber\Timber`. This is useful because we won’t have to write `Timber\Timber::render()`, but can still use `Timber::render()`.

#### Solution

- Remove method `Timber::backwards_compatibility()`. Move alias for `Timber\Timber` to `Timber::init()`.
- Remove class aliases for `Timber\PostCollection`.

#### Impact

There will be fatal errors for developers who use non-namespaced class names in PHP. A section in the upgrade guide that will be added in a separate pull request can help.

#### Testing

I ran tests and removed `TestTimberPostCollection::testBackwards()`, because there’s no backwards compatibility anymore.

### DIFF
--- a/lib/PostCollection.php
+++ b/lib/PostCollection.php
@@ -133,6 +133,3 @@ class PostCollection extends \ArrayObject {
 		return $posts;
 	}
 }
-
-class_alias('Timber\PostCollection', 'Timber\PostsCollection');
-class_alias('Timber\PostCollection', 'TimberPostsCollection');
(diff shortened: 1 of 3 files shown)


jarednova:
Indeed, it's au revoir to some old friends. To the future! (but still with `Timber::render`)

PR #1700, opened 2018-03-30
2.x Update methods to retrieve site option

gchtr:
**Ticket**: #1580 

#### Issue

Up until now, site option values could be retrieved through `{{ site.option_name }}` or `{{ site.meta('option_name') }}`. The term `meta` doesn’t fit very well for a site option, because the value doesn’t live in a meta database table, but in `wp_options`. It would be more intuitive to have the following methods available for retrieving site options:

- `{{ site.option_name }}` (leave as is)
- `{{ site.option('option_name') }}`

#### Solution

- Add `option()` method that retrieves a site option through `__get()`.
- Deprecate `meta()` method in favor of `option()`.
- Update DocBlocks.

#### Testing

- Added test for new `option()` method.
- Updated test for `meta()` to catch deprecation.

### DIFF
--- a/lib/Site.php
+++ b/lib/Site.php
@@ -229,18 +229,57 @@ class Site extends Core implements CoreInterface {
 	}
 
 	/**
-	 * @param string  $field
-	 * @return mixed
+	 * Get the value for a site option.
+	 *
+	 * @api
+	 * @example
+	 * ```twig
+	 * Published on: {{ post.date|date(site.date_format) }}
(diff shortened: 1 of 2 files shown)


jarednova:
Looks good! While I love the idea of `meta` everywhere; this is where people expect the WP convention of "Terms/Users/Posts have meta, Sites have options" (but we still have meta as a backup for now) — merged!

Issue #1629, opened 2018-01-06
Change is_external() function in MenuItem.php to respect "Open in new tab" checkbox


gserafini:
### Expected behavior
If a user clicks the "Open in new tab" checkbox in Dashboard > Appearance > Menus, link should be opened in a new tab, regardless of if it in the same domain or subdomain or not, if the helper `is_external` function code is used.

### Actual behavior
Menu links that have "Open in new tab" do not open in a new tab.  The `is_external` function is hard-coded to return false in the case of a subdomain, and true for all other domains that are different from the base domain.

### Steps to reproduce behavior
1. Open a menu in Dashboard > Appearance > Menus
2. Add a link
3. Click the "Open in new tab" checkbox (to view this, make sure that Targets is selected in the Screen Options
4. Load page that includes a Timber navigation menu with the suggested usage code: `<a href="{{ item.link }}" target="{{ item.is_external ? '_blank' : '_self' }}">`
5. Click the link that you specified to open in a new tab.  If it is an internal or subdomain link it will open in the same window, and external custom URLs (excluding subdomains) will always open in a new window, regardless of the WordPress setting for that link.

### Suggested solution
Update the is_external() function call in lib/MenuItem.php to check for the existence of the checkbox preference and respect it if it's set.  Get rid of old behavior that assumes that all non-subdomain links should be opened in a new window.  Or... make a new function that does this below, maybe called is_new_tab() or something.

~Line 282 in MenuItem.php:

```
	public function is_external() {
		if ( isset ( $this->_menu_item_target ) && $this->_menu_item_target ==  "_blank") {
			return true;
		}
		return false;
	}
```

### What version of WordPress, PHP and Timber are you using?
WordPress 4.9.1 (latest), PHP 7.0.26, Timber 1.6.0

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Upgraded to newest version via plugin updater in WordPress dashboard.

gchtr:
https://github.com/timber/timber/issues/1629

Hey @gserafini, thanks a lot for taking the time to properly fill out the issue template!

I very much like your idea to respect the menus setting for opening a menu item in a new tab.

I think that the idea behind `is_external()` is really to tell whether the link is to an external site (other domain or subdomain). It might for example also be used to display an icon next to the link for better accessibility. So I’d go for your suggested alternative solution to use another function name for your idea.

[...]

PR #1701, opened 2018-04-01
2.x Introduce is_target_blank method for MenuItem

gchtr:
**Ticket**: #1629

#### Issue

There is no easy way to check whether the «Open link in new tab» option is checked on a menu item in the backend.

#### Solution

- Introduce a new method `MenuItem::is_target_blank()` that checks whether the post meta entry `_menu_item_target` is set to "blank".
- Add fallback return value of `null` for `MenuItem::meta()`.
- Add documentation for `is_target_blank()` and `is_external()` to use them in combination.

#### Impact

None.

#### Usage Changes

Item targets can now be checked with

```twig
<a href="{{ item.link }}" {{ item.is_target_blank ? 'target="_blank"' }}>
```

And in combination with `MenuItem::is_external()`:

```twig
<a href="{{ item.link }}" {{ item.is_target_blank or item.is_external ? 'target="_blank"' }}>
```

#### Testing

I added a test to check for the different states:

- No `_menu_item_target` meta value set on a menu item
- Meta value "blank" for `_menu_item_target`.
- Meta value "" for `_menu_item_target`.

### DIFF
--- a/lib/MenuItem.php
+++ b/lib/MenuItem.php
@@ -264,6 +264,27 @@ class MenuItem extends Core implements CoreInterface {
 	}
 
 	/**
+	 * Checks whether the «Open in new tab» option checked in the menu item options.
+	 *
+	 * @example
+	 * ```twig
+	 * <a href="{{ item.link }}" {{ item.is_target_blank ? 'target="_blank"' }}>
+	 * ```
+	 *
+	 * In combination with `is_external()`
+	 *
+	 * ```twig
+	 * <a href="{{ item.link }}" {{ item.is_target_blank or item.is_external ? 'target="_blank"' }}>
+	 * ```
+	 *
+	 * @return bool Whether the menu item has the «Open in new tab» option checked in the menu item
+	 *              options.
+	 */
+	public function is_target_blank() {
+		return '_blank' === $this->meta( '_menu_item_target' );
+	}
+
+	/**
 	 * Get the type of the menu item.
 	 *
 	 * Depending on what is the menu item links to. Can be `post_type` for posts, pages and custom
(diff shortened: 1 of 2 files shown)


jarednova:
Hey @gchtr, thanks as always for the PR and tests. I'm wondering if the `{{ item.target }}` approach you had suggested in #1629 might be a more universal solution though. The Twig result would be...

Item targets can now be checked with
```twig
<a href="{{ item.link }}" target="{{ item.target }}">
```
And in combination with MenuItem::is_external():
```twig
<a href="{{ item.link }}" {{ item.target is '_blank' or item.is_external ? 'target="_blank"' }}>
```

I realize it's a small distinction, but it seems (slightly) more versatile. While I agree with @gserafini that the possibilities are pretty limited, this would eliminate the need for a conditional in menus (only requiring it if someone wanted the external or blank logic). What say you?

gchtr:
@jarednova The only reason I didn’t choose to use `target="{{ item.target }}"` is that `target="_self"` would be an obsolete attribute. I also thought about having

```twig
<a href="{{ item.link }}" {{ item.target }}>
```

which could add `target="_blank"` if necessary, but throughout Timber, we only ever add the value of an attribute, and not a whole attribute (right ?). And I didn’t think about reducing **conditionals in menus**, which is a really good reason. So yeah, I see, we should probably go with:

```twig
<a href="{{ item.link }}" target="{{ item.target }}">
```


But why don’t we add both? Would we add redundancy if developers could still use `item.is_target_blank()`?

```twig
<a href="{{ item.link }}" {{ item.is_target_blank or item.is_external ? 'target="_blank"' }}>
```

And I see, before we merge this, I should probably update the menu guides and add `target="{{ item.target }}` where necessary.

jarednova:
I'm good with both! (and the gif that goes along with it). Let's do it

PR #1708, opened 2018-04-23
WP filters return object

drzraf:
#### Issue
Under some configuration, adding a filter like the one below:
```php
    $twig->addFilter( new Twig_SimpleFilter( 'bar', function ( $value ) { return "foo"; } ) );
} );
```
Fails with `Call to a member function addFunction() on null`

#### Solution
WordPress filters return an object otherwise NULL is assumed what would replace the `$twig` object.

#### Impact
Fix documentation and respect WP best practices.


jarednova:
Indeed, thank you @drzraf !

Issue #1709, opened 2018-04-28
Feature Request: overrideable way to instantiate MenuItems


acobster:
I'd like to be able to override how `MenuItem` instances are created. Right now, `new $this->MenuItemClass($item)` is bound up inside `Menu::order_children`. If it were in its own method, I could override that in order to, for instance, polymorphically create different kinds of `MenuItem`s:

```php
class CustomMenu extends Timber\Menu {
  protected function create_menu_item($item) {
    if ($this->is_login_nav_item($item)) {
     return new LoginMenuItem($item);
    }

    return new $this->MenuItemClass($item);
  }
}
```

In this example, `LoginMenuItem` might return a different title/URL depending on the user's login state. But we don't necessarily want all items to be instances of that class.

gchtr:
Thanks for providing a pull request for this 💪. I guess we can close this now.

PR #1710, opened 2018-04-28
#1709 break $MenuItemClass instantiation into own method

acobster:
**Ticket**: #1709 

#### Issue

I'd like to make the instantiation of `$MenuItemClass` instances more modular and overrideable in subclasses of `Menu`.

#### Solution

Create a new protected method, `create_menu_item`, that is solely responsible for instantiating Timber-flavored menu items.

Refactor `order_children` to call `create_menu_item` instead of instantiating directly.

#### Impact

All the polymorphism.

#### Usage Changes

If I've done everything right, this should be a pure refactor with no effect on behavior.

#### Considerations

This PR is literally the pinnacle of perfection and there is no possible way to improve it.

srsly tho idk, p simple

#### Testing

If this breaks any unit tests, I'll buy someone a beer.

### DIFF
--- a/lib/Menu.php
+++ b/lib/Menu.php
@@ -223,7 +223,7 @@ class Menu extends Core {
 					$old_menu_item = $item;
 					$item = new $this->PostClass($item);
 				}
-				$menu_item = new $this->MenuItemClass($item);
+				$menu_item = $this->create_menu_item($item);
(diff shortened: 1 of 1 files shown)


jarednova:
@acobster looks like you don't owe me a beer! Once I realized what was going on, the cleanliness of this solution really "clicked" — nice work!

Issue #1465, opened 2017-06-25
Gulp task to generate POT file and .php clone files for WPML


luism-s:
This isn't an issue, but more of a procedure suggestion. I don't know what would the best place for something like this  be, so I'll post this here.

I was having an hard time adapting POEdit to scan .twig files, so I figured that I needed a different approach in order to register the strings from .twig files in WPML.

So I coded a gulp script with the following logic:
- Iterates over all given .twig files
- Search and Replace for gettext functions in .the twig files and wraps them around php tags
- Outputs each file as .php
- Send files to a  cache folder
- Iterate over all .php files (cache included)
- Generate .pot 
- ta  daaaa

Dependencies:
````
npm install --save-dev del gulp gulp-if run-sequence gulp-rename gulp-replace gulp-wp-pot
````

Code (as if the script is in the theme's root):

````

/**
 * Gettext Scanner Script for Twig Projects
 * v1.0
 * 
 * Developed by luism-s
 * https://github.com/luism-s
 */

/**
 * Usage:
 * `gulp pot`
 *
 * Logic:
 *
 * - Iterates over all given .twig files
 * - Search and replace for gettext functions in Twig files and wraps them around PHP tags
 * - Outputs each file as .php into a cache folder
 * - Scan all .php files for gettext functions using 'gulp-wp-pot' (cache included)
 * - Generate .pot file
 */

/**
 * Dependencies:
 *
 * `npm install --save-dev gulp gulp-wp-pot gulp-replace gulp-rename del gulp-if run-sequence`
 */
var gulp    = require('gulp');
var wpPot   = require('gulp-wp-pot');
var replace = require('gulp-replace');
var rename  = require('gulp-rename');
var del     = require('del');
var gulpif  = require('gulp-if');
var runSequence  = require('run-sequence');

/**
 * Configuration Options
 * 
 * All paths are as if this script is 
 * located in the root of the theme and all the Twig
 * files are located under /views
 */
var config = {
  "text_domain" : "theme-test",       // Replace with your domain
  "twig_files"  : "views/**/*.twig",  // Twig Files
  "php_files"   : "**/*.php",         // PHP Files
  "cacheFolder" : "views/cache",      // Cache Folder
  "destFolder"  : "languages",        // Folder where .pot file will be saved
  "keepCache"   : true                // Delete cache files after script finishes
};

/**
 * Main Task
 */
gulp.task('pot', function(callback) {
  runSequence('compile-twigs', 'generate-pot', callback);
});

/**
[...]

PR #1711, opened 2018-05-01
Documentation: Add new method for generating POT files

gchtr:
**Ticket**: #1465 

Add a section in the internalization guide about a method to generate POT files with a gulp script. This might be helpful for developers who do not want or can’t work with Poedit.

Thanks to @luism-s for coming up with the idea and for sharing it!


jarednova:
Awesome, thanks @gchtr and @luism-s !

PR #1712, opened 2018-05-09
Previewing changes made to ACF fields

braican:
**Related to Ticket**: #836  

#### Issue
When previewing a post that utilizes ACF fields, it appears that Timber is inconsistently handling the revision ID when populating the custom field values in the `TimberPost` class. To get ACF fields to show up properly when previewing a post, you need to use `{{ post.get_field('custom_field') }}`, as using `{{ post.custom_field }}` won't output the previewed value.

#### Solution
In [PR 876](https://github.com/timber/timber/pull/876) some code was added to check if the user is previewing a post, and then returning the ID of the revision that is used as the preview. However, this check was only happening if no post ID was passed to the `TimberPost` instance.

Adding another check to the `determine_id()` method to see if the user is previewing a post, regardless of whether or not the ID was passed, seems to fix the issue.

Worth noting is that this check will also only run if the current `queried_object_id` is the same as the determined `$pid` in the logic above. This will prevent this code from running on *all* calls to the `TimberPost` class that happen on a previewed page.

#### Impact
Shouldn't be much...if the user is not previewing a page then this additional code won't run.

#### Usage Changes
None.

#### Considerations
It's possible (probable?) that the preview check in the first `if` clause in the `determine_id()` method can be removed entirely. I'm thinking it can but I just don't know enough of the codebase to know for sure.

#### Testing
Unit test has been added.

### DIFF
--- a/lib/Post.php
+++ b/lib/Post.php
@@ -252,11 +252,11 @@ class Post extends Core implements CoreInterface {
 			$pid = $pid_from_loop;
 		}
 		if (
-			isset($wp_query->queried_object_id)
-			&& ($wp_query->queried_object_id === $pid || (is_object($pid) && $wp_query->queried_object_id === $pid->ID))
-			&& isset($_GET['preview'])
+			isset($_GET['preview'])
 			&& isset($_GET['preview_nonce'])
 			&& wp_verify_nonce($_GET['preview_nonce'], 'post_preview_'.$wp_query->queried_object_id)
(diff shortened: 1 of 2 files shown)


jarednova:
Thanks again @braican. I reorganized things in #1713 so that there's a `Post::is_previewing()` method that can contain that logic. Thanks for the tests!!!!!

PR #1715, opened 2018-05-17
2.x Introduce target method for MenuItem

gchtr:
**Ticket**: #1629 and #1701

#### Issue

Picking up on the discussion that we had in #1701 whether it would make sense to have an `item.target` method.

#### Solution

- Added a new method `item.target` for menu items that can be used instead of `item.is_target_blank` and adds a little more flexibility for developers.
- Updated documentation for menu items. See the updates to the menu guide for what’s possible.

#### Impact

None.

#### Usage Changes

Makes handling menu item targets more versatile.

#### Testing

Yes. I added tests, they ran successfully.

### DIFF
--- a/docs/guides/menus.md
+++ b/docs/guides/menus.md
@@ -49,7 +49,7 @@ Optionally, you can send additional options to `Timber\Menu`. Current only `dept
 
 ```php
 $args = array(
-	'depth' => 2,
+    'depth' => 2,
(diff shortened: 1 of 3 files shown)


jarednova:
Thanks @gchtr! I think this one is sufficiently covered now. Merging and closing #1629

Issue #1704, opened 2018-04-11
Feature Request -> Timber\Menu->getCurrent()


waltercruz:
### Expected behavior
Hi! I think that adding a getCurrent to Timber Menu could be a good idea. What do you guys think about it?

waltercruz:
It would return the current menu from Timber\Menu-> items

I'm currently making a single page application, and the menu items of the page are wordpress items. When I click though items, it moves to the target (anchor) and changes the class of the current section to current-menu-item. 

I'm using vue.js to do this, and I need to initialize the first state of the menu though vue.

It's a similar pattern to "Utilising beforeMount and data attributes to inject data" on this article: https://nehalist.io/directly-injecting-data-to-vue-apps-with-symfony-twig/

By current pattern is like this:

`{% for item in items %}`
`// print menu items`
`{% if item.current %}<input type="hidden" ref="active" data-value="{{item.slug}}" />{% endif %}`
`{% endfor %}`

It feels weird that I need to test for current on each iteraction of the loop, would be nice to have this outside of the loop, I think.

What do you guys think about it?

acobster:
@waltercruz I think what you are looking for is something like:

```
<input type="hidden" data-value="{{ menu.get_current_item.slug }}">
```

Right?

(Here I'm using the name `get_current_item`, since `get_current` is ambiguous - it seems like it might refer to the current _menu_, whatever that would mean. :smiley: )

I actually implemented this in a library plugin I'm working on that's built on top of Timber, but I'm thinking more and more that this would be a great bit of logic to push down into Timber layer. All that is to say, if I understand what you're asking for, +1!

gchtr:
@waltercruz 

> It feels weird that I need to test for current on each iteraction of the loop, would be nice to have this outside of the loop, I think.

I agree. I actually also stumbled over a use case for this recently, when I wanted to display child menu items of the current page outside the loop of the main menu.

@waltercruz I think it would be a great addition for Timber. If you already have a solution that you’re willing to share, could you do a pull request for it? Naming it `get_current_item` would make sense to me, too.

waltercruz:
As I was in a hurry, I did with a loop by now.

If I think about a good implementation I will submit it

gchtr:
@waltercruz Sorry, I first referenced you for the second paragraph, but actually it was intended for @acobster, so I changed the mention.

But anyways, if anyone of you wants to come up with a solution, that would be great!

acobster:
[...]

PR #1716, opened 2018-05-19
Add Menu::get_current_item() helper

acobster:
**Ticket**: #1704 

#### Issue

Current solutions for getting the current menu item outside the main nav loop involve looping over all menu items explicitly. This is clunky.

#### Solution

Add a `Menu::get_current_item()` method that returns the current item, or the closest ancestor.

#### Impact

Timber is more awesome now.

#### Usage Changes

No breaking changes; strictly new code. I did add a note to the changelog for 2.0. Oh and MOAR DOCS. :)

#### Considerations

It'd be nice to add a `$depth` parameter in the future, to limit the depth of the tree traversal: e.g. calling `get_current_item(1)` (or 0?) would only scan the top menu level for the current/ancestor item. (I think the `depth` constructor option is 1-based? In any case it should be consistent with that.)

Once we have that it'd be trivial to implement a `get_current_top_level_item` helper, which is a use-case I encounter all the time. It might also be nice to add a `get_current_item_parent` helper.

#### Testing

https://www.daedtech.com/wp-content/uploads/2012/12/TestAllTheThings.jpg

### DIFF
--- a/docs/guides/menus.md
+++ b/docs/guides/menus.md
@@ -136,6 +136,28 @@ Here’s an example to display child menu items only if it’s the menu item is 
 
 Other properties that are available are `current_item_parent` for direct parents of a menu item and `current_item_ancestor` for when you have deeper nesting.
 
+### Getting the current menu item outside the loop
+
+Say you want to display sibling menu items of the current page, for example in
+a sidebar. You can use the `get_current_item()` helper to achieve this:
+
+**Twig**
+
+```twig
+<div class="sidebar">
+  <a href="{{ menu.get_current_item.link }}">
+    {{ menu.get_current_item.title }}
+  </a>
+  <ul>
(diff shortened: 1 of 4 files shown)


acobster:
I just knocked out the `$depth` param and `get_current_top_level_item()`. I may circle back to this to refactor for readability (recursion, anyone?), but unit tests are passing and I'm feeling good about the documentation. Still might be nice to have a `get_current_item_parent()` helper at some point.

jarednova:
Thanks @acobster! Taking a look at this now ...

jarednova:
Thanks again @acobster ! I made a few tweaks in #1724 ...

* Removed `get_` from public function names
* Made `null` the default to return all values instead of `-1`

... this is now merged this into `2.x`!

PR #1729, opened 2018-06-13
Term_id field name in Term::get_posts

shvlv:
Use standard field name https://codex.wordpress.org/Class_Reference/WP_Query#Taxonomy_Parameters

**Ticket**: # 

#### Issue

Using id field in WP_Query args impact in third party plugin (hooking at 'pre_get_posts' for example)

#### Solution

Change `id` ot `term_id` - standard name.

### DIFF
--- a/lib/Term.php
+++ b/lib/Term.php
@@ -260,7 +260,7 @@ class Term extends Core implements CoreInterface {
 			$PostClass = $this->PostClass;
 		}
 		$default_tax_query = array(array(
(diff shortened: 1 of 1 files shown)


jarednova:
Thanks for this catch @shvlv — I'm surprised that it worked in the first place. It seems that when sent something that's not a valid "field", WP defaults to `term_id` which is why it worked previously (even though it was wrong) — merged!

PR #1726, opened 2018-06-08
SVG resize

palmiak:
**Ticket**: https://github.com/timber/timber/issues/1703

The problem is described in the ticket. In short Timber's Resize sometimes try to resize SVG file which makes an nonexistant image. To fix this I've added a is_svg filter.

I've added it in resize, tojpg, letterbox and retina.

PR #1736, opened 2018-07-05
1726/svg check

jarednova:
Tests and follow-up for #1726


jarednova:
@palmiak thank you again for this contrib! Fixing some last style issues and will then merge. 

### Future possibilities:
- [ ] Allow SVGs to be converted to raster images (I attempted, but it seems to require additional libraries that we can't count on)
- [ ] Allow PDFs and other non-raster image media types to be handled with similar logic

palmiak:
I saw that SVG could also be saved as plain text. I'll add this today.

pon., 9 lip 2018, 21:14 użytkownik Jared Novack <notifications@github.com>
napisał:

jarednova:
@palmiak — yep, one step ahead of you on the `text/plain`. As far as PDFs, etc. I think that's a next step from here after merging. The methods you created were non-public so I'm comfortable refactoring in the future to account for those — are you interested in picking that up as yr next PR from here?

palmiak:
Of course I am :)

PR #1737, opened 2018-07-08
Update PHPDoc @return for $id and thumbnail()

0x6368656174:
#### Issue
Wrong PHPDoc `@return` types for `Post.id` and `Post.thumbail()`

#### Solution
Fix PHPDoc `@return` types

#### Impact
No impact


jarednova:
awesome, thank you for those fixes @0x6368656174 !

Issue #1739, opened 2018-07-09
Theme::path() does not work correctly, if the URL specifies a port


0x6368656174:
### Expected behavior
`Theme::path ()` on the site that is running on http://localhost:4200 should return '/wp-content/themes/my-timber-theme'.

### Actual behavior
Returns 'http://localhost:4200/wp-content/themes/my-timber-theme'.

The problem is related to the fact that `Timber::path()` uses `URLHelper::get_rel_url()` https://github.com/timber/timber/blob/9eae5dbd8e899eb9e947a4abbdfb1c425a66c98d/lib/Theme.php#L117, which in its code compares the host returned by `parse_url()` to the host returned by `self::get_host()` https://github.com/timber/timber/blob/9eae5dbd8e899eb9e947a4abbdfb1c425a66c98d/lib/URLHelper.php#L91-L93 and this check will always be `false` if the address specifies a port. `parse_url()['host']` returns only the host without the port, and `self::get_host()` returns `$_SERVER ['HTTP_HOST']`, which is taken from the URL and contains the port https://github.com/timber/timber/blob/9eae5dbd8e899eb9e947a4abbdfb1c425a66c98d/lib/URLHelper.php#L114-L116.

### Steps to reproduce behavior
Specify the port in the address bar.

### What version of WordPress, PHP and Timber are you using?
Wordpress 4.9.7, Timber 1.7.1

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Composer

PR #1740, opened 2018-07-09
Them::path() use `force = true` to work with specifying the port

0x6368656174:
#### Issue
https://github.com/timber/timber/issues/1739

#### Solution
In order to turn off the `parse_url()['host'] `and `self::get_host()` check, we pass ` force = true`, so that `Timber::path ()` works correctly when the site is opened with the port number ,

#### Impact
It should not be, because in `URLHelper::get_rel_url()` it is passed `Theme::link()`

#### Usage Changes
No changes


jarednova:
Thanks @0x6368656174 ! I just added a test in #1741 to cover the behavior. Merged!

PR #1741, opened 2018-07-10
1740/theme path test

jarednova:
Test for #1740


jarednova:
Thanks @0x6368656174 ! I just added a test to cover the behavior

Issue #1381, opened 2017-04-04
Warning: call_user_func_array() expects parameter 1 to be a valid callback, no array or string given in C:\xampp\htdocs\marine\wordpress\wp-content\plugins\timber-library\lib\Twig.php on line 268


joethuoo:
Is this a support or usage question? Please post to [StackOverflow](http://stackoverflow.com/questions/tagged/timber) using the Timber tag

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)

WHY DO I GET THIS? Warning: call_user_func_array() expects parameter 1 to be a valid callback, no array or string given in C:\xampp\htdocs\marine\wordpress\wp-content\plugins\timber-library\lib\Twig.php on line 268

jarednova:
Hey @joethuoo thanks for posting, but I'm afraid I need more info in order to recreate the issue. Could you post more information so I can reproduce?

joethuoo:
I
![screenshot 1](https://cloud.githubusercontent.com/assets/3135231/24764449/0857799e-1afd-11e7-9d75-ce87ccc84548.png)
 installed Timber via installing the plugin on WordPress admin. 
The actual behavior: Everything appears as it should, because i haven't started the hooking the posts. This is just from adding my html css js and images from my designer.
I've only worked on the header and footer.
Nothing has been tinkered on from the back-end PHP files.
I am using (WordPress version 4.7.3) PHP 5.6.30 on xampp, Timber 1.2.4

frecklebit:
I'm getting the same thing.

frecklebit:
I opened the lib/Twig.php file and added an @ to suppress the warning but this is something that should be looked into.

joethuoo:
@frecklebit I had this solution from stackoverflow, but i can't tell what happened there. Further should be looked into. 
 http://stackoverflow.com/questions/43263238/warning-call-user-func-array-expects-parameter-1-to-be-a-valid-callback-timb?noredirect=1#comment73729489_43263238

bderstine:
I had the same error and googling brought me here, so I actually answered this on the stackoverflow page too, but the real issue I was running into was trying to use the fn() function to check a wordpress function, I needed to add single quotes around the name of the function. Error gone!

`{% if fn('wp_is_mobile') %}`

gchtr:
[...]

PR #1746, opened 2018-07-12
added woocommerce error detail

Mikewalker24:
#### Issue
See Issue #1381. This PR adds the text of the error message so that google searches will lead to this guide.


jarednova:
Sometimes the simplest changes are the most profound. Thank you @Mikewalker24, and congrats on your first contrib.!

Issue #1748, opened 2018-07-17
Not able to init in TimberImage with iid=//www.gravatar.com/avatar/0da2f1a9340d6bf196b510f6c107a248?s=96&#038;r=g&#038;d=mm


henrywright:
We are getting a message in our error log which relates to Timber.

> Not able to init in TimberImage with iid=//www.gravatar.com/avatar/0da2f1a9340d6bf196b510f6c107a248?s=96&#038;r=g&#038;d=mm

The message seems to be generated in the `init()` method in timber-library/lib/image.php

We are using WP 4.9.7, PHP 7.2 and Timber 1.7.1

jarednova:
@henrywright I think it's recognizing that as a server path as opposed to a URL — can you see if there's a way to prefix the gravatar URL with `http` or `https`?

henrywright:
Can a server path have a `//` in it? If not then you could just do this:
```
if ( strstr( $iid, '//' ) ) {
    $this->init_with_url( $iid );
    return;
}
```

jarednova:
Thanks @henrywright for the issue and PR! Just merged in #1749 with @mahagr's adjustment

henrywright:
Thanks @jarednova 

@mahagr's suggestion is a good improvement

PR #1749, opened 2018-07-17
Check for // instead of ://

henrywright:
This PR relates to #1748

### DIFF
--- a/lib/Image.php
+++ b/lib/Image.php
@@ -231,7 +231,7 @@ class Image extends Post implements CoreInterface {
 		}
 
 		if ( !is_numeric($iid) && is_string($iid) ) {
-			if ( strstr($iid, '://') ) {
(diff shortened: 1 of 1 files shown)


lib/Image.php
```diff
@@ -231,7 +231,7 @@ public function init( $iid = false ) {
 		}
 
 		if ( !is_numeric($iid) && is_string($iid) ) {
-			if ( strstr($iid, '://') ) {
+			if ( strstr($iid, '//') ) {
```
mahagr: It should really be something like "starts with // or contains ://":

```php
if ( strpos($iid, '//') === 0 || strstr($iid, '://') ) {
```

jarednova:
I agree @mahagr. Thanks for your review! I'm integrating your suggestion into this PR (tests running now....)

henrywright:
Good suggestion @mahagr

PR #1750, opened 2018-07-18
Integration of Square Payment API to CPT using Timber and ACF

jake46a:
**Ticket**: # 

#### Issue
Integration of payment processor with Timber

#### Solution
Simple Integration using Square API PHP SDK

#### Impact
NONE


Levdbas:
Hi @jake46a, thank you for your work back then. Unfortunately we are not going to integrate this in Timber since we really want to keep our core as minimal as possible in the future.

PR #1753, opened 2018-07-20
Don't overwrap WordPress i18n functions

drzraf:
**Ticket**: # 

#### Issue
Normally, when the argument of a Twig Function is a Callable, Twig creates such call (see [Twig's `compileCallable()`](https://github.com/twigphp/Twig/blob/2.x/lib/Twig/Node/Expression/Call.php#L43)).
Eg:
```
{{__("foo")}}
=> echo twig_escape_filter($this->env, call_user_func_array($this->env->getFunction('__')->getCallable(), array("foo")), "html", null, true)
```

This has the side effect that effect resulting template loose the xgettext friendly syntax.

#### Solution
By using feeding string to Twig Function, the result is clearer (to xgettext):
```
{{__("foo")}}
=> echo twig_escape_filter($this->env, __("foo"), "html", null, true);
````

#### Impact
Since wrapping these functions is of no use (same arguments), and since compiled templates containing these functions  may very well have to be looked into for gettext*() function calls, it's better to avoid it.
No impact is expected (other than reducing function calls)

#### Usage Changes
no

#### Testing
no


jarednova:
WOW is that cleaner. I added a test in #1759 to confirm everything still works — and it does. Merged!

PR #1755, opened 2018-07-24
Fixed URLHelper on https as no port is needed

pascalknecht:
#### Issue
The URL Helper adds the port 443 if the website is served via https

#### Solution
I added a condition to not include the port if the page is served via https

#### Impact
Should have none

#### Usage Changes
No


pascalknecht:
Perfect @jarednova . Thanks for merging.

Issue #1673, opened 2018-02-26
Force Timber to use Twig 2.x?


gsalkin:
Is there a way to pass a flag or force Timber to use Twig 2.x? My installation with Timber 1.7 continues to use Twig 1.34.1. Would like to take advantage of some new features in the latest version of the templating language.

jarednova:
@gsalkin this feels like a composer thing. Based on my tests, composer will use Twig 2.x when you install on > PHP 7.1. How are you installing?

gsalkin:
The initial install was via composer but updates are via WordPress admin. I can re-install via composer but I'm running PHP 7.0.x. That a problem?

jarednova:
@gsalkin ahhhh, that makes sense. I _think_ what happened is you had things right at first, but then WP.org over-writes with its version (which I keep low to help people still on PHP 5.x). You should be able to do a fresh install and get Twig 2.x since you're running PHP 7 (and yes PHP 7.0 _should_ be okay).

I brings up a good Q for people who install via Composer. Should we allow upgrades via the WP Admin?

gsalkin:
@jarednova Ideally we'd all be using Composer, right? But several clients see an update for Timber and run the update via WP Admin. I'd recommend removing Timber upgrades from WP Admin. 

A possible goal for #1617 and future version of 2.x would be creating parity between Composer and WP Admin installs (announce users _must_ have >= 7.0 installed on their servers). However, most Timber users would be more advanced and already have a Composer workflow.

jarednova:
Good note @gsalkin — I created #1688 as a future to-do (always open to PRs for anyone who wants to take a stab at it)

gchtr:
@gsalkin Out of curiosity: When you install Timber through Composer, why did you decide to install it the plugins folder? We always install Timber as a dependency of the theme, so that it’s not listed in the plugins list, exactly because we want to update Timber through Composer and not through WordPress updates. Timber is often tightly coupled to the theme. Updating it through WordPress could break the theme, because WordPress plugin updates don’t care about semantic versioning.

I’m just thinking, is this something that we should recommend in the installation guide (not to install Timber in the plugins folder if it’s installed through Composer) and as a fallback have the check for how Timber was installed and then block updates for Composer installs? Doesn’t the plugin option exists for developers who are not familiar with Composer or don’t want to work with it?

gsalkin:
[...]

PR #1756, opened 2018-07-24
composer: type = wordpress-plugin

drzraf:
#### Issue
When using bedrock it common to set custom plugin path according to package "type" like so:
```
    "extra": {
	"installer-paths": {
	    "web/app/mu-plugins/{$name}/": ["type:wordpress-muplugin"],
	    "web/app/plugins/{$name}/": ["type:wordpress-plugin"],
	    "web/app/themes/{$name}/": ["type:wordpress-theme"]
	}
    }
```
This work when using wpackagist, but not directly using (GitHub-based) upstream, especially when intending to use `dev-master`

#### Solution
Timber is a "wordpress-plugin". Just use that as its type.

#### Impact
None, hopefully.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,6 @@
 {
   "name": "timber/timber",
-  "type": "library",
+  "type": "wordpress-plugin",
   "description": "Plugin to write WordPress themes w Object-Oriented Code and the Twig Template Engine",
   "keywords": [
     "timber",


gchtr:
I’d feel good about changing this to `wordpress-plugin`. But it’s a little more complicated. Timber doesn’t have to be installed a plugin. It can also be considered as a theme library (though that’s probably an edge case). When Timber is used as a dependency of a theme and is not activated as a plugin, then it would be installed to `wp-content/themes/{theme-name}/wp-content/plugins/`, if we `wordpress-plugin` as the type. This could be set back to the `vendor/` folder with a custom installer path in the `composer.json` of the theme. But we would also have to tell all developers that use it as a theme dependency to do so.

In my opinion, we don’t need to decide whether Timber is a library, a theme layer or a plugin, because Timber is really versatile and can be used as any of these. And probably we don’t need to argue about the best installation method. I’d say it’s more about which installation methods we want to offer and about providing the proper guidance in the documentation. There’s a [related issue](https://github.com/timber/timber/issues/1673) about this.

chrisvanpatten:
What about an option where `timber/timber` continues to exist in its current form, as a traditional PHP library, but a new repo is added, `timber/timber-plugin`, which is set as a `wordpress-plugin` type that just depends on `timber/timber`?

drzraf:
**If**, at some point, Timber would provide a `wp-cli` command (to manipulate/process/... twig in accordance with theme settings), where would it goes? `plugin.php` at the root of the repository?

Side note: `johnpbloch/wordpress` should be dropped from `require-dev`. It's up to top-level `composer.json` to manage wordpress & plugins dependencies, and `bedrock` plugin does `require-dev(johnpbloch/wordpress)`

philiprabbett:
I too use Timber as a library rather than a plugin and prefer it that way.

If folks want to install the library as a plugin from Github, then why not use composer package as follows (not tested)

```
    {
      "type": "package",
      "package": {
        "name": "timber/timber",
        "version": "dev-master",
        "type": "wordpress-plugin",
        "dist": {
          "type": "zip",
          "url": "https://github.com/timber/timber/archive/master.zip"
        },
        "require": {
    "php": ">=5.3.0|7.*",
    "twig/twig": "1.34.*|2.*",
    "upstatement/routes": "0.4",
    "composer/installers": "~1.0",
    "asm89/twig-cache-extension": "~1.0"
        }
      }
    }
```

chrisvanpatten:
@drzraf Even if you manage to install this repo as a `wordpress-plugin`, there's no root `plugin.php` in the repo, so I don't even think WordPress would be able to process it, unless I'm missing something…

jarednova:
Thanks to @drzraf for the contrib and to everyone weighing-in. In short, while Timber _can_ be used as a plugin, this repo's default is as a theme dependency.

I'm all for a mirroring GH repo that allows a distribution so that someone _can_ use as a `wordpress-plugin', but for now I'm closing this PR. Please respond here if anyone's interested in picking that up as a `@todo`

PR #1758, opened 2018-07-25
Fix pagination troubleshootings

Mahjouba91:
I am currently developing a search.php page to display search results, I have this kind of code :

```php
<?php
global $paged;
if ( ! isset( $paged ) || ! $paged ) {
	$paged = 1;
}
$templates = array( 'search.twig', 'archive.twig', 'index.twig' );
$args = [
	'post_status' => ['publish'],
	'posts_per_page' => get_option('posts_per_page'),
	's' => get_search_query(),
	'paged' => $paged
];

$context['posts'] = new Timber\PostQuery($args);
$context['pagination'] = Timber\Pagination::get_pagination( array() );

Timber::render( $templates, $context );
```

the pagination object is at null if I don't put this line of code : 
```$context['pagination'] = Timber\Pagination::get_pagination( array() );```

So it's really annoying. Maybe you have further explanations ?

### CODE REVIEW

**Comment by jarednova** [comment]
Thanks @Mahjouba91 — that's totally weird and not intended (obvious, I hope). Perhaps it could be a quirk unique to search? I'll investigate and see if I can figure out what might be happening

**Comment by jarednova** [comment]
@Mahjouba91 thanks for submitting this PR/Question. In looking at it with greater depth I realize that I misunderstood you originally. The preferred way to get pagination is to use whatever `PostQuery`'s `pagination` object. Here's the info in the docs:

https://timber.github.io/docs/guides/pagination/

I had thought your Q was about default arguments for `get_pagination`. Specifically, it'd make sense to me that this ...

```php
$context['pagination'] = Timber\Pagination::get_pagination( array() );
```

Should default to an empty array so that you can simply do:

```php
$context['pagination'] = Timber\Pagination::get_pagination();
```

I'll submit a PR to resolve this

Issue #1639, opened 2018-01-17
Improve compatibility with third-party plugins through "the_post" action


gchtr:
Now, here’s a problem I encountered when working with ACF and WooCommerce.

## Current behavior

When an ACF field returns an array of post objects, Timber tries to convert these `WP_Post` objects into `Timber\Post` objects:

https://github.com/timber/timber/blob/0d72eee300bf4e52ccb6aba34afa2eb4af6cf188/lib/Post.php#L735

This line was added in https://github.com/timber/timber/pull/969, solving a compatibility issue with Visual Composer (https://github.com/timber/timber/pull/958).

There’s one thing though that might cause weird behavior when working with other plugins like WooCommerce. In `Timber\Post::get_info`, a `the_post` action is called, where plugins might hook in, e.g. to setup their global data:

https://github.com/timber/timber/blob/0d72eee300bf4e52ccb6aba34afa2eb4af6cf188/lib/Post.php#L529-L536

WooCommerce hooks into the `the_post` to setup its `$product` global. However, it shouldn’t set that global when we loop over ACF objects. In my case, on a single product view, I have a custom list of products that I show before the related products section. That list of products is defined through ACF. Now when Timber loops over that list, the `$product` global is updated for each of the items in the list. When WooCommerce displays the related products, it takes the `$product` global as a reference and derives the related product from it. Unfortunately, this is always the last product of my ACF list. The related products section seems to be displaying random products.

## Solutions

My problem is not that I can’t find a solution that works for me. It took a long time to track down what’s going on and I’m looking for a solution that will not cause any headache for other developers and that works out of the box, and I see multiple directions that we could go.

### Solution 1: Make ACF filters not call `the_post`.

We could use a filter to disable `the_post` hook when ACF filters converts `WP_Post` objects into `Timber\Post` objects:

```php
if ( apply_filters( 'timber/post/setup_global', '__return_true' ) ) {
    do_action_ref_array( 'the_post', array( &$post, &$GLOBALS['wp_query'] ) );
}
```

Maybe we could also say that `the_post` shouldn’t be called when `Timber\Post::convert()` is used.

### Solution 2: Add a new integration for Visual Composer.

The new integration could call `the_post` in the proper place. This would seem like the simplest solution. But on the other hand, `the_post` is an integral part of WordPress, so we probably should integrate it into Timber as well.

Which leads me to solution 3:

### Solution 3: Find a proper place for `the_post`.

[...]

PR #1766, opened 2018-08-15
2.x Refactor context and improve compatibility with third party plugins

gchtr:
**Ticket**: #1639

This pull request refactors Timber’s context and improves compatibility with third party plugins.

## Improving compatibility for third party plugins

As lined out in #1639, Timber’s current behaviour can lead to compatibility problems with plugins such as BuddyPress or WooCommerce, because…

- `the_post` hook is called whenever a `Timber\Post` object is instantiated.
- Loop variables are not set for singular templates.

I figured that a way to fix both problems would be to use Timber’s context function, because that’s the function we always call in a template. Using a WordPress hook like `template_redirect` could be tricky, because developers probably will set different posts as the main post for singular templates.

That’s why I added the necessary logic to mimick WordPress’s behavior for singular templates:

- Sets `$wp->query->in_the_loop = true`.
- Calls the `loop_start` hook.
- Calls `$wp_query->setup_postdata()`.

And I removed the call to `the_post` inside `Timber\Post::get_info()` and moved it to a context function.

## A context guide

I added a new Context Guide in `docs/guides/context.md` that will explain the concept of the context in Timber. It’s probably best to read through it in addition to what is written here.

## Rename function to `context()`

To be consistent with the naming that we applied in the rest of Timber’s classes, I renamed the `get_context()` function to `context()`. The `get_context()` function will be marked as deprecated.

## New functions

Next to `Timber::context()`, I added three functions to better handle the different context caches.

- `Timber::context_global()` – Used for getting the global context.
- `Timber::context_post()` - Used for getting the context for a singular post.
- `Timber::context_posts()` – Used for getting the context a post collection for an archive.

## Template based contexts

The context changes based on which template is displayed. For singular templates, it will set `post` in the context, and for archive templates it will set `posts` in the context.

This way, Timber will make assumptions about what you’re gonna do in your template. Instead of setting up your template like this:

**archive.php**

```php
<?php
$context          = Timber::get_context();
$context['posts'] = new Timber\PostQuery();

Timber::render( 'archive.twig', $context );
```

You could now set it up like this:

**archive.php**

```php
<?php
// Minimal style
Timber::render( 'archive.twig', Timber::get_context() );
```

(Little known fact: You wouldn’t have to use `$context['posts'] = new Timber\PostQuery();` in the current version of Timber either. Timber did always set `posts` in the context.)

Of course there are a lot of cases where we need to adapt the default. That’s where context arguments are coming in.

## Arguments for the context

The context now accepts an arguments array. Currently, there are three arguments supported: `post`, `posts` and `cancel_default_query`.

### posts

The `posts` argument can be an array of posts, a `Timber\PostQuery` or an array of arguments that will be passed to `Timber\PostQuery`. If it’s set to `false`, Timber will not set `posts` in the context.

**archive.php**

```php
$context = Timber::context( array(
    'posts' => array(
        'post_parent' => 0,
    ),
) );

Timber::render( 'archive.twig', $context );
```

Up until now, there was no easy way to *adapt the default WordPress template query*. Now, if `posts` is an array of arguments, it will **merge** this array with the default arguments that WordPress uses for its template query. So, in the example above, you will get the default list of posts for an archive, just without child posts.

It’s also possible to pass a `Timber\PostQuery` object to `posts`. Then, no merging will happen.

### cancel_default_query

By default, parameters passed with `posts` will merge with the default WordPress post query. If the `cancel_default_query` parameter is set to `true`, merging will be disabled. Instead, the default query will be overwritten with the parameters passed in `posts`.

### post

This can be a post ID, a `WP_Post` object, a `Timber\Post` object or a class instance that inherits from `Timber\Post`.

**single.php**

```php
<?php

// Use a custom class for post.
Timber::render( 'single.twig', Timber::context(
    'post' => new Extended_Post(),
) );
```

You now might be tempted to try something like this:

```php
<?php
// Don’t do this!

$context         = Timber::context( array( 'post' => false ) );
$context['post'] = new Extended_Post();

Timber::render( 'single.twig', $context );
```

While this might seem fine, it’s probably **not a good idea**. Setting `post` after running `Timber::context()` will not set up the necessary globals for compatibility with third party plugins.

As outlined in #1639, we can’t set up the globals inside `Timber\Post`, because that leads to other problems. It’s important to tell Timber for which post it should set up globals, in case a different post ID is used.

But how do we ensure that developers won’t use the example above, because it has become so prevalent in most Timber examples? And what if figuring out the post to display depends on what the context returns?

The only solution I can come up with now could be to use `Timber::context_post()` separately:

```php
<?php
$context         = Timber::context( array( 'post' => false ) );
$context['post'] = Timber::context_post( array( 'post' => new Extended_Post() ) );

Timber::render( 'single.twig', $context );
```

## Caching

`Timber::context()` can be called more than once in a page load, for example inside hooks that display only a small component of a page. That’s why it has a cache. But because the context can now be dynamic, I see two ways to go:

### Option 1 – Only cache global context

If we only cache the global context and don’t cache template contexts, we’d probably have to advise developers to always set the `post` and `posts` parameters to `false` when they use `Timber::context()` inside actions, because otherwise additional queries would run.

Or we could make it a standard to only use the global context via `context_global()`:

```php
add_shortcode( 'global_address', function() {
    return Timber::compile( 'global_address.twig', Timber::context_global() );
} );
```

### Option 2 – Cache everything separately

If we introduce separate caches for global variables and template contexts, there would be four caches:

- `$context_cache`: A cache for all global context variables (`site`, `request`, `theme`, `user`, `http_host`, `wp_title`, `body_class`) as well as context data set through the `timber_context` filter.
- `$context_chache_post`: A cache for post data in singular contexts.
- `$context_chache_posts`: A cache for post collection data in archive contexts.
- `$context_args`: A cache for the arguments passed to context.

The result for `Timber::context()` will be put together from these caches.

It should still be possible to run `Timber::context()` with different arguments after a cache is already set. To be able to ignore the post and posts cache by passing in new arguments, there’s the `$context_args` cache. This cache will save the arguments from a call to `Timber::context()`. Whenever arguments are passed to `Timber::context()`, it will compare the arguments with the arguments saved in this cache and will either **return the cache** or **the result for the new arguments, ignoring the cache**. If a new query runs based on different arguments, this query **won’t be saved** in the cache. The cache from the first run will persist, including the query saved in there.

Does this logic make sense? How should the context cache behave?

### Personal opinion

I’d settle for Option 1, because I think it might be easier to describe and understand. In my 3+ years developing themes with Timber, I only ever required the global context inside hooks.

## New filters

I first thought about using a lot of filters to change the context defaults, but then I realized it’s not really necessary. One new filter would suffice:

`timber/context/args`

With this filter, we can change the arguments that are passed to `Timber::context()`. When added globally in **functions.php**, it will affect all queries. But it could also be added in a template file like **single.php** or **archive.php** to affect only certain templates. Here’s an example:

**functions.php**

```php
// Globally disable `post` and `posts` in context
add_filter( 'timber/context/args', function( $args ) {
    $args['post'] = false;
    $args['posts'] = false;

    return $args;
} );
```

You could also use some logic here:

```php
// Globally disable `post` and `posts` in context
add_filter( 'timber/context/args', function( $args ) {
    if ( is_post_type_archive( 'book' ) ) {
        $args['posts'] = false;
    }

    return $args;
} );
```

## Discussion

The changes introduced here might be opinionated, but I thought it’s probably easiest if I propose a solution that we can then discuss.

- [x] Should we call it `cancel_default_query`, `run_default_query`, `overwrite_default_query` or `merge_default_query`?
- [x] Should `cancel_default_query` (or any other name we choose) also be used to tell Timber to not set `post` in the context?
- [x] Does the caching logic introduced for the context make sense? How should the context cache behave? Should we only cache the global context?

## Testing

I added tests, but testing is a little hard, because Unit Testing doesn’t work well with [Conditional Tags](https://codex.wordpress.org/Conditional_Tags), because it’s not really meant to be used for this.

## Todo

- [x] Complete the Context Guide when discussion is finished.
- [x] Update documentation where `Timber::get_context()` is used instead of `Timber::context()`.
- [x] Improve inline documentation for parameters, filters, etc.
- [x] Add changes to upgrade guide.

### DIFF
--- a/docs/getting-started/theming.md
+++ b/docs/getting-started/theming.md
@@ -139,7 +139,7 @@ Let’s crack open **index.php** and see what’s inside:
 
 ```php
 <?php
-$context = Timber::get_context();
+$context = Timber::context();
 $context['posts'] = new Timber\PostQuery();
 
 Timber::render( 'index.twig', $context );
@@ -151,7 +151,7 @@ This is where we are going to handle the logic that powers our index file. Let�
 
 ```php
 <?php
-$context = Timber::get_context();
+$context = Timber::context();
 ```
 
 This is going to return an object with a lot of the common things we need across the site. Things like the site name, the description or the navigation menu you’ll want to start with each time (even if you over-write them later). You can do a `print_r( $context );` to see what’s inside or open-up [**Timber.php**](https://github.com/timber/timber/blob/master/lib/Timber.php) to inspect for yourself.
--- a/docs/guides/acf-cookbook.md
+++ b/docs/guides/acf-cookbook.md
@@ -41,7 +41,7 @@ $post = new Timber\Post();
 if (isset($post->hero_image) && strlen($post->hero_image)){
 	$post->hero_image = new Timber\Image($post->hero_image);
 }
-$data = Timber::get_context();
+$data = Timber::context();
 $data['post'] = $post;
 Timber::render('single.twig', $data);
 ```
--- a/docs/guides/cheatsheet.md
+++ b/docs/guides/cheatsheet.md
@@ -5,10 +5,10 @@ menu:
     parent: "guides"
 ---
 
-Here are some helpful conversions for functions you’re probably well familiar with in WordPress and their Timber equivalents. These assume a PHP file with the `Timber::get_context();` function at the top. For example:
+Here are some helpful conversions for functions you’re probably well familiar with in WordPress and their Timber equivalents. These assume a PHP file with the `Timber::context();` function at the top. For example:
 
 ```php
-$context = Timber::get_context();
+$context = Timber::context();
 $context['post'] = new Timber\Post();
 Timber::render( 'single.twig', $context );
 ```
--- a/docs/guides/cheatsheet.md
+++ b/docs/guides/cheatsheet.md
@@ -9,7 +9,7 @@ Here are some helpful conversions for functions you’re probably well familiar 
 
 ```php
 $context = Timber::context();
-$context['post'] = new Timber\Post();
+
 Timber::render( 'single.twig', $context );
 ```
 
--- a/docs/guides/context.md
+++ b/docs/guides/context.md
@@ -7,7 +7,7 @@ menu:
 
 The context is one of the most important concepts to understand in Timber. Think of the context as the set of variables that are passed to your Twig template.
 
-In the following example, `$data` is an array of values. Each of the values will be available in the Twig template with the key of the value as a variable name.
+In the following example, `$data` is an associative array of values. Each of the values will be available in the Twig template with the key as a variable name.
 
 **single.php**
 
@@ -29,7 +29,7 @@ Timber::render( 'single.twig', $data );
 <p>{{ message }}</p>
 ```
 
-Of course you don’t have to figure all the variables you need for yourself. Timber will provide you with a set of useful variables when you call `Timber::context()`. 
+Of course you don’t have to figure out all the variables you need for yourself. Timber will provide you with a set of useful variables when you call `Timber::context()`. 
 
 **single.php**
 
@@ -41,7 +41,7 @@ $context = Timber::context();
 Timber::render( 'single.twig', $context );
 ```
 
-Follow this guide to get an overview of what’s in the context, or use `var_dump( $context );` in PHP or `{{ dump() }}` in Twig to display the contents right on your site.
+Follow this guide to get an overview of what’s in the context, or use `var_dump( $context );` in PHP or `{{ dump() }}` in Twig to display the contents in your browser.
 
 ## Global context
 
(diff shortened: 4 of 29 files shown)


gchtr:
@jarednova Could I get your overall opinion and your opinion on the points in the Discussion section before completing this pull request?

jarednova:
Sure thing! will wade into today

jarednova:
@gchtr this is a great solution to (2) very old problems. My outstanding Qs are around the areas you highlight:

### 1. To cancel or not to cancel?
My thinking here is because the user is that 95% of the time people will just use the default query, and most of that other 5% will be about query modifying ("I want 20 posts per page instead of the default"). Thus, `cancel` or `overwrite` make the most sense to me. I lean slightly towards "overwrite" (since if you're sending an array to `posts` this is the literal effect of your action) but I can see how that could be confused with "overwrite the defaults with my values here." Thus, I land at `cancel_default_query` being the most clear (though still imperfect)

### 2. Cache
I have to admit, I'm rusty on some much of the caching use/logic. I _think_ I understand the advantage to option 2, but also think like it seems more complicated than the potential benefit. As you note, even as pro Timber developers (indeed, we make the darn thing) it's an edge case beyond what we'd even need. I'd prefer to err on the side of simpler and wait for the people to demand a more in-depth solution.

Overall this accomplishes ...

- [x] Make Timber play nicer with popular plugins
- [x] Make the API for `context` simpler/cleaner/morefun

I'd call that a huge 🎉 . Let's clean up the rest of the `context` vs. `get_context` stuff and the rest of the todo to pull this into **2.x** — great work!

gchtr:
Okay, thanks for having a look at this. I think it’s now ready to be merged.

- The context now works with option 1, so Timber will only cache the global context.
- I improved the Context Guide to describe in more detail how it works.
- I changed the parameter for `Timber::context_post()`. It will not accept the same arguments array that can be passed to `Timber::context()`, but a post-like argument, which means a post ID, a `WP_Post` object or a `Timber\Post` object.
- I updated all examples I could find where the context is used.

jarednova:
🎉 
this is a great add — thanks for your work as always

Issue #1528, opened 2017-08-22
Issue in do_action implementation for twig


nkmaurya-dev:
Hi Team, 

I have found that the **Timber is passing context** by default to every **do_action hook**. 

So if any callback function has any parameter or first parameter in case of multiple , that would be replaced as Timber context. It must be not happen, because any plugin or theme would have written them for their own use.

The timber is adding the **action** function to the **Twig** and passing the context there. I think it should not or in some way, if we can override that method to not the pass the context to any hook by default.

Please see the method add_timber_functions line: 37 and below is the do_action hook code snippet:
See: https://github.com/timber/timber/blob/master/lib/Twig.php

```php
$twig->addFunction(new Twig_Function('action', function( $context ) {
	$args = func_get_args();
	array_shift($args);
	$args[] = $context;
	call_user_func_array('do_action', $args);
}, array('needs_context' => true)));
```

For testing , if I did comment the line: 42, it means no context is passing to the callback functions and everything works well.

I'm explaining here the detailed explanation of code flow. Its just an example of that issue.

In tease-product.twig, I have used the hook as:

`{% do action( 'woocommerce_after_shop_loop_item') %}`

and the plugin [YITH WooCommerce Quick View](https://wordpress.org/plugins/yith-woocommerce-quick-view/) used this hook in their class as:

```
add_action( 'woocommerce_after_shop_loop_item', array( $this, 'yith_add_quick_view_button' ), 15 );
```

where `$this` : current class object
`yith_add_quick_view_button` : the callback function

So function declaration is:

```php
public function yith_add_quick_view_button( $product_id = 0, $label = '', $return = false ) {
	global $product;
                
	// get product id
	! $product_id && $product_id = yit_get_prop( $product, 'id', true );
	// get label
	! $label && $label = $this->get_button_label();

	$button = '<a href="#" class="button yith-wcqv-button" data-product_id="' . $product_id . '">' . $label . '</a>';
    $button = apply_filters( 'yith_add_quick_view_button_html', $button, $label, $product );

	if( $return ) {
		return $button;
	}

	echo $button;
}
```

Now, the first parameter i.e $product_id = 0 is automatically changed to Timber context, instead it should be 0 , if no value is passed for first parameter. Timber context is an array so it is causing the error in button HTML reasonably. 

We can't force to pass the context as argument to do_action callback function. 

Please suggest me here.

WordPress 4.8.1, 
PHP 7.1, 
Timber 1.4.1 (installed via WordPress.org)

[...]

PR #1773, opened 2018-08-30
2.x Remove action context

gchtr:
**Ticket**: #1528

## Issue

Timber adds an `action` function for Twig like this:

```php
$twig->addFunction( new Twig_Function( 'action', function( $context ) {
    $args = func_get_args();
    array_shift( $args );
    $args[] = $context;
    call_user_func_array( 'do_action', $args) ;
}, array( 'needs_context' => true ) ) );
```

This function will always add the context as the last parameter in the hook, so that it can be used like this:

```php
<?php
add_action( 'my_action_with_args', 'my_function_with_args', 10, 3 );

function my_function_with_args( $foo, $bar, $context ){
    echo 'I say ' . $foo . ' and ' . $bar;
    echo 'For the post with title ' . $context['post']->post_title;
}
```

As I see it, this was added so that developers can add their own actions and receive the context, should they need it in the hook.

Like mentioned in the ticket, there can be problems: When you don’t pass in any parameters, Timber will automatically add the context as the first parameter. This can lead to unexpected behavior, because the action function receives a parameter although you might not have passed one when you used `{% do action() %}`.

Consider this example, where a plugin might add an action that receives a string:

```php
<?php
add_action( 'aye_aye', function( $string ) {
    echo 'I say ' . $string;
}, 10, 1 );
```

You could call the action like this, without an argument:

```twig
{% do action('aye_aye') %}
```

Timber would now pass a `$context` array to the action, which would lead to an array-to-string conversion error, because the action echoes out the string, which is an array in this case.

When calling `do_action`, Timber doesn’t know which functions will be called and how many arguments they will receive. Maybe we could read this out, but should we? And do we have to?

In my opinion, we should simply make it possible to call `do_action` and not try to do some magic here. The required arguments for `do_action` should just be passed in manually.

## Solution

- Delete functionality that adds the context automatically.
- Delete tests.
- Update documentation.

## Impact

This will break calls with `{% do action() %}` that relied on the context in an argument. That’s why it’s a good place to add it to version 2.0.

## Usage Changes

Hooks that expect the context to be passed as the last argument will have to be adapted so that the required arguments can be passed manually. So the example mentioned earlier would have to be rewritten like this:

```php
<?php
add_action( 'my_action_with_args', 'my_function_with_args', 10, 3 );

function my_function_with_args( $foo, $bar, $post ){
    echo 'I say ' . $foo . ' and ' . $bar;
    echo 'For the post with title ' . $post->post_title;
}
```

The post will be passed manually as a third argument.

```twig
{% do action( 'my_action_with_args', 'Yippie', 'Yeah', post ) %}
```

## Considerations

Did I forget to consider something?

## Testing

I removed the relevant tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,11 +21,11 @@ before_script:
     - composer require satooshi/php-coveralls:1.0.*
 
 script:
-    - if ( [ "$TRAVIS_BRANCH" == "master" ] || [ "$TRAVIS_BRANCH" == "2.x"] ) && [ "$TRAVIS_PHP_VERSION" == 7.1 ]; then mkdir -p build/logs; vendor/phpunit/phpunit/phpunit --coverage-clover build/logs/clover.xml -c phpunit.xml; else vendor/phpunit/phpunit/phpunit -c phpunit-nocover.xml; fi
+    - if [ "$TRAVIS_BRANCH" == "master" ] || [ "$TRAVIS_BRANCH" == "2.x"] && [ "$TRAVIS_PHP_VERSION" == 7.1 ]; then mkdir -p build/logs; vendor/phpunit/phpunit/phpunit --coverage-clover build/logs/clover.xml -c phpunit.xml; else vendor/phpunit/phpunit/phpunit -c phpunit-nocover.xml; fi
 
 after_script:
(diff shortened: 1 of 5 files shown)


jarednova:
Thanks @gchtr, reviewing now!

jarednova:
Looks good! I made two small changes...

- Updated the docs to show an example of using a "template" variable in an action
- Used this as an opportunity to run coverage tests on things merging into 2.x (I was curious to see if we lost test coverage %%, we didn't)

gchtr:
Cool thing about the example in the documentation!

jarednova:
📜📜📜📜📜

PR #1781, opened 2018-09-06
Update PostPreview Dokumentation

mark-wff:
A little bit more documentation because I had to read the source-Code to understand how to use it.


jarednova:
Indeed, that makes that part way clearer/easier — even matches what's in the PHPDocs for bonus clarity:
http://php.net/manual/en/function.strip-tags.php

PR #1783, opened 2018-09-07
Added teardown method, added new postsiterator

pascalknecht:
- Posts can now clean up after themselves with a teardown method
- PostsIterator calls the teardown method of the post before looping to next item


jarednova:
Awesome to see this cranking!

PR #1785, opened 2018-09-10
2.x context revision

pascalknecht:
- Added Interface for setup & teardown method
- Added test for `$wp_query->in_the_loop` (Not yet working)


jarednova:
🎉 I won't get in the way of this between you and @gchtr, but awesome to see this moving along!

Issue #1557, opened 2017-09-20
Clearly document the escaping strategy of Timber


johnbillion:
The Timber library does't clearly state its strategy for escaping. This means developers who are new to Twig but are familiar with other templating engines might expect that a `{{` double brace expression `}}` has its output automatically escaped. For example, [Handlebars](http://handlebarsjs.com/) automatically escapes expressions in double braces, and triple braces need to be used to output raw HTML.

The [Escapers documentation](https://timber.github.io/docs/guides/escapers/) doesn't really clarify things. #615 gives us the definitive answer, as does reading the code in the `Timber\Timber\get_twig()` method.

The readme file or the documentation should be updated to clearly state Timber's escaping strategy, along with how to enabled `autoescape` via adding `Timber::$autoescape = true;` to your theme's `functions.php`.

mahagr:
In addition to this, it would be nice if the examples used `|raw` filter on variables which are known to be escaped and `|e` when they should be escaped. This would be a good practice as then all the examples would work with both settings.

johnbillion:
I've just discovered that vanilla Twig _does_ escape by default, and Timber disables this. Ref: https://twig.symfony.com/doc/2.x/templates.html#html-escaping

This increases the need for documentation that clarifies the situation.

jarednova:
@johnbillion thanks for the heads-up on this. There are reasons good and bad for this default (mostly bad). But regardless, I agree this should be clearly documented.

I've added #1562 to resolve the documentation gap. 

However, given your experience (and the fact that this is Twig's default behavior) I think it makes sense to consider making escaping the _default_. However, I'm nervous about rolling out a potential breaking change like that. Do you have any advice or recommendation here?

johnbillion:
That's a big change to make :-)

I'm currently investigating some weird behaviour on a client site with autoescaping enabled. I'll report back soon.

johnbillion:
It looks like the issue I mentioned above is an anomaly, but we're still looking into it.

In the meantime, yes I could support switching to escaping by default in Timber as I think it's expected behaviour given that's how Twig works.

Given that Timber seems to support semver, a version 2.0 could be released which switches to escaping by default as a breaking change. This would potentially cause issues for users who are using the Timber Library plugin via the w.org plugin directory, though. There's not really a good way to deal with that.

mahagr:
[...]

PR #1786, opened 2018-09-12
2.x Improve Escaping Guide

gchtr:
**Ticket**: #1557

#### Issue

Following the discussion in the linked issue, I think the Escapers guide can be optimized.

#### Solution

- Renamed guide from **Escapers** to **Escaping**, because the guide is also about escaping in general and not only about the escapers in Twig. The guide will still be accessible under the old URL, but will redirect to the new one.
- Tried to use simpler language in some paragraphs.
- Added more links to WordPress documentation about escaping.
- Documented more clearly that Timber doesn’t escape by default.
- Optimized heading structure.
- Optimized code examples by using fenced code blocks.

#### Impact

None.

#### Usage Changes

None.

#### Considerations

- We should really consider adding escapers to the code examples used throughout Timber.

#### Testing

Not needed.


jarednova:
Thanks @gchtr! #1787 is a great first issue for someone to pick up. Big impact w/o needing to know anything about PHPUnit. Found one little mistake (that looks like it had been there for a while — the example for `js_escape` was incomplete), otherwise all good!

danielpost:
Not sure if this is the right thread for this but I was having issues with using the autoescaper—as it turns out `$autoescape = true` [has been deprecated](https://github.com/twigphp/Twig/issues/2386) and is now changed to HTML: `$autoescape = 'html'`.

If you'd like I can submit a PR reflecting this change in the docs. Furthermore, it might be nice to add a small check in Timber to update `true` to `'html'` automatically for backwards compatibility, but on the other it's not a big issue to introduce a change like this in 2.x.

gchtr:
@danielpost Usually, it might be better to open a new issue for this, because it might get lost in an already closed pull request.

Very good catch! 👍 I didn’t know that. As I see, the feature is [deprecated since Twig 1.21](https://twig.symfony.com/doc/1.x/deprecated.html#miscellaneous):

> As of Twig 1.21, setting the environment option `autoescape` to `true` is deprecated and will be removed in 2.0. Use `html` instead.

So I guess even for for Twig version > 1.21, it would be better to use `html`. In **composer.json**, we require version 1.34.x: https://github.com/timber/timber/blob/42cd28149e1c7c9894bc9eb2cff34e533b2c8553/composer.json#L32

In my understanding, this means that

- we can take this into the current docs for Timber, and not only for 2.x.
- we would indeed need the check for backwards compatibility in 1.x, but not in 2.x.
- we can add this to the [Upgrade Guide for 2.x](https://github.com/timber/timber/blob/2.x-upgrade-guide/docs/upgrade-guides/2.0.md), so it’s clear that from 2.x, `html` is required.

It would be great if you could open a pull request for this! Or even two: one against the `master` branch or the changes in `1.x` and one for the `2.x` branch.

PR #1788, opened 2018-09-12
2.x Add Attachment class

gchtr:
**Ticket**: #1580

## Issue

Timber uses the `Image` class a representation for images as well as attachments. While the `Image` class also works for working with attachments such as PDF files, it’s rather unintuitive to do so.

## Solution

Add an `Attachment` class that `Image` builds upon. 

- Moved all relevant functions from the `Image` to the `Attachment` class.
- Removed `get_post_meta()` function to make use of the `get_meta_values()` function in the `Post` class.
- Optimized code to use `Post::import_custom()` function instead of own logic.
- Renamed internal function `get_dimensions()` to `get_dimension()` to better reflect what it does.
- Renamed internal function `get_dimensions_loaded()` to `get_dimension_loaded()` to better reflect what it does.
- Renamed internal function `_maybe_secure_url()` to `maybe_secure_url()`.
- Renamed internal function `$_dimensions` to `$dimensions`.
- Removed unused protected property `$_can_edit`.
- Changed method order in both classes.
- Renamed `$file_types` to `$image_file_types`, because image file types are not the only file types for attachments.
- Added more inline documentation.
- Applied coding standards for functions in `Attachment`.

### New functions

- Added new function `size_raw()` that gets the filesize in bytes.
- Added new function `size()` that gets filesize in a human readable format. E.g. `16 KB` instead of `16555 bytes`.
- Added new function `extension()` that gets the extension of the attached file.
- Added new `Attachment` function for Twig.

## Impact

None. Using `Timber\Image` instead of `Timber\Attachment` still works fine, because `Timber\Image` extends `Timber\Attachment`.

In the future, there might be classes for other file types that can extend the `Attachment` class.

## Usage Changes

For attachments that are not images, the `Attachment` class can be used in PHP and Twig instead of the `Image` class.

```php
$context['attachment'] = new Attachment( 54 );
```

Example for using `Attachment::size()`:

```twig
<a class="download" href="{{ attachment.src }}" download="{{ attachment.title }}">
    <span class="download-title">{{ attachment.title }}</span>
    <span class="download-info">(Download, {{ attachment.size }})</span>
</a>
```

Example for using `Attachment` in Twig, `Attachment::size_raw()` and `Attachment::extension()`:

```twig
<table>
    {% for attachment in Attachment(attachment_ids) %}
        <tr>
            <td>{{ attachment.title }}</td>
            <td>{{ attachment.extension }}</td>
            <td>{{ attachment.size_raw }} bytes</td>
        </tr>
    {% endfor %}
</table>
```

## Considerations

- [x] The `Attachment::init()` function still contains code that would only be revelant for an image:

https://github.com/timber/timber/blob/2e6e7663bd77929974ded644996dea8386b791c3/lib/Attachment.php#L199-L208

This part should probably be moved over to `Image::init()`. But I didn’t know how to decouple this properly.

@pascalknecht I know you already thought a lot about Timber’s image classes. Do you have something to add to this pull request in general (that doesn’t involve changing the whole underlying structure ;), which we would probably take on in a later version of Timber)?

## Testing

- Moved and updated relevant tests over to new **test-timber-attachment.php** file.
- Renamed `TimberImage_UnitTestCase` to `TimberAttachment_UnitTestCase`
- Updated some test function names.
- Added tests for new functions.


jarednova:
IT'S FINALLY HERE!!!!!

Thanks @gchtr I'll work on reviewing!

jarednova:
@gchtr the section that you highlight as maybe moving from `Attachment::init` to `Image::init` might be ready for deprecation. I _think_ what it does is guess that if you send Image a Post with a `thumbnail_id` on it, what you intend is to get the post's thumbnail, not the post itself.

This strikes me as one of those "overly defensive/forgiving" positions I took way way back that might not make sense any more. I'm going to see about removing those sections entirely and see what the outcome in the tests is

jarednova:
@gchtr I created #1789 to handle the deprecated functionality. Because it's a significant change to your PR, I wanted to make it its own PR so we could handle/resolve that independently.

Otherwise: the PR looks great. As you consider #1789 I'll go through and double-check some of the smaller pieces. The documentation on Attachment feels like such a nice start to work from in the future.

I just had one question: what was your methodology behind re-ordering the methods? public vs. protected? alphabetical? something else? My strong opinion here is just that those should follow a consistent logic so we can apply the same methodology to other classes

gchtr:
@jarednova Alright, I’ll have a look at that pull request then.

About the reordering: I realized that I put a little more thought into the order of the methods in `Image` than for the methods in `Attachment`. But I have to admit it was an intuitive action 🙈. I mostly followed the methodology of *Initializers* > *Public and most relevant methods* > *Less relevant* > *Private*, but wasn’t too consistent, either. The relevancy can be quite subjective though 🤔. I don’t have a strong opinion on this, because in PHPStorm I can display a method list that I can sort alphabetically.

I don’t know if there’s a general ordering principle for methods in classes. It would probably be more relevant for the [Reference in the documentation](https://timber.github.io/docs/reference/). Currently, the methods are sorted alphabetically, which I thought makes them easier to find. But we could also rely on the search functionality there and apply the same order as they are found in the class.

It would be nice if we could define some short guidelines for this.

lib/Attachment.php
```diff
@@ -0,0 +1,556 @@
...
+					// Assume a relative path.
+					if ( strstr( strtolower( $iid ), $type ) ) {
+						$this->init_with_relative_path( $iid );
+
+						return null;
+					}
+				}
+			}
+		} elseif ( $iid instanceof \WP_Post ) {
+			$ref  = new \ReflectionClass( $this );
```
pascalknecht: @jarednova @gchtr  Maybe that's too forgiving too?

lib/Attachment.php
```diff
@@ -0,0 +1,556 @@
...
+		 */
+		if ( $iid instanceof Attachment ) {
+			$iid = (int) $iid->ID;
+		} elseif ( is_array( $iid ) && isset( $iid['ID'] ) ) {
+			// Assume ACF image array.
+			$iid = $iid['ID'];
+		}
+
+		if ( ! is_numeric( $iid ) && is_string( $iid ) ) {
+			if ( strstr( $iid, '://' ) ) {
```
pascalknecht: I think this should go in another class. We mix WordPress attachments with external images here. I think of another `ExternalImage` class which does not extend `Image` as they don't share the same properties. Any thoughts? @gchtr @jarednova

lib/Attachment.php
```diff
@@ -0,0 +1,556 @@
...
+	 * @internal
+	 * @link https://developer.wordpress.org/reference/functions/wp_upload_dir/
+	 *
+	 * @return array
+	 */
+	public static function wp_upload_dir() {
+		static $wp_upload_dir = false;
+
+		if ( ! $wp_upload_dir ) {
+			$wp_upload_dir = wp_upload_dir();
```
pascalknecht: Maybe refactor to `wp_get_upload_dir`. See https://github.com/WordPress/WordPress/blob/master/wp-includes/functions.php#L2003 for more information

pascalknecht: Also no static cache like implemented would be needed as WP does this by itself. See https://github.com/WordPress/WordPress/blob/master/wp-includes/functions.php#L2057

tests/test-timber-attachment.php
```diff
@@ -0,0 +1,188 @@
+<?php
+
+class TestTimberAttachment extends TimberAttachment_UnitTestCase {
+
+/* ----------------
+ * Helper functions
+ ---------------- */
+
+ 	static function replace_attachment( $old_id, $new_id ) {
+		$uploadDir = wp_upload_dir();
```
pascalknecht: I would use `wp_get_upload_dir` for performance reasons instead

lib/Attachment.php
gchtr: @pascalknecht @jarednova For me, it would make sense that an image that is loaded through an URL or a path wouldn’t have the same methods and properties as an image that is an actual attachment. I like the concept, but I’m not yet sure about the name «External», maybe `ImageAsset`? And we couldn’t return a new class here, right? Is this where object factories would come in handy?

pascalknecht: @gchtr I would probably implement a new class. The advantage of object factories is mostly that you can decouble the instantiation of objects from the classes so no real benefit here I think. I would suggest implementing a new class `ImageAsset` or `ExternalImage` which doesn't have all the empty properties it currently has.

jarednova:
@gchtr I think we're in good shape. The last change I made was a re-organization of the tests. Because `TestTimberImage` extended `TestTimberAttachment`, the tests in `TestTimberAttachment` were executing twice.

I think we should proceed with merging into 2x with three outstanding to-dos spun up for the future...
- [ ] Address the possibility of an `ExternalImage` class as raised by @pascalknecht 
- [ ] Apply the `maybe_convert_array` to the other "Object" methods in `Timber\Twig`
- [ ] Decide and apply a consistent methodology for method ordering (maybe look to WP Coding Standards here?)

All three of those are outside the scope of this ticket, so I think they become to-dos from here so we don't risk this working getting stale. I'll wait for the tests to finish and plan on merging once it's morning over in🇨🇭so @pascalknecht and @gchtr have a last chance to weigh-in

pascalknecht:
@jarednova The class still uses `wp_upload_dir` even thought @gchtr refactored that. Did you maybe miss that in the merge conflicts? See #eb49f7f2e42916f0dd412a7cb8ce4b05e8b01c57
Otherwise ready to merge from my side.

gchtr:
@jarednova Yep, the changes for `wp_upload_dir()` seem to have slipped in again: https://github.com/timber/timber/pull/1788/commits/eb49f7f2e42916f0dd412a7cb8ce4b05e8b01c57.
Otherwise, it looks good to me 👍. Good thing about the separate todos 👍!

jarednova:
Awesome, thanks for that catch. Restored the stuff from @gchtr — merging!

PR #1789, opened 2018-09-14
Create determine_id and a overriding function in Image 

jarednova:
**Ticket**: #1788 

#### Issue
This re-organizes the `Attachment::init` method and separates out a `determine_id` method (same as `Post`) to handle the logic of "wait, what attachment is the user _really_ trying to get?

This allows us to also isolate the remaining piece of `Image` code (which handles posts with `thumbnail_id`s) and move it into the `Image` class itself.

#### Considerations
The piece of isolated `Image` code for determining ID (which extrapolates that when a user sends a "regular" Post they probably intend to get the `thumbnail_id` associated with said post) is overly-opinionated and could cause unexpected behavior. I move that we deprecate this functionality.

#### Testing
`TestTimberImage::testTimberImageFromPost` currently passes, but once we remove the deprecated functionality, it fails.
```php
	function testTimberImageFromPost() {
		$post = $this->get_post_with_image();
		$image = $post->thumbnail();
		$post = get_post($post->ID);

		$str = '{{ Image(post).src }}';
		$result = Timber::compile_string( $str, array('post' => $post) );

		$this->assertEquals($image->src(), $result);
	}
```
this shows the functionality it preserves though. As you can see, I don't see it to be particularly helpful or useful

### DIFF
--- a/lib/Attachment.php
+++ b/lib/Attachment.php
@@ -163,7 +163,7 @@ class Attachment extends Post implements CoreInterface {
 		 * thus there's nothing in the DB for us to do here 
 		 */
 		if ( null === $iid ) {
-			return null;
+			return;
 		}
 
 		$attachment_info = $this->get_attachment_info( $iid );
@@ -191,8 +191,6 @@ class Attachment extends Post implements CoreInterface {
 
 			$this->id = $this->ID;
 		}
-
-		return null;
 	}
 
 	/**
@@ -211,14 +209,16 @@ class Attachment extends Post implements CoreInterface {
 		}
 
 		/**
-		 * If passed a Timber\Attachment object, grab the ID and continue. Otherwise, try to check
-		 * for an ACF image array an take the ID from that array.
+		 * If passed a Timber\Attachment or WP_Post object, grab the ID and continue. Otherwise, try
+		 * to check for an ACF image array an take the ID from that array.
 		 */
-		if ( $iid instanceof Attachment ) {
-			$iid = (int) $iid->ID;
+		if ( $iid instanceof Attachment
+		    || ( $iid instanceof \WP_Post && 'attachment' === $iid->post_type )
+		) {
+		    $iid = (int) $iid->ID;
 		} elseif ( is_array( $iid ) && isset( $iid['ID'] ) ) {
-			// Assume ACF image array.
-			$iid = $iid['ID'];
+		    // Assume ACF image array.
+		    $iid = $iid['ID'];
 		}
 
 		if ( ! is_numeric( $iid ) && is_string( $iid ) ) {
--- a/lib/Attachment.php
+++ b/lib/Attachment.php
@@ -215,7 +215,7 @@ class Attachment extends Post implements CoreInterface {
 		if ( $iid instanceof Attachment
 		    || ( $iid instanceof \WP_Post && 'attachment' === $iid->post_type )
 		) {
-		    $iid = (int) $iid->ID;
+		    return (int) $iid->ID;
 		} elseif ( is_array( $iid ) && isset( $iid['ID'] ) ) {
 		    // Assume ACF image array.
 		    $iid = $iid['ID'];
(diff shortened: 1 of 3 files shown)


lib/Attachment.php
```diff
@@ -156,6 +156,53 @@ public function __toString() {
...
 	 */
 	public function init( $iid = null ) {
+		$iid = $this->determine_id( $iid );
+
+		/**
+		 * determine_id returns null when the attachment is a file path,
+		 * thus there's nothing in the DB for us to do here 
+		 */
+		if ( null === $iid ) {
+			return null;
```
gchtr: The `init()` function probably shouldn’t return anything, right? I think we should just `return` here. Then we can also remove the `return null;` statement at the end of this function.

lib/Attachment.php
```diff
@@ -156,6 +156,53 @@ public function __toString() {
...
+		}
+
+		$attachment_info = $this->get_attachment_info( $iid );
```
gchtr: Would it be clearer if we moved this part to a separate method, `init_from_attachment()` maybe?

lib/Attachment.php
```diff
@@ -156,6 +156,53 @@ public function __toString() {
...
+	}
+
+	/**
+	 * Tries to figure out the attachment id you want or otherwise handle when
+	 * a string or other data is sent (object, file path, etc.)
+	 * @internal
+	 * @param mixed a value to test against
+	 * @return int|null the numberic id we should be using for this post object 
+	 */
+	protected function determine_id( $iid ) {
```
gchtr: Should the `determine_id()` method also check for an instance of `WP_Post`? It could very well be a WordPress attachment we try to init with, right?

```php
/**
 * If passed a Timber\Attachment or WP_Post object, grab the ID and continue. Otherwise, try
 * to check for an ACF image array an take the ID from that array.
 */
if ( $iid instanceof Attachment
    || ( $iid instanceof \WP_Post && 'attachment' === $iid->post_type )
) {
    $iid = (int) $iid->ID;
} elseif ( is_array( $iid ) && isset( $iid['ID'] ) ) {
    // Assume ACF image array.
    $iid = $iid['ID'];
}
```

gchtr:
Okay, now it’s clearer what that function does. But I think it’s really an edge case and hardly anyone would expect Timber to do that. It’s probably more intuitive to write:

```twig
{% if post.thumbnail %}
    <img src="{{ Image(post.thumbnail).src }}">
{% endif %}
```

So I’d be up for deprecating that functionality 👍

lib/Attachment.php
pascalknecht: What about returning null on failure and itself on creation. One thing I currently don't like about `Image()` is that it will return an object even though the constructor argument might be entirely false. Think about `Image(99999)`. This is not correct but will not cancel or throw an error. I think the `Image()` function should return `null` if the initialisation fails and hence Twig will render nothing on output. What do you think?

lib/Attachment.php
pascalknecht: In my opinion no. It is too forgiving for the user as it doesn't make sense to ouput a post as an image. Any thoughts?

lib/Attachment.php
jarednova: Agreed. Being "forgiving" with `null`s was a bad move, let's fix it

lib/Attachment.php
jarednova: @pascalknecht I think the scenario @gchtr outlines is less about outputting a post as an image, but rather instances when the data passed to the function was for some reason sent inside a `WP_Post` class (for example, as the result of a `WP_Query`)

lib/Attachment.php
jarednova: I'd prefer to keep this section in the main `init` section. All `inits` are not created equal and the "regular" course of action for an Attachment should be `init`ing as an attachment. It's the other cases: when the data is a URL, file path, etc., that are "special" and deserve to be processed elsewhere

jarednova:
Thanks @gchtr and @pascalknecht — I've incorporated all those notes into the PR. Of course, I managed to screw something up in trying to resolve a merge conflict so I started in clean in #1794. Once those tests pass I'll merge into #1788 and we can wrap things up from there

PR #1791, opened 2018-09-17
Add support for path namespaces.

jenkoian:
#### Issue
When looking into timber for a project I'm currently working on, I was looking to make use of path namespaces (e.g. https://symfony.com/doc/current/templating/namespaced_paths.html#registering-your-own-namespaces) but it seems that Timber doesn't currently support this.

#### Solution
This patch introduces support for path namespace, at time of writing only as part of user defined locations, however this could easily be added to to support namespace in theme dirs as well if you wanted.

#### Impact
It should be backwards compatible, so it shouldn't have any negative impact on existing users.

**Update** This will have an affect on users of some filters, see https://github.com/timber/timber/pull/1791#issuecomment-423122266

#### Usage Changes
Users who wish to make use of path namespaces can now do so, e.g. with a styleguide namespace

```php
Timber::$locations = array(
    '/Users/jared/Sandbox/templates',
    '~/Sites/timber-templates/',
    ABSPATH.'/wp-content/templates',
    ABSPATCH.'/wp-content/styleguide' => 'styleguide'
);
```

#### Considerations
I've not spent too much time on this patch, I'm hoping for some guidance/advice if there are any edge cases I may not have considered as I am very new to this library.

~It's also worth noting that I've just checked for a namespace presence and ignored if none there, but we could easily change to give each location the default namespace (`__MAIN__`) if none is configured, to cut down on an if statement.~

**Update** This now does indeed set the default namespace.

#### Testing
~I included one basic unit test, but would be happy to add more if you think this patch is worth including.~

**Update** Have included several useful tests.


jarednova:
@jenkoian thanks for the contrib! I'll confess we're deep into work on 2.x so I won't be able to start a review right away.

My biggest concern is just double-tripple checking that the use of a namespace will not disrupt rendering a `.twig` file from a non-namespaced path (ie. inside the `/templates` or `/views` directory. Can you write a test that demonstrates and verifies this?

jenkoian:
@jarednova thanks for taking a look at the PR! Yeah I figured it was all hands to the pump re. 2.x release (what a great 2.x feature namespace support would be btw! 😃). I'll add some more tests later today.

jenkoian:
@jarednova I've added a few more tests, let me know if that's what you had in mind 👍

jarednova:
Yep! Those tests show namespaced and non-namespaced template calls living side-by-side in harmony. I think we're good here, but I'd still like @gchtr or @pascalknecht to give it a look as well to make sure there's something we're not missing/considering

jenkoian:
Looks like build failed due to a composer issue on one of the jobs, are you able to restart the job?

pascalknecht:
@jarednova @jenkoian Codewise it looks good to me. Only objection to merge for me is documentation which is currently missing and code style which is currently not according to WordPress Guidelines.

jarednova:
@pascalknecht agreed! @jenkoian can you format according to WordPress Coding Standards and update the docs? I think this would belong in the `template-locations.md` file:

https://github.com/timber/timber/blob/master/docs/guides/template-locations.md
https://timber.github.io/docs/guides/template-locations/

gchtr:
Oh, this is a really nice addition! I just have one question: In the linked Symfony documentation, the pattern to describe namespaces is defined through `path => namespace`. Here, we would add it as `namespace => path`. Is this a deliberate solution that was chosen because of technical reasons? Or would it make sense to have `'ABSPATH.'/wp-content/styleguide' => styleguide'` instead of `'styleguide' => ABSPATH.'/wp-content/styleguide'` to follow Symfony’s style?

chrisvanpatten:
I'd echo @gchtr about following the Symfony convention. Anywhere Timber can follow existing conventions would be fantastic!

jenkoian:
Thanks for your comments!

> Codewise it looks good to me. Only objection to merge for me is documentation which is currently missing and code style which is currently not according to WordPress Guidelines.

Docs now added, have amended code style, although it doesn't look like this is enforced throughout the lib? 

> Oh, this is a really nice addition! I just have one question: In the linked Symfony documentation, the pattern to describe namespaces is defined through path => namespace. Here, we would add it as namespace => path. Is this a deliberate solution that was chosen because of technical reasons? Or would it make sense to have 'ABSPATH.'/wp-content/styleguide' => styleguide' instead of 'styleguide' => ABSPATH.'/wp-content/styleguide' to follow Symfony’s style?

Excellent spot! 🦅 👁 I initially planned to follow what symfony do (i.e. path => namespace) but I didn't think it made as much sense here as we're not configuring via yaml. For example, in Symfony you can imagine..

```yaml
'path/to/templates1': namespace1
'path/to/templates2': namespace2
'path/to/templates3': ~
```

Would map to PHP as..

```php
array(
    'path/to/templates1' => 'namespace1',
    'path/to/templates2' => 'namespace2',
    'path/to/templates3' => null,
);
```

Which makes looping over etc. really easy. However, if you imagine the following...

```php
Timber::$locations = array(
    'path/to/templates1' => 'namespace1',
    'path/to/templates2' => 'namespace2',
    'path/to/templates3',
);
```

We'd end up with...

```php
array(
    'path/to/templates1' => 'namespace1',
    'path/to/templates2' => 'namespace2',
    0 => 'path/to/templates3',
);
```

So we end up with the path being the key in some instances and the value in others, we can of course code around that, but it seemed simpler to me to just flip them around. 

**However** I noticed Symfony supports multiple paths per namespace, which obviously won't work if we have the namespace as our key as we can't have duplicate keys. So, I've gone ahead and made this change. It's quite involved unfortunately, but all the tests pass. Only concern I have is that it will potentially change the following filters:

`timber/loader/paths`
`timber/loader/loader`
`timber_locations`
`timber/locations`

As previously they would get a list of paths like so:

```php
array(
    'path/of/templates1',
    'path/of/templates2'
);
```

Where as now it'll be:

```php
array(
    '__MAIN__' => array(
        'path/of/templates1'
    ),
    'namespace' => array(
        'path/of/templates2'
    )
);
```


Finally, I've added a test for multiple paths per namespace and I've also thought of nested namespaces, so I've added a test to cover that too.

Cheers!

lib/Loader.php
```diff
@@ -130,16 +131,29 @@ protected function template_exists( $name ) {
...
-		$fs = apply_filters('timber/loader/loader', $fs);
+		$fs = new \Twig_Loader_Filesystem( array(), $rootPath );
+		foreach ( $paths as $namespace => $path_locations ) {
+			if ( is_array( $path_locations ) ) {
+				array_map( function ( $path ) use ( $fs, $namespace ) {
+					$fs->addPath( $path, $namespace );
+				}, $path_locations );
+			} else {
+				$fs->addPath( $path_locations, $namespace );
+			}
```
jenkoian: I'm being overly defensive here as `$path_locations` will always be an array, as it affects code coverage I'll remove this if/else.

lib/Timber.php
```diff
@@ -439,22 +439,32 @@ public static function get_sidebar( $sidebar = 'sidebar.php', $data = array() )
...
+						include trailingslashit( $uri ) . $sidebar;
+						$found = true;
+					}
+				}
+			} else {
+				if ( file_exists( trailingslashit( $uri_locations ) . $sidebar ) ) {
+					include trailingslashit( $uri_locations ) . $sidebar;
+					$found = true;
+					break;
+				}
```
jenkoian: I'm being overly defensive here as `$uri_locations ` will always be an array, as it affects code coverage I'll remove this if/else.

pascalknecht:
Ok so it will affect the filters you mentioned: 
`
timber/loader/paths
timber/loader/loader
timber_locations
timber/locations
`

We will need to include this in the upgrade guide to 2.0 then. Agreed @jarednova ?

lib/LocationManager.php
```diff
@@ -33,19 +45,19 @@ public static function get_locations( $caller = false ) {
...
+		$roots      = array_map( 'realpath', $roots );
+		$roots      = array_unique( $roots );
 		foreach ( $roots as $root ) {
-			if ( !is_dir($root) ) {
+			if ( ! is_dir( $root ) ) {
 				continue;
 			}
-			$theme_locs[] = $root;
-			$root         = trailingslashit($root);
+			$theme_locs[ FilesystemLoader::MAIN_NAMESPACE ][] = $root;
```
jenkoian: This might be undesirable if a user is using a custom loader (via a filter) and have customised the main namespace name. Should a define a constant on the `Loader ` class within Timber rather than use the one from the Twig Filesystem Loader?

jenkoian: Addressed in https://github.com/timber/timber/pull/1791/commits/f31adba1691712e44102f46611a2b523b2ae04c8

docs/guides/template-locations.md
```diff
@@ -23,6 +23,31 @@ Timber::$locations = array(
...
+```php
+<?php
+Timber::$locations = array(
+    '/Users/jared/Sandbox/templates',
+    '~/Sites/timber-templates/',
+    ABSPATH.'/wp-content/templates',
+    ABSPATH.'/wp-content/styleguide' => 'styleguide'
+);
+```
+
```
gchtr: Here I would add an example for how this can be used in Twig with the `@` character, pretty much in the same way that the [Symfony Docs](https://symfony.com/doc/current/templating/namespaced_paths.html#registering-your-own-namespaces) do it.

docs/guides/template-locations.md
```diff
@@ -23,6 +23,31 @@ Timber::$locations = array(
 );
 ```
 
+You can also use namespaces in your locations too, just define it as the value next to a path, for example:
```
gchtr: I think I’d put this section in an own section with a separate heading and move the section just before `## Changing the default folder for Twig files`

```
## Register your own namespaces

…
```

docs/guides/template-locations.md
jenkoian: Addressed in https://github.com/timber/timber/pull/1791/commits/9abb875f985827b741fc50006408848253c02c06

docs/guides/template-locations.md
jenkoian: Addressed in https://github.com/timber/timber/pull/1791/commits/9abb875f985827b741fc50006408848253c02c06

docs/guides/template-locations.md
gchtr: Cool! 👍

docs/guides/template-locations.md
gchtr: Oops, maybe I wasn’t clear enough about where to place it. I meant that this heading could be added as a heading for the whole section that you added to the docs 😅. Sorry!

I’d add it before this line: `
You can also use namespaces in your locations too, just define it as the value next to a path, for example:`

jenkoian: Got it, thanks. Updated in https://github.com/timber/timber/pull/1791/commits/2f22f6ea44e75ffd0c0b728c4201a0d90f3eb898

gchtr:
Cool stuff 😎. This looks good for me in terms of documentation. 

I have to admit that the Loader isn’t my strongest suit, so I’d be happy if someone else could look at the code in detail, @jarednova maybe?

If the filters are going to change, we definitely need to add this in the Upgrade Guide for 2.0. I normally do this separately, as soon as a feature is merged. Currently, this pull request wants to merge into `master`. I guess when the existing filters will change, we would have to pull this into the `2.x` branch.

jenkoian:
Do you want me to rebase against 2.x ?

lib/Loader.php
jenkoian: Have removed the else, think the if is fine.

lib/LocationManager.php
```diff
@@ -12,17 +12,27 @@ class LocationManager {
...
-		$locs = array_merge($locs, self::get_locations_user());
-		$locs = array_merge($locs, self::get_locations_caller($caller));
+		$locs = array_merge_recursive( $locs, self::get_locations_user() );
+		$locs = array_merge_recursive( $locs, self::get_locations_caller( $caller ) );
 		//remove themes from caller
-		$locs = array_diff($locs, self::get_locations_theme());
-		$locs = array_merge($locs, self::get_locations_theme());
-		$locs = array_merge($locs, self::get_locations_caller($caller));
-		$locs = array_unique($locs);
+		//$locs = array_diff_assoc( $locs, self::get_locations_theme() );
```
jenkoian: Will remove this.

jenkoian: Just to explain, I don't think the diff is needed, it's handled in the array merge.

jarednova:
@jenkoian thanks so much for hanging with us through all these Qs. I think we're very close, but I get caught on the issue of the filters and how it could potentially break compatibility with anyone who's using `timber/loader/paths` or some of the others. 

I just wrote a test in 9f4ed02d98b61f25679a1d66b3bdf8d640552826 to validate the current behavior and see how it compares to this PR (on my local the new test fails against this PR). I think there are two routes to go to merge this in:

1. Merge into `1.x` and write some functions to adapt old filter behavior into the how things are handled with the namespaces
2. Merge into `2.x` and handle write the info in the upgrade guide.

My vote is for #2; but I want to see what @jenkoian thinks of this

pascalknecht:
Also voting for option 2

jenkoian:
Awesome test! I'll take a closer look and get back to you 👍

jenkoian:
Ok had chance to review the failing test. Yes agree with your two options, I am happy to add the backwards compatibility workaround if you'd like, but yeah would probably vote for option 2. I guess one thing on that, is it would require the user to add a namespace, even if using the default, e.g.

`$paths[Timber\Loader::MAIN_NAMESPACE][] = __DIR__.'/october/';`

as opposed to simply

`$paths[] = __DIR__.'/october/';`

They would also need to ensure they're appending to an existing array (or passing an array if using a new namespace they've created). 

Not sure if we deem this too much of a departure even for 2.x or not? I'm not sure how used these filters are for the community? I guess as long as we document the filter and it's usage it should be ok, I can't seem to find much docs on the filter at the mo, so perhaps that's something else I can add?

gchtr:
I also vote for option 2. In addition to that, could we get rid of the following pattern, where locations are assigned through a static variable?

```php
Timber::$locations = array(
    'path/to/templates1' => 'namespace1',
    'path/to/templates2' => 'namespace2',
    'path/to/templates3',
);
```

Instead, I’d only use the `timber/locations` filter and make this the primary filter for custom locations.

> They would also need to ensure they're appending to an existing array (or passing an array if using a new namespace they've created).

This is also a reason why I think using a filter only would be better, because that’s the established behavior for filters: You filter something, or you append to something existing. If we provide good examples, I don’t see a problem.

For the deprecation, I’d do it like this:

- Let `Timber::$locations` work like before, but issue a deprecation warning through `Helper::deprecated()`.
- Add a check for the filters, whether the resulting array is associative or not. If it’s not associative, then do not apply the namespacing functionality and issue a deprecation warning through `Helper::deprecated()`.

@jenkoian There’s not so much documentation about the filters yet, but if you’re going to rebase against 2.x, then you’re gonna see that there’s the base for the upcoming filter documentation (read more about this [here](https://github.com/timber/timber/blob/866fe8726487c444b31033f917d4be4d6c6bf6d6/CONTRIBUTING.md#documenting-hooks).

https://github.com/timber/timber/blob/7efcfb222ed8bf88c3c37f9468feef5b5af7c35e/lib/LocationManager.php#L25-L34

I’d be very happy if you could complete that with descriptions and examples, etc. We’d still have the main documentation in the Template Locations Guide. The filter documentation is more some kind of technical reference.

jenkoian:
@gchtr that is an awesome approach (love this project!). When I get some time i'll add the filter documentation. 

I've not rebased this against 2.x yet, do you want me to do that now, or should I open up a new PR with this against 2.x for comparison?

jarednova:
@jenkoian please open a fresh PR against 2.x (referencing this PR), I think things will be easiest to track there — thanks!

jenkoian:
Ok opened against 2.x here: https://github.com/timber/timber/pull/1811 struggling for time a bit at the mo, so not had chance to make the recommended changes. Will do so as soon as I can!

jarednova:
Great, thanks @jenkoian! I'm going to close this PR so we can follow-up on #1811 and resolve that last change there

PR #1792, opened 2018-09-18
Update to newest PHPUnit v6

jarednova:
Testing if PHPUnit v6 will work with Travis


jarednova:
Closing b/c I ended up making this a change for `master` (since 1.x will benefit as well)

PR #1796, opened 2018-09-23
docs: Fix wrong capitalization with the term "PhpStorm"

mirucon:
Just fixed wrong capitalization with PhpStorm in the docs :)


jarednova:
Thanks for the catch @mirucon. @gchtr can you review and confirm?

gchtr:
Hey @mirucon! Thank you for your pull request! We like contributions that fix small issues, like the correct way to spell PhpStorm. It’s probably the same struggle to write WordPress the correct way, but in the end, it’s the small details that count, too! 😎🎉

Issue #1797, opened 2018-09-25
`mime_content_type` not supported in some PHP Installs


jarednova:
I received an email bug report about v 1.8.0 and [the introduction of `mime_content_type`](https://github.com/timber/timber/compare/1.7.1...1.8.0) in `ImageHelper::is_svg`:

```
[21-Sep-2018 14:14:14 UTC] PHP Fatal error:  Uncaught Error: Call to undefined function mime_content_type() in /home/ruisenreuring/public_html/web/app/plugins/timber-library/lib/ImageHelper.php:154

Stack trace:
#0 /home/ruisenreuring/public_html/web/app/plugins/timber-library/lib/Image/Operation/Resize.php(176): Timber\ImageHelper::is_svg('/home/ruisenreu...')
#1 /home/ruisenreuring/public_html/web/app/plugins/timber-library/lib/ImageHelper.php(579): Timber\Image\Operation\Resize->run('/home/ruisenreu...', '/home/ruisenreu...')
#2 /home/ruisenreuring/public_html/web/app/plugins/timber-library/lib/ImageHelper.php(71): Timber\ImageHelper::_operate('https://www.rui...', Object(Timber\Image\Operation\Resize), false)
#3 /home/ruisenreuring/public_html/web/app/plugins/timber-library/vendor/twig/twig/lib/Twig/Environment.php(462) : eval()'d code(41): Timber\ImageHelper::resize('https://www.rui...', 58, 30, 'center')
#4 /home/ruisenreuring/public_html/web/app/plugins/timber-library/vendor/twig/twig/lib/Twig/Template.php(432): __TwigTemplate_042318cf4d6b07fb3d85 in /home/ruisenreuring/public_html/web/app/plugins/timber-library/lib/ImageHelper.php on line 154
```

Apparently `mime_content_type` might require a PECL installation in certain PHP versions or explicitly enabling in `php.ini`:

https://stackoverflow.com/questions/14809054/mime-content-type-function-not-working

Given that this will be distributed to WP.org we should look for a method that's 100% compatible with supported PHP versions

jarednova:
Closed by #1798

PR #1798, opened 2018-09-25
Add homemade mime_content_type method to ImageHelper

jarednova:
**Ticket**: #1797 

#### Issue
Version 1.8.0's `is_svg` method includes checking for whether or not a file is an SVG and includes a call to PHP's `mime_content_type`. Unfortunatley, this PHP method is not available in all installs and might require a PECL installation or enabling via PHP.ini:

https://stackoverflow.com/questions/14809054/mime-content-type-function-not-working

#### Solution
I rolled my own `mime_content_type` method using the `finfo` class.

#### Impact
I fear the performance on this might not be as good as the native method, but I haven't tested to validate

#### Usage Changes
None

#### Considerations
- Maybe `ImageHelper::\_mime_content_type` could call the native if it's available and only use the "homemade" component if it's not
- Maybe this is all stupid, and we should just check for the presence of `.svg` or `.SVG` in the file name and call it a day

#### Testing
All tests pass

### DIFF
--- a/lib/ImageHelper.php
+++ b/lib/ImageHelper.php
@@ -157,11 +157,22 @@ class ImageHelper {
     	return $ret;
 	}
 
-	static function _mime_content_type($filename) {
+	/**
+	 * 
+	 * Reads a file's mime type. This is a hack b/c some installs of PHP don't enable this function
+	 * by default. See #1798 for more info
+	 * @since 1.8.1
+	 * @param string $filename to test.
+	 * @return string|boolean mime type if found (eg. `image/svg` or `text/plain`) false if not
+	 */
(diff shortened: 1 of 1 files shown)


jarednova:
For now I'm going to go with ...

_Maybe ImageHelper::\_mime_content_type could call the native if it's available and only use the "homemade" component if it's not_

Maybe in the future we can ditch it all for just a simple string check for `.svg`. I'm guessing that has performance benefits, but haven't tested to validate. I'm going to merge and plan to release as Timber 1.8.1 unless anyone has any suggestions or flags

pascalknecht:
No objections from my side against merging @jarednova

PR #1802, opened 2018-09-27
allow additional filtering on Timber\Post::terms() retrieval when calling wp…

mdahlke:
…_get_post_terms

#### Issue
Getting the terms from a `Timber\Post` object using `Timber\Post::terms()` can not be filtered by terms that do not have a parent (they are the top-most-level term)

#### Solution
Add an additional argument to the method `Timber\Post::terms` that can be passed to the function `wp_get_post_terms` that is called within the method

#### Impact
Will have:
1. Next to no impact on codebase
2. Be backwards compatible
3. Allow much better filtering for the user

#### Usage Changes
Calling `Timber\Post::terms` now accepts a 4th argument that is an array of "Term query parameters", see WP_Term_Query::__construct() for supported arguments.


gchtr:
@mdahlke Thanks for this pull request! I agree that we need this. Just today I ran into the same issue and had to solve it with a separate function. I’m very glad you raised this issue and provided a proposal for a solution 👍. It’s really great if we have so many developers who bring in their thoughts! 😎 But sometimes a little change like this can have a bigger impact than you initially thought or raise a bigger discussion.

Adding another parameter to the function might seem like the easiest solution. But then we’d have one more optional parameter in that function. I’m always hesitant about adding too many optional parameters, because then some questions arise. How should we define which parameter is the most important? If somebody wants to change only the last parameter, does he always need to check the defaults for the other parameters and pass them as well?  
There are other areas in Timber where we have the same problem. I always propose to use an arguments array instead, because that seems like a good compromise in a situation like this. In this very particular case of *getting terms*, I’d even want to make a strong case for this:

When using `Timber::get_terms()`, we already need to pass the taxonomy along with other parameters. We use the same pattern that is used for WordPress’s [`get_terms()`](https://developer.wordpress.org/reference/functions/get_terms/) function. Couldn’t we use the same pattern here, so that we merge the `tax` and `args` into one array?

```php
$terms = $post->terms( array(
    'taxonomy' => 'custom_tax',
    'orderby'  => 'count',
) );
```
I think it’s not important to know for a developer that in the back, Timber uses `wp_get_post_terms()`, which takes different parameters than `get_terms()`.

Of course this would be a breaking change and we’d have to merge it into the `2.x` branch instead. When we do that, I’d even go further and say that we use only one parameter:

```php
$terms = $post->terms( array(
    'query' => [
        'taxonomy' => 'custom_tax',
        'orderby'  => 'count',
    ],
    'merge'      => false,
    'term_class' => 'My_Term_Class'
) );
```
An arguments array allows you to only pass the parameters that you want to change. Remember, we just merged in a pattern like this in #1778 for `Timber\PostQuery`:

```php
$persons = new Timber\PostQuery( array(
    'query' => array(
        'post_type' => 'person',
    ),
    'post_class' => 'Person',
) );
```

Did I miss something? Any other opinions (@pascalknecht, @jarednova, @chrisvanpatten)?

jarednova:
@gchtr I totally agree. I like the pattern of #1778 where we have ~3x arguments. This could be implemented here in a way that doesn't break backwards compatibility via an adapter function (not sure if that's what it's called, but basically [function overloading](http://php.net/manual/en/language.oop5.overloading.php)) that will run a modified function when that first argument is a two-dimensional array.

@mdahlke: is that something you'd be up for?

mdahlke:
@jarednova I think that @gchtr solution is a much better way of implementing this than my quick and dirty solution. I say go that direction with it instead of just arbitrarily adding another argument.

gchtr:
@jarednova @mdahlke @pascalknecht Okay, I’m going to try to add this without breaking backwards compatibility 👍.

jarednova:
Thanks again for the PR @mdahlke! @gchtr ended up with a broader solution in #1845, but this is what got the ball rolling. So, while we're not going to merge in this PR know that it shall burn in our hearts forever — thanks again for contributing to 🌲 !

Issue #1800, opened 2018-09-25
Support/Bug: Password protected page functionality


erbanach:
I am using WordPress along with Timber to build a custom theme. I need a page of my website to be password protected. I have changed the settings in the WordPress admin area for this page but the out of the box WordPress functionality does not seem to work with Timber. I searched around a bit and found the following code snippets, but they do not work. What am I doing wrong here?

In my `page.php` logic:

```
    $context = Timber::get_context();
    $post = new TimberPost();
    $context['page'] = $post;
    $context['global'] = get_fields('options');
    $context['protected'] = post_password_required($post->ID); // here i am trying to check if the current page is password protected and adding the check to the context
    $context['password_form'] = get_the_password_form(); // grabbing the function to display the password form and adding it to the context
    Timber::render( array(
    'pages/page-' . $post->post_name . '.twig',
    'pages/page.twig'
    ), $context );
```

In my Twig template, I tried:

```
    {% if protected %}
    {{ password_form }}
    {% else %}
    <!-- rest of page template here -->
    {% endif %}
```

This `if` check does not seem to work, it always says that the page is not password protected even if it has been specified otherwise in the WP admin. The `{{ password_form }}` does display the password form correctly. How can I get this working?

I also tried this in `page.php`:

```
    $context = Timber::get_context();
    $post = new TimberPost();
    $context['page'] = $post;
    $context['global'] = get_fields('options');

    if ( post_password_required($post->ID) ) {
      Timber::render('page-password.twig', $context);
      } else {
      Timber::render(array('pages/page-' . $post->post_name . '.twig', 'pages/page.twig' ), $context);
      }
```

With `page-password.twig` displaying the password form. This if check also did not work. This might be a Timber bug because I believe those functions should be working.

[Stack overflow question here.](https://stackoverflow.com/questions/52505820/password-protected-page-functionality-not-working-out-of-the-box-with-timber-twi)

marciojc:
@erbanach you are seeing the page as admin? you must be logged out to see the form.

jarednova:
saw that there's a good amount of activity over on StackOverflow, moving there ...

https://stackoverflow.com/questions/52505820/password-protected-page-functionality-not-working-out-of-the-box-with-timber-twi

gchtr:
No matter what the final solution for this will be, I think we could add a section in the documentation for how to do password protected pages in Timber, maybe in the [WordPress Integration Guide](https://timber.github.io/docs/guides/wp-integration/)?

[...]

PR #1804, opened 2018-10-01
Add docs section about password protected posts

gchtr:
**Ticket**: #1800

## Issue

There’s no standard way to handle password-protected posts.

## Solution

Add section in the WordPress Integration Guide for how to secure password protected posts with a password form. Thanks to @tnottu for recommending a nice solution that I added here.

## Impact

None.

## Usage Changes

None.

## Considerations

The documented filter could be added to the starter theme as well.

## Testing

None.


jarednova:
Thanks @gchtr — I just added a full example using the `post_password_required()` method

PR #1805, opened 2018-10-02
Remove broken link to Watson-Ruby

teddybradford:
The URL `http://goosecode.com/watson/` no longer exists. It seems the project has been discontinued or is no longer maintained.


jarednova:
@gchtr this was added back in 36c2126616f7f3a589a894845c09bf8d3c5e95e1. Can you see if there's a replacement available? My Googling didn't turn up anything

gchtr:
In that linked commit, I only improved the docs syntax, but I couldn’t find when the link was actually added. But Watson Ruby is available here:

- <https://github.com/nhmood/watson-ruby>
- and <http://nhmood.github.io/watson-ruby/>

It doesn’t list Twig in the [supported languages](https://github.com/nhmood/watson-ruby#supported-languages), so I can’t say if this would still work.

jarednova:
Thanks @gchtr and @teddybradford. I ended-up preserving the block (with the updated link that @gchtr provided). I confirmed that it still works with Twig!

PR #1803, opened 2018-09-27
WIP: Documentation

marciojc:
#### Issue
Update documentation

#### Usage Changes
Change the documentation

Issue #1795, opened 2018-09-19
Make phpcs a dev requirement in composer.json


gchtr:
Picking up on the comments in #1561.

We should probably make `phpcs` a dev requirement in **composer.json**. We already check for the coding standards in new pull requests, so it would make sense to have `phpcs` ready when somebody wants to contribute code and run code sniffing as well.

jarednova:
Totally! Would that be something that runs in Travis? or is that another external integration (like Coveralls or Scrutinizer) that we'd need to setup?

marciojc:
What about the wpcs?

marciojc:
@jarednova and @gchtr about the phpcs, do you want to have all the files covered? 
As this https://github.com/marciojc/timber/commit/36286c561ef43ca98557b504e919b834268cfd2a?

gchtr:
@jarednova I’m not sure. Here’s an example where they use it in Travis: <https://github.com/BracketSpace/Notification/blob/develop/.travis.yml>. But it could also be set up in Scrutinizer: <https://scrutinizer-ci.com/docs/tools/php/code-sniffer/>. I think we already discussed that in https://github.com/timber/timber/pull/1603#issuecomment-354039295.

@marciojc Yes, `phpcs` is supposed to be used with the WordPress Coding Standards. We have a custom config file (<https://github.com/timber/timber/blob/master/phpcs.xml>), where we try to stay as close to the WordPress Coding Standards as possible (see #1603). The standards are not applied to the codebase yet, though. We’ll do it in one bunch, before we release version 2.0. After that release, we want to be more strict about the standards. So, for the moment, we’d rather not have pull requests that fix coding standards in the code itself. **But!** The coding standards will also require DocBlocks to be complete and I like the work you did for the DocBlocks in https://github.com/marciojc/timber/commit/36286c561ef43ca98557b504e919b834268cfd2a 😎. We could really use some help for the inline documentation, there’s a lot that can be improved. Method summaries and descriptions, examples, type hints, parameter and return descriptions, etc. Would you be willing to help us there? I could give you some guidance about how it will be translated into the [Reference](https://timber.github.io/docs/reference/).

marciojc:
@gchtr yes I can do the inline documentation. Is there a list of priorities?

gchtr:
Closed by #1933.

PR #1807, opened 2018-10-02
Document inline documentation standards in CONTRIBUTING.md

gchtr:
**Tickets**: #1795, #1803

## Issue

I felt like there’s not enough info about how Inline Documentation is translated to the Reference in the documentation.

## Solution

- Added section about inline documentation that matches how inline documentation is translated to the Reference in 2.x. This includes the info that I initially put into [Documenting hooks](https://github.com/timber/timber/wiki/Documenting-hooks) in the wiki.
- Made some smaller improvements.
- Added a Table of Contents for faster navigation.

This might be interesting for @marciojc.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,22 +1,44 @@
-## Here are ways to get involved:
+# Contributing to Timber
+
+Hey there! We’re really happy you’ve found your way here 💪. Timber is a community project that is developed by people from all over the world. We appreciate any help.
+
+Here are ways to get involved:
 
 1. [Star](https://github.com/timber/timber/stargazers) the project!
-2. Answer questions that come through [GitHub issues](https://github.com/timber/timber/issues?state=open)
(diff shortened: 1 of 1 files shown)


marciojc:
@gchtr already see some updates that I need to do at my branch.

jarednova:
@gchtr good thinking! (wow this makes everything feel so much more official). I made one slight tweak and have one Q:

On line 220 you've got `Filters ...` and then on 222 you have the description. I think it might be helpful to illustrate a concrete example of how these two relate (or if they should be collapsed into a single description line like "Filters the status of a particularly named transient so a user can override the behavior on a case-by-case basis"). Obviously it's damn good as it is, this is the only part that left me with a question in my head

gchtr:
@jarednova Cool. Yeah, using a real example would be more helpful indeed. I just added updated it.

Probably there will be more to add and improve in the future, but I think for now we should be good 👍

jarednova:
Lookin' good, merged!

Issue #1752, opened 2018-07-19
Page Preview should not return a revision post


jnweaver:
### Expected behavior

Within the WP editor, when clicking the `Preview Changes` button (or the `Preview`button if post is in *draft* status), the preview should render the modified published or draft post.

(Wordpress previews work by modifying the edited posts' content with the autosaved revision posts' content. The post object's post_type remains as `post` instead of `revision`. See the [_set_preview()](https://github.com/WordPress/WordPress/blob/master/wp-includes/revision.php#L551) function in WP core.)

### Actual behavior

Timber, as of 1.7.1 with these commits (c119c16e2de92eed8dc920e21143ec03b09aef4e and e34a702c367b5a710a09e153878f532c77805ce1) now returns a post object whose post_type is `revision` instead of the expected `post`. Noticeable implications of this arise in particular with `post.class` in a twig file. Wordpress's `get_post_class()` function adds CSS classes based on several conditionals that fail for posts with post_type `revision`. So styling based on these classes will not appear when previewing a post. This is related to #1728 and possibly has other impacts on previewing posts.

### Steps to reproduce behavior

Trace the `post_type` in preview. In a vanilla Wordpress install with the 2017 theme, the `post_type` is `post`. Add Timber, the same preview returns a post with `post_type` of `revision`.

### What version of WordPress, PHP and Timber are you using?

Wordpress 4.9.7, PHP 5.6 and Timber 1.7.1. I've also replicated the issue running Timber 2.x-dev.

Reverting to 1.7.0 fixes the issue. Or in 2.x-dev, commenting out [line 304](https://github.com/timber/timber/blob/2.x/lib/Post.php#L304) also resolves it.

I'm not sure I understand why the change was made to return a revision post object so I can't say if reverting this change would re-introduce other issues.

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)

Composer.

chrisvanpatten:
@jarednova Not sure what the process is for milestoning/labeling ticket but I'd love to see this solved in 2.x if possible. It seems like maybe we could do something similar to core's `_set_preview` to mix together the revision and the parent post data. Might also impact @gchtr's work on post context.

chrisvanpatten:
For anyone else running into the issue, here's my (temporary) solution…

1. Extend `\Timber\Post` with your own model
2. Implement a function like this:

```
/**
 * Attempt to safely get an ID
 *
 * @return int
 */
public function getSafeId()
{
[...]

PR #1809, opened 2018-10-04
1752/revision class

jarednova:
**Ticket**: #1752 

#### Issue
Previewing of posts (from the WordPress editor) can sometimes return unexpected results — with a mix of data from the original post vs. the revised post

#### Solution
As opposed to fetching data from the revision, this PR starts with data from the original post and only overwrites a few specific pieces of revised data including ...
- Title
- Main Content
- Custom fields

though this seems limited, this represents the data that the user is truly revising (things like author, a post's CSS classes, etc. are fixed).  

#### Impact
I've yet to test some second-order data (ex: previews extracted from content) or other data (tags) to see how those are impacted by this chang

#### Considerations
For `2.x` this could impact how we handle custom field data. I've found that it's MUCH easier/cleaner here if those fields are retrieved at runtime (ie `{{ post.meta("my_field") }}`) vs. initialization (where the data is attached to the `Timber\Post` object before it's even fetched). 

There's also a conflict on how the term "preview" is used. Sometimes "preview" refers to a revision, sometimes it's Timber's assembly of the post's "excerpt." — this might be worth rectifying in `2.x`

#### Testing
I moved the tests related to post revisions into its own file.


jarednova:
I'd love feedback from any contributor, but especially insights from @jnweaver or @chrisvanpatten as they were active on #1752

gchtr:
I can’t say much about the preview stuff, but I can say something about the meta considerations.

I’m very much okay with retrieving custom fields only on runtime. This removes the possibility to see what values are present when dumping an object in Twig. But then again, there were some issues where people thought this was the standard way of getting custom field data, not knowing that it’s only raw data. Based on the discussion we had in https://github.com/timber/timber/pull/1618#issuecomment-354627351, would we have to add a possibility to retrieve raw meta values then?

jnweaver:
@jarednova Thanks! This does resolve my issue from #1752 . I tested and confirmed.

gchtr:
@jarednova A combination of `raw` and `meta` is exactly what I thought of, too! So `meta_raw` or `raw_meta` would be much better than `custom`, I guess. But let’s handle this in a separate issue/pull request. I’m gonna add a task to the 2.x project for this.

PR #1811, opened 2018-10-08
Add twig namespace support 2.x

jenkoian:
This PR is https://github.com/timber/timber/pull/1791 opened against `2.x`


jenkoian:
Tests are passing! Take a look, I've added some docs and things too. Wondering if it needs more though, particularly around other related things like `Timber::$dirname` etc.

jarednova:
@jenkoian thanks so much for your work on this. I went through this afternoon and found a few small tweaks to make (#1815) but overall this is a really really strong PR

This is now merged into 2.x! Now, what's next?

jenkoian:
Woohoo, thanks!

jenkoian:
Just out of interest, do you have a roadmap or anything as to when `2.x` will be released?

gchtr:
For this pull request I only have one question:

We have a `timber/locations` filter:

https://github.com/timber/timber/blob/5101648dd78bdb0b3546864e6674cd284327af9c/lib/LocationManager.php#L27-L36

What’s the difference to the `timber/loader/paths` filter, which is now used in the examples? From reading the code, I don’t really see a difference apart from that `timber/locations` is loaded first. Could we get rid of one? I’d say we keep `timber/locations`, because when we talk about «Template Locations», it might more clear to use `timber/locations` instead of `timber/loader/paths`. Isn’t the Loader something that we want to keep somewhat hidden from the public API?

@jenkoian Thanks again for this great pull request. We could really use more of this! 😎 We’re actively working on `2.x` and are tracking progress in the [Timber 2.0 project](https://github.com/timber/timber/projects/4). We can’t say when it will be released, because there are still quite a few breaking things to figure out. But if you want to help, you’re very welcome to do so.

acobster:
@gchtr AFAICT the difference is [this line](https://github.com/timber/timber/blob/7a73a5d11ffd6f05358aed9e0138752512a8399c/lib/Loader.php#L134), where either `ABSPATH` or `/` is appended to the list of paths, in order to accommodate `open_basedir` ini directives (see https://github.com/timber/timber/commit/fbc6d29f). Not sure what the reason is for having two different filters to handle this (@jarednova ?) but I'd definitely be in favor of deprecating `timber/loader/paths` in favor of `timber/locations`, and possibly moving that `open_basedir` logic from `Loader::get_loader()` to `LocationManager::get_locations()` since it's more concerned with _locations_ than with the loader itself. IMHO that's totally tangential to this PR, though, and definitely deserves its own.

Nice work again, @jenkoian ! I had totally forgotten about Twig's namespace feature. :)

jenkoian:
Thanks all, have to admit that `timber/loader/paths` and `timber/locations` thing confused me too 😄 

Hoping to convert a project to use Timber but was blocked by lack of namespace support. Now that's in 2.x I'll continue the conversion based on that and happy to raise any patches as and when I come across anything 👍

PR #1813, opened 2018-10-09
Set ignore_sticky_posts=true as default wp_query argument

### LINKED ISSUE #1812: Timber::get_posts() return too many posts when using sticky_posts
### Expected behavior
Given an array of wp_query arguments with a `numberposts` value set, `Timber::get_posts()` should return that number of posts maximum, irrespective of the other query arguments.
Also: `Timber::get_posts()` should return the exact same posts as WordPress' `get_posts()` given the same arguments.

### Actual behavior
`Timber::get_posts()` returns more than `numberposts` posts when using sticky posts. It looks like the sticky posts get added to the posts-array afterwards, increasing the number of posts by the number of sticky posts.
If `ignore_sticky_posts=true` is added to the query arguments, the correct number of posts is given. 
Note that using WordPress' `get_posts()` does return the correct amount because `ignore_sticky_posts=true` is defined as default in `get_posts()` (`wp-includes.php/post.php:1759`).

- Create a query-arguments array with a `numberposts` value e.g.
```
$latestPostsArgs = array(
    'post_type' => 'post',
    'numberposts' => 10,
    'post_status' => 'publish'
);
```
- Make sure you have more than `numberposts` posts of said post types
- Mark a few of these posts as sticky post
- Fetch the posts e.g. 
`$posts = Timber::get_posts($latestPostsArgs);`
- Note that the number of posts returned is greater than the `numberposts` value because of the sticky posts that get added:
`echo count($posts); //prints 13, in my case`
- Also note that when using WordPress' `get_posts()` using the same query arguments, the correct number of posts is returned:
`$wpPosts = get_posts(latestPostsArgs);`
`echo count($wpPosts); //prints 10`
- Also note that if we add `'ignore_sticky_posts'=>true` to the query arguments, the correct number of posts is returned

WordPress 4.9.8, PHP 7.2.8, Timber 1.8.1

Composer/Packagist

### Proposed fix ###
Because WordPress' `get_posts()` sets `ignore_sticky_posts=true` as default, I think the same should be applied to `Timber::get_posts()`. The best way to do so may be by adding another "fix"-action in `QueryIterator.php`'s constructor.
I will add a PR in a few moments with a proposed fix.

bartvanraaij:
... to prevent fetching too many posts when using stickies.

This fixes #1812 and makes sure that `Timber::get_posts()` produces the same output as WP `get_posts()` when using sticky posts.

**Ticket**: #1812

#### Issue
See my bug report: when using sticky posts; the incorrect amount of posts in returned.

#### Solution
I've added a method `set_query_defaults` to `QueryIterator` which is hooked at  `pre_get_posts`. The method sets `ignore_sticky_posts` to `true` if not set otherwise.
This is the same behaviour as WP `get_posts`, see `wp-includes/post.php:1759`.

#### Impact
Other than fixing this bug, no impact, presumably.

#### Testing
I've added  `testGettingPostsWithStickiesReturnsCorrectAmountOfPosts`  in `test-timber-post-getter.php`
If you comment out the new `add_action set_query_defaults` at `QueryIterator.php:26` the test will fail. With my fix, it will pass.

### DIFF
--- a/lib/PostGetter.php
+++ b/lib/PostGetter.php
@@ -68,8 +68,8 @@ class PostGetter {
 	 * Sets some default values for those parameters for the query when not set. WordPress's get_posts sets a few of
 	 * these parameters true by default (compared to WP_Query), we should do the same.
 	 * @internal
-	 * @param WP_Query $query
-	 * @return WP_Query
+	 * @param \WP_Query $query
+	 * @return \WP_Query
 	 */
 	public static function set_query_defaults( $query ) {
 		if ( isset($query->query) && !isset( $query->query['ignore_sticky_posts']) ) {
--- a/lib/PostGetter.php
+++ b/lib/PostGetter.php
@@ -52,7 +52,7 @@ class PostGetter {
 	}
 
 	public static function get_posts( $query = false, $PostClass = '\Timber\Post', $return_collection = false ) {
-		add_filter( 'pre_get_posts', array('Timber\PostGetter', 'set_query_defaults') );
+		add_filter('pre_get_posts', array('Timber\PostGetter', 'set_query_defaults'));
 		$posts = self::query_posts($query, $PostClass);
 		return apply_filters('timber_post_getter_get_posts', $posts->get_posts($return_collection));
 	}
@@ -72,16 +72,16 @@ class PostGetter {
 	 * @return \WP_Query
 	 */
 	public static function set_query_defaults( $query ) {
-		if ( isset($query->query) && !isset( $query->query['ignore_sticky_posts']) ) {
+		if ( isset($query->query) && !isset($query->query['ignore_sticky_posts']) ) {
 			$query->set('ignore_sticky_posts', true);
 		}
-		if ( isset($query->query) && !isset( $query->query['supress_filters']) ) {
+		if ( isset($query->query) && !isset($query->query['supress_filters']) ) {
 			$query->set('supress_filters', true);
 		}
-		if ( isset($query->query) && !isset( $query->query['no_found_rows']) ) {
+		if ( isset($query->query) && !isset($query->query['no_found_rows']) ) {
 			$query->set('no_found_rows', true);
 		}
-		remove_filter( 'pre_get_posts', array('Timber\PostGetter', 'set_query_defaults') );
+		remove_filter('pre_get_posts', array('Timber\PostGetter', 'set_query_defaults'));
 		return $query;
 	}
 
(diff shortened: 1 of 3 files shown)


jarednova:
@bartvanraaij thanks for this PR. I think this reveals one of the differences between WP's `get_posts` vs. `WP_Query`. 

It appears that WP's `get_posts` ignores stickys, while `WP_Query` respects it.

I (think) merging this would fix the `get_posts` diffs between Timber and WP, but in the process create a disparity between `WP_Query` and `Timber\PostQuery`.

Can you take a look to see if there's a way (perhaps using a `$defaults` array) to have both methods match WordPress default behavior?

bartvanraaij:
Thanks for the feedback. I will take a look at that. 
I’m at a conference this week (IPC München) so it will be next week.

palmiak:
@jarednova @bartvanraaij If we want to make them working the same shouldn't we also add:
- `no_found_rows`
- `supress_filters`
?

jarednova:
@palmiak agreed, while we're here we may as well get the other parts of `get_posts` vs `WP_Query` to align

jarednova:
@palmiak I made some adjustments based on @bartvanraaij's code. I think this is in a good place, but I'd love your eyes in a review so we can make sure — thanks!

palmiak:
It looks good @jarednova . I just saw one thing in 2.0 branch related with this, but I'll open an issue for this.

jarednova:
@palmiak yeah, good call on how/if this comes into 2.0 — let's discuss it there

bartvanraaij:
Thank you @jarednova . This totally slipped my attention.

jarednova:
all good @bartvanraaij ! Thanks again for your contribution. I'll release later this week to Composer and WP.org!

lib/PostGetter.php
```diff
@@ -63,6 +64,27 @@ public static function query_post( $query = false, $PostClass = '\Timber\Post' )
...
+	 * @param \WP_Query $query
+	 * @return \WP_Query
+	 */
+	public static function set_query_defaults( $query ) {
+		if ( isset($query->query) && !isset($query->query['ignore_sticky_posts']) ) {
+			$query->set('ignore_sticky_posts', true);
+		}
+		if ( isset($query->query) && !isset($query->query['supress_filters']) ) {
+			$query->set('supress_filters', true);
+		}
```
gchtr: It should be `suppress_filters` instead of `supress_filters`!!! 🚨

Issue #1529, opened 2017-08-22
Incompatibility with TimberMenu


nkmaurya-dev:
Hi Team,

I'm using TimberMenu for my navigation layout. But non of plugins, which are built for tweaking navigation , are working with navigation layout.

I have seen the TImberMenu class and observed that there is no filter hook is used which are required by plugins, like 
 **1. wp_nav_menu_items
 2. wp_nav_menu_{$menu->slug}_items**

So please suggest me here that, how plugins can tweak my layout.

WordPress 4.8.1,
PHP 7.1,
Timber 1.4.1 (installed via WordPress.org)

Thanks,
Neeraj Maurya

gchtr:
These two filter that you mentioned are applied in the `wp_nav_menu()` function, which generates **HTML output for a menu**. As we can read in the filter description, they "filter the HTML list content for navigation menus".

With `Timber\Menu`, we don’t want to get the final HTML, but build our own HTML with Twig, by looping over arrays of menu items. The goal of this is that we don’t have to work with a Walker class to tweak certain HTML output, which can be quite complicated. That’s why Timber uses the `wp_get_nav_menu_items()` function to get menu items based on the menu ID. This function applies the `wp_get_nav_menu_items` filter, which is the preferred way to filter menu items present in a menu.

If you want to tweak the HTML output of a menu with a plugin, then probably `Timber\Menu` is not what you’re looking for. Remember that with Timber you can still use WordPress functions, including the standard way of displaying menus with `wp_nav_menu`.

So which plugins did you try to use? Or what do you want to achieve exactly?

blended-dev:
Hi gchtr,

Thanks for the response.

Yeah, I'm totally agree with you regarding filters and the TimberMenu concept, as I have been already observed these.

Actually, I have a template for navigation which has complete HTML/Twig structure and I'm including this template into extended template could say page.twig. I'm passing the complete menu item object with the help of TimberMenu from page.twig to navigation template where looping is happening. So just like any generic WordPress theme, I'm building a theme with the help of Timber/Twig,  to submit on **.org**  or could sell on **themeforest**. 

Its obvious that anyone can install plugins according to their own requirement. So I was just listing the plugins for maximum support from my theme and I have found that no menu related plugin is working with my theme.

I have mentioned those filters, because most of plugins are using those hooks for tweaking the navigation. Let me put an example here. Suppose a user is going to use the plugin **WooCommerce Menu Cart**, it would not work due to non existence of those hooks.

[...]

PR #1814, opened 2018-10-12
Added filters for plugins to hook into TimberMenu

### LINKED ISSUE #1732: Problems with WooCommerce menu item active element
### Expected behavior
I wanna use `{% if item.current %}active{% else %}nav-item{% endif %}` on WooCommerce menu items. But active element in WooCommerce main page didn't work.
Also tried `{% if item.slug == post.post_name %}active{% endif %}` but same thing, didn't work.
All other page type menu item active elements (page.twig, also single product, product archive, cart and so on) works perfectly, but not WooCommerce main page.
When in example Cart page parent page is WooCommerce main page then Cart active element's also didn't work, but when I use Cart page without parent page then it's working perfectly.

So... If I visit WooCommerce shop main page (/shop), then this page active element didn't work.

### Actual behavior
My menu navigation twig file:
https://i.imgur.com/SlLOito.png

WordPress 4.9.6, PHP 7.2, WooCommerce 3.4.2 and Timber 1.7.1

### AND WHEN I INSTALL WOOCOMMERCE STEP BY STEP:
https://timber.github.io/docs/guides/woocommerce/
THEN I GET 500 ERRORS WHEN I VISIT WOOCOMMERCE MAIN PAGE...

WP.org / starter theme from Github

pascalknecht:
Fixes #1732


pascalknecht:
@jarednova @gchtr Let me know what you think about any problems which might occur from this change. This fixes the problem of WooCommerce not adding a active class as they use a filter to do so. See https://github.com/woocommerce/woocommerce/blob/d0491072e8a7d0a4535db0ec2bae4a6d43ff64fd/includes/wc-page-functions.php#L154 for reference.

gchtr:
This is great! The base problem here is that most of the themes use the `wp_nav_menu()` function to render a navigation menu, and many plugins hook into this function, right? The `wp_nav_menu()` function uses 7 filters:

1. `wp_nav_menu_args` – Filters the arguments used to display a navigation menu.
2. `pre_wp_nav_menu` – Filters whether to short-circuit the wp_nav_menu() output.
3. `wp_nav_menu_container_allowedtags` – Filters the list of HTML tags that are valid for use as menu containers.
4. `wp_nav_menu_objects` – Filters the sorted list of menu item objects before generating the menu's HTML.
5. `wp_nav_menu_items` – Filters the HTML list content for navigation menus.
6. `wp_nav_menu_{$menu->slug}_items` – Filters the HTML list content for a specific navigation menu.
7. `wp_nav_menu` – Filters the HTML content for navigation menus.

Timber doesn’t hook into any of these. But there are quite a few plugins out there that allow users to customize a menu. In #1529, which is kind of related to this, the [WP Menu Cart](https://wordpress.org/plugins/wp-menu-cart/) plugin is mentioned, which hooks into the `wp_nav_menu_{$menu->slug}_items` filter.

I guess with this pull request, it’s quite easy to add support for the `wp_nav_menu_objects` filter. I don’t see any problems with this, currently 👍.

Now, I had a quick thought: The title of this pull request is «Added filters for plugins to hook into TimberMenu». Would it be possible to add support for filters 5 and 6, `wp_nav_menu_items` and `wp_nav_menu_{$menu->slug}_items` as well?

Because these two filter HTML, it might be a little trickier. We could use [a proper Twig example](https://timber.github.io/docs/guides/menus/#displaying-the-menu-items). If I’m not totally mistaken, the following might work. We’d have to provide the `default_args` and the `slug` as properties of `Timber\Menu`.

```twig
{% for item in menu.items %}
    <li class="nav-main-item {{ item.classes|join(' ') }}">
        <a class="nav-main-link" href="{{ item.link }}">{{ item.title }}</a>
        {% if item.children %}
            <ul class="nav-drop">
                {% for child in item.children %}
                    <li class="nav-drop-item">
                        <a href="{{ child.link }}">{{ child.title }}</a>
                    </li>
                {% endfor %}
            </ul>
        {% endif %}
    </li>
    {{ ''|apply_filters('wp_nav_menu_items', '', menu.default_args) }}
    {{ ''|apply_filters('wp_nav_menu_' ~ menu.slug ~ 'items', '', menu.default_args) }}
{% endfor %}
```

Or we could add a custom iterator like we did for the `PostsIterator`, where we would add a hook after the loop has ended:

https://github.com/timber/timber/blob/5101648dd78bdb0b3546864e6674cd284327af9c/lib/PostsIterator.php#L51-L55

I think we don’t need to add support for the other filters. What do you guys think?

pascalknecht:
@gchtr yes exactly. Many plugins hook into this kind of functions. Here a quick feedback to your other suggestions (didn't yet have a deep dive into all filters):

wp_nav_menu_args – Filters the arguments used to display a navigation menu. --> Will not be needed as we don't have menu args in Timber.
pre_wp_nav_menu – Filters whether to short-circuit the wp_nav_menu() output. -->Could maybe be implemented. Not sure if needed though.
wp_nav_menu_container_allowedtags – Filters the list of HTML tags that are valid for use as menu containers. --> Will not be needed
wp_nav_menu_objects – Filters the sorted list of menu item objects before generating the menu's HTML. --> Definitely needed as WooCommerce and other plugins hook into.
wp_nav_menu_items – Filters the HTML list content for navigation menus. --> We haven't digged into these kind of html filters yet. Will probably only be needed by theme developers who need a lot of 3rd party plugins to be compatible. We will need to think about if and how we would want to do that.
wp_nav_menu_{$menu->slug}_items – Filters the HTML list content for a specific navigation menu. --> Same as wp_nav_menu_items
wp_nav_menu – Filters the HTML content for navigation menus. --> Save as wp_nav_menu_items

In my opinion we could probably merge this pull request and have a look at whether or not to implement the other filters in another ticket?

gchtr:
@pascalknecht So we’re on the same page here 👍. I’ll open a separate issue for the compatibility with the two other filters.

For this pull request, I applied coding standards and added comments. I’ll leave it to @jarednova to say whether we also need tests for this and to merge this in.

jarednova:
@pascalknecht this is awesome. Anything that can solve other plugin / WooCommerce compatibility will a huge boon. I think one test to verify the basic behavior would be a good move (def. not worth covering every default property, etc.). After that I think we're good to 👍

pascalknecht:
@jarednova @gchtr  Here we go. Added a little test to check if the filter works. Does that look allright to you?

gchtr:
This looks good to me! Because this is a pull request against the master branch, I’ll leave it to @jarednova to merge this one in.

jarednova:
Good to me as well, thanks @pascalknecht !

PR #1819, opened 2018-10-16
Responsive image attributes

maxxwv:
#### Issue
I didn't see an easy way to get the srcset and sizes attributes for a Timber\Image object that was as simple as `post.thumbnail.src`.

#### Solution
It seemed like a good idea to add `srcset` and `sizes` as public methods, allowing `post.thumbnail.srcset` and `post.thumbnail.sizes`. Unfortunately, `.sizes` didn't work but `img_sizes` did.

#### Impact
As far as I can tell, none.

#### Usage Changes
Should be no impact.

#### Considerations
For whatever reason, creating a size() method in the Image object wasn't working. It didn't throw an error, just that `post.thumbnail.sizes` didn't ever call the method. It'd be a more intuitive call than `img_sizes`, which would be nice.

#### Testing
There's really nothing to fail, it's just a method to get the data that's more in keeping with the rest of the project.

### DIFF
--- a/lib/Image.php
+++ b/lib/Image.php
@@ -467,8 +467,8 @@ class Image extends Post implements CoreInterface {
 	 */
 	public function srcset( $size = "full" ) {
 		$srcset = wp_get_attachment_image_srcset($this->ID, $size);
-		$srcset = apply_filters('timber/image/srcset', $srcsset, $this->ID);
-		$srcset = apply_filters('timber_image_srcset', $srcsset, $this->ID);
+		$srcset = apply_filters('timber/image/srcset', $srcset, $this->ID);
+		$srcset = apply_filters('timber_image_srcset', $srcset, $this->ID);
 		return $srcset;
 	}
 	
--- a/lib/Image.php
+++ b/lib/Image.php
@@ -466,7 +466,10 @@ class Image extends Post implements CoreInterface {
 	 *	@return bool|string
 	 */
 	public function srcset( $size = "full" ) {
-		return wp_get_attachment_image_srcset($this->ID, $size);
+		$srcset = wp_get_attachment_image_srcset($this->ID, $size);
+		$srcset = apply_filters('timber/image/srcset', $srcsset, $this->ID);
+		$srcset = apply_filters('timber_image_srcset', $srcsset, $this->ID);
+		return $srcset;
 	}
 	
 	/**
@@ -483,7 +486,10 @@ class Image extends Post implements CoreInterface {
 	 *	@return bool|string
 	 */
 	public function img_sizes( $size = "full" ) {
-		return wp_get_attachment_image_sizes($this->ID, $size);
+		$img_sizes = wp_get_attachment_image_sizes($this->ID, $size);
+		$img_sizes = apply_filters('timber/image/img_sizes', $img_sizes, $this->ID);
+		$img_sizes = apply_filters('timber_image_img_sizes', $img_sizes, $this->ID);
+		return $img_sizes;
 	}
 	
 	/**
--- a/lib/Image.php
+++ b/lib/Image.php
@@ -453,6 +453,40 @@ class Image extends Post implements CoreInterface {
 	}
 
 	/**
+	 * @param string $size a size known to WordPress (like "medium")
+	 * @api
+	 * @example
+	 * ```twig
+	 * <h1>{{ post.title }}</h1>
+	 * <img src="{{ post.thumbnail.src }}" srcset="{{ post.thumnbail.srcset }}" />
+	 * ```
+	 * ```html
+	 * <img src="http://example.org/wp-content/uploads/2018/10/pic.jpg" srcset="http://example.org/wp-content/uploads/2018/10/pic.jpg 1024w, http://example.org/wp-content/uploads/2018/10/pic-600x338.jpg 600w, http://example.org/wp-content/uploads/2018/10/pic-300x169.jpg 300w "
+	 * ```
+	 *	@return bool|string
+	 */
+	public function srcset( $size = "full" ) {
+		return wp_get_attachment_image_srcset($this->ID, $size);
+	}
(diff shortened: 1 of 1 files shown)


lib/Image.php
```diff
@@ -452,6 +452,46 @@ public function src( $size = 'full' ) {
...
+	 * ```
+	 * ```html
+	 * <img src="http://example.org/wp-content/uploads/2018/10/pic.jpg" srcset="http://example.org/wp-content/uploads/2018/10/pic.jpg 1024w, http://example.org/wp-content/uploads/2018/10/pic-600x338.jpg 600w, http://example.org/wp-content/uploads/2018/10/pic-300x169.jpg 300w sizes="(max-width: 1024px) 100vw, 102" />
+	 * ```
+	 *	@return bool|string
+	 */
+	public function img_sizes( $size = "full" ) {
+		$img_sizes = wp_get_attachment_image_sizes($this->ID, $size);
+		$img_sizes = apply_filters('timber/image/img_sizes', $img_sizes, $this->ID);
+		$img_sizes = apply_filters('timber_image_img_sizes', $img_sizes, $this->ID);
```
jarednova: We're deprecating this filter style, so you can remove
```suggestion
```

jarednova:
Thank you @maxxwv! This looks like a good implementation of the WP methods. Before we merge, I want to make sure we have test coverage (even though it's a pretty simple thing, I want to make sure the functionality is always checked). Is that something you'd be up for trying?

lib/Image.php
```diff
@@ -452,6 +452,46 @@ public function src( $size = 'full' ) {
...
+	 * ```
+	 * ```html
+	 * <img src="http://example.org/wp-content/uploads/2018/10/pic.jpg" srcset="http://example.org/wp-content/uploads/2018/10/pic.jpg 1024w, http://example.org/wp-content/uploads/2018/10/pic-600x338.jpg 600w, http://example.org/wp-content/uploads/2018/10/pic-300x169.jpg 300w" />
+	 * ```
+	 *	@return bool|string
+	 */
+	public function srcset( $size = "full" ) {
+		$srcset = wp_get_attachment_image_srcset($this->ID, $size);
+		$srcset = apply_filters('timber/image/srcset', $srcset, $this->ID);
+		$srcset = apply_filters('timber_image_srcset', $srcset, $this->ID);
```
```suggestion
```

gchtr: Old filter style, see Jared’s review.

lib/Image.php
```diff
@@ -452,6 +452,46 @@ public function src( $size = 'full' ) {
...
+	 * <img src="{{ post.thumbnail.src }}" srcset="{{ post.thumnbail.srcset }}" />
+	 * ```
+	 * ```html
+	 * <img src="http://example.org/wp-content/uploads/2018/10/pic.jpg" srcset="http://example.org/wp-content/uploads/2018/10/pic.jpg 1024w, http://example.org/wp-content/uploads/2018/10/pic-600x338.jpg 600w, http://example.org/wp-content/uploads/2018/10/pic-300x169.jpg 300w" />
+	 * ```
+	 *	@return bool|string
+	 */
+	public function srcset( $size = "full" ) {
+		$srcset = wp_get_attachment_image_srcset($this->ID, $size);
+		$srcset = apply_filters('timber/image/srcset', $srcset, $this->ID);
```
gchtr: I’m not sure if we should add this filter, because the `wp_get_attachment_image_srcset()` function already provides 3 filters to filter the resulting srcset. What would be the benefit of having this against using the filters in `wp_get_attachment_image_srcset()`?

lib/Image.php
```diff
@@ -452,6 +452,46 @@ public function src( $size = 'full' ) {
...
+	 * <img src="{{ post.thumbnail.src }}" srcset="{{ post.thumnbail.srcset }}" sizes="{{ post.thumbnail.sizes }}" />
+	 * ```
+	 * ```html
+	 * <img src="http://example.org/wp-content/uploads/2018/10/pic.jpg" srcset="http://example.org/wp-content/uploads/2018/10/pic.jpg 1024w, http://example.org/wp-content/uploads/2018/10/pic-600x338.jpg 600w, http://example.org/wp-content/uploads/2018/10/pic-300x169.jpg 300w sizes="(max-width: 1024px) 100vw, 102" />
+	 * ```
+	 *	@return bool|string
+	 */
+	public function img_sizes( $size = "full" ) {
+		$img_sizes = wp_get_attachment_image_sizes($this->ID, $size);
+		$img_sizes = apply_filters('timber/image/img_sizes', $img_sizes, $this->ID);
```
gchtr: Also here, I’m not sure if we need this filter at all, because `wp_get_attachment_image_sizes()` calls `wp_calculate_image_sizes()` in the last line, which in turn has a `wp_calculate_image_sizes` filter in the last line. So the `timber/image/img_sizes` filter would be called immediately after `wp_calculate_image_sizes`. What would be the benefit of this?

maxxwv:
Thanks for the comments!

@jarednova - Sure, I'll take a stab at some basic tests, though honestly it's been a while since I've done unit testing so I can't guarantee much there. I'll look at the test-timber-image.php file and maybe the resize tests for examples and inspiration, unless there's a more appropriate place.

@gchtr - I'm fine with pulling out those filters. I just put them in to match the src() interface more closely, to be honest. I'll pull them out in the next revision.

I was also thinking it could be handy to have a method that runs both these methods and returns a single string of the attributes, as neither is really all that useful on their own. Thoughts or opinions?

maxxwv:
@gchtr @jarednova - not sure what to do about the coverage failure, but everything else should be good to go. Let me know if there's anything else you need from me.

drzraf:
The phpdoc example provided uses `sizes="{{ post.thumbnail.sizes }}"`  instead of `sizes="{{ post.thumbnail.img_sizes }}"`

PR #1829, opened 2018-10-19
2.x Documentation updates

gchtr:
This is a pull request that makes some updates to the documentation for 2.x.

## Changes

- Updated some smaller syntax issues.
- Added new Guide **Custom Fields**.
- Added a `docs/_index.md` file to generate a list of subpages for the main documentation index.
- Renamed **Testing** to **Unit Testing**.

## Different order for doc pages

I added a **weight** argument to the FrontMatter block in guides to support a custom order in the documentation. The guides will not be ordered by name anymore, but by weight.

The new order will be as follows:

- Context
- Template Locations
- WordPress Integration
- Twig Cookbook
- Text Cookbook
- Escaping
- Custom Page Templates
- Custom Fields
- Menus
- Functions
- Filters
- Image Cookbook
- Pagination
- Sidebars
- Performance/Caching
- Internationalization
- Cheatsheet
- Extending Timber
- Debugging
- Unit Testing
- Routing

## An Integrations section

While Timber grew, support for several third party plugins was added. But not everyone might use all the available integrations in all the projects. In the documentation, there’s now a separate section for integrations.

Some integrations might have to be loaded into Timber separately in the future, like the [new WooCommerce integration](https://github.com/mindkomm/timber-integration-woocommerce). And the documentation for them can grow so big that one documentation page or each integration wouldn’t be enough. That’s why the integration section supports subpages. Here’s a quick preview for how this might look like:

## Considerations

There’s still a Routing Guide present. When we remove routes from version 2, we should also remove the Routing Guide.


gchtr:
@jarednova I’d merge this into `2.x` now. Work on the documentation is not finished. But it’s easier to make pull requests against `2.x` if the current documentation and the Upgrade Guide for 2.x is already in the `2.x` branch.

I added a new issue to remind myself to resolve the remaining `@todo`s: https://github.com/timber/timber/issues/1860.

jarednova:
Sounds good to me, merging!

Issue #1645, opened 2018-01-26
Timber and Gutenberg: asking for future plans


gorgoglionemeister:
Is there a Timber compatibility plan with the new Gutenberg editor? Each Gutenberg block generates HTML both in the backend and in the frontend. In short, lot of HTML that can not be edited (or only partially) and that you get as is. How can Timber work in this context?

palmiak:
I think there shouldn't be any problem with Timber. `post.content` will work just the way it works now.

On the other hand - blocks can be rendered with PHP or JS. The JS part doesn't concern us. 

The PHP part looks just like in this example https://github.com/WordPress/gutenberg/blob/master/docs/blocks-dynamic.md and here timber would be very usefull. And I think it will work - I'll try to provide you with working example in a couple of days.

gorgoglionemeister:
Great! Thank you. Do you think it will be easier to extend pre-existing blocks or write new ones with Timber? I also think that the HTML rendered with JS must be the same as that rendered with PHP, for consistency.

palmiak:
After some playing around I see that there is a problem with php rendering - I mean it works but you can't pass the block attributes as seperate variables (you can pass only block options).

We'll have to wait for this as It looks they are working on it. When pass variables will be working you will be able to make block parts in timber.

On the other hand the JS part doesn't look so bad as I thought it will be (I'm not a fan of JS). I made some basic editable blocks and they worked. I'm trying to rebuild my blog's content (which is now built on ACF flexible content) to Gutenberg.

Also - everything is changing very fast so all the missing things can appear in the next update.

gorgoglionemeister:
OK, please can you share what you will do when it is ready?

addedlovely:
Great, thanks for checking it out. Adding my comment here for future notifications :)

IngoVals:
[Found this on twitter](https://twitter.com/Maximebj/status/960203611950276609)

DrecDroid:
Hi, today trying to replace content `img`s with `amp-img` I discovered gutenberg. As a Timber lover I've been trying how to mix Timber and Gutenberg, and that tweet resolved all my questions.

Basically, you have two ways of rendering gutenberg blocks, statically and dynamically.

With the static rendering you render the same way in the editor and the frontend, so you only can use javascript for the rendering, this makes the things a bit complicated because you have to go back and forth from json to html. 

[...]

PR #1835, opened 2018-10-27
ACF Blocks documentation

palmiak:
**Ticket**: https://github.com/timber/timber/issues/1645

This the documentation for ACF Block in Timber.

I added the universal example and the examples for groups and repeaters - I don't think any other types are needed. Although if you think I should add some more I'll add them.

### DIFF
--- a/docs/guides/acf-gutenberg-cookbook.md
+++ b/docs/guides/acf-gutenberg-cookbook.md
@@ -1,0 +1,88 @@
+---
+title: "Gutenberg with ACF Blocks Cookbook"
+menu:
+  main:
+    parent: "guides"
+---
+
+## What are the ACF Blocks
+ACF Blocks are the easiest way to create content blocks for [Gutenberg](https://wordpress.org/gutenberg/). You don't need any advanced JavaScript knowledge to create new blocks. If you want to learn more about them read the article on [advancedcustomfields.com](https://www.advancedcustomfields.com/blog/acf-5-8-introducing-acf-blocks-for-gutenberg/).
+
+## How to use ACF Blocks with Timber
+Before you can start using ACF Blocks, you must install the Advanced Custom Fields 5.8.0-beta or later.
+
+To create your content block, first you have to register it in `functions.php` or in a seperate plugin:
+```php
+add_action('acf/init', 'my_acf_init');
+function my_acf_init() {
+	
+	// check function exists
+	if( function_exists('acf_register_block') ) {
+		
+		// register a new block
+		acf_register_block(array(
+			'name'				=> 'example_block',
+			'title'				=> __('Example Block'),
+			'description'		=> __('A custom example block.'),
+			'render_callback'	=> 'my_acf_block_render_callback',
+			'category'			=> 'formatting',
+			'icon'				=> 'admin-comments',
+			'keywords'		    => array( 'example' ),
+		));
+	}
+}
+```
+
+next you you have to create a `render_callback`:
+```php
+function my_acf_block_render_callback( $block ) {
+    // store block values
+    $vars['block'] = $block;
+
+    // store field values
+    $vars['fields'] = get_fields(); 
+
+    // render the block
+    Timber::render( '/template-parts/block/example-block.twig', $vars );
+}
+
+```
+you create an extra array called `$vars` with two values:
+- **block** - with all data like block title, alignment etc
+- **fields** - all custom fields - also all the fields created in **ACF**
+
+finally you can create the template `template-parts/block/example-block.twig`:
+```twig
+{#
+/**
+ * Block Name: Example block
+ *
+ * This is the template that displays the example block.
+ */
+#}
+<div id="example-{{ block.id }}" class="wrapper">
+    <h1>{{ fields.title }}</h1>
+    <p>{{ fields.description }}</p>
+</div>
+<style type="text/css">
+	#testimonial-{{ block.id }} {
+		background: {{ fields.background_color }};
+		color: {{ fields.text_color }};
+	}
+</style>
+
+```
+
+## Using repeaters
+```
+{% for field in fields.repeater %}
+    Title: {{ field.title }} <br/>
+    Url: {{ field.url }}
+{% endfor %}
+```
+
+## Using groups
+```
+Title: {{ fields.group.title }} <br/>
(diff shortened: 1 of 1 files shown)


gchtr:
Hey @palmiak 

Very good! Thanks a lot for taking up the Gutenberg documentation.

I think the title «Gutenberg with ACF Blocks Cookbook» is a little too specific. I’d go for a filename **gutenberg.md** with the title «Gutenberg». So it would be a Gutenberg guide instead of a guide for how Timber works with ACF Blocks. Then, I’d add a H2 called «ACF Blocks», where you explain everything about ACF Blocks.

In a section before that, we could explain how Timber works with Gutenberg *in general*, like you do in https://github.com/timber/timber/issues/1645#issuecomment-360952899. Something like:

> Timber works with Gutenberg out of the box. If you use `{{ post.content }}`, Timber will render all the Gutenberg blocks.

In case the Gutenberg guide becomes too long at some point in the future, we could think about splitting it up into multiple guides then.

What do you think?

palmiak:
Great thoughts. The part about post.content is so obvious I forgot about it
and it's very important.

3 lis 2018 11:55 "Lukas Gächter" <notifications@github.com> napisał(a):

Hey @palmiak <https://github.com/palmiak>

Very good! Thanks a lot for taking up the Gutenberg documentation.

I think the title «Gutenberg with ACF Blocks Cookbook» is a little too
specific. I’d go for a filename *gutenberg.md* with the title «Gutenberg».
So it would be a Gutenberg guide instead of a guide for how Timber works
with ACF Blocks. Then, I’d add a H2 called «ACF Blocks», where you explain
everything about ACF Blocks.

In a section before that, we could explain how Timber works with Gutenberg *in
general*, like you do in #1645 (comment)
<https://github.com/timber/timber/issues/1645#issuecomment-360952899>.
Something like:

Timber works with Gutenberg out of the box. If you use {{ post.content }},
Timber will render all the Gutenberg blocks.

In case the Gutenberg guide becomes too long at some point in the future,
we could think about splitting it up into multiple guides then.

What do you think?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/timber/timber/pull/1835#issuecomment-435578622>, or mute
the thread

palmiak:
@gchtr I changed the file name, the title, structure (h2s, h3s)  and added the part about Gutenberg with plain Timber.

docs/guides/gutenberg.md
```diff
@@ -0,0 +1,91 @@
...
+<style type="text/css">
+	#testimonial-{{ block.id }} {
+		background: {{ fields.background_color }};
+		color: {{ fields.text_color }};
+	}
+</style>
+
+```
+
+### Using repeaters
```
gchtr:
```suggestion
### Using repeaters

```

docs/guides/gutenberg.md
```diff
@@ -0,0 +1,91 @@
...
+```
+{% for field in fields.repeater %}
+    Title: {{ field.title }} <br/>
+    Url: {{ field.url }}
+{% endfor %}
+```
+
+### Using groups
```
gchtr:
```suggestion
### Using groups

```

docs/guides/gutenberg.md
```diff
@@ -0,0 +1,91 @@
...
+ * Block Name: Example block
+ *
+ * This is the template that displays the example block.
+ */
+#}
+<div id="example-{{ block.id }}" class="wrapper">
+    <h1>{{ fields.title }}</h1>
+    <p>{{ fields.description }}</p>
+</div>
+<style type="text/css">
```
gchtr: I think inline styles are a good example. But I would also like to add a hint for how you can register your CSS files to work with Gutenberg so that CSS doesn’t have to necessarily be written into Twig files.

@pascalknecht has written a short tip for this: https://pascalknecht.ch/gutenberg-quick-tips-01-editor-styles/.

docs/guides/gutenberg.md
```diff
@@ -0,0 +1,91 @@
...
+	#testimonial-{{ block.id }} {
+		background: {{ fields.background_color }};
+		color: {{ fields.text_color }};
+	}
+</style>
+
```
gchtr:
```suggestion
```

palmiak:
Thanks @gchtr 👍 Awesome review.

To do:
- [ ] check the indentation
- [ ] add example of editor css / js
- [ ] add `supports` param

palmiak:
@gchtr I have checked the indentation (4 spaces everywhere), fixed array to 5.x style in one place and added the stuff about external assets.

gchtr:
As far as I can tell, this looks good! 👍

If we see that there’s more documentation needed, we can always come back to it.

docs/guides/gutenberg.md
```diff
@@ -0,0 +1,116 @@
...
+        ) );
+    }
+}
+```
+
+Next, you you have to create your `render_callback()` function:
+
+```php
+function my_acf_block_render_callback( $block ) {
+    $context = array();
```
pascalknecht: This implementation creates a high amount of confusion for developers as the cannot use `theme` or `site` variables afterwards. Maybe use `$context = Timber::get_context()` here? This should not negatively impact performance right @gchtr ?

pascalknecht: cc @palmiak

palmiak: @pascalknecht very good point. To be honest I was thinking about this but most times when I'm building blocks (Gutenberg or Flexible Content) I'm trying to pass only what's necessary, but your right in terms of user confusion. 

I also think we should add this. Less problems and less issues.

gchtr: In Timber 1.x, you can call `Timber::get_context()` again, without losing performance 👍. So I’d add this as well.

For Timber 2.x, we’d have to use [`Timber::context_global()`](https://github.com/timber/timber/blob/2.x/docs/guides/context.md#context-cache). You could add a second pull request with this change for 2.x once this is merged in. Otherwise, we might forget it again.

gchtr: Pull request submitted #1872!

gchtr:
The guide is now live: https://timber.github.io/docs/guides/gutenberg/

Issue #1838, opened 2018-11-05
Parse error: syntax error, unexpected '[' 


vanekj:
### Expected behavior
It should display the website, but after upgrade it leads to an PHP error.

### Actual behavior
`Parse error: syntax error, unexpected '[' in /data/www/wp-content/plugins/timber-library/lib/ImageHelper.php on line 155`

Using an `array()` fixed the issue for me.

### Steps to reproduce behavior
Use the latest version of Timber when running an older PHP version (eg. 5.3 in my case)

### What version of WordPress, PHP and Timber are you using?
- WordPress 4.9.8
- PHP 5.3
- Timber 1.8.1

### How did you install Timber?
Upgraded to newest version via plugin updater in WordPress dashboard

palmiak:
@vanekj I've created a pull request to fix this - https://github.com/timber/timber/pull/1839 . Check is this fixes your problem.

vanekj:
Yep, this solution fixes the issue

PR #1839, opened 2018-11-05
Patch 1838

palmiak:
**Ticket**: #1838 

#### Issue
No php 5.3 support

#### Solution
Changed `[]` to `array()`.

#### Impact
It will work on older php versions.


jarednova:
Thanks @palmiak — just had to make a quick fix to an (unrelated) change to how YouTube embeds are output from WP. Merged!

PR #1841, opened 2018-11-07
Update autoescape option

danielpost:
#### Issue
As discussed [here](https://github.com/timber/timber/pull/1786#issuecomment-433835735), `autoescape = true` has been deprecated since Twig 1.21 in favor of `autoescape = 'html'`.

#### Solution
This PR updates the docs to reflect this change, and adds a fallback to automatically change `Timber::$autoescape = true` to `Timber::$autoescape = 'html'` if applicable.

#### Impact
No impact.

#### Usage Changes
It is recommended to update `Timber::$autoescape = true` to `Timber::$autoescape = 'html'` to ensure compatibility with Timber 2.x.

#### Considerations
N/A

#### Testing
N/A

PR #1842, opened 2018-11-07
2.x: Update escaping docs

danielpost:
Almost identical to https://github.com/timber/timber/pull/1841, so please look there for details.

Differences:
- This PR doesn't add a fallback if `autoescape` is set to `true`.
- This PR adds a note about the change in the Upgrade Guide.


gchtr:
@danielpost Cool, thank you for your help! 👍 I added some improvements to be more explicit in the changes.

When the fallback is merged into master, we’ll have it in 2.x as well.

PR #1843, opened 2018-11-08
Optimize SVG mime check

gchtr:
**Ticket**: #1797, #1798 

## Issue

Reading a file type through either [`mime_content_type()`](http://php.net/manual/en/function.mime-content-type.php) or the `finfo` class requires the **fileinfo extension** to be loaded. Falling back from `mime_content_type()` to `finfo` like introduced in #1798 maybe isn’t helpful if the reason that `mime_content_type()` doesn’t exist is the missing extension.

I just happened to run into a fatal error on a site that didn’t have the extension installed.

## Solution

Checking mime types is tricky. Fortunately, WordPress can do it for us. There are two tickets in core ([39950](https://core.trac.wordpress.org/ticket/39550) and [39552](https://core.trac.wordpress.org/ticket/39552)) that made changes to `wp_check_filetype_and_ext()`. That function now uses `finfo` if the extension is available and falls back to a check for the extension in the filename by default. Because SVG images are not allowed by default in WordPress, we have to pass a default mime type for SVG images to it.

- Replaced the check in `Timber\ImageHelper::is_svg()` with a check to the native `wp_check_filetype_and_ext()` function.
- Removed the recently introduced `_mime_content_type()` method.
- Added a test to check the `Timber\ImageHelper::is_svg()` function.

## Impact

I don’t know if we can remove the `_mime_content_type()` method again, or if we have to deprecated it properly. It was only introduced in 1.8.1 and was thought to be used internally.

## Usage Changes

Nope.

## Considerations

Performance, maybe?

## Testing

I added a new test together with the Timber logo as an SVG to test out the function.


jarednova:
What, WP had a function for that this whole time???? I'll take a look!

jarednova:
Still kicking myself for writing that huge workaround when such an easy fix was sitting there. Won't make that mistake again. Thanks for the fix @gchtr !

gchtr:
@jarednova You bet! WordPress has a function for everything! And believe me, I have been there, too. And I had to rewrite it three times until I came to the solution here 😅.

palmiak:
@gchtr I just hit this bug too - thanks for fixing it :)

@jarednova could you tag this as version 1.8.2 or something? Because now I had to change version number do dev-master in package.json to update this.

jarednova:
@palmiak yup! will take care of that right now

Issue #1828, opened 2018-10-18
Apply `maybe_convert_array` to all class functions in Timber\Twig


jarednova:
As created in #1788, implement the `maybe_convert_array` method throughout `Timber\Twig`

https://github.com/timber/timber/blob/2.x/lib/Twig.php#L91-L93

seancrater:
@jarednova Think this might be a decent first contribution? If so, want to keep running with `pid` as opposed to `post_id` for argument naming convention?

jarednova:
def. `post_id` please! `clarity > brevity`

seancrater:
I’m going to work this out tonight! Assuming the PR should be against the 2.x branch too 😉

seancrater:
Created that PR whenever you have a chance @gchtr and @jarednova, thanks!

PR #1846, opened 2018-11-11
Implementing maybe_convert_array to all class functions in Timber/Twig

seancrater:
**Ticket**: #1828 

#### Issue
Currently the `maybe_convert_array` utility method is only applied in one instance when there are multiple class instances that could take advantage of it rather than explicitly repeating the same logic over and over.

#### Solution
Replaces repetitive code with more sustainable method that had been created in a prior PR.

#### Impact
Will make code more maintainable and should have no effect on the behavior since the logic is the same, just re-used rather than re-declared.

#### Usage Changes
N/A, Twig Functions will still behave the same 😄 

#### Considerations
I renamed a couple of the arguments in favor of consistency in code styling.

#### Testing
Existing unit tests should cover this PR since none of the actual functionality has changed.

### DIFF
--- a/lib/Twig.php
+++ b/lib/Twig.php
@@ -163,7 +163,14 @@ class Twig {
 			$TermClass = $processed_args['TermClass'];
 		}
 
-		return self::maybe_convert_array( $term_id, $TermClass );
+		if ( is_array($term_id) && !Helper::is_array_assoc($term_id) ) {
+			foreach ( $term_id as &$p ) {
+				$p = new $TermClass($p, $taxonomy);	
+			}	
+			return $term_id;
+		}
+
+		return new $TermClass($term_id, $taxonomy);
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


lib/Twig.php
```diff
@@ -168,26 +150,20 @@ public static function maybe_convert_array( $post_id, $class ) {
...
 			$taxonomy = $processed_args['taxonomy'];
 			$TermClass = $processed_args['TermClass'];
 		}
 
-		if ( is_array($tid) && !Helper::is_array_assoc($tid) ) {
-			foreach ( $tid as &$p ) {
-				$p = new $TermClass($p, $taxonomy);
-			}
-			return $tid;
-		}
```
seancrater: Just noticed I unintentionally removed the taxonomy arg entirely...since maybe_convert_array only currently supports one arg, should we expand it to take an array of args perhaps?

jarednova:
Awesome, let's pull this into 2.x — thanks @seancrater !

Issue #1849, opened 2018-11-19
webp not working


bobhuf:
### Expected behavior
webp generation

### Actual behavior
jpg is generated in stead...

### Steps to reproduce behavior
Used the basic example you [posted on twitter](https://twitter.com/TimberWP/status/1042066512813744129):
```
{% set image = Image(50) %}
<picture>
  <source srcset="{{ image.src | towebp }}" type="image/webp">
  <img src="{{ image.src }}" alt="{{ image.alt }}">
</picture>
```

### What version of WordPress, PHP and Timber are you using?
WordPress 4.9.8, PHP 7.1.12, Timber 1.8.1

palmiak:
Hi :)

First of all:
``` 
var_dump( gd_info() );
```
Let's be sure your server has webp support.

pascalknecht:
Yes your server probably does not support the image extension needed for the conversion. You may want to check with your hoster.

jarednova:
Thanks for the report @bobhuf! It seems that this is a server issue with your `gd` or `imagick`. That said, perhaps we can add something to the error logs alerting the dev to WebP not being supported

bobhuf:
You are all right, my local setup (MAMP) has GD enabled but without webp support. Everything works as expected on our dev servers. Maybe add some more info about requirements to the docs as well?

jarednova:
I just submitted #1854 to resolve the documentation gap

jarednova:
Closed by #1854

PR #1854, opened 2018-11-25
ref #1849 -- add docs to clarify WEBP behaviors

jarednova:
Added some documentation clarification based on problems supported in #1849

### DIFF
--- a/lib/Image/Operation/ToWebp.php
+++ b/lib/Image/Operation/ToWebp.php
@@ -45,6 +45,11 @@ class ToWebp extends ImageOperation {
             return false;
         }
 
+        // Attempt to check if SVG.
+		if ( ImageHelper::is_svg($load_filename) ) {
+			return false;
+		}
+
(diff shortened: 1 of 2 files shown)


jarednova:
Could someone give this a second look and let me know if you see any issues/errors/shortfalls?

palmiak:
@jarednova I think it's ok - just one thing shouldn't toWebp have `is_svg` check?

lib/Image/Operation/ToWebp.php
```diff
@@ -2,9 +2,12 @@
 
 namespace Timber\Image\Operation;
 
+use Timber\Helper as Helper;
 use Timber\Image\Operation as ImageOperation;
 
 /**
+ * This class is used to process webp images. Not all server configurations support webp. 
+ * If webp is not enabled, Timber will generate webp images instead
```
szepeviktor, line 10: @jarednova We may have a brain bug 🐞 here :)

PR #1857, opened 2018-11-26
Removing Dependency Status

palmiak:
There is something wrong with dependancy status badge - https://screenshots.firefox.com/uz9fx7fkydqkGfop/github.com

And it redirects to https://afiliados.medicapanamericana.com/user/projects/574e40e6e298f30048059b9f - it looks wrong.


jarednova:
Yup! I'd say that one looks like it's gone

PR #1859, opened 2018-11-27
2.x Remove deprecated functionality in $post->terms()

gchtr:
**Ticket**: #1845

## Issue

In #1845, I said I was going to remove the deprecated functionality for `$post::terms()`.

## Solution

- Remove deprecated `$merge` and `$term_class` parameters.
- Remove fallbacks to `post_tag` when `tag` or `tags` is used.
- Remove fallback to `category` when `category` is used.
- Update tests.

## Impact

Backwards compatibility with 1.x will be removed.

## Usage Changes

Only an arguments array or a taxonomy string or an array of taxonomy strings will be accepted as a parameter.

## Considerations

None.

## Testing

I updated the tests. They ran successfully.


jarednova:
👋 to arbitrary argument ordering. How we won't miss you. I just fixed-up an error log message, otherwise looks good!

Issue #1827, opened 2018-10-18
Deprecate `get_field`


jarednova:
- [x] Deprecate `Core::get_field()` (in favor of meta)
- [x] Deprecate any other class's `::get_field()` (in favor of meta)
- [x] Ensure docs reflect `meta` > `get_field`
- [x] Add to upgrade guide
- [ ] ~Implement defensive [`__get()` solution](https://github.com/timber/timber/pull/1618#issuecomment-356441760) proposed by @gchtr~

palmiak:
But what about ACF integration? Beacuse most user, when using ACF, would probably like to use `post.get_field` - maybe we should add the alias only when ACF is installed. What do you think?

palmiak:
@gchtr ok - I fully understand and I think this is a great approach - especially the separation of integrations.

But let's go a bit further with ACF example. `get_field` has 3 parameters - `id`, `field_name` and `format_value` - in case i would like to set `format_value` to false how should I approach this? 
What do you think about adding the second parameter to `meta()` called `$args` with this, it would be easier for everybody to add his own parameters for integrations `post.meta( 'some_field', { 'format': false } )`.

palmiak:
@gchtr I did some quick research on the plugins that are similar to ACF (creating metaboxes and displaying them). 

**Plugins that use only one parameter:**
- Custom Field Suite
- ultimate-fields

**Plugins that use more than one parameter:**
- ACF - but as you saw we can use the `raw_meta` - although IMO it would be more logical to use the same function and different arguments
- Toolset is using `echo(types_render_field( 'field-slug', array( 'arg1' => 'val1', 'arg2' => 'val2' ) ));`
- Pods - `pods_field ( $pod, $id = false, $name = null, $single = false )`
- Carbon Fields - `carbon_get_post_meta( $id, $name, $container_id = '' )`
- Metabox.io - `rwmb_meta( $field_id, $args, $post_id )`
- Papi - `papi_get_field( $meta, $slug, $default )`

**Plugins using native `get_post_meta`:**
- CMB2
- http://fieldmanager.org/
- Piklist
- Custom Post Types and Custom Fields creator – WCK

In short - two most popular plugins are using more than one argument - so I think it would be safer and easier for integration creators if we could provider them with the possibility to pass all the parameters. 

[...]

Issue #1823, opened 2018-10-17
Use raw meta values in a `raw_meta` property only, and not as properties on the class itself.


jarednova:
See comments in #1809.

### Acceptance Criteria
- [x] `Post::raw_meta` returns an array of properties with the data stored in the `wp_postmeta` table. The data should be unprocessed by ACF and other plugins (probably the same as what's returned from WP's native functions)
- [x] Can utilize existing performance overhead (no new queries, etc.)
- [x] Preserve current behavior where fetching data using magic `__get` DOES return the ACF-processed data (current behavior)

gchtr:
@jarednova I have some questions, because I’m not sure what you envisioned here exactly. 

## raw_meta and/or custom?

In https://github.com/timber/timber/pull/1809, you write:

> For `2.x` this could impact how we handle custom field data. I've found that it's MUCH easier/cleaner here if those fields are retrieved at runtime (ie `{{ post.meta("my_field") }}`) vs. initialization (where the data is attached to the `Timber\Post` object before it's even fetched).

> I think it's still sufficient to leave those raw values inside the `custom` property. Is there another word that makes more sense? `{{ post.raw_meta }}` ?

Will `raw_meta` replace the `custom` property? If yes, should we properly deprecate the `custom` property or make this a breaking change and just delete the `custom` property? Or is `raw_meta()` a method that accesses the `custom` property? Or is `raw_meta` both a method and a property? Should single values be accessible through `post.raw_meta('field_name')` as well as `post.raw_meta.field_name`?

## Populate the object with raw meta values on initialization?

In https://github.com/timber/timber/pull/1618#issuecomment-354575795, I wrote:

> So the values are still loaded from the database and added to the `custom` property. When using `__get()`, is Timber (or WordPress) retrieving a cached value? Otherwise, should `Timber\Core::__get()` or the `meta()` method check for a custom property and return that value?

To which you answered:

> you are right about the `custom` property. As it turns out, those values are cached so we get them for "free" (at least, that was based on my earlier tests about a year ago so assuming nothing changed).

Currently, we don’t check for an already retrieved value in `Timber\Post::meta()`. We always get it from the database through `get_post_meta()`. Do we really get raw values for free? What if somebody only wants to access one meta value of a post that should be filtered in the template? Will the population of the object with raw values run for nothing?

I see two ways to go here:

1. Deprecate that the post object gets populated with raw meta values on initialization and only use `raw_meta()` to access raw properties. Let WordPress do all the performance optimization.
[...]

PR #1830, opened 2018-10-21
Deprecate get field

marciojc:
**Ticket**: #1827 

#### Issue
- Deprecate Core::get_field() (in favor of meta)
- Deprecate any other class's ::get_field() (in favor of meta)
- Ensure docs reflect meta > get_field

PR #1861, opened 2018-11-27
WIP: 2.x Deprecate get_field() in favor of meta()

gchtr:
**Ticket**: #1827

### MetaInterface

Added an interface `MetaInterface` as proposed by @pascalknecht. The goal is to have less inheritance and more «contracts». All classes that want to implement fetching meta values from the database should implement `MetaInterface`.

I made `Post`, `Term`, `User`, `Comment` and `MenuItem` implement this class. This means that any methods for handling meta values were removed from `Core` and `CoreInterface`.

Because every class now implements their own methods for `meta()` and the deprecated `get_field()` method, this has the additional benefit that in the generated reference documentation, all these methods will be properly documented 🎉.

### Integrations

Added a parameter `$args` to `timber/{object}/pre_meta` and `timber/{object}/meta` (where object can be `post`, `term`, `user`, `comment`).

I added the `$args` parameter before the `$object` parameter, because I think it might be used more often than the object parameter.

```php
/**
 * Filters the value for a post meta field.
 *
 * This filter is used by the ACF Integration.
 *
 * @see   \Timber\Post::meta()
 * @since 2.0.0
 *
 * @param string       $value      The field value.
 * @param int          $post_id    The post ID.
 * @param string       $field_name The name of the meta field to get the value for.
 * @param array        $args       An array of arguments.
 * @param \Timber\Post $post       The post object.
 */
$value = apply_filters(
    'timber/post/meta',
    $value,
    $this->ID,
    $field_name,
    $args,
    $this
);
```

### ACF Integration

I updated the ACF integration:

- Changed ACF integration to use `timber/post/pre_meta` instead of `timber/post/meta`. This will prevent running `get_post_meta()`.
- Changed ACF integration to use `timber/term/pre_meta` instead of `timber/term/meta`. This will prevent running `get_term_meta()`.

This makes it possible to use the third argument for [`get_field()`](https://www.advancedcustomfields.com/resources/get_field/) in the following way:

```twig
{{ post.meta('field_name', { format_value: true }) }}
```

### Post

Removed unnecessary warning when no value was provided for `meta()`:

```php
if ( null === $field_name ) {
    Helper::warn('You have not set what meta field you want to retrive this can cause strange behavior and is not recommended');
}
```

Removed unnecessary warning when `meta` field name was used

```php
if ( 'meta' === $field_name ) {
    Helper::warn('You’re trying to retrieve a meta field named "meta". This can cause strange behavior and is not recommended.');
}
```

We can’t use the defensive method for `__get()`, because Twig accesses an object method directly, when it’s public and it exists. That’s why we can’t add a warning here.

### Term

- Added new `timber/term/pre_meta` filter to make it behave like `$post->meta()`.

### Comment

- Renamed `timber/comment/get_meta` to `timber/comment/meta` to make it follow the same pattern as in the other classes.
- Removed shirt-circuit and applied behavior of `$post->meta()`

### Meta caching

In `Term`, there’s still a check at the beginning of `meta()`:

```php
if ( isset($this->$field_name) ) {
    return $this->$field_name;
}
```

This is a small caching mechanism. Do we need this? Depending on what we do in https://github.com/timber/timber/issues/1823, we don’t. Otherwise, we should probably add it to the other classes as well.

### DIFF
--- a/lib/Comment.php
+++ b/lib/Comment.php
@@ -485,14 +485,21 @@ class Comment extends Core implements CoreInterface, MetaInterface {
 		 *
 		 * @since 2.0.0
 		 *
-		 * @param string          $value      Passing a non-null value will short-circuit
-		 *                                    `Comment::meta()`, returning the value instead.
-		 *                                    Default null.
+		 * @param string $value               The field value. Passing a non-null value will skip
+		 *                                    fetching the value from the database. Default null.
 		 * @param int             $comment_id The comment ID.
 		 * @param string          $field_name The name of the meta field to get the value for.
+		 * @param array           $args       An array of arguments.
 		 * @param \Timber\Comment $comment    The comment object.
 		 */
-		$value = apply_filters( 'timber/comment/pre_meta', null, $this->ID, $field_name, $this );
+		$value = apply_filters(
+			'timber/comment/pre_meta',
+			null,
+			$this->ID,
+			$field_name,
+			$args,
+			$this
+		);
 
 		/**
 		 * Filters the value for a comment meta field before it is fetched from the database.
@@ -506,12 +513,9 @@ class Comment extends Core implements CoreInterface, MetaInterface {
 			'timber/comment/pre_meta'
 		);
 
-		// Short-circuit
-		if ( $value ) {
-			return $value;
+		if ( null === $value ) {
+			$value = get_comment_meta($this->ID, $field_name, true);
 		}
-
-		$value = get_comment_meta($this->ID, $field_name, true);
 
 		/**
 		 * Filters the value for a comment meta field.
@@ -523,9 +527,17 @@ class Comment extends Core implements CoreInterface, MetaInterface {
 		 * @param string          $value      The field value.
 		 * @param int             $comment_id The comment ID.
 		 * @param string          $field_name The name of the meta field to get the value for.
+		 * @param array           $args       An array of arguments.
 		 * @param \Timber\Comment $comment    The comment object.
 		 */
-		$value = apply_filters( 'timber/comment/get_meta', $value, $this->ID, $field_name, $this );
+		$value = apply_filters(
+			'timber/comment/get_meta',
+			$value,
+			$this->ID,
+			$field_name,
+			$args,
+			$this
+		);
 
 		/**
 		 * Filters the value for a comment meta field.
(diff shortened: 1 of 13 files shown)


gchtr:
@jarednova This pull request is based on #1830. As soon a it’s merged in, that pull request will be automerged as well.

lib/Comment.php
```diff
@@ -31,7 +31,7 @@
  * <p class="comment-attribution">- Sullivan Ballou</p>
  * ```
  */
-class Comment extends Core implements CoreInterface {
+class Comment extends Core implements CoreInterface, MetaInterface {
```
pascalknecht: What do we want to do with that CoreInterface in the future? It only has a very limited functionality right now.

lib/Comment.php
```diff
@@ -472,24 +472,34 @@ protected function get_meta_values( $comment_id = null ) {
 	 * @api
 	 *
 	 * @param string $field_name The field name for which you want to get the value.
+     * @param array  $args       An array of arguments for getting the meta value. Third-party
+	 *                           integrations can use this argument to make their API arguments
+	 *                           available in Timber. Default empty.
 	 * @return mixed The meta field value.
 	 */
-	public function meta( $field_name ) {
+	public function meta( $field_name, $args = array() ) {
```
pascalknecht: I don't yet understand why we need this `args` argument. What is it for?

lib/Comment.php
gchtr: I’ve thought about this, too. See this discussion: https://github.com/timber/timber/issues/478#issuecomment-442256153.

lib/Comment.php
palmiak: @pascalknecht I'll answer this :) 

First of all look at https://github.com/timber/timber/issues/1827#issuecomment-431487553 - this is a list of plugins similar to ACF - some of them use more than one argument (even ACF have two arguments). Thanks to `$args` it will be easier to create integration for other plugins that will be using `meta`.

gchtr:
I updated the description of this pull request. Before I can continue to work on this, we need to take decisions on #478 and as well as the related #1823.

jarednova:
@gchtr thanks for the bump. I just followed-up on #478 and #1823 — do you have what you need now?

gchtr:
@jarednova Yep, I have the answers I need 👍

gchtr:
I removed the logic in `Timber\Term` that assigned value of a field to a custom dynamic property.

I think this is now ready for merging (@jarednova). The remaining todos for meta values in 2.0 will be solved with a pull request for #1823.

jarednova:
Woohoo! This is a big one. Excited to tear into it!

PR #1870, opened 2018-12-04
Added documentation for post.preview

freakpants:
#### Issue
post.preview is not documented and it's usage isn't completely obvious.

#### Solution
Added documentation

#### Impact
none

### DIFF
--- a/lib/Post.php
+++ b/lib/Post.php
@@ -404,6 +404,19 @@ class Post extends Core implements CoreInterface {
 	}
 
 	/**
+	 * get a preview of your post, if you have an excerpt it will use that,
+	 * otherwise it will pull from the post_content.
+	 * If there's a <!-- more --> tag it will use that to mark where to pull through.
+	 * @example
+	 * ```twig
+	 * <p>{{post.preview.length(50).read_more("Continue Reading")}}</p>
+	 * ```
+	 * @method length(int) The number of words that WP should use to make the tease. 
+	 * @method chars(int) The number of characters that WP should use to make the tease.
+	 * @method force(bool) What happens if your custom post excerpt is longer then the length requested? By default (`$force = false`) it will use the full `post_excerpt`. However, you can set this to true to *force* your excerpt to be of the desired length
+	 * @method read_more(bool|string) The text you want to use on the 'readmore' link or no text (false)
+	 * @method strip(bool|string) true for default, false for none, string for list of custom attributes
+	 * @method end(string) The text to end the preview with (defaults to ...)
 	 * @return PostPreview
 	 */
 	public function preview() {
--- a/lib/Post.php
+++ b/lib/Post.php
@@ -404,12 +404,13 @@ class Post extends Core implements CoreInterface {
 	}
 
 	/**
-	 * get a preview of your post, if you have an excerpt it will use that,
+	 * Get a preview of your post, if you have an excerpt it will use that,
 	 * otherwise it will pull from the post_content.
+	 * This uses Timber's PostPreview object.
 	 * If there's a <!-- more --> tag it will use that to mark where to pull through.
 	 * @example
 	 * ```twig
-	 * <p>{{post.preview.length(50).read_more("Continue Reading")}}</p>
+	 * <p>{{ post.preview.length(50).read_more("Continue Reading") }}</p>
 	 * ```
 	 * @method length(int) The number of words that WP should use to make the tease. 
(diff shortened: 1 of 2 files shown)


pascalknecht:
Looks good to me. @gchtr will this work properly with the automatic documentation generation?

pascalknecht:
@gchtr This was on my initiative. I think we should still include that in the docs as a reference for 1.x as some people will still support 1.x for some time if only small changes are made. Maybe we will even need a switch in the documentation of Timber for that as we now approach a new version of Timber and people will still need 1.x documentation to not use future documentation. Any thoughts on this?

pascalknecht:
See for example https://laravel.com/docs/5.7/installation --> One can change the version at the top right

gchtr:
@pascalknecht Valid point! I’ll check what I can do for the 1.x version of the preview documentation.

> Maybe we will even need a switch in the documentation of Timber for that as we now approach a new version of Timber and people will still need 1.x documentation to not use future documentation.

I’ve already accounted for this: https://github.com/timber/docs/tree/2.x#release-a-new-version. Versioned docs are a must 😉.

acobster:
@gchtr yes, I'm working on a PR related to this and expect to finish in the next week or so. Furthermore, the `preview` method is being deprecated in favor of `excerpt` (due to "preview" being a somewhat overloaded term), which _will_ take an array of arguments.

I think if we wanted to document this in the meantime, it'd probably be best to add a link via a `@see` tag to the `PostPreview` documentation, beef up the documentation in that class directly, and perhaps provide an overarching guide for dealing with post excerpts.

jarednova:
I've made an update to this PR to clean up some style things and add some `@see` tags, etc. As others have noted here, this isn't the "final" answer — but we want to make sure the docs reflect practices for version 1.x. Before merging, I just wanted to confirm:

### Last thing to resolve:
- [ ] resolve `@method` versus another annotation — I looked through the PHPDoc reference and didn't see anything else that'd satisfy how we're using it here. Not totally surprising. My construction of this was a bit ... interesting and unconventional (but of course, seemed like a good idea at the time).

### Todos
- [ ] enable linking of `@see` tags (@gchtr is this something you're in the midst of? I can work on linking up all the manual `@see`s even if you're planning on doing an "automatic" linker)
- [ ] Allow for versioned docs
- [ ] Replace `preview` with `except` in 2.0 (and rename the `PostPreview` object as `PostExcerpt` or Excerpt` @acobster

jarednova:
I realize my note on `@method` might be incomplete. Basically, our options are to use `@method` (which PHPDoc says is a no-no) or invent our own (`@function`?). In my mind, neither is good — let alone, ideal. 

My take: While we'd be using `@method` incorrectly, at least it's an officially recognized tag. It also is true to what those actually are (methods of that object). `@function` has the double-whammy of both being unrecognized AND inaccurate. Maybe there's another term that I'm not thinking of that would be better and not run into conflict though

gchtr:
@jarednova I wouldn’t use neither `@method` nor `@function` tags, but do what Coby proposed as well:

> I think if we wanted to document this in the meantime, it'd probably be best to add a link via a `@see` tag to the `PostPreview` documentation, beef up the documentation in that class directly, and perhaps provide an overarching guide for dealing with post excerpts.

I think we should add more documentation for the methods in `PostPreview` and provide examples for all of them. Adding more descriptions and examples is probably even better for actually reading the documentation than adding some kind of list for methods.

Let me come up with a pull request that shows what I mean…

> * [ ] enable linking of `@see` tags (@gchtr is this something you're in the midst of? I can work on linking up all the manual `@see`s even if you're planning on doing an "automatic" linker)
> * [x] Allow for versioned docs

Yeah, I’d love to add better linking between docs, but it’s a little harder than I initially thought. I’ll probably try to tackle this in the next two weeks.

Versioned docs will work as of version 2. I already implemented this in the [2.x branch of the docs](https://github.com/timber/docs/tree/2.x). When we release version 2 in the docs, we can first archive the documentation for version 1, which will be available under `https://timber.github.io/docs/v1`, while the current documentation can be found under `https://timber.github.io/docs/`.

gchtr:
> Let me come up with a pull request that shows what I mean…

Here you go: https://github.com/timber/timber/pull/1882.

jarednova:
This is great @gchtr. I'm going to close this in favor of what you've got in #1882. @freakpants thanks so much for your original PR here. As you can see, it sparked a lot of good conversation to get us to a solid and future-friendly fix.

Issue #1874, opened 2018-12-06
Truncate filter does not work for character based languages


kirdia:
Truncate filter does not work for character based languages, even though excerpt filter (which uses wp_trim_words) works correctly.

Truncate and excerpt filters seems like providing identical funcionality, but I will open a PR to sync trim_words function with that of wp_trim_words

### Expected behavior
{{ '寒くなってきましたね。14日には北海道でも記録的に遅い初雪が降ったそ'|truncate( 5, true ) }}

Output:
寒くなってきましたね。14日には北海道でも記録的に遅い初雪が降ったそ… 

### Actual behavior
Output:
寒くなって… 

### What version of WordPress, PHP and Timber are you using?
WordPress 4.9.8, PHP 7.0.28, Timber 1.83

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Composer

jarednova:
Some questions were left on #1875

jarednova:
I've merged a test into `master` through #1880 that confirms that `truncate` and `wp_trim_words` are currently in-sync. 

@kirdia I'd love your feedback on whether the output is correct based on the rules for WP and East Asian languages in the newest version

Levdbas:
Since we haven't heard back from the reporter, I will close this issue.

PR #1875, opened 2018-12-06
Sync trim_words with wp_trim_words

kirdia:
Sync functionality/translation context with [wp_trim_words](https://developer.wordpress.org/reference/functions/wp_trim_words/)

**Ticket**: #1874

#### Issue
Fix trim_words function / sync with wordpress core functionality

#### Solution
trim_words function will work correctly for character based languages

#### Usage Changes
None

### DIFF
--- a/functions/timber-helper.php
+++ b/functions/timber-helper.php
@@ -175,7 +175,7 @@ class TimberHelper {
         $text = strip_tags( $text, $allowed_tag_string );
         /* translators: If your word count is based on single characters (East Asian characters),
         enter 'characters'. Otherwise, enter 'words'. Do not translate into your own language. */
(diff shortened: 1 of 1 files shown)


jarednova:
@kirdia thanks for your PR! I wrote some tests to cover the scenarios you outlined in #1874:

https://github.com/timber/timber/pull/1880

I found that the results were the same for both Timber's `truncate`  and `wp_trim_words`. Is there a setting I need to configure in the tests/WP to make sure that it knows to evaluate East Asian (i.e. `set_option("blog_charset")` type thing? I want to make sure I'm simulating the right conditions)

jarednova:
I'm going to close this PR for now so we can pick-up discussion on #1874 for what the best outcome here is. FWIW this PR is against an old version of Timber (the modified function now lives in `lib/TextHelper.php`

Issue #1825, opened 2018-10-17
Deprecate Timber::get_pagination


jarednova:
- [x] `Timber::get_pagination` includes `@deprecated` tag
- [x] `Timber::get_pagination` points to docs for proper pagination method: https://timber.github.io/docs/guides/pagination/

gchtr:
Resolved in #1877.

PR #1877, opened 2018-12-09
Deprecate Timber::get_pagination

marciojc:
Add deprecated and link tags

#1825

### DIFF
--- a/lib/Timber.php
+++ b/lib/Timber.php
@@ -730,6 +730,8 @@ class Timber {
 	 * Get pagination.
 	 *
 	 * @api
+	 * @deprecated 2.0.0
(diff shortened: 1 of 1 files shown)


jarednova:
This looks spot-on to me! But since this is a 2.x thing, I'll pass to @gchtr for his official 👍

gchtr:
Cool, thanks @marciojc! 👍 We now [ignore coverage for code marked with `@deprecated` tags](https://github.com/timber/timber/pull/1894) automatically, so there’s no need for an extra `@codeCoverageIgnore` tag. But I added an extra call to `Helper::deprecated()` in #1909.

PR #1878, opened 2018-12-18
Enable Post::convert() to convert WP_Posts, WP_Term and WP_User

luism-s:
#### Issue
Similar to  https://github.com/timber/timber/pull/1439

On a relationship, repeater and Post Object field, Timber returns an array with instances of TimberPost (only). I find it useful to also return a TimberTerm or TimberUser if get_field returns a WP_Term or WP_User.

#### Solution
Make a helper function that checks for the instance of the given class and returns the equivalent timber class. Ex. Given a WP_Post, a \Timber\Post will be returned

#### Impact
get_field() on a repeater/taxonomy/etc which returns WP_Term's while on Twig will now return a \Timber\Term, same for WP_User. No impact for WP_Post, as it works the same way as before.

I find that the second argument of convert() was no longer necessary.

#### Usage Changes
Second argument of convert() was no longer necessary, so it was removed.

#### Considerations
The helper function could be extended to support attachments.

#### Testing
A unit test was added @ test-timber-helper.php

```
function testConvertWPObject() {

	// Test WP_Post -> \Timber\Post
	$post_id = $this->factory->post->create();
	$wp_post = get_post( $post_id );
	$timber_post = \Timber\Helper::convert_wp_object($wp_post);
	$this->assertTrue($timber_post instanceof \TimberPost);

	// Test WP_Term -> \Timber\Term
	$term_id = $this->factory->term->create();
	$wp_term = get_term( $term_id );
	$timber_term = \Timber\Helper::convert_wp_object($wp_term);
	$this->assertTrue($timber_term instanceof \TimberTerm);

	// Test WP_User -> \Timber\User
	$user_id = $this->factory->user->create();
	$wp_user = get_user_by('id', $user_id);
	$timber_user = \Timber\Helper::convert_wp_object($wp_user);
	$this->assertTrue($timber_user instanceof \TimberUser);

	// Test strange input
	$random_int = 2018;
	$convert_int = \Timber\Helper::convert_wp_object($random_int);
	$this->assertTrue($convert_int === $random_int);

	$array = array();
	$convert_array = \Timber\Helper::convert_wp_object($array);
	$this->assertTrue(is_array($convert_array));
}
```

### DIFF
--- a/lib/Helper.php
+++ b/lib/Helper.php
@@ -493,11 +493,11 @@ class Helper {
 	public static function convert_wp_object( $obj ) {
 
 		if ( $obj instanceof \WP_Post ) {
-			return new Timber\Post($obj->ID);
+			return new \Timber\Post($obj->ID);
 		} elseif ( $obj instanceof \WP_Term ) {
-			return new Timber\Term($obj->term_id);
+			return new \Timber\Term($obj->term_id);
 		} elseif ( $obj instanceof \WP_User ) {
-			return new Timber\User($obj->id);
+			return new \Timber\User($obj->ID);
 		}
 
 		return $obj;
--- a/lib/Helper.php
+++ b/lib/Helper.php
@@ -480,4 +480,26 @@ class Helper {
 		Helper::warn('TimberHelper::get_current_url() is deprecated and will be removed in future versions, use Timber\URLHelper::get_current_url()');
 		return URLHelper::get_current_url();
 	}
+
+	/**
+	 * Converts a WP object (WP_Post, WP_Term) into his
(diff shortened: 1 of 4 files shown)


luism-s:
Some svg logic is screwing with my build :(

jarednova:
Thanks @luism-s! This is a handy fix that helps tame those objects. I just tweaked the tests to use the "normal" paths for those objects (`\Timber\Post` vs. `\TimberPost`)

jarednova:
Indeed! There may be uses for this on #1861 and others that have to do the `WP_Post` --> `Timber/Post`, `WP_Term`, etc. dance — thanks for getting us started!

PR #1880, opened 2018-12-20
1875/trim words

jarednova:
Tests to investigate #1874 and #1875


jarednova:
Merging this in. This tests merely confirms that Timber's `truncate` filter matches `wp_trim_words`. There's good fodder for whether the filter should be renamed to match WP's language, but I'll leave that for more discussion on #1874

Issue #1642, opened 2018-01-24
Possible XSS attacks using pagination links


tapetersen:
### Expected behavior

Pagination should be consistent with escaping it's output as most timber and wordpress functions do.
Especially as it depends on query vars for it's output which can very easily be attacker controlled.

### Actual behavior

Pagination doesn't escape links after return from add_query_var() which opens XSS attacks if not escaped explicitly in template with `{{ page.link|e('html_attr') }}`. This is further a problem as the example at https://timber.github.io/docs/guides/pagination/ does not do this

### Steps to reproduce behavior
An url (found by scanning tool) such as the one below results in an alert on the page (not in chrome since the built in reflective xss protection finds it) if the page contains for example the snippet from the guide.

This is consistent with the behavior documented for `add_query_var`
```
/*
...
 * Important: The return value of add_query_arg() is not escaped by default. Output should be
 * late-escaped with esc_url() or similar to help prevent vulnerability to cross-site scripting
 * (XSS) attacks.
...
 */
```
/arhive-page/?wx9um%2522%253e%253cscript%253ealert%25281%2529%253c%252fscript%
253eaq86s=1

### What version of WordPress, PHP and Timber are you using?
WordPress 4.9.1, PHP 7-1, Timber 1.5 -->

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Composer main repo (not the packagist plugin)

### Proposed solution
Escape link with esc_attr before putting it in as it's most commonly used that way. This may break sites that currently escapes explicitly in the template but would propablly be the best long term solution.

Minimum would be to at least update the documentation to mention that it needs to be escaped explicitly and update the code sample for it.

gchtr:
Related: https://github.com/timber/timber/issues/1557

tapetersen:
In relation to #1557 this is related as the strategy to not use autoescaping may be the right choice as most other things coming from wordpress and Timber helpers such as TimberPost and in this case pagination is pre escaped and therefore most code assumes it can be used without escaping. For example the titles coming from the items are escaped. Because of that reason I would suggest to escape the links as well especially as they are such an easy XSS target (directly url controlled as opposed to the titles and other things that need to go through content).

jarednova:
Closed by #1881

PR #1881, opened 2018-12-20
Escape pagination

jarednova:
Review @gchtr's changes on pagination


jarednova:
Lookin' good @gchtr. I ran your tests against `master` and confirmed that this resolves the issues. This should close out #1642. I'm going to update the readme, etc. so we can push to Packagist and WP.org ASAP

PR #1884, opened 2018-12-20
Gutenberg docs - small tweaks

palmiak:
First off, hello! Thanks for submitting a PR. We love/welcome PRs (especially if it's your first). Have any questions? Read [this section in CONTRIBUTING.md](https://github.com/timber/timber/blob/master/CONTRIBUTING.md#pull-requests) 

#### Issue

No issue - just changed the name of enqueued style and added `is_preview` parameter with example.

PR #1890, opened 2019-01-03
Fix spacing issues in code examples

gchtr:
**Ticket**: #1884

## Issue

There are some spacing issues in the code examples when looking at the [live documentation](https://timber.github.io/docs/guides/gutenberg/).

bildschirmfoto 2019-01-03 um 16 52 44

## Solution

This pull request fixes these spacing issues so that the guide is easier to read.

I also applied a bailout pattern for the `function_exists( 'acf_register_block' )` check. In my opinion, nested if-statements are harder to read than a bailout pattern.

```php
// Bail out if function doesn’t exist.
if ( ! function_exists( 'acf_register_block' ) ) {
    return;
}
```

### DIFF
--- a/docs/guides/gutenberg.md
+++ b/docs/guides/gutenberg.md
@@ -22,21 +22,23 @@ To create a content block, you first have to register it in **functions.php** or
 
 ```php
 add_action( 'acf/init', 'my_acf_init' );
+
 function my_acf_init() {
-    // Check function exists.
-    if( function_exists('acf_register_block') ) {
-        
-        // Register a new block.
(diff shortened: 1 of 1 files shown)


gchtr:
@jarednova Yes, I would always use spaces in Markdown, also see https://github.com/timber/timber/blob/master/CONTRIBUTING.md#code-examples:

> Timber uses tabs for indentation, but you should always use spaces for indentation in code examples, because the resulting Markdown will have a more consistent styling.

The only downside to this is that when people will copy code over from a code example, it will have spaces instead of tabs.

PR #1894, opened 2019-01-07
Added some @codeCoverageIgnores and other clean-up

jarednova:
#### Issue
Because of deprecations and a number of other small items our coverage score has fallen

#### Solution
Remove some zombie code that's not used and unable to be tested, apply `@codeCoverageIgnore` docblock tag to deprecated methods

#### Impact
Coverage % should go up (though probably still a bit beneath `master)

#### Considerations
* I'm 99% sure that `Timber\CommentThread::mecount();` was a typo and was supposed to be `Timber\CommentThread::count();` — this method is not in `master

* I'm unable to create a test for where a good chunk of `Timber\Term::get_tid` is used. I removed what seems like zombie code. 

#### Testing
Added a few new tests to cover the special instances mentioned above; otherwise ignoring deprecated methods

### DIFF
--- a/lib/Comment.php
+++ b/lib/Comment.php
@@ -351,7 +351,8 @@ class Comment extends Core implements CoreInterface, MetaInterface {
 	 * Gets a comment meta value.
 	 *
 	 * @api
-	 * @deprecated 2.0.0, use `{{ comment.meta('field_name) }}` instead
+	 * @deprecated 2.0.0, use `{{ comment.meta('field_name) }}` instead.
+	 * @codeCoverageIgnore
 	 *
 	 * @param string $field_name The field name for which you want to get the value.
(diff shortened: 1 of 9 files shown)


jarednova:
Nevermind on the "fix" from `Timber\CommentThread::mecount` ==> `Timber\CommentThread::count`. While that IS fixing a typo — it also brings up all kinds of issues that we should handle on a separate thread

jarednova:
I was Googling and found perhaps an easier solution:

https://github.com/sebastianbergmann/php-code-coverage/issues/351

Looks like there's an option in PHPUnit to recognize all `@deprecated` blocks as ignores. I'll give that a whirl to see if we can kill all birds with one quick change there

gchtr:
Oh, this is great! 😎

Issue #1866, opened 2018-12-03
comment.author.role is blank


stvnrynlds:
### Expected behavior
Per documentation on https://timber.github.io/docs/reference/timber-comment/#author, using `{{comment.author.role}}` within a comment loop should print the applicable role for a commenter.

### Actual behavior
`{{comment.author.role}}` prints nothing.

#### twig input
```twig
<h3>Comments by...</h3>
<ol>
{% for comment in post.comments %}
    <li>{{comment.author.name}}, who is a {{comment.author.role}}</li>
{% endfor %}
</ol>
```
#### html result
```html
<h3>Comments by...</h3>
<ol>
    <li>Jared Novack, who is a </li>
    <li>Katie Ricci, who is a </li>
    <li>Rebecca Pearl, who is a </li>
</ol>
```

### What version of WordPress, PHP and Timber are you using?
Version 4.9.8, PHP 7.2.0, Timber 1.8.3

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Upgraded to newest version via plugin updater in WordPress dashboard.

gchtr:
Clearly something’s missing 😅. Maybe it was never there.

@palmiak I think the relevant commits are these two:

- https://github.com/timber/timber/commit/14655bbb1e1f266184e6bb2ba6ed04906c152010
- https://github.com/timber/timber/commit/c2a0309fa7f23a615bd67723f9f1cbea567d4094

When we start thinking about how to make roles accessible in the Timber user object, I’d add it as a method. Simply accessing the roles array won’t work well, because this array contains "internal" slugs. They are not the real names of the roles. And they don’t work with internationalization. They will always be `administrator`, `editor` and `author`, and never `Administrator`, `Redakteur` or `Autor`, when we take a German example.

I’m thinking of this:

```php
/**
 * Gets the role of a user.
 *
 * @api
 * @since 1.8.4
 *
 * @return string|null
 */
public function role() {
    if ( empty( $this->roles ) ) {
        return null;
    }

    $wp_roles = wp_roles();
    $names    = $wp_roles->get_names();
    $role     = current( $this->roles );

    if ( isset( $names[ $role ] ) ) {
        return translate_user_role( $names[ $role ] );
    }

    return $role;
}
```

But then again, would somebody want to access the slug? Would a second method make sense? Maybe `role_slug()`? Or should `role()` have a parameter so that we can decide whether it returns a slug or not?

```php
$user->role(); // Editor
[...]

PR #1898, opened 2019-01-10
Adds roles() and can() method

palmiak:
**Ticket**: #1866

#### Issue
Adds to new methods to `User`:
- `roles()` - an array with role slugs and names
- `can()` - checks if user is having a capability

#### Solution
Just added the code after discussion with @gchtr in #1866  

#### Impact
Shouldn't have any. It's something new.

#### Usage Changes
Everything is described in the examples provided in inline documentation.

### DIFF
--- a/lib/Comment.php
+++ b/lib/Comment.php
@@ -86,9 +86,9 @@ class Comment extends Core implements CoreInterface {
 	 * ```html
 	 * <h3>Comments by...</h3>
 	 * <ol>
-	 * 	<li>Jared Novack, who is a contributor</li>
-	 * 	<li>Katie Ricci, who is a subscriber</li>
-	 * 	<li>Rebecca Pearl, who is a author</li>
+	 *  <li>Jared Novack, who is a contributor</li>
+	 *  <li>Katie Ricci, who is a subscriber</li>
+	 *  <li>Rebecca Pearl, who is a author</li>
 	 * </ol>
 	 * ```
 	 * @return User
@@ -130,13 +130,13 @@ class Comment extends Core implements CoreInterface {
 		}
 
 		$email = $this->avatar_email();
-		
+
 		$args = array('size' => $size, 'default' => $default);
 		$args = apply_filters('pre_get_avatar_data', $args, $email);
 		if ( isset($args['url']) ) {
 			return $args['url'];
 		}
-		
+
 		$email_hash = '';
 		if ( !empty($email) ) {
 			$email_hash = md5(strtolower(trim($email)));
@@ -206,7 +206,7 @@ class Comment extends Core implements CoreInterface {
 	 * @return boolean
 	 */
 	public function approved() {
-		return Helper::is_true($this->comment_approved);		
+		return Helper::is_true($this->comment_approved);
 	}
 
 	/**
--- a/lib/Comment.php
+++ b/lib/Comment.php
@@ -79,7 +79,7 @@ class Comment extends Core implements CoreInterface {
 	 * <h3>Comments by...</h3>
 	 * <ol>
 	 * {% for comment in post.comments %}
-	 * 	<li>{{comment.author.name}}, who is a {{comment.author.role}}</li>
+	 * 	<li>{{comment.author.name}}, who is a {{comment.author.roles|join(', ')}}</li>
 	 * {% endfor %}
 	 * </ol>
 	 * ```
--- a/lib/Comment.php
+++ b/lib/Comment.php
@@ -79,7 +79,7 @@ class Comment extends Core implements CoreInterface {
 	 * <h3>Comments by...</h3>
 	 * <ol>
 	 * {% for comment in post.comments %}
-	 * 	<li>{{comment.author.name}}, who is a {{comment.author.roles|join(', ')}}</li>
+	 *     <li>{{comment.author.name}}, who has the following roles: {{comment.author.roles|join(', ')}}</li>
 	 * {% endfor %}
 	 * </ol>
 	 * ```
--- a/lib/User.php
+++ b/lib/User.php
@@ -247,13 +247,13 @@ class User extends Core implements CoreInterface {
 	/**
 	 * Creates an associative array with user role slugs and their translated names.
 	 *
-	 * @api
+	 * @internal
 	 * @since 1.8.5
-	 * @param array $roles user roles
+	 * @param array $roles user roles.
 	 * @return array|null
 	 */
 	protected function get_roles( $roles ) {
-		if ( empty( $roles ) ) {
+		if ( empty($roles) ) {
 			return null;
 		}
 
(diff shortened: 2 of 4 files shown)


lib/User.php
```diff
@@ -229,4 +243,81 @@ public function path() {
...
+	/**
+	 * Checks whether a user has a capability.
+	 *
+	 * @api
+	 * @since 1.8.5
+	 *
+	 * @param string $capability The capability to check.
+	 * @example
+	 * ```twig
+	 * {% if post.author.can('editor') %}
```
gchtr: I wouldn’t use an example with a role here, but with a capability. The documentation for [`WP_User::has_cap()`](https://developer.wordpress.org/reference/classes/wp_user/has_cap/) says:

> While checking against a role in place of a capability is supported in part, this practice is discouraged as it may produce unreliable results.

For an example, I would use the use case provided in the issue:

```php
/**
 * @example
 * Give moderation users another CSS class to style them differently.
 *
 * ```twig
 * <span class="comment-author {{ comment.author.can('moderate_comments') ? 'comment-author--is-moderator }}">
 *     {{ comment.author.name }}
 * </span>
 * ```
 */
```

lib/User.php
```diff
@@ -229,4 +243,81 @@ public function path() {
...
+	 *   {{ role }} 
+	 * {% endfor %}
+	 * ```
+	 * ```twig
+	 * <h2>Role name</h2>
+	 * {{ post.author.roles|join(', ') }}
+	 * ```
+	 * ```twig
+	 * {% for slug in post.author.roles|keys %}
+	 *   {{ slug }}
```
gchtr:
```suggestion
	 *     {{ slug }} 
```

lib/User.php
```diff
@@ -229,4 +243,81 @@ public function path() {
...
+	/**
+	 * Gets the user roles.
+	 *
+	 * @api
+	 * @since 1.8.5
+	 * @example
+	 * ```twig
+	 * <h2>Role name</h2>
+	 * {% for role in post.author.roles %}
+	 *   {{ role }}
```
gchtr:
```suggestion
	 *     {{ role }} 
```

lib/User.php
```diff
@@ -229,4 +243,81 @@ public function path() {
...
+		if ( empty( $roles ) ) {
+			return null;
+		}
+
+		$wp_roles = wp_roles();
+		$names    = $wp_roles->get_names();
+
+		$values = array();
+
+		foreach( $roles as $role ) {
```
gchtr:
```suggestion
		foreach ( $roles as $role ) {
```

lib/User.php
```diff
@@ -229,4 +243,81 @@ public function path() {
 	public function slug() {
 		return $this->user_nicename;
 	}
+
+	/**
+	 * Create the array with user roles names and slugs.
```
gchtr:
```suggestion
	 * Creates an associative array with user role slugs and their translated names.
```

lib/User.php
```diff
@@ -229,4 +243,81 @@ public function path() {
...
+	 *
+	 * @api
+	 * @since 1.8.5
+	 *
+	 * @return array|null
+	 */
+	public function get_roles( $roles ) {
```
gchtr: Shouldn’t this method be protected? Roles should only be accessed through the `roles()` method. Instead of the `@api` tag, it would then need an `@internal` tag. And there’s also a `@param` tag missing.

lib/User.php
```diff
@@ -229,4 +243,81 @@ public function path() {
...
+	 * {% for role in post.author.roles %}
+	 *   {{ role }} 
+	 * {% endfor %}
+	 * ```
+	 * ```twig
+	 * <h2>Role name</h2>
+	 * {{ post.author.roles|join(', ') }}
+	 * ```
+	 * ```twig
+	 * {% for slug in post.author.roles|keys %}
```
gchtr:
```suggestion
	 * {% for slug, name in post.author.roles %}
```

lib/Comment.php
```diff
@@ -79,7 +79,7 @@ public function init( $cid ) {
 	 * <h3>Comments by...</h3>
 	 * <ol>
 	 * {% for comment in post.comments %}
-	 * 	<li>{{comment.author.name}}, who is a {{comment.author.role}}</li>
+	 * 	<li>{{comment.author.name}}, who is a {{comment.author.roles|join(', ')}}</li>
```
gchtr:
```suggestion
	 *     <li>{{comment.author.name}}, who has the following roles: {{comment.author.roles|join(', ')}}</li>
```

lib/User.php
```diff
@@ -229,4 +243,81 @@ public function path() {
...
+
+			$values[$role] = $name;
+		}
+		
+		return $values;
+	}
+
+	/**
+	 * Gets the user roles.
+	 *
```
gchtr:
```suggestion
	 * Roles shouldn’t be used to check whether a user has a capability. Use roles only for
	 * displaying purposes. For example, if you want to display the name of the subscription a user
	 * has on the site behind a paywall.
	 *
	 * If you want to check for capabilities, use `{{ user.can('capability') }}`. If you only want
	 * to check whether a user is logged in, you can use `{% if user %}`.
	 *
```

lib/User.php
```diff
@@ -229,4 +243,81 @@ public function path() {
...
+	 * @return array|null
+	 */
+	public function roles() {		
+		return $this->roles;
+	}
+
+	/**
+	 * Checks whether a user has a capability.
+	 *
```
gchtr:
```suggestion
	 *
	 * Don’t use role slugs for capability checks. While checking against a role in place of a
	 * capability is supported in part, this practice is discouraged as it may produce unreliable
	 * results. This includes cases where you want to check whether a user is registered. If you
	 * want to check whether a user is a Subscriber, use `{{ user.can('read') }}`. If you only want
	 * to check whether a user is logged in, you can use `{% if user %}`.
	 *
```

lib/User.php
```diff
@@ -229,4 +243,97 @@ public function path() {
...
+	 * results. This includes cases where you want to check whether a user is registered. If you
+	 * want to check whether a user is a Subscriber, use `{{ user.can('read') }}`. If you only want
+	 * to check whether a user is logged in, you can use `{% if user %}`.
+	 *
+	 * @api
+	 * @since 1.8.5
+	 *
+	 * @param string $capability The capability to check.
+	 *
+         * @example
```
gchtr: Argh, what happened to the indentation here? 😅

palmiak:
Forgot to add WiP 😀 That happens when you want contribute to relax and
everything goes wrong. I'll clean this tomorrow

czw., 24 sty 2019, 21:48: Lukas Gächter <notifications@github.com>
napisał(a):

palmiak:
@gchtr now it should be ok :)

jarednova:
More than okay @palmiak! I really love the real-world examples and guidance you provided in the documentation (ex: `Don’t use role slugs for capability checks. While checking against a role in place of a capability is supported in part, this practice is discouraged ...`) — that helps not just explain what a method does, but why you'd want to use it. 

There's a decrease coverage here, but it's a sanity check for a protected method. I just added some ignore annotations and then will merge

PR #1900, opened 2019-01-12
2.x Remove deprecated checks when fetching embedded media from post content

gchtr:
**Ticket**: https://github.com/timber/timber/issues/1823#issuecomment-453276787

## Issue

With the new approach to fetch meta values outlined in #1823, there’s no need to check for existing custom meta values.

## Solution

Remove the checks.

## Impact

Probably (hopefully) none. The `$custom` property will be protected in the future (upcoming pull request), so proper error messages will be shown when updating to 2.x.

## Usage Changes

None.

## Considerations

None.

## Testing

Removed deprecated tests.


jarednova:
Yep, goodbye to the bloat!

PR #1901, opened 2019-01-12
2.x ACF integration cleanups

gchtr:
## Issue

In preparation for the pull request of the pull request for #1823, I found that there are two methods in the ACF integration that practically do nothing. The `post_get_meta()` method additionally has a filter. They were added in https://github.com/timber/timber/commit/9f04ff62bfb4716beffc00d5731ee3b24dc256c2 and in https://github.com/timber/timber/commit/933f8ded923d150be316d7bf62b4cfd92f3fb3f8.

## Solution

Remove the methods and the filter call.

## Impact

None.

## Usage Changes

None.

## Considerations

None.

## Testing

Ran tests. All good.


jarednova:
Looks good to me! I just updated the tests to properly catch the deprecation in `Timber::get_pagination`. Speaking of deprecated. Since we deprecated `timber/term/meta/set` a while back, I took this as an opportunity to convert to a standard format with version number. Does that edit look good to you?

gchtr:
> Since we deprecated `timber/term/meta/set` a while back, I took this as an opportunity to convert to a standard format with version number. Does that edit look good to you?

Yeah, that looks fine! Because it’s `add_filter()` here and not `apply_filters()`, it won’t shop up in the documentation. But the `@deprecated` tag will surely help us to find all the functionality we can remove in a next major version 👍.

PR #1902, opened 2019-01-12
Quick Scrutinizer Fixes

jarednova:
Address a few issues flagged by Scrutinizer


jarednova:
This takes care of a few issues flagged by Scrutinizer:

- In `Menu` we were calling an instance method as a static method
- Various `@param`s weren't fully documenting the accepted data types
- A piece of unused code in `Menu` declaring an array in the midst of a method

PR #1903, opened 2019-01-13
2.x Streamline get_meta_values() for classes that implement MetaInterface

gchtr:
## Issue

- The `Term` class is still missing a `timber/term/pre_get_meta_values` filter.
- Preventing the automatic fetching of all meta values only works by passing a non-empty array. If a developer wanted to disable fetching meta values, they’d have to invent an array item, like `array( 'something' )`. That’s not very intuitive.
- Some issues in DocBlocks.

### Documentation

I improved the DocBlocks for the filters and added some more descriptions.

### get_post_meta() instead of get_post_custom()

I changed the method for fetching all meta data for posts from `get_post_custom()` to `get_post_meta()`. The difference is that `get_post_custom()` will fetch the post ID of the current post if not provided, while `get_post_meta()` requires a post ID to be present. In the context of a `Timber\Post` object, we always require a post ID, so it’s safe to use `get_post_meta()`. This is in line with the methods that the other classes use: `get_term_meta()`, `get_user_meta()` and `get_comment_meta()`.

### Update to pre-filtering logic

I streamlined the logic for pre-filtering so that when the `timber/{object_type}/pre_get_meta_values` is used, it will **skip fetching the meta values from the database** when the filter:

- returns `false`
- or returns a non-empty array

Prior to this pull request, it was only possible to skip pre-fetching via a non-empty array. So you’d have to invent an array item like `array( 'something' )` to disable pre-fetching.

Now, if developers want to completely disable fetching meta values on object initialization, they can return `false` in this filter:

```php
add_filter( 'timber/post/pre_get_meta_values', '__return_false' );
```

Or they can still return an array that already provides meta values.

## Impact

For the pre-filtering logic, until now, passing in a non-array value would trigger fetching the meta values. Now, passing an empty value that is not `false` will trigger fetching the meta values.

## Usage Changes

- See **Impact**.
- New filter `timber/term/pre_get_meta_values` becomes available.

Also see updates to Update Guide for 2.0.

## Considerations

Currently, there are four different filters:

- `timber/post/pre_get_meta_values`
- `timber/term/pre_get_meta_values`
- `timber/user/pre_get_meta_values`
- `timber/comment/pre_get_meta_values`

We could also write this with the same filter for each class, passing in the object type as an argument instead of passing it in the filter name:

```php
$meta_values = apply_filters( 'timber/pre_get_meta_values', $meta_values, 'post', $post_id, $this );
$meta_values = apply_filters( 'timber/pre_get_meta_values', $meta_values, 'term', $term_id, $this );
$meta_values = apply_filters( 'timber/pre_get_meta_values', $meta_values, 'user', $user_id, $this );
$meta_values = apply_filters( 'timber/pre_get_meta_values', $meta_values, 'comment', $comment_id, $this );
```

This way, one could disable prefetching meta values for all classes:

```php
add_filter( 'timber/pre_get_meta_values', '__return_false' );
```

But then again, this would be a slightly different approach to all the other meta filters, which wouldn’t be consistent either.

## Testing

I ran tests with coverage. All seems good, coverage shouldn’t change. But as far as I see it, not all meta filters are covered with tests. Some tests exist for the deprecated filters after fetching values from the database. I guess this should be handled in a separate issue/pull request.

### DIFF
--- a/docs/upgrade-guides/2.0.md
+++ b/docs/upgrade-guides/2.0.md
@@ -339,6 +339,12 @@ For the following filters, we updated the logic a little bit:
 
 In 1.x, you could return a non-empty array in these filters to skip fetching meta values from the database. In 2.x, you can also return `false` to achieve the same.
 
+```php
+add_filter('timber/post/pre_get_meta_values', function($custom_data, $post_id, $post){
+    return false;
+}, 10, 3);
+```
+
 ## New functions
 
 **Timber\Timber**
--- a/docs/upgrade-guides/2.0.md
+++ b/docs/upgrade-guides/2.0.md
@@ -340,9 +340,7 @@ For the following filters, we updated the logic a little bit:
 In 1.x, you could return a non-empty array in these filters to skip fetching meta values from the database. In 2.x, you can also return `false` to achieve the same.
 
 ```php
-add_filter('timber/post/pre_get_meta_values', function($custom_data, $post_id, $post){
-    return false;
-}, 10, 3);
+add_filter( 'timber/post/pre_get_meta_values', '__return_false' );
 ```
 
 ## New functions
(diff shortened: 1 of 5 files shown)


jarednova:
@gchtr great stuff! I went through and added examples for the various `@todo` examples. Can you confirm those?

In the sprit of streamlining, I also applied a uniform scheme to the variable names across these (virtually) identical code blocks. So various `$customs`, `$metas` are now all `${{object_type}}_meta`, `${{object_type}}_id`.

Back to you for merging!

gchtr:
Cool! I found some small coding standard issues in the examples as well as missing parameter count arguments for `add_filter()`. Thanks for taking on the examples! Next to all the code and writing pull request descriptions, I’m sometimes a little lazy about coming up with examples 😅.

It makes total sense to rename these variables, because they will end up in the filter documentation in the exact way they appear in the code. So, `$post_id` or `$term_id` is way easier to understand than `$pid` and `$tid` 👍.

PR #1913, opened 2019-01-27
Jan 26/scrutinizer fixes

jarednova:
#### Issue
Scrutinizer flagged a number of documentation and commenting issues in the current 2.x branch. This resolves a number of them. 

Most significantly though, I discovered that in deprecating a bunch of Twig functions in `lib/Twig.php` functionality was removed. In 2.0 Twig code like:
```twig
{{ TimberPost(maybe_post).title }}
```

should now be written as:
```twig
{{ Post(maybe_post).title }}
```

However, these methods (`TimberPost`, `TimberTerm`, etc.) had not been properly deprecated until now. Thus, I think we should send the deprecation notice (which 2.x currently does) though still maintain present 1.x functionality (for now)

#### Solution
Restore functionality for these deprecated methods and update tests accordingly

#### Usage Changes
None

#### Considerations
We should totally kill these functions in 3.x now that they've been properly deprecated

#### Testing
Tests are updated

### DIFF
--- a/lib/Helper.php
+++ b/lib/Helper.php
@@ -150,7 +150,7 @@ class Helper {
 	 * @param string $slug
 	 */
 	public static function _unlock_transient( $slug ) {
-		delete_transient($slug.'_lock', true);
+		delete_transient($slug.'_lock');
 	}
 
 	/**
@@ -434,7 +434,6 @@ class Helper {
 			return false;
 		}
 		throw new \InvalidArgumentException('$array is not an array, got:');
-		Helper::error_log($array);
 	}
 
 	/**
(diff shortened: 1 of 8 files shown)


jarednova:
This is ready for your formal review @gchtr

jarednova:
I did some research and it seems that while phpdoc says either is fine, in PHP apparently "bool" and "int" are the official names and "boolean" and "integer" are aliases: https://stackoverflow.com/questions/44009037/php-bool-vs-boolean-type-hinting

I made a [card in the 2.0 project](https://github.com/timber/timber/projects/4#card-17807743) to go through and make everything `int` and `bool` as part of the Final Pass

Issue #1899, opened 2019-01-11
strpos() Error in PHP 7.3


kchevalier-mg:
Hi Guys. First off, thanks for working on this great project. We've incorporated it into our base WordPress theme for many of our new projects, you can see that base theme here: https://github.com/mindgruve/mg-press

I've encountered an issue when testing on PHP 7.3. Running a `resize` filter on an image `src` property throws a Twig_Error_Runtime exception:

```
An exception has been thrown during the rendering of a template ("strpos(): Non-string needles will be interpreted as strings in the future. Use an explicit chr() call to preserve the current behavior").
```

The Twig markup looks like this:

```
<img src="{{ post.thumbnail.src|resize(1200, 300) }}">
```

And the stack trace points to line 457 (strpos) in file `timber-library/lib/ImageHelper.php`:

```
    protected static function is_in_theme_dir( $path ) {
        $root = realpath(get_stylesheet_directory_uri());
        if ( 0 === strpos($path, $root) ) {
            return true;
        }
    }
```

`get_stylesheet_directory_uri()` should return a URI, and I'm pretty sure that `realpath()` is intended to work with filepaths only, so $root is always `null`, and now PHP 7.3 is complaining. This looks like a bug. If I switch that to `get_stylesheet_directory()` instead it works as expected.

### Expected behavior
Resized image.

### Actual behavior
Twig_Error_Runtime exception thrown.

### Steps to reproduce behavior
Add to Twig template:

```
<img src="{{ post.thumbnail.src|resize(1200, 300) }}">
```

Upgrade to PHP 7.3.

### What version of WordPress, PHP and Timber are you using?
WordPress 5.0.3, PHP 7.3, Timber 1.8.4

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Composer

albertonieto:
Same issue here, please advice.

jarednova:
Thanks @albertonieto and @kchevalier — at this point I've only used PHP 7.1 for testing. I'll switch to PHP 7.3 to see if I can replicate issue

palmiak:
@albertonieto @kchevalier I added a PR to fix this problem

fluff:
That PR does not look right at all. `chr()` returns a single character based on an ascii code, but that's not what you want to `strpos` search into. 

[...]

PR #1915, opened 2019-01-30
Fixes strpos for php 7.3

palmiak:
**Ticket**: #1899 

#### Issue
Problem with php 7.3 and strpos - described in ticket.

#### Solution
As stated in PHP docs:
```
If needle is not a string, it is converted to an integer and applied as the ordinal value of a character. This behavior is deprecated as of PHP 7.3.0, and relying on it is highly discouraged. Depending on the intended behavior, the needle should either be explicitly cast to string, or an explicit call to chr() should be performed. 
```

so I added the `chr` function.

#### Impact

None.

#### Usage Changes
None


fluff:
Adding `chr()` does not seem right at all. Please see my comment in https://github.com/timber/timber/issues/1899#issuecomment-461635642

palmiak:
Just leaving this here to test later:
```
protected static function is_in_theme_dir( $path ) {
    $root = realpath(get_stylesheet_directory());

    if ( false === $root ) {
        return false;
    }

    if ( 0 === strpos($path, (string)$root) ) {
        return true;
    } else {
        return false;
    }
}
```

lib/ImageHelper.php
```diff
@@ -328,7 +328,7 @@ protected static function process_delete_generated_files( $filename, $ext, $dir,
 	 */
 	public static function get_server_location( $url ) {
 		// if we're already an absolute dir, just return
-		if ( 0 === strpos($url, ABSPATH) ) {
+		if ( 0 === strpos($url, chr(ABSPATH)) ) {
```
fluff: I don't think this `chr()` should be here either?

palmiak: Saw that too - pushed fix already. Thanks.

palmiak:
@jarednova I'm looking at the Travis test results and don't quite understand all the errors. I think that the problem is mostly caused by WP 4.7.9 + PHP 7.3 not the fix itself. 

Do you have any idea how to fix those bugs?

jarednova:
@palmiak according to this post on [WordPress and PHP 7.3](https://make.wordpress.org/core/2018/10/15/wordpress-and-php-7-3/), they aimed for full 7.3 support in the next release, which at that time would have been 5.0. 

Because 4.7.x gets us to +80% in WP version support I'd like to take an honest whack at this one and see what happens if we go up to 4.7.12 for the 4.7 version we test against. Let's see what that does ....

palmiak:
@jarednova 4.7.12 didn't changed anything 😞 We can assume that 4.7.* won't be compatible with 7.3. 

Can we remove just the tests for 4.7 + php 7.3 because they'll always give us error.

jarednova:
Yeah, it appears the WordPress gods have not smiled upon us. I'd like to still test against a "lower" WP version so I just upped that to 4.8.x to see if that makes Travis happy

jarednova:
Ended up going back down to WP 4.7.12 but excluding PHP 7.3 for that combo. All tests are good now!

Issue #1332, opened 2017-02-23
Feature Request: twig filter to resize & fit - no cropping


tomasvanrijsse:
I'd like to resize my images without cropping them, just making them fit.
We got used to declaring image sizes in twig but seem to lack this feature.

For example like this:

    <img src="{{ product.thumbnail|resize(1000,500,'fit') }}" />

although that might be a bit confusing to as an alternative:

    <img src="{{ product.thumbnail|fit(1000,500) }}" />

jarednova:
Hi @tomasvanrijsse thanks for this idea. Just to make sure you're aware, there's a filter called `letterbox` that might accomplish what you're looking for:

`<img src="{{ production.thumbnail | letterbox(1000, 500, '#000000') }}">`

This transforms an image like...

![colbert](https://cloud.githubusercontent.com/assets/1298086/23265402/77072c2a-f9b2-11e6-91a7-24273f3dae23.jpg)

into...

![colbert-letterbox](https://cloud.githubusercontent.com/assets/1298086/23265401/7702ba28-f9b2-11e6-8975-63f413b0e317.jpg)

With `fit`, what is the expected output for this [source image](https://cloud.githubusercontent.com/assets/1298086/23265402/77072c2a-f9b2-11e6-91a7-24273f3dae23.jpg) fit with `<img src="{{ product.thumbnail|fit(1000,500) }}" />`

tomasvanrijsse:
The expected output for images smaller then the requested limitations would be the original dimensions. That would make sense in the grammatical sense because the original image _fits_ the given dimensions.

And the website I'm currently developing has images with black backgrounds and images with white backgrounds. The result would look appropriate if only I would know what background color to apply.

danburzo:
If I understand correctly, I think Timber needs in the first place a way to _proportionally_ resize images so that they fit a certain width and height. 

There are a couple of things that are currently missing from the docs, but looking at the code it seems they might be possible already:

* The docs mention you can resize an image proportionally to a specific width by omitting the _height_ attribute in the `resize` filter. You may be able to resize proportionally to a specific height if you call `resize(0, height)` (I haven't tested yet).

* There seems to be a third parameter that controls the crop mode, with the allowed values: 'default', 'center', 'top', 'bottom', 'left', 'right', 'top-center', 'bottom-center'. In addition, you can send `false` as the crop argument (but again I haven't tested).

[...]

Issue #280, opened 2014-07-03
Image resize relative to height


alexej-d:
Would it be possible to implement something like this:
`<img src="{{item.thumbnail|resize(false, 200)}}">`
So that one can resize all images in a row to have the same height, but different widths according to their aspect ratio?
Thanks in advance!

marciojcoelho:
Have you try
```
<img src="{{item.thumbnail|resize(0, 200)}}">
```

tmyie:
`resize(0, 200)` works great.

alexej-d:
Guys, thank you, but the question is already a year old. By the the time I asked this was not possible… Nevertheless thank you ;)

Issue #1435, opened 2017-05-25
Timber Resize stretching images


njbarrett:
I'm having issues with Timber trying to resize my 100px image to 400px if I use the `|resize(400,0,'center')` filter.

I'm not sure if this is intended, because from my previous experience with Timber it did not try to "upscale" smaller images to fit the resize width and height.

Is there a way to tell Timber to not upscale smaller images?

certainlyakey:
I second that... There should be at least option to set if you want to prevent images being upscaled.

jarednova:
@njbarrett @certainlyakey I agree as well — do either of you have ideas on what that would look like in Twig? I'm hesitant to keep adding arguments to `resize` — is there another word that would make sense? `resize_but_dont_upscale` seems a bit much

certainlyakey:
Hmm, I see no option other than adding one more argument. Introducing a new separate filter would probably complicate things, given that it would still need all the same arguments as `resize` currently has.

njbarrett:
It depends whether the default functionality should be to upscale or not.. IMO it shouldn't but I guess a lot of sites could have been built with that in mind.
In a perfect world:
`|resize` - resizes but doesnt upscale
`|upscale` - resizes but does upscale

My 2 cents.

To get around this currently, I am doing an if statement to check the image size is not less than what im trying to resize to, e.g

```
                    {% if module.image.width > 800 %}
                        <img src="{{ module.image.src|resize(800, 0, 'center') }}" alt="{{ module.image.alt }}" title="{{ module.image.title }}" />
                    {% else %}
                        <img src="{{ module.image.src }}" alt="{{ module.image.alt }}" title="{{ module.image.title }}" />
                    {% endif %}
```

FlyingDR:
At my sight there should be at least a way to control resizing process. For example [\Timber\Image\Operation\Resize::get_target_sizes()](https://github.com/timber/timber/blob/master/lib/Image/Operation/Resize.php#L85) may have Wordpress filters to decide if image should be upscaled. Something like:
```php
// ... calculation of target image size into get_target_sizes()
if (($w > $src_w || $h > $src_h) && (!apply_filters('timber/image/should_upscale', $image, $w, $h))) {
  // Return image size structure with default information
}
```
It will also be useful to have a filter to let developer to control  final image size / position, like:
```php
// ... target image size / position are calculated and stored in $result
return apply_filters('timber/image/resize', $image, $result);
```
These filters will give developers much better control over images resizing process

[...]

PR #1916, opened 2019-02-07
Add possibility to disable cropping in resize function #1332

idflood:
#### Issue
https://github.com/timber/timber/issues/280
https://github.com/timber/timber/issues/1332

#### Solution
Allow passing `false` as third argument of the resize function. With this the image will be resized to fit in the specified dimension while preserving the original aspect ratio and without cropping.

#### Impact
Adds less than 5 lines of code : )

#### Usage Changes
The 3rd parameter to the resize can be set `false` to disable cropping.

    

#### Considerations
The current code had a condition `if (!$crop)` where I add the crop attributes (Line 110). Not sure if it was really used since it was not a documented option. But it was probably there for a reason I can't imagine.

### DIFF
--- a/lib/Image/Operation/Resize.php
+++ b/lib/Image/Operation/Resize.php
@@ -106,7 +106,8 @@ class Resize extends ImageOperation {
 			return array(
 				'x' => 0, 'y' => 0,
 				'src_w' => $src_w, 'src_h' => $src_h,
-				'target_w' => $w, 'target_h' => $h
+				'target_w' => $w, 'target_h' => $h,
+				'crop' => false
 			);
 		}
 		// Get ratios
@@ -188,13 +189,17 @@ class Resize extends ImageOperation {
 			}
 
 			$crop = self::get_target_sizes($image);
-			$image->crop( 	$crop['x'],
-							$crop['y'],
-							$crop['src_w'],
-							$crop['src_h'],
-							$crop['target_w'],
-							$crop['target_h']
-			);
+			if (!isset($crop['crop']) || $crop['crop'] !== false) {
+				$image->crop( 	$crop['x'],
+								$crop['y'],
+								$crop['src_w'],
+								$crop['src_h'],
+								$crop['target_w'],
+								$crop['target_h']
+				);
+			else {
+				$image->resize($crop['target_w'], $crop['target_h']);
+			}
 			$quality = apply_filters( 'wp_editor_set_quality', 82, 'image/jpeg');
 			$image->set_quality($quality);
 			$result = $image->save($save_filename);


palmiak:
@idflood if you would generate:
```
<img src="{{ product.thumbnail|resize(1000,500,false) }}" />
```
what would be the file name? Wouldn't it be a duplicate of the name:
```
<img src="{{ product.thumbnail|resize(1000,500) }}" />
```

Also maybe it would be better to change the third param to array as I mentioned https://github.com/timber/timber/issues/1435#issuecomment-450886255 . From what I remember @gchtr is also a big fan of params as arrays. 

It would open a way to create new functionality like quality, upscale (or not) etc etc.

gchtr:
My opinion on how to define how images are processed with arguments and an overview of what needs to considered when touching image functions can be found here: https://github.com/timber/timber/issues/857#issuecomment-364782702.

Recently I’ve been working on other things and this became a little stale. But I think we could tackle it now (meaning, we don’t have to wait for 2.x to come out) and add argument arrays while keeping backwards compatibility with the current parameters. We keep hitting the same end and argument arrays seem to be a way to go. I’m happy to hear about other opinions or better solutions.

@idflood We can definitely use your logic for this, just with another API.

Does somebody want to take this over or do you want to leave it to me? I think I could find some time soon.

palmiak:
@gchtr I would be interested in changing the third parameter to array + backward compatibility. It would be a great start and base for some further stuff:

If I'm thinking correctly this will need:
- checking is the third param is array, string or none
- changing the name generator function (make it more universal for all the future stuff)

I'll create an pull request to discuss the second part.

Having this we'll be able to add:
- crop / no cropp
- upscale or not
- quality per resize

gchtr:
@palmiak I’d be happy if you could take this over! I have some big updates to the documentation I’m still working on.

I think I wouldn’t change only the third parameter to an array – I would change it so that `resize()` accepts only a single argument array, including parameters for width and height. So instead of 

> * checking is the third param is array, string or none

I’d go for 

* Checking if param for resize is an arguments array.

For backwards compatibility, we can still check what’s coming in. For version 1.x, we could still keep the second parameter. In version 2.x, we’d kick that out as well.

> * changing the name generator function (make it more universal for all the future stuff)

I didn’t think about file names yet, but I think you’re on the right track there!

> I'll create a pull request to discuss the second part.

Good idea. A pull request with a `wip` label can help others to see what’s being worked on and give them a chance to chip in with comments and ideas 👍.

> Having this we'll be able to add:
> 
> * crop / no cropp
> * upscale or not
> * quality per resize

In the future, we should be able to control all necessary image manipulations:

```php
<img src="{{ img|resize({
    width: 300,
    height: 200,
    quality: 75,
    allow_upscale: false,
    fit: true,
    crop: 'something',
    letterbox: '#fff',
    format: 'webp',
    profile: 'rgb'
}) }}">
```

Levdbas:
Hi @idflood , thank you for your ticket back then. Since we are not going to actively put new features in Timber 1, I am going to close your PR. I do going to add this to a new ticket where we track several requests regarding Image manipulation so we can decide how we will handle images in upcoming Timber 2 updates.

Issue #1921, opened 2019-02-15
Deprecate remaining update() methods


jarednova:
As documented on the [2.0 Project](https://github.com/timber/timber/projects/4), we're fully deprecating `update` methods. Why? They were never fully implemented leaving them only to cover metadata and nothing else. Timber is primarily a "read" tool (cleaner themes/templates and what not) — "fixing" how WP handles the create/update/delete process is literally 3x the scope

chrisvanpatten:
Also very easy to add yourself — in a way that makes sense for each use-case — by extending Timber's models :)

jarednova:
that too!

jarednova:
Closed by #1922

PR #1922, opened 2019-02-15
1921/deprecate update methods

jarednova:
**Ticket**: #1921 

#### Issue
There were some random `update()` methods on the main objects that let you update meta data — but that was about it.

#### Solution
Deprecate 'em!

#### Impact
People will get deprecation notices

#### Considerations
Perhaps one day, a project will come about to unify all of WP's various read/update/delete methodology, but it will not be this one

#### Testing
Updated the relevant tests to just use WP native functions. Updated one test to watch for the deprecation notices

### DIFF
--- a/lib/Core.php
+++ b/lib/Core.php
@@ -86,14 +86,16 @@ abstract class Core {
 		}
 	}
 
-
 	/**
-	 * @deprecated 2.0.0 use `\update_metadata` instead.
-	 * @param string  $key
-	 * @param mixed   $value
+	 * Updates metadata for the object.
+	 *
+	 * @deprecated 2.0.0 Use `update_metadata()` instead.
+	 *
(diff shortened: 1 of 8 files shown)


jarednova:
@gchtr this is ready for review. Coverage decreased, but only because those methods are now not counted as covered because they're deprecated

jarednova:
yup, agree with all those! just had to make one tweak so those would match the tests. Mergin'!

adamtomat:
@jarednova I think it's a good move that these are getting deprecated. You said:

> Perhaps one day, a project will come about to unify all of WP's various read/update/delete methodology, but it will not be this one

Are you imagining Timber picking this up again one day? It could be something we can look at for Lumberjack if not - staring with inserting/updating/deleting posts & meta.

PR #1923, opened 2019-02-16
Update documentation.

maxxwv:
https://github.com/timber/timber/pull/1819#issuecomment-460633755

Documentation was incorrectly using post.thumbnail.sizes instead of post.thumbnail.img_sizes.


jarednova:
whoops, thanks for the catch!

PR #1928, opened 2019-02-22
User avatar resize

### LINKED ISSUE #1730: No way to change avatar size
### Expected behavior
I would like to be able to change the size of the avatar pulled by `post.author.avatar`.

### Actual behavior
The avatar property always pulls the default gravatar size of 96x96.

Use `post.author.avatar` anywhere in the single.twig

WordPress 4.9.6, PHP 7.2.5, Timber 1.7.1

WP.org

--- 
Looks like the method is just pulling the deafult avatar without allowing to set any arguments for `get_avatar_url` function:
https://github.com/timber/timber/blob/master/lib/User.php#L148

palmiak:
**Ticket**: #1730

#### Issue
Author avatar only used default size. As described in the issue "Now, I know I can do `post.author.avatar.src | resize('128x128')` but it won't stretch the image if base size is 96px."

#### Solution
I just added a possibility to pass arguments to `get_avatar_url`

#### Impact
It's fully backward compatible.

#### Usage Changes
The only change is that user can pass an array as argument - it's documented, so everything should be clear.

#### Testing
There wasn't any tests for avatar before - I'm not sure is there anything to test.

### DIFF
--- a/lib/Image.php
+++ b/lib/Image.php
@@ -214,7 +214,7 @@ class Image extends Post implements CoreInterface {
 
 	/**
 	 * @internal
-	 * @param int $iid
+	 * @param int|bool|string $iid
 	 */
 	public function init( $iid = false ) {
 		//Make sure we actually have something to work with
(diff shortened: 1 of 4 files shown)


palmiak:
Great stuff @jarednova . I didn't thought about coAuthors. Thanks.

PR #1930, opened 2019-02-22
Fix DocBlock issues with new User::avatar() method

gchtr:
**Ticket**: #1928

## Issue

Some smaller things that can be improved for the new `User::avatar()` function.

## Solution

Fix some issues in the DocBlock:

- Title cases.
- DocBlock tag order.
- Removed unnecessary quotes for hash notation in Twig (`{ size: 150 }`).
- Removed closing tag for ``, because `` is a void element and [must not have an end tag](https://www.w3.org/TR/2012/WD-html-markup-20120320/img.html#img-tags).
- Use Markdown for URL so it will actually be a link in the documentation.
- Fix return type. Because `Timber\Image::__toString()` returns the image source, directly outputting an image without `src()` (like in the example) works.

## Impact

None.

#### Usage Changes

None.

## Considerations

None.

## Testing

None.

### DIFF
--- a/lib/User.php
+++ b/lib/User.php
@@ -339,7 +339,7 @@ class User extends Core implements CoreInterface {
 	 * Get a user avatar with a width and height of 150px:
 	 *
 	 * ```twig
-	 * <img src="{{ post.author.avatar({ size: 150 }) }}"/>
+	 * <img src="{{ post.author.avatar({ size: 150 }) }}">
 	 * ```
 	 *
 	 * @param null|array $args Parameters for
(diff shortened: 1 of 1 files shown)


jarednova:
Ahhh geeze, thanks for saving our butts @gchtr. That's much cleaner + complete — merging!

palmiak:
Great work @gchtr - just one question:
`return new Image( get_avatar_url( $this->id, $args ) );`

either we have a bug in phpcs or it should:
`return new Image(get_avatar_url($this->id, $args));`

After your fix I get "Space after opening parenthesis of function call prohibited" from phpcs.

Issue #1917, opened 2019-02-11
`term.posts` not sorted by `menu_order` by default


hacknug:
I'm using `Timber::get_terms` to get non-empty terms for a given taxonomy. Each one of those terms is correctly sorted (given their `menu_order`). After looping over this, I need to print the posts using the current term. This is were things don't work _as expected_.

Somewhat related to #1366 maybe?

### Expected behavior
I'd expect `work.posts` to be returned according to their `menu_order` just like the default WP_Query would do when using one of these plugins:
- https://es.wordpress.org/plugins/intuitive-custom-post-order/
- https://es.wordpress.org/plugins/simple-custom-post-order/

### Actual behavior
Returned posts are ordered by some other parameter, I guess it must be because the class has some kind of default arguments in case they're not provided.

### Steps to reproduce behavior
Here's a simplified version of what I'm doing:

```php
$data['works'] = Timber::get_terms(
	array(
		'taxonomy'   => 'category',
		'hide_empty' => true,
	)
);
```

```twig
{% for work in works %}
    <ol>
        {% for project in work.posts %}
            <li>{{ project.title }}</li>
        {% endfor %}
    </ol>
{% endfor %}
```

### What version of WordPress, PHP and Timber are you using?
WordPress 5.0.3, PHP 7.2, Timber 1.8.4.

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Upgraded to newest version via plugin updater in WordPress dashboard.

palmiak:
There is a way to make it working:
```
public function posts( $numberposts_or_args = 10, $post_type_or_class = 'any', $post_class = '' ) {
	return $this->get_posts($numberposts_or_args, $post_type_or_class, $post_class);
}
```

so you can pass `$args` to this method:
```
{% for work in works %}
    <ol>
        {% for project in work.posts( ARGS_GOES_HERE ) %}
            <li>{{ project.title }}</li>
        {% endfor %}
    </ol>
{% endfor %}
```

gchtr:
I just had exactly the same use case this morning. But I had no problems using the [Simple Custom Post Order](https://wordpress.org/plugins/simple-custom-post-order/) plugin. When I change the menu order for the posts, they appear in the correct order when looping with `$term->posts()`.

[...]

PR #1931, opened 2019-02-22
Update tests and documentation for Term::posts()

gchtr:
**Ticket**: #1917

## Issue

The documentation for `Term::posts()` could need some love. Currently, it’s quite hard what the function accepts without having a very close look at the code.

## Solution

Add more documentation.

## Impact

Everybody happy.

## Usage Changes

None.

## Considerations

For 2.x, we could still consider to change the 1st and 2nd parameters so that they don’t accept different values (like "either post type or class"). Again, I suggest an arguments array that we also applied in #1845.

## Testing

None.


lib/Term.php
```diff
@@ -414,19 +414,48 @@ public function path() {
...
+	 *                                       arguments for the post query that this method is going.
+	 *                                       to perform. Default `10`.
+	 * @param string $post_type_or_class     Optional. Either the post type to get or the name of
+	 *                                       post class to use for the returned posts. Default
+	 *                                       `any`.
+	 * @param string $post_class             Optional. The name of the post class to use for the
+	 *                                       returned posts. Default `Timber\Post`.
 	 * @return array|bool|null
 	 */
 	public function posts( $numberposts_or_args = 10, $post_type_or_class = 'any', $post_class = '' ) {
```
palmiak: Much better variable names - but what would say about using them in `get_posts()` too?

palmiak:
It's much clearer now 👍 

As I written in the comments I'm not sure about one thing in `get_posts` - if I would like to use:
```
term.posts( { posts_per_page: 5, orderby: 'menu_order' }, 'page', `myClass` )
```
and I have class `page` user might think it should works like this:
- post_type will be set to `page`
- postClass will be set to `myClass`
but it won't. It will use:
- post_type `post`
- postClass `page`

I think there should be some extra logic:
```
if ( $PostClass == '' && class_exists($post_type) ) {
	$PostClass = $post_type;
}
```
and
```
if ( $post_type != '' ){
    $args['post_type'] = $post_type;
} elseif ( $post_type == '' && !isset($args['post_type']) ) {
    $args['post_type'] = 'any';
} 
```
or am I overthinking this?

lib/Term.php
gchtr: Good idea! I just updated variable names.

gchtr:
@palmiak You are right! The logic that you described doesn’t work. I added some more tests to cover the different use cases. I think the reason that your use case fails is that in Timber, there’s a somewhat hidden logic that automatically uses a class for a post when it has the same name as the requested post type (as described in https://github.com/timber/timber/issues/799#issuecomment-192445207).

So when there’s a class `page` and we select the post type `page`, `Timber::get_posts()` will always use the `page` class, no matter what we pass in the third parameter of `Term::posts()`. Even if we would change the logic with your suggestions. But this is probably a use case most uses will very rarely run into. To me, that behavior seems intended, though I think it’s too much magic.

It’s also something that might change with object factories, according to @pascalknecht.

palmiak:
@gchtr I fully understand. I think that for now we won't make it any better. Let's wait for @pascalknecht change and then we'll think should what to do next.

Great work as always Lukas 👍

gchtr:
Some tests failed because of some autoloading issue. I submitted #1940 to fix that.

gchtr:
@palmiak Now the tests work! I’d still need an approving review from you to be able to merge this 😉.

palmiak:
I'll try to do this tomorrow or on monday

pt., 8 mar 2019, 21:19 użytkownik Lukas Gächter <notifications@github.com>
napisał:

gchtr:
Could @jarednova or @palmiak approve this pull request so I can finally merge it? 😉

jarednova:
Thanks @gchtr and @palmiak !

PR #1932, opened 2019-02-25
Remove some Timber/User overhead

jarednova:
#### Issue
In reviewing #1928 I noticed a bunch of excess code that was mucking-up the `Timber\User` class. Stuff that should be removed for 2.x, but left in place for 1.x users (in case there was any edge-edge-edge case stuff happening).

#### Solution
Remove it!

#### Impact
There could be some backwards compatibility issues for someone (for example) who was relying on `User::name` being a property vs. methods or some weird stuff I added long long ago (substituting "name" and "display name" for example.

#### Testing
No new tests, but all should still pass


jarednova:
Indeed, thanks for that add!

PR #1934, opened 2019-02-27
Improve GitHub templates

gchtr:
- I’ve found that I always have to change the heading levels in pull requests, if I add a little more text than usual. Let’s use `<h2>` for all the headings! They are not humongous, but add more structure and readability, in my opinion.
- Recently, I saw that some issues and pull requests were opened with text that should probably be commented out, so I added some HTML comments.
- Call me pedantic 😜.


jarednova:
Yup, much more Markdown happiness — let's do it!

PR #1936, opened 2019-03-03
Update WooCommerce guide with missing step

stefenphelps:
#### Issue
This fixes the issue with the shop page not displaying any products.

#### Solution
Declare WooCommerce support in your functions.php file

#### Impact
None, just a documentation update.

#### Considerations
It may be worth considering adding some basic woocommerce support directly in the Timber Starter Theme

### DIFF
--- a/docs/guides/woocommerce.md
+++ b/docs/guides/woocommerce.md
@@ -6,7 +6,17 @@ menu:
 ---
 
 ## Point of entry - main WooCommerce PHP file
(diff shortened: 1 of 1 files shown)


jarednova:
Thanks @stefenphelps! Is there a link to any other WC getting started docs that would be helpful to add as well? 

I think we should keep things to a minimum "Hello World" example (ie. everything you need to get started) w/o trying to re-create the entire setup process. It'd be good just to point people to a good resource for getting up-and-running

stefenphelps:
@jarednova yes, WC has a docs page "[Theme Developer Handbook](https://docs.woocommerce.com/document/woocommerce-theme-developer-handbook/)" that has all the relevant info regarding theme support.

Issue #1784, opened 2018-09-10
Update Starter Theme for 2.x


jarednova:
- [ ] Ensure calls to all Timber classes are namespaced
- [ ] Ensure there are no deprecation warnings from any methods/classes called

marciojc:
Hi,
I did some improves, https://github.com/timber/starter-theme/pull/72.

jarednova:
Awesome, thanks @marciojc ! I'll check them out in the next few days

jarednova:
Thanks again @marciojc — I've now merged your changes into the starter theme. I'm going to leave this issue open as we work to get the starter theme to use the new 2.x syntax and classes

marciojc:
Thanks @jarednova. I will continue working on updating the starter theme when possible.

gchtr:
With the updates in https://github.com/timber/starter-theme/pull/142, I think we can close this.

PR #1939, opened 2019-03-05
Alias Timber::get_context as Timber::context

jarednova:
**Ticket**: #1784 

## Issue
Amongst the 2.0 API changes is `Timber::get_context` ==> `Timber::context();`. To help users prepare I wanted to update the starter theme to use `Timber::context();`. However, because this method doesn't exist yet, it triggers a fatal error

## Solution
Make an alias for `Timber::get_context()` at `Timber::context();`. This will allow me to update the starter theme. My goal is to minimize the number of deprecations a user has to deal with when/if they upgrade their theme to 2.0

## Impact
No impact on any existing code, just a bonus method out there

## Usage Changes / Considerations
We could consider actively encouraging users to start using `Timber::context();` by updating the docs and adding a deprecation notice to `get_context`. For now, I've left it at the lowest common denominator of "it's here, but no pressure man".

This aliased method is NOT what's coming in 2.0 which a more efficient and organized method. But this could help ease the transition.

## Testing
Tests pass. I didn't write a new test to cover `Timber::context()`

### DIFF
--- a/lib/Timber.php
+++ b/lib/Timber.php
@@ -226,6 +226,18 @@ class Timber {
 	================================ */
 
 	/**
+	 * Alias for Timber::get_context() which is deprecated in 2.0
+	 * 
+	 * This will allow us to update the starter theme to use the ::context method and better prepare * users for the upgrade (even if the details of what the method returns differs slightly)
+	 * @see \Timber\Timber::get_context()
+	 * @api
+	 * @return array
+	 */
+	public static function context() {
+		self::get_context();
+	}
+
+	/**
 	 * Get context.
 	 * @api
 	 * @return array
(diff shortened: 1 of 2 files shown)


jarednova:
@gchtr @palmiak I'd love your thoughts on this. It's not perfect, but I think it can help starter theme users begin using the correct 2.0 methods and reduce the amount of frustration a user might experience from a bunch of deprecation notices.

This might be less about people who literally upgrade their theme, but more about what's "learned" for any new user as of today

jarednova:
Great, and thanks for saving my 🥓 on this missing `return`. I threw in a test to make sure I don't make any other mistakes like that

PR #1942, opened 2019-03-07
Image parameters

palmiak:
This is a WiP version there are some thing missing yet:
- [x] - letterbox filter - it should be done the same way
- [x] - documentation - just fixed some variable descriptions for now, but there is work to do yet
- [x] - no tests yet

I just wanted to be sure this is the good way to. Or maybe I should consider something more.

**Ticket**: #1916 

## Issue

In #1916 we discussed with @gchtr about changing parameters in Image function from witdth, height, crop to array. With this we'll have a posibility to add other parameters in the future.

## Solution

I've changed **$w** to **$parameters_or_w** and if it is an array I'll check for existing parameters. 

If it's not an array it will work like before. Althought in some time soon we could depracate this backward compatibility part.

## Impact

It's fully backward compatible. This PR doesn't have any big impact yet.

## Usage Changes

The main change is instead of:
```
{{ Image( image_id )|resize(200,200,'center') }} 
```
we now will be using:
```
{{ Image( image_id )|resize({width: 600, height: 600, crop: 'none'}) }}
```


lib/Image/Operation/Resize.php
```diff
@@ -17,27 +17,51 @@
  */
 class Resize extends ImageOperation {
 
-   private $w, $h, $crop;
+	private $parameters_or_w, $h, $crop;
```
gchtr: Can we list properties separately, with a DocBlock for each parameter?

lib/Image/Operation/Resize.php
```diff
@@ -17,27 +17,51 @@
...
-		// Sanitize crop position
-		$allowed_crop_positions = array('default', 'center', 'top', 'bottom', 'left', 'right', 'top-center', 'bottom-center');
-		if ( $crop !== false && !in_array($crop, $allowed_crop_positions) ) {
-			$crop = $allowed_crop_positions[0];
+	public function __construct( $parameters_or_w, $h, $crop ) {
+
+		$allowed_crop_positions = array( 'default', 'center', 'top', 'bottom', 'left', 'right', 'top-center', 'bottom-center' );
+
+		if ( is_array($parameters_or_w) ) {
+			if ( isset($parameters_or_w['width']) ) {
```
gchtr: Would it maybe help to work with `wp_parse_args()` here instead of a list of if-statements?

```php
$args = wp_parse_args( $parameters_or_w, array(
    'width' => 0,
	'height' => 0,
	// …
) );

$this->w = $args['width'];
$this->h = $args['height'];
// …
```

lib/Image/Operation/Resize.php
```diff
@@ -49,7 +73,7 @@ public function filename( $src_filename, $src_extension ) {
 		if ( $this->h ) {
 			$h = $this->h;
 		}
-		$result = $src_filename.'-'.$w.'x'.$h.'-c-'.($this->crop ? $this->crop : 'f'); // Crop will be either user named or f (false)
+		$result = $src_filename . '-' . $w . 'x' . $h . '-c-' . ( $this->crop ? $this->crop : 'f' ); // Crop will be either user named or f (false)
```
gchtr: Let’s not add too many coding standard fixes. We wanted to do that in one swoop before releasing 2.x, because they might distract from the actual changes. 😉

lib/Image/Operation/Resize.php
gchtr: It would also be great to have support for using `w` and `h` instead of `width` and `height`, see https://github.com/timber/timber/issues/857#issuecomment-366797326.

gchtr (commented):
This is exactly what I meant 👍.

In addition to the `letterbox` argument, I can remember there [was a pull request](https://github.com/timber/timber/pull/1657) for a `quality` parameter which we could also implement here.

Do you need more info from me here or is this enough to keep on working on this?

palmiak:
@gchtr thanks for the input- I'll hope that I'll find some time to finish all of this during the weekend.

lib/Image/Operation/Resize.php
palmiak: they are seperate now. Probably the doctype part will need some polishing thought.

lib/Image/Operation/Resize.php
palmiak: `wp_parse_args` is used now - it was really a better solution. Thanks @gchtr . I also changed the variables to `w` and `h`.

lib/Image/Operation/Resize.php
palmiak: About the quality param - I'll skip it for now. Let's finish this one and than we can start adding new filters.

palmiak:
Changes:
- changes variables names to `h` and `v`
- added param array to letterbox
- added tests for resize

ToDo:
- tests for LetterBox
- examples in documentation

palmiak:
@gchtr @jarednova one problem with codeowners emerged - I wanted to merge master into branch from github level and it needs approval. I can do from CLI level, but github was the easier way :)

jarednova:
d'oh! yeah, looks like that's one of the drawbacks. Let me see if I can upgrade your permissions to do that ...

lib/Image/Operation/Resize.php
gchtr: Oh, I’m sorry! I read my comment again and it’s not what I wanted to say. Maybe I was too tired.

> It would also be great to have support for using `w` and `h` instead of `width` and `height`

What I meant to say was: In addition to the `width` and `height` parameters, it would also be great to have `w` and `h` shorthands.

So I would keep the variable names `width` and `height` – they are surely more readable, but add something like:

```php
$shorthands = array(
    'w' => 'width',
    'h' => 'height',
);

foreach ( $shorthands as $shorthand => $name ) {
    if ( isset( $args[ $shorthand ] ) ) {
        $this->$name = $args[ $shorthand ];
    }
}
```

What do you think? Does that make sense?

palmiak: I thought about it a bit and I would go the other way around. 

```
$shorthands = array(
	'w' => 'width',
	'h' => 'height',
	'crop' => 'c',
);

foreach ( $shorthands as $shorthand => $name ) {
	if ( isset( $args[ $shorthand ] ) ) {
		$this->$shorthand = $args[ $shorthand ];
	} elseif ( isset( $args[ $name ] ) ) {
		$this->$shorthand = $args[ $name ];
	}
}
```

Doing it this way will keep the code more backward compatible - because now we use `$this->w` so changing now everything to `$this->width` would mean it would be a good thing to change it everywhere in this helper to keep the naming convention.

The only downside is the `crop => c` instead of `c => crop`. Probably all the new params will also be added in reverse way. It bothers me bit, but I still think it's a bit clearer than having two loops for normal and reveresed params. I think that a short note near w and h should be enought.

gchtr: Oh, okay. I think I missed that the properties are already called `$w` and `$h`. I think the most important thing is to make both the full name and the shorthands work. So: 👍.

jarednova:
@palmiak I notice the WIP label is still on this. Is that correct, or should I take a look at review and merging?

palmiak:
I have plans to finish it this week. At last a short moment to do some cool
stuff.

pon., 1 lip 2019, 20:03 użytkownik Jared Novack <notifications@github.com>
napisał:

Levdbas:
Hi @palmiak , Thank you for your work back in the day on this ticket. Since we are not going to actively put new features in Timber 1, I am going to close your PR. I do going to add this to a new ticket where we track several requests regarding Image manipulation so we can decide how we will handle images in upcoming Timber 2 updates.

Issue #1824, opened 2018-10-17
Deprecate Timber::fetch() and its filter timber_compile_result


jarednova:
The current behavior and function layering doesn't really do anything but add extra functions/filters. This should be simplified for 2.x

palmiak:
I just took a look a this and have some questions:
- `timber_compile_result` won't have a replacement, as what needs to be filtered will be filtered in `compile` 
- why does render have both echo and return? I thought it's for tests but they use `ob_start();`

jarednova:
**`timber_compile_result` won't have a replacement, as what needs to be filtered will be filtered in `compile`**
We'll definitely want to deprecate that filter, if only for its "old style" format. I think it makes sense to have _some_ filter for the output of the Twig compilation in there though

**Why does render have both echo and return? I thought it's for tests but they use `ob_start();`**
I think just the messiness of the paleolithic-era Timber cod. I can't think of any good reason to leave them there, can you? This seems like the right time to pull them out

jarednova:
@palmiak is this a ticket you'd like to pick up?

palmiak:
Of course - I'll have a WordCamp Nordic contributor's day to deal with this and some others I left :)

PR #1943, opened 2019-03-07
Deprecate fetch

palmiak:
**Ticket**: #1824 

## Issue

Just remove the code no one probably used :)

## Solution

Added depracations for function and filter.

## Impact

If someone used those functions he'll get a warning. Also I created a new filter `timber/compile/render` that should replace the `timber_compile_result`.

## Usage Changes

The new filter `timber/compile/render` - it works just as `timber_compile_result`.

## Considerations

I'm not sure about the name of the new filter `timber/compile/render`. Any propositions?

## Testing

Not yet, I'll add them if everything else is ok.


lib/Timber.php
```diff
@@ -640,7 +659,6 @@ public static function fetch( $filenames, $data = array(), $expires = false, $ca
 	public static function render( $filenames, $data = array(), $expires = false, $cache_mode = Loader::CACHE_USE_DEFAULT ) {
 		$output = self::fetch($filenames, $data, $expires, $cache_mode);
```
jarednova: I think we just need to replace this with `self::compile` and that failing test will resolve

lib/Timber.php
```diff
@@ -528,9 +528,19 @@ public static function compile( $filenames, $data = array(), $expires = false, $
 				);
 			}
 
-			$output = $loader->render($file, $data, $expires, $cache_mode);
+			$output = $loader->render( $file, $data, $expires, $cache_mode );
 		}
 
+		/**
+		 * Replaces the depraceted `timber_compile_result`.
```
gchtr: The DocBlock needs to start with "Filters" so that it will be picked up for the [Hook Documentation](https://github.com/timber/timber/blob/master/CONTRIBUTING.md#documenting-hooks).

```suggestion
		 * Filters the compiled result before it is returned in `Timber::compile()`.
		 *
```

lib/Timber.php
```diff
@@ -528,9 +528,19 @@ public static function compile( $filenames, $data = array(), $expires = false, $
...
+		 * It adds the posibility to filter the output ready for render.
+		 *
+		 * @since 2.0.0
+		 *
+		 * @param string $output
+		 */
+		$output = apply_filters( 'timber/compile/output', $output );
```
gchtr: Isn’t "output" more related to "echoing" something? But the compile method would return a html instead of echoing it, right? I’d find `timber/compile/result` not that bad. But naming filters is really hard 😅. What does @jarednova as a native English speaker think?

gchtr: I think with

> @todo In case this isn’t deprecated for 2.0.0, update filter hook name

further down in the code I meant that we should use `/` instead of `_` for the hook name separation.

jarednova:
@palmiak is it cool with you if I finish this up? I think this one just needs some of those suggestions pulled in and then we're done. That way you can focus your attention on #1942

Issue #1929, opened 2019-02-22
Handle image generation from remote sources and AWS storage. sideload_image function not compatible with non-local storage systems.


petertwise:
Our site pulls in images from a JSON feed and processes them with timber. So the originating URL is a non-local URL from the parent company's CDN. We are also using WPEngine for hosting and are using their proprietary LargeFS system, which pushes images older than 10 days to an S3 bucket.

* Images URLs are produced via Timber/Twig using usually either the ImageHelper::resize or ImageHelper::letterbox functions.
* Both functions call ImageHelper _operate to generate the image
* _operate checks if the image is external, which it will always be for images from the JSON feed / API. If it is external it will run ImageHelper::sideload_image on the URL.
* sideload_image will check if the file exists using the get_sideloaded_file_loc function. But this function uses wp_upload_dir() 'path' which will always return the server path , not a URL.
* Conclusion: Any image with an original URL being pulled from the Workman API will constantly be going around in a circle where it is regenerated when the page is visited, then 10 days later moved to AWS/LargeFS and then when the page is visited again, Timber sees it is missing from the server and regenerates it on the server again.
* Additional suspicion that I have not confirmed: Because of the /year/month/file URL structure in WordPress I think a huge amount of the storage on AWS is unused duplicate files. Each circular go around where images are needlessly regenerated puts the new image in the current year/month directory and then eventually pushes it to AWS and then starts over again.

Example page: https://www.storey.com/books/the-pesto-cookbook/

Twig image code:
```
<img src="{{book.cover_image | default('/wp-content/themes/storey/static/img/placeholders/placeholder_tease.jpg') | resize(600) }}" alt="">
```

**Describe the solution you'd like**

It would be great to have some option or modification to the sideload_image function to force it to check for the existence of the existing absolute file URL instead of only checking the server path to see if an image exists before recreating it.

It would also be great to get some advice regarding the /year/month/file URL structure and how to avoid creating a lot of duplicate images in different year/month directories.

**Describe alternatives you've considered**

Modifying the custom theme to bypass Timber image functions and just use core WP image resizing.

**Additional context**

Related but not identical issues:

* https://github.com/timber/timber/issues/510
* https://github.com/timber/timber/issues/1156

squarecandy:
https://github.com/timber/timber/blob/master/lib/ImageHelper.php#L577
This always returns true for us because the images URLs being passed in are on an external CDN.

https://github.com/timber/timber/blob/master/lib/ImageHelper.php#L368
[...]

PR #1944, opened 2019-03-07
Improve notes about CDN limitations

gchtr:
**Ticket**: #1925

Let’s rewrite the CDN limitation section again based on the new information provided by @palmiak. It looks like when Maciej talked about «Push CDN», he was talking about the same as the following section by Jared:

> However, there may be some CDN setups that work fine with the above filters (for example: one that’s merely watching the contents of the `wp-content/uploads` directory and copying it to the CDN). For example, the filters seem to work fine with the default setups for WPEngine and Pantheon.

I changed the text to first hint about the differences between Push and Pull CDN and then describe how this relates to Timber.

### DIFF
--- a/docs/guides/cookbook-images.md
+++ b/docs/guides/cookbook-images.md
@@ -137,12 +137,21 @@ You can now use all the above functions to transform your custom images in the s
 
 ## Limitations when working with a CDN
 
-Timber’s image functions may be somewhat limited when using a CDN. If you use a plugin like [WP Offload Media Lite](https://wordpress.org/plugins/amazon-s3-and-cloudfront/) to store your images on a CDN, you should be aware of these limitations.
+Timber’s image functions may be somewhat limited when using a CDN.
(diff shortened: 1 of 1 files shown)


gchtr:
@jarednova Cool, yeah, additional headings surely help! I just added them.

However, from looking at #1929, there are also issues when working with WPEngine, which you mentioned should work. I’m guessing this needs yet another round of clarifications. Do we have any knowledge about CDN plugins that work flawlessly?

palmiak:
@gchtr I only used W3Total Cache with maxCDN (which is a typical pull zone). I don't know anything about those hosting related CDNs (like WPEngine) we would need to test them one by one.

jarednova:
Yeah, this info is definitely an improvement. A good @todo to come out of it would be to take a few popular WP hosts and describe their current compatibility. Even if it's a limited list, it might encourage others (or who knows, the hosts themselves?) to state where they are

gchtr:
Cool! I will start a new issue with a basic table of what we currently now.

PR #1951, opened 2019-03-15
Ref #1936 – Update WooCommerce guide

gchtr:
**Ticket**: #1936 

- New section about Theme Developer Handbook mentioned in https://github.com/timber/timber/pull/1936#issuecomment-470272909
- Some small fixes, like bold file names and the language tag for the fenced code block.

### DIFF
--- a/docs/guides/woocommerce.md
+++ b/docs/guides/woocommerce.md
@@ -6,22 +6,25 @@ menu:
 ---
 
 ## Point of entry - main WooCommerce PHP file
-The first step to get your WooCommerce project integrated with Timber is declaring WooCommerce support in your theme's `functions.php` file like so:
 
-```
-function mytheme_add_woocommerce_support() {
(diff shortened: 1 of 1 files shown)


marciojc:
@gchtr normally I declare the `add_theme_support( 'woocommerce' )` inside the timber setup function.

https://github.com/marciojc/storefront-timber/blob/master/functions.php#L150

gchtr:
> @gchtr normally I declare the `add_theme_support( 'woocommerce' )` inside the timber setup function.
> 
> [marciojc/storefront-timber:functions.php@`master`#L150](https://github.com/marciojc/storefront-timber/blob/master/functions.php#L150)

Yes, that’s totally fine. I myself prefer to use a separate class for all functionality related to WooCommerce. In the example, we mostly about using **functions.php**. It basically means: «You can put this into your theme, but you can decide yourself how you want to structure your code. Just don’t put it into a template file like single.php or archive.php.»

PR #1954, opened 2019-03-20
Add support for non-cookied comment awaiting moderation message

codeclarified:
## Issue
In WordPress 4.9.6 with the GDPR updates, WordPress no longer displays the "comment awaiting moderation" message without the user agreeing to accept cookies when posting a comment. This could leave an anonymous visitor confused about whether or not their comment was properly submitted to the site. In WordPress 5.1 an update was made to display the visitor's comment along with the moderation message immediately after a comment was submitted - https://core.trac.wordpress.org/changeset/44659 however Timber does not yet support this update.

## Solution
This PR adds some handling to the `comments()` method in `Timber\Posts` to check if `wp_get_unapproved_comment_author_email()` exists and if so, check if the comment thread should display unapproved comments made by the visitor.

## Impact
This should allow anonymous commenters on WP sites running 5.1+ to see a comment awaiting moderation message after submitting a comment, whether or not they choose to accept cookies. This will only display on the next page load as it uses a hash in the redirection URL from the comment submission. This should not affect backward compatibility for older versions of WordPress as it is checked after the current checks for unapproved comments, and is wrapped in a `function_exists()` call to prevent a call to undefined function error.

## Usage Changes
None

## Considerations
Beyond the scope of this PR, but perhaps making a filter available for the `$args` array used in the `CommentThread->init()` call?

## Testing

Have not written any and not entirely sure how to approach for this honestly.
- Anonymous user submits comment form, without selecting to accept cookies
- Comment with ID matching the ID in the redirection query string should be returned in the comment list

### DIFF
--- a/lib/Comment.php
+++ b/lib/Comment.php
@@ -427,10 +427,4 @@ class Comment extends Core implements CoreInterface {
 		return str_replace('&#038;', '&amp;', esc_url($out));
 	}
 
-	public static function get_unapproved_comment_author_email(){
-		if ( function_exists('wp_get_unapproved_comment_author_email') ) {
-			return wp_get_unapproved_comment_author_email();
-		}
-	}
-
 }
--- a/lib/Comment.php
+++ b/lib/Comment.php
@@ -427,4 +427,11 @@ class Comment extends Core implements CoreInterface {
 		return str_replace('&#038;', '&amp;', esc_url($out));
 	}
 
+	public static function get_unapproved_comment_author_email(){
+		if ( function_exists('wp_get_unapproved_comment_author_email') ) {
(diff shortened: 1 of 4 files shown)


jarednova:
Thank you @codeclarified! I didn't realize we had a GDPR change hiding out in there. 

I think you've got the right idea on what a potential unit test should cover. Are you up for giving it a shot? Here's our guide to running tests locally:

... I'm happy to help with pointers along the way

jarednova:
Thanks @codeclarified — in the end, all that's needed is a few tests to verify the behaviors. Glad we can be up-to-speed with WP 5.1!

PR #1955, opened 2019-03-20
Update documentation to use Timber::context

marciojc:
As Timber::get_context() will be deprecated, I updated the documentation to the new Timber::context()

### DIFF
--- a/docs/guides/wp-integration.md
+++ b/docs/guides/wp-integration.md
@@ -156,7 +156,7 @@ public function widget( $args, $instance ) {
 
 Well, if it works for widgets, why shouldn't it work for shortcodes? Of course it does!
 
-Let’s implement a `[youtube]` shortcode which embeds a youtube video.  
+Let’s implement a `[youtube]` shortcode which embeds a youtube video.
 For the desired usage of `[youtube id=xxxx]`, we only need a few lines of code:
 
 ```php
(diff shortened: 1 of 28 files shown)


marciojc:
@jarednova should I also change the tests?

jarednova:
@marciojc yes please! that'd be great

jarednova:
The changes on `get_context` ==> `context` all look good to me. But I also noticed changes in the `.md` files from spaces to tabs. @gchtr does this present any issues with the 2.x merge?

marciojc:
sorry @jarednova, already fix that and update the editorconfig.

docs/guides/wp-integration.md
```diff
@@ -257,15 +257,15 @@ function get_password_protected_template( $template ) {
     }
 
     return $template;
-}, 99 );
+};
```
jarednova: whoops! thanks for removing that extra `)`

jarednova:
@marciojc thanks for this edit! On the tabs vs. spaces front in the docs, we're going to do a final pass once Timber 2.0 is done. But before that, we're avoiding stylistic changes like that so we don't create a bunch of potential merge conflicts

gchtr:
> But I also noticed changes in the `.md` files from spaces to tabs. @gchtr does this present any issues with the 2.x merge?

There’s not a "real" issue with tabs for indentation in Markdown files. I just figured that when we use spaces, the indentation is more consistent and reliable. So it’s mainly readability that we’d care about here. The downside is that when people copy code over from the documentation, it will have spaces instead of tabs. But this shouldn’t be a problem if you use a proper editor or IDE ;).

PR #1960, opened 2019-04-01
ref #1813 -- suppress spelling fixed

jarednova:
@gchtr caught a typo in #1813 which this fixes


jarednova:
I'm quick merging this into master; no reason for the typo to hang-out there

Issue #1953, opened 2019-03-20
timber's twig loader loads mailpoet's twig instance and fails due to conflicting twig versions


ghost:
## Expected behavior
Load the correct twig instance

## Actual behavior
Loading MailPoet's twig instance

## Steps to reproduce behavior
With Timber and Mailpoet installed, activating a plugin using Timber fails with 

> Declaration of Asm89\Twig\CacheExtension\TokenParser\Cache::parse(Twig_Token $token) must be compatible with Twig\TokenParser\TokenParserInterface::parse(Twig\Token $token) in /var/www/project/esgo40-std/wp-content/plugins/timber-library/vendor/asm89/twig-cache-extension/lib/Asm89/Twig/CacheExtension/TokenParser/Cache.php on line 0

Probably simply due to the fact that MailPoet's twig instance is found first. 

## What version of WordPress, PHP and Timber are you using?
Wordpress 5.1.1
Timber 1.9.2 with twigs version 1.34.4
MailPoet 3 (New) 3.21.1 with twigs version 1.38.0 (as included in the Avada theme)
MyPlugin utilizing Timber

## How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Installed as plugin from WP.org

palmiak:
https://github.com/timber/timber/issues/1950 - we also had this problem.

BTW @jarednova @pascalknecht @gchtr why do we have ` "twig/twig": "1.34.*|2.*",` and not ` "twig/twig": "1.38.*|2.*",` with https://github.com/twigphp/twig-cache-extension/pull/6#issuecomment-472064244 we could adopt the newer version.

pahandel:
Thanks for the hint!
The resulting fatal error seems to look like the same and maybe there's something wrong with MailPoet's twig (although it doesn't throw any errors when used within MailPoet)
My first thought was that Timber's and MailPoet's twig instances interacted somehow resulting in that error. Probably a wrong assumption.
Anyhow, the fact remains that Timber is loading another plugin's twig instance instead of it's own if there is one - at least if it's alphabetically 'below'.

gchtr:
@palmiak I guess we could optimize the Twig dependencies with the [Caret Operator](https://getcomposer.org/doc/articles/versions.md#caret-version-range-) to always get the most current minor version:

```
"twig/twig": "^1.34|^2.0",
```

It also says so in Composer’s documentation:

> This is the recommended operator for maximum interoperability when writing library code.

[...]

Issue #1941, opened 2019-03-07
Use namespaced classes for Twig


gchtr:
According to [this announcement](https://symfony.com/blog/new-in-twig-namespaced-classes), non-namespaced Twig classes **will be deprecated with the next 2.x version of Twig**, and **will be removed** in Twig 3.x.

This means that in [`Timber\Twig`](https://github.com/timber/timber/blob/master/lib/Twig.php), we should probably use namespaced Twig classes. Instead of:

```php
namespace Timber;

class Twig {
    public function __construct() {
        add_action( 'timber/twig/filters', array( $this, 'add_timber_filters' ) );
    }

    public function add_timber_filters( $twig ) {
        $twig->addFilter( new \Twig_SimpleFilter(
            'resize',
            array( 'Timber\ImageHelper', 'resize' )
        ) );
    }
}
```

We should probably use:

```php
namespace Timber;

use Twig\TwigFilter;

class Twig {
    public function __construct() {
        add_action( 'timber/twig/filters', array( $this, 'add_timber_filters' ) );
    }

    public function add_timber_filters( $twig ) {
        $twig->addFilter( new TwigFilter(
            'resize',
            array( 'Timber\ImageHelper', 'resize' )
        ) );
    }
}
```

## Questions

- Will this work with Twig 1.x?
- What kind of backwards-compatibility should we keep? What should be the minimum version required for Twig in Timber 2.x? Should we bump it to Twig 2.x?
- How does [Twig_Function.php](https://github.com/timber/timber/blob/master/lib/Twig_Function.php) play into this? Do we need to use `Timber\Twig_Function` in `Timber\Twig` as well? Do we need the same checks for `Twig\TwigFilter` to keep backwards compatibility?

## Todo

- [ ] Update to namespaced classes in code while providing backwards compatibility.
- [ ] Update to namespaced classes in documentation and examples.

jenkoian:
Do you think Timber should mirror the versions of Twig? Similar to what Elastica does with ElasticSearch https://github.com/ruflin/Elastica#versions--dependencies if so it might help answer some of the questions. 

As someone waiting eagerly for the release of `2.x` I'd vote for whatever doesn't delay that any further :)

jarednova:
**Will this work with Twig 1.x?**
I'm guessing not. But even if it does, I think for 2.x, we shouldn't let that worry us ...

[...]

PR #1963, opened 2019-04-04
Improve Twig compatibility

gchtr:
**Ticket**: #1941, #1953 

## Issue

Twig recently changed to [namespaced classes](https://symfony.com/blog/new-in-twig-namespaced-classes) and will deprecate all other class names in the future.

We already have backwards compatibility for Twig Functions through `Timber\Twig_Function`, but not yet for Twig Filters.

## Solution

This pull request:

- Adds `Timber\Twig_Filter` as an alias that can be used for filters in the same way as `Timber\Twig_Function` can be use for functions.
- Updates the documentation to make use of these classes.
- Changes the `version_compare()` checks to checks for `class_exists()`. The `version_compare()` function relied on the existence of `Twig_Environment`, which will change to `Twig\Environment` in the future. So we can’t use it for the check, because it might not exist. By checking for the class that we want to use, we should be safe, right?
- Removes `$twig->addExtension(new Twig_Extension_StringLoader());` from the example, because I’ve seen it added quite a few times in the wild, even though it’s probably not necessary for that project.

## Impact

There should be no impact. 

## Usage Changes

`Timber\Twig_Filter` can be used instead of `Twig_Filter` or `Twig_SimpleFilter`.

## Considerations

- Instead of using `Twig_Extension_StringLoader` as an example for an extension, we should probably add an example for an extension that might actually be useful.
- In Timber 2.x, we will probably remove all the backwards compatibility and change all the examples to use the new namespaced classes (separate pull request ahead).

## Testing

I updated tests where necessary.


jarednova:
Thank you @gchtr! Since the new class uses all the same patterns as `Timber\Twig_Function` it's pretty easy to see the symmetry. The clarified docs are also helpful — merging!

PR #1964, opened 2019-04-04
Remove deprecated files that were used to generate documentation

gchtr:
## Issue

Just some housekeeping with files that were used to generate documentation.

## Solution

Delete them.

## Impact

None.

## Usage Changes

None.

## Considerations

None?

## Testing

None.

PR #1965, opened 2019-04-04
2.x Update Twig to namespaced classes, drop backwards compatibility

gchtr:
**Ticket**: #1941, #1964 

## Issue

Twig compatibility proved to be quite a hassle with all the different versions of Twig and PHP. And as of 2.4, Twig comes with namespaced classes, that will become the standard in newer versions.

## Solution

For 2.x, we can make it easier by only supporting Twig >= 2.4.

This pull request:

- Updates Twig’s Composer dependency
    - drops support for Twig 1.x.
    - Changes to [twigphp/cache-extension](https://github.com/twigphp/twig-cache-extension), which unfortunately doesn’t have an official release with a semantic version yet.
- Updates all Twig classes to their namespaced versions.
- Drops backwards compatibility through `Timber\Twig_Filter` and `Timber\Twig_Function`.
- Updates documentation and the Upgrade Guide.

## Impact

- Everybody will have to update their Twig classes.

## Usage Changes

See Impact.

## Considerations

Should we keep `Timber\Twig_Function` and `Timber\Twig_Filter` so that devs can just keep using what they have always been using?

## Testing

We probably need to update Travis and remove all the included specials for different versions of Twig. @jarednova, can you do that?


pascalknecht:
@gchtr Looks good to me 👍

gchtr:
I updated some classes already found in #1977.

jarednova:
@gchtr I removed a bunch of extra Travis stuff. Everything's passing — is that what you meant by "included specials for different versions of Twig"?

gchtr:
@jarednova I guess I mean’t that when we have `--prefer-lowest --prefer-stable` in Travis, it would have selected Twig version 1.x over the 2.x version. But that’s actually obsolete with #2006, so I think you can ignore my request now.

I’m giving this free for merging.

gchtr:
@jarednova Damn these merge conflicts! 🐱🐭

jarednova:
NOW I think we got it ...

PR #1969, opened 2019-04-11
feat: Term::from handling multiple terms

Grafikart:
This change will help with converting WP_Term[] into Timber\Term[]

## Issue
Term api is a bit too limited and it's not possible (to my knowledge) to get the result from get_categories() into a list of Timber Term

## Solution
The from method could handle arrays
```php
Term::from(get_categories(['parent' => 0]))
``` 

## Impact
No impact or breaking changes


jarednova:
Thanks @Grafikart — I made some spacing tweaks and ensured that arrays can take the second `$taxonomy` argument. Merging!

Grafikart:
Thanks ;)

Issue #1974, opened 2019-04-22
Issue with Menu in WPML


jarednova:
Recently we had issue in our support forum related to your plugin and WPML compatibility, here: https://wpml.org/forums/topic/language-switch-as-last-item-in-menu-inherits-custom-class-names/

Can you please check it out, I am providing fix as well.:

In this file /wp-content/plugins/timber-library/lib/Menu.php
After line 253 change this:

```php
if ( isset($item->ID) ) {
                if ( is_object($item) && get_class($item) == 'WP_Post' ) {
                    $old_menu_item = $item;
                    $item = new $this->PostClass($item);
                }
                $menu_item = $this->create_menu_item($item);
                if ( isset($old_menu_item) ) {
                    $menu_item->import_classes($old_menu_item);
                }
                $index[$item->ID] = $menu_item;
            }

```
Into this:

 ```php
      if ( isset($item->ID) ) {
                if ( is_object($item) && get_class($item) == 'WP_Post' ) {
                    $old_menu_item = $item;
                    $item = new $this->PostClass($item);
                }
                if ( get_class($item) == 'WPML_LS_Menu_Item' ) {
                    $old_menu_item = $item;
                }
                $menu_item = $this->create_menu_item($item);
                if ( isset($old_menu_item) ) {
                    $menu_item->import_classes($old_menu_item);
                }
                $index[$item->ID] = $menu_item;
            }
   
```

mgarabedian:
I have  similar issue, where _order_children_ duplicates classes if the menu item is not of type 'WP_Post'.  I do not understand the purpose of importing the last menu's classes.

PR #1975, opened 2019-04-22
ref #1974 -- apply fix suggested by WPML team

jarednova:
**Ticket**: #1974 

## Issue
According to [this thread on WPML](https://wpml.org/forums/topic/language-switch-as-last-item-in-menu-inherits-custom-class-names/) there's an issue where if WPML's language switch is the last item in a menu, it "steals" the custom class name from the previous menu item.

## Solution
The WPML team sent a fix that tests for items that match WPML's class and over-writes previous assignments in the Menu ordering function.

## Impact
Because there's a test against WPML classes, I expect no fallout in non-WPML installs.

## Usage Changes
None

## Considerations
I'm afraid of "hiding" another plugin's functionality inside Timber code — but I also see this as a bit of a devil to try and make a filter/hook for. 

## Testing
No new tests. I'm relying on the WPML team here. I don't believe they make their library available for free, so we can't include it against PHPUnit (without some major acrobatics). We could fake it with a mock `WPML_LS_Menu_Item` class, but that rabbit hole could go quite deep


jarednova:
I've reached out to the WPML team for their help reviewing this patch before we do our own team review

jarednova:
closed to focus on #2009

PR #1972, opened 2019-04-21
Twig deprecated features.

MouadDB:
Fixing 'Deprecated' notices like "Using the "Twig_Token" class is deprecated since Twig version 2.7, use "Twig\Token" instead."
https://twig.symfony.com/doc/2.x/deprecated.html

PR #1977, opened 2019-04-24
Ref #1972 – Twig deprecated features

gchtr:
**Ticket**: #1972

## Issue

We missed some deprecated Twig class names in #1963.

## Solution

Use the new namespaced Twig class names. We should be save to do this when we require Twig versions that already support the new class names Timber’s composer.json. For this, I upgraded the required Twig version from `^2.0` to `^2.4` (Namespaces were introduced in 2.4: https://github.com/twigphp/Twig/blob/328d2aeba0b92002570b769706f500089ef6e1db/CHANGELOG#L161).

## Impact

There should be no impact.

## Usage Changes

None.

## Considerations

None.

## Testing

Ran tests locally.

### DIFF
--- a/lib/Image.php
+++ b/lib/Image.php
@@ -458,7 +458,7 @@ class Image extends Post implements CoreInterface {
 	 * @example
 	 * ```twig
 	 * <h1>{{ post.title }}</h1>
-	 * <img src="{{ post.thumbnail.src }}" srcset="{{ post.thumnbail.srcset }}" />
+	 * <img src="{{ post.thumbnail.src }}" srcset="{{ post.thumbnail.srcset }}" />
 	 * ```
 	 * ```html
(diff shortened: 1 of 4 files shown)


lib/Image.php
```diff
@@ -458,7 +458,7 @@ public function src( $size = 'full' ) {
 	 * @example
 	 * ```twig
 	 * <h1>{{ post.title }}</h1>
-	 * <img src="{{ post.thumbnail.src }}" srcset="{{ post.thumbnail.srcset }}" />
+	 * <img src="{{ post.thumbnail.src }}" srcset="{{ post.thumnbail.srcset }}" />
```
chrisvanpatten: typo!

gchtr: Thanks, Chris!

PR #1983, opened 2019-04-30
Fix php unit setup

### LINKED ISSUE #1982: Make PHPunit setup more robust outside of VVV

Running the supplied `bin/install-wp-tests.sh` within VVV works, but this takes probably the 'happiest' path through the script.

It looks like this started out as a copy of the script from [WP CLI Scaffold Command](https://github.com/wp-cli/scaffold-command/blob/master/templates/install-wp-tests.sh) which has since also received some updates that fix the `sed` problems on Mac OS.

Features which would be great
- use the system temp directory
- ensure the paths are all correct and do not feature trailing slashes, even if set via environment variables
- fix sed command syntax for `darwin` machines
- update the download logic to deal with betas and RC's like in the above linked script

PR will be submitted to resolve all issues

Can use in VVV, this is for people running things like laravel valet etc, overall it makes it more robust to run anywhere.

aj-adl:
**Ticket**: 

Resolves #1982 

## Issue

- sed commands fail on Mac OSX
- It's easy to set or end up with dud paths featuring double slashes
- tests and wordpress install were hardcoded to `/tmp` when WP CLI has moved toe `$TMPDIR`

## Solution
- Removed whitespace in `sed -i` command for `darwin` machines
- added in download logic from [Original Script / Inspiration](https://github.com/wp-cli/scaffold-command/blob/master/templates/install-wp-tests.sh)
- added -L flag to curl requests to help with issues opening `.tar.gz` files downloaded via curl
- switch to `$TMPDIR` for all default paths

## Impact

My understanding is it will have no impact for those running VVV, but would make setting up PHP unit tests on a Mac significantly easier (Laravel Valet users or similar)

## Usage Changes

No Usage changes for VVV users.

Mac OS users can now get phpunit running easily, just change the DB details to suit.

```bash
git clone https://github.com/timber/timber.git
cd timber
composer install
bash bin/install-wp-tests.sh db_name root "" localhost latest
./vendor/bin/phpunit
```

## Considerations

I think some of the docs around this could use a slight update?

## Testing
No formal tests, as it's regarding setup up testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,12 +8,15 @@ php:
     - 7.3
 
 env:
-    - WP_VERSION=4.7.12 WP_MULTISITE=0 PREFER_LOWEST="--prefer-lowest --prefer-stable"
-    - WP_VERSION=4.7.12 WP_MULTISITE=0 PREFER_LOWEST=""
-    - WP_VERSION=4.7.12 WP_MULTISITE=1 PREFER_LOWEST=""
-    - WP_VERSION=latest WP_MULTISITE=0 PREFER_LOWEST=""
-    - WP_VERSION=latest WP_MULTISITE=1 PREFER_LOWEST=""
-    - WP_VERSION=latest WP_MULTISITE=1 PREFER_LOWEST="--prefer-lowest --prefer-stable"
+    global:
+        - TMPDIR=/tmp
+    matrix:
+        - WP_VERSION=4.7.12 WP_MULTISITE=0 PREFER_LOWEST="--prefer-lowest --prefer-stable"
+        - WP_VERSION=4.7.12 WP_MULTISITE=0 PREFER_LOWEST=""
+        - WP_VERSION=4.7.12 WP_MULTISITE=1 PREFER_LOWEST=""
+        - WP_VERSION=latest WP_MULTISITE=0 PREFER_LOWEST=""
+        - WP_VERSION=latest WP_MULTISITE=1 PREFER_LOWEST=""
+        - WP_VERSION=latest WP_MULTISITE=1 PREFER_LOWEST="--prefer-lowest --prefer-stable"
 
 matrix:
     exclude:
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
   },
   "require": {
     "php": ">=5.3.0|7.*",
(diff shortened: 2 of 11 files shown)


gchtr:
Hey @aj-adl 

I mostly use PHPUnit without VVV and I’m eager whether this pull request fixes some of the error messages that I saw and always ignored.

I still have have an issue though when I run `phpunit`:

```
PHP Warning:  require_once(/tmp/wordpress-tests-lib/includes/functions.php): failed to open stream: No such file or directory in /web/wordpress/timber/tests/bootstrap.php on line 6
```

Of course this makes total sense, when we look at how **tests/bootstrap.php** starts:

```php
<?php

$_tests_dir = getenv('WP_TESTS_DIR');
if ( !$_tests_dir ) $_tests_dir = '/tmp/wordpress-tests-lib';

require_once $_tests_dir . '/includes/functions.php';
```

If `WP_TESTS_DIR` is not set, the `/tmp/wordpress_tests-lib` folder will be used, which doesn’t exist now, because the systems temporary folder is used. If I use `export WP_TESTS_DIR=/var/folders/6r/y1mbslqx5fl9sh1f3ldyp7_h0000gn/T/wordpress-tests-lib`, everything runs fine.

The problem is that we don’t set `WP_TESTS_DIR` anywhere when installing the tests. What do you think would be the best solution to set this so that it’s persistent across different terminal sessions?

Would it maybe be the easiest solution to update the start of **tests/bootstrap.php** like this?

```php
<?php

$_tests_dir = getenv( 'WP_TESTS_DIR' );

if ( ! $_tests_dir ) {
	$_tests_dir = rtrim( getenv( 'TMPDIR' ), '/' ) . '/wordpress-tests-lib';
}

require_once $_tests_dir . '/includes/functions.php';
```

aj-adl:
Hey @gchtr great catch there, I think i may have set that variable manually when going through all of this. 

Your proposed fix works well, keen to see this merged!

jarednova:
Thanks so much for all your work on this @aj-adl. I just pulled in @gchtr's edits in #1995. I would have that would autoclose this issue since it has the same commits, but I guess not. Right now this is 100% caught-up with `master` so I'm closing — thanks again for your first contrib!

PR #1984, opened 2019-04-30
Template trim fix

### LINKED ISSUE #1705: space -> WSOD
Hi

archive-custom_type.php (my template file in theme):

```
<?php
Timber::render( [ 'archive-custom_type.twig ' ], $context );
```
Give ^^^ white screen error. 

The error reason: see space after the twig template filename.

-------------
```
<?php
Timber::render( [ 'archive-custom_type.twig' ], $context );
```

Works!

--------------

I spend a lot of time to find the error. There was nothing in error log.

I suggest using trim() function in render function to remove space around template's filenames.

Thanks
Adam

aj-adl:
**Ticket**: Resolves #1705 

## Issue
A single whitespace before or after a template name (with the string) can cause the whole site to WSOD. While linters etc are great, if we can be a little more forgiving why not?

## Solution

Use trim within the template loader. 

## Impact
There should be no negative impacts, as filenames that begin or end in whitespace would generally be invalid, and everyone would be writing them without the whitespace, as it would cause a WSOD.

## Usage Changes
Timber's `render` and `compile` functions will now trim whitespace around template names before attempting to locate them.

## Considerations

Is it worth showing a warning when this happens? We could just silently allow it, but a warning could also be useful, perhaps via one of the helper functions.

If the dev is using a variable for the template name, they might like to know it's "wrong" as it could be causing other issues.

```php

$template = 'single.twig ';

if ( $template === 'single.twig') {
    // these things wont be done but the dev will wonder why as single.twig WILL render
}

// Will work if this PR gets merged
Timber::render( $template, $context );
````

## Testing

Test included in `tests/test-timber-loader.php`

### DIFF
--- a/lib/Loader.php
+++ b/lib/Loader.php
@@ -103,6 +103,9 @@ class Loader {
 
 		// Run through template array
 		foreach ( $templates as $template ) {
+
+			// Remove any whitespace around the template name
+			$template = trim( $template );
 			// Use the Twig loader to test for existance
 			if ( $loader->exists($template) ) {
 				// Return name of existing template
--- a/tests/test-timber-loader.php
+++ b/tests/test-timber-loader.php
@@ -20,6 +20,11 @@
 			$str = Timber::compile( array('assets/barf.twig', 'assets/lonestar.twig') );
 			$this->assertFalse($str);
 		}
+		
+		function testWhitespaceTrimForTemplate(){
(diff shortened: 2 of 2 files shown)


jarednova:
@aj-adl — this is definitely better than the old WSOD. Really appreciate your PR and the test with it!

PR #1986, opened 2019-05-04
Introduce static analysis + fixes

szepeviktor:
Hello!

This is PHPStan for Timber and some fixes.

Try running `vendor/bin/phpstan analyze -l 1` (Level 1) if you want to discover more errors.
Level 2 is minimum suggested, but robust code can be reached on Level 7.

Current Level 1 errors:

```
 ------ --------------------------------------------------------------------
  Line   Comment.php
 ------ --------------------------------------------------------------------
  209    Access to an undefined property Timber\Comment::$comment_approved.
  275    Access to an undefined property Timber\Comment::$comment_parent.
 ------ --------------------------------------------------------------------

 ------ ------------------------------------------------------------
  Line   Core.php
 ------ ------------------------------------------------------------
  126    Call to an undefined method Timber\Core::get_meta_field().
 ------ ------------------------------------------------------------

 ------ --------------------------------------------------------------------------------
  Line   Image.php
 ------ --------------------------------------------------------------------------------
  99     Timber\Image::__construct() does not call parent constructor from Timber\Post.
 ------ --------------------------------------------------------------------------------

 ------ -------------------------------------------------------------------
  Line   ImageHelper.php
 ------ -------------------------------------------------------------------
  152    Variable $file_path in isset() always exists and is not nullable.
 ------ -------------------------------------------------------------------

 ------ ---------------------------------------------------------------------------------
  Line   Integrations/CoAuthorsPlusUser.php
 ------ ---------------------------------------------------------------------------------
  40     Access to an undefined property Timber\Integrations\CoAuthorsPlusUser::$avatar.
 ------ ---------------------------------------------------------------------------------

 ------ --------------------------------------------------------------------
  Line   MenuItem.php
 ------ --------------------------------------------------------------------
  79     Access to an undefined property Timber\MenuItem::$name.
  301    Access to an undefined property Timber\MenuItem::$_menu_item_type.
 ------ --------------------------------------------------------------------

 ------ ---------------------------------------------------------
  Line   Site.php
 ------ ---------------------------------------------------------
  180    Access to an undefined property Timber\Site::$pingback.
 ------ ---------------------------------------------------------

 ------ -----------------------------------------------------
  Line   User.php
 ------ -----------------------------------------------------
  154    Access to an undefined property Timber\User::$name.
 ------ -----------------------------------------------------
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,8 @@
     "wp-coding-standards/wpcs": "^2.0",
     "wpackagist-plugin/advanced-custom-fields": "5.*",
     "wpackagist-plugin/co-authors-plus": "3.2.*",
-    "dealerdirect/phpcodesniffer-composer-installer": "^0.5.0"
+    "dealerdirect/phpcodesniffer-composer-installer": "^0.5.0",
+    "szepeviktor/phpstan-wordpress": "^0.1.0"
   },
   "suggest": {
     "satooshi/php-coveralls": "1.0.* for code coverage"
--- a/lib/Admin.php
+++ b/lib/Admin.php
@@ -94,11 +94,11 @@ class Admin {
 	 *
 	 *	@codeCoverageIgnore
 	 *
-	 *  @type	function
-	 *  @date	4/22/16
+	 *  @type function
+	 *  @date 4/22/16
 	 *
-	 *  @param	{array}		$plugin_data
-	 *  @param	{object}	$r
+	 *  @param array $plugin_data
+	 *  @param object $r
 	 */
 	public static function in_plugin_update_message( $plugin_data, $r ) {
 		$current_version = $plugin_data['Version'];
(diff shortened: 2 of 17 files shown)


szepeviktor:
> szepeviktor/phpstan-wordpress 0.1.1 requires php ~7.1 -> your PHP version (5.6.30) does not satisfy that requirement.

Static analysis does not run on PHP 5. Maybe moving from composer.json to `composer require` command in the CI would solve it.

jarednova:
@szepeviktor thanks for the PR and running this tool. It caught some stuff we should def. fix. Here's what I think we should do here:

1) Resubmit this as a PR against 2.x. The package requires PHP 7 anyway, so you'd avoid the PHP 5 issues we see here. I think we still will want to do some thinking about adding another tool in the build process (which could perhaps slow down test setup and running), but that will be a better place to consider.
2) Submit a new PR against `master` with just the fixes themselves. We'll still need to be mindful of potential merge conflicts with 2.x (we're avoiding big codebase sweeps for spacing, etc.) — but there are a lot of good catches here that would be nice to be caught up on

gchtr:
I agree with @jarednova. I’m interested in having a tool like PHPStan available in Timber, even if it’s only available as a Composer script.

@szepeviktor Would you mind adding a separate pull request with the all the necessary files to run PHPStan, but without any fixes applied? And could you do that against the `2.x` branch?

jarednova:
Closing in favor of #2149

Issue #1987, opened 2019-05-06
Changes in 1.9.4 break WooCommerce My Account Page


aj-adl:
Changes in #1813 break the WooCommerce My Account Orders Page by setting defaults on the Query that WooCommerce is not expecting - specifically `'no_found_rows' => true`.

When `no_found_rows` is `true` the query object's `found_posts` will ALWAYS be 0 unless we hook in and alter it.

When using the `[woocommerce_my_account]` shortcode to display the WooCommerce account section they check `found_posts`

See [this template file](https://github.com/woocommerce/woocommerce/blob/master/templates/myaccount/orders.php#L23-L27) and [this data store function](https://github.com/woocommerce/woocommerce/blob/master/includes/data-stores/class-wc-order-data-store-cpt.php#L772-L780) for the relevant lines of WooCommerce code.

## Expected behavior
I would expect that the order section displays.

## Actual behavior
The orders section displayed no orders, even if there are customer orders in the DB.

## Steps to reproduce behavior
 Install the latest timber + WooCommerce.

Visit the account section, usually `http://mysite.com/my-account/`

Visit the orders tab, no orders will be displayed..

Overall, I feel like setting defaults on EVERY WordPress query, as this PR has done, is very dangerous. Anecdotally, we've had to downgrade Timber on other sites recently due to misc issues and weird behaviour, I'd suspect that these issues too were cause by these changes.

Fine to set those defaults for all Timber Queries, but the current method sets it for ALL queries. 

## What version of WordPress, PHP and Timber are you using?
WordPress 5.1.x, Timber 1.9.4, WooCommerce 3.6.2+

## How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Composer

gchtr:
> Overall, I feel like setting defaults on EVERY WordPress query, as this PR has done, is very dangerous. Anecdotally, we've had to downgrade Timber on other sites recently due to misc issues and weird behaviour, I'd suspect that these issues too were cause by these changes.

I agree. Timber sets it for queries that come **after you use `Timber::get_posts()`**. It shouldn’t be the case when you use `new Timber\PostQuery()`. For now, can you try to use `new Timber\PostQuery()` for the page that displays your WooCommerce account page?

Adding a filter that affects global queries can have very strange effects in different environments. So it can very well be that your other issues are caused by this as well.

@jarednova What do you think, should we reverse this breaking change? There have been other issues, like #1961. Or can we find another solution?

Related: #1957

[...]

PR #1989, opened 2019-05-06
Filter for not forcing the WordPress way of get_posts.

palmiak:
**Ticket**: #1987

## Issue
Adds an additional filter that will give devs a chance to use get_posts in classic Timber way or in WP way.

## Solution
Adds an extra if + filter that can be set to true or false.

## Impact
It should solve issues like #1987 and on the hand will provide with quite easy way to solve #1812 

## Usage Changes
```
add_filter( 'timber/post_getter/wp_get_posts', '__return_true' );
add_filter( 'timber/post_getter/wp_get_posts', '__return_false' );
```

## Considerations
This is a temporary fix until we'll figure out what to do with `Timber::get_posts()`.

Also - I'm not sure about the filter name. Making those name is hard :(

## Testing
I added add_filter in tests so the test would run in WP way.

### DIFF
--- a/lib/PostGetter.php
+++ b/lib/PostGetter.php
@@ -53,14 +53,27 @@ class PostGetter {
 
 	public static function get_posts( $query = false, $PostClass = '\Timber\Post', $return_collection = false ) {
 		/**
-		 * Checks if we should use **get_posts()** in a Timber or WordPress way.
+		 * Filters whether Timber::get_posts() should mirror WordPress’s get_posts() function.
 		 *
-		 * @since x.x.x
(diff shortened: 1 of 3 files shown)


palmiak:
I like the mirror name 😀 You should add to your CV "senior filter and
variable name creator"

wt., 7 maj 2019, 08:25 użytkownik Lukas Gächter <notifications@github.com>
napisał:

palmiak:
I see that there are travis-errors but I think they are related with #1990

gchtr:
@palmiak Oh yes! I opened #1994 to fix this.

mibu31:
This is a great solution, please resolve/merge 👍

PR #1995, opened 2019-05-14
Ref #1983 – Fix bug when WP_TESTS_DIR is not defined

gchtr:
**Ticket**: #1983

## Issue

As discussed with @aj-adl in https://github.com/timber/timber/pull/1983#issuecomment-490613098.

## Solution

This pull request adds a fallback for when the `WP_TESTS_DIR` is not defined, which considers the `TMPDIR` environment variable.

As soon as this pull request is merged, #1983 should be auto-merged as well.

## Impact

More failsafe testing setups.

## Usage Changes

None.

## Considerations

None.

## Testing

Manually tested in my local non-VVV testing setup.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,12 +8,15 @@ php:
     - 7.3
 
 env:
-    - WP_VERSION=4.7.12 WP_MULTISITE=0 PREFER_LOWEST="--prefer-lowest --prefer-stable"
-    - WP_VERSION=4.7.12 WP_MULTISITE=0 PREFER_LOWEST=""
-    - WP_VERSION=4.7.12 WP_MULTISITE=1 PREFER_LOWEST=""
-    - WP_VERSION=latest WP_MULTISITE=0 PREFER_LOWEST=""
-    - WP_VERSION=latest WP_MULTISITE=1 PREFER_LOWEST=""
-    - WP_VERSION=latest WP_MULTISITE=1 PREFER_LOWEST="--prefer-lowest --prefer-stable"
+    global:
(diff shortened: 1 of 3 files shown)


gchtr:
Looks like Travis is not happy at all about this change 😅. I’ll have a look.

jarednova:
It looks like there's something here that's breaking Travis (#1983 works). I'm getting this error in the Travis tests:

`
PHP Warning:  require_once(/wordpress-tests-lib/includes/functions.php): failed to open stream: No such file or directory in /home/travis/build/timber/timber/tests/bootstrap.php on line 9
`

I can do some digging later today or this weekend to isolate what's going on. As far as I can tell, the only diff between the PRs is 6dd12393ba0e91878ca97dcd1c5f4a0311647608

gchtr:
@jarednova Yep. You can find more info about it here: https://github.com/timber/timber/pull/1983#issuecomment-490613098. If you start looking into this, please assign yourself to this pull request, so I see you look into it. Otherwise we both work on the same thing ;).

jarednova:
@gchtr cool, didn't realize the thing you were seeing locally was the same thing Travis now sees. I think it makes sense for you to stay on this for now since you're familiar with that error. I've got plenty of other tickets calling my name right now

gchtr:
@jarednova It turns out Travis didn’t have a `$TMPDIR` global, so we need to set it. I read some documentation and tried it out. It seems to work! 🎉

jarednova:
@gchtr awesome! glad it was such a small fix. Let's pull it in!

PR #1996, opened 2019-05-14
2.x Remove unneeded use statements

gchtr:
## Issue

When we are in the same namespace, we don’t have to use `use` statements for classes in the same namespace. E.g., when `Timber\Term` wants to use `Timber\Post`, it can use `new Post()` without a `use` statement.

## Solution

- Remove all unneeded `use` that were flagged by PhpStorm.
- Remove unneeded fully qualified names used in the code (not including tests).
- Optimize class names in `ImageHelper` class.

## Impact

None. Or is there some PHP backwards compatibility that I don’t know of and that has to be considered?

## Usage Changes

None.

## Considerations

None.

## Testing

I ran the tests. This seems to work fine except for the failing tests that are already handled in different pull requests.


lib/ImageHelper.php
```diff
@@ -2,15 +2,8 @@
 
 namespace Timber;
 
-use Timber\Image;
-use Timber\Image\Operation\ToJpg;
```
acobster: These `Timber\Image\Operation` declarations are redundant if you're also using the fully qualified class name as we are [here](https://github.com/timber/timber/blob/d3b301ff85507c2739a7ecbd686b69e0d5ca2f09/lib/ImageHelper.php#L684), but note you can also `use` just a namespace rather than a fully qualified class. So for example we can include a single `use Timber\Image\Operation` here and then usage is simply `new Operation\ToJpg()` or `new Operation\Letterbox`.

gchtr: Yes, of course! Makes total sense for `Timber\Image\Operation` 👍. I updated the class names in the `ImageHelper` class.

PR #1997, opened 2019-05-14
2.x Deprecate Twig environment option variables

gchtr:
## Issue

Setting options for the Twig Environment seems weird. For example, caching has to be enabled by setting `Timber::$cache = true`. And only a couple of the [available options](https://twig.symfony.com/doc/2.x/api.html#environment-options) are supported.

## Solution

This pull request does a couple of things:

### Introduces `timber/twig/environment/options` filter

With this filter, we can set the [options](https://twig.symfony.com/doc/2.x/api.html#environment-options) that are used when an instance of `\Twig\Environment` is created.

Here’s an example:

```php
add_filter( 'timber/twig/environment/options', 'update_twig_environment_options' );

/**
 * Updates Twig environment options.
 *
 * @link https://twig.symfony.com/doc/2.x/api.html#environment-options
 *
 * @param array $options An array of environment options.
 *
 * @return array
 */
function update_twig_environment_options( $options ) {
    $options['cache']       = true;
    $options['auto_reload'] = true;

    return $options;
}
```

### Deprecates setting options via variables on the `Timber` class

Deprecates the following variables:

- `Timber::$twig_cache`
- `Timber::$cache` (alias for `Timber::$twig_cache`)
- `Timber::$autoescape`

When these variables contain non-default values, a call to `Helper::deprecated()` is used to inform developers about the change.

All these options can be set via the `timber/twig/environment/options` filter. Backwards compatibility is still there and can be remove in a future major version.

### Deprecates the `timber/cache/location` filter

When using the `timber/twig/environment/options` filter, a `cache` option can be passed where you can set your own absolute path to where cache files should be stored. This would be the same that the `timber/cache/location` filter would achieve. That’s why I propose to deprecated this filter.

If you use `true` for the `cache` option in the `timber/twig/environment/options` filter, the cache will still use the default location `Timber_LOC . '/cache/twig'`.

### Updates documentation

I updated the documentation and the Upgrade Guide where it was appropriate.

## Impact

The old way of defining options for the Twig Environment still works, but will trigger deprecation warnings.

## Usage Changes

The new `timber/twig/environment/options` will have to be used in the future.

### Filter name

Is `timber/twig/environment/options` a good name for the new filter?

### Setting Twig cache in development

I’m confused about the following lines in the Performance docs:

>  You can look in your your `/vendor/timber/timber/cache` directory to see what these files look like.
>
> This does not cache the _contents_ of the variables. This is recommended as a last-step in the production process. Once enabled, any change you make to a `.twig` file (just tweaking the HTML for example) will not go live until the cache is flushed.

Because for me, when I use the `auto_reload` option, everything seems to work fine. Here’s the description for the `auto_reload` option from the [Twig docs](https://twig.symfony.com/doc/2.x/api.html#environment-options):

> When developing with Twig, it's useful to recompile the template whenever the source code changes. If you don't provide a value for the `auto_reload` option, it will be determined automatically based on the debug value.

Maybe we need to update the docs for this?

## Testing

For all the relevant functions, I added `@expectedDeprecated` tags and created duplicate functions without the deprecated option variables.


docs/guides/performance.md
```diff
@@ -99,13 +99,24 @@ Every time you render a `.twig` file, Twig compiles all the HTML tags and variab
...
+
+```php
+add_filter( 'timber/twig/environment/options', function( $options ) {
+	$options['cache'] = '/absolute/path/to/twig_cache';
+
+    return $options;
+} );
+```
 
 This does not cache the _contents_ of the variables. This is recommended as a last-step in the production process. Once enabled, any change you make to a `.twig` file (just tweaking the HTML for example) will not go live until the cache is flushed.
```
jarednova:
```suggestion
This does not cache the _contents_ of the variables. But rather, the structure of the Twig files themselves (i.e. the HTML and where those variables appear in your template). Once enabled, any change you make to a `.twig` file (just tweaking the HTML for example) will not go live until the cache is flushed.

Thus, during development, you should enable the option for `auto_reload`:
```php
  add_filter( 'timber/twig/environment/options', function( $options ) {
    $options['cache']       = true;
    $options['auto_reload'] = true;
    return $options;
  });
```
... you can set `$options['auto_reload'] = false` when in production

```

gchtr: Yes, I wasn’t aware of the `auto_reload` feature either before working on this pull request. The example I added in the method example matches what I tested out. And it worked flawlessly 😎.

PR #1994, opened 2019-05-14
Fix failing test

gchtr:
**Ticket**: #1900

Same as #1900, but for the `master` branch.

PR #1999, opened 2019-05-14
Added the correct expected result to PostWithVideo

marciojc:
testPostWithVideo was failing, wrong expected result

### DIFF
--- a/tests/test-timber-post.php
+++ b/tests/test-timber-post.php
@@ -1003,7 +1003,7 @@
 			$pid = $this->factory->post->create(array('post_content' => $quote));
 			$post = new Timber\Post($pid);
 			$expected = array(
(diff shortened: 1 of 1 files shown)


palmiak:
I think this is a duplicate of https://github.com/timber/timber/pull/1994

gchtr:
Yes, this is a duplicate. Merely adding the title attribute wouldn’t help, because with older versions of WordPress, the title attribute will not be added. That’s why the build for this pull request still fails.

I’m sorry @marciojc. Thank you very much for your effort of looking into the failing test, but I guess I still have to close this in favor of #1994.

marciojc:
No problem @gchtr  didn't see the #1990 or #1994

Issue #1971, opened 2019-04-21
Comment threads/nesting - Still no examples? How do we do this?


randymwebdev:
I've seen people on here asking since 2015 for examples on how to get threaded/nested comments working, and still not a single example or clear solution has been posted. Apparently CommentThread has been added, but how do we use this? Any example please?

The starter pack theme for timber doesn't support threaded comments. And even when I use
    `{{ function('comments_template') }}`
I don't get any threaded/nested comment behavior. It's like Timber is stopping it. Any ideas?

randymwebdev:
I found out I needed the wordpress **comment-reply.js** file to run. I never included that in my custom theme.

`function theme_queue_js(){
if ( (!is_admin()) && is_singular() && comments_open() && get_option('thread_comments') )
  wp_enqueue_script( 'comment-reply' );
}
add_action('wp_print_scripts', 'theme_queue_js');`

adding that to my functions.php file fixed the issue. But still documentation on comments could be a little better, or at least show a working example.

jarednova:
I agree @randymwebdev — something I'll come up with an example for and add to the docs

PR #2002, opened 2019-05-17
Added big example to top of CommentThread class — with nesting

jarednova:
**Ticket**: #1971

## Issue
As noted, the only example of how to do nested comments is buried in the tests

## Solution
Put an example in!

## Impact
Expands the example in `Timber\Post::comments` showing a nested example. The same example appears in the a new docblock for `Timber\CommentThread`.

## Considerations
Is this the right place? Originally I had it just in CommentThread — but since that's not really intended to be a public class, it felt wrong. This solution provides where I think people are most likely to discover it: with `{{ post.comments }}` which they interface with, and `CommentThread.php` in case someone is looking through the files

### DIFF
--- a/lib/CommentThread.php
+++ b/lib/CommentThread.php
@@ -4,6 +4,35 @@ namespace Timber;
 
 use Timber\Comment;
 
+/**
+ * This object is a special type of array that hold WordPress comments as Timber\Comment objects. 
+ * You propbably won't use this directly. This object is returned when calling `{{ post.comments }}` 
+ * in Twig.
+ *
+ * @example
+ * ```twig
+ * <div id="post-comments">
+ *   <h4>Comments on {{ post.title }}</h4>
+ *   <ul>
+ *     {% for comment in post.comments %}
+ *       {% include 'comment.twig' %}
+ *     {% endfor %}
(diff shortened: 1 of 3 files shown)


lib/CommentThread.php
```diff
@@ -54,6 +91,19 @@ public function order( $order = 'ASC' ) {
 		return $this;
 	}
 
+	/**
+	 * @internal
+	 * @experimental
```
gchtr: The `@experimental` tag is not really an official tag, or is it?

jarednova: No, those are something I saw at some point and brought over — but not official. I think we should scrub them all (and probably the associated functions in 2.0)

jarednova:
@palmiak can I get a follow-up review on this? I added info on using `comment-reply.js` in this update so we should be totally covered now

lib/CommentThread.php
gchtr: I just added a [project card](https://github.com/timber/timber/projects/4#card-22187846) for this.

jarednova:
@palmiak yes. My hope is that this can finally be put to 🛏 — thanks!

Issue #1688, opened 2018-03-08
Provide proper upgrade guidance in admin depending on installation method


jarednova:
This is a todo based on @gsalkin's comments in #1673 — basically we should properly message upgrade procedure in WP Admin based on how Timber was installed (Composer vs. WP.org)

gsalkin:
I would modify requirements for `timber-library` 2.0 to also require Twig 2.x. Currently `timber-library` still calls for the deprecated version of Twig. Now that most installations support PHP 7.x, I would move users to have that installed.

Creates better parity between the Github version and WordPress plugin.

gchtr:
A quick note: Bumping the minimum Twig version to 2.x would probably break compatibility with WPML. WPML still uses Twig 1.32.0 (for compatibility with PHP5) and also loads Twig through Composer. Currently, the only solution I found to make WPML work with Timber is to install Timber as a plugin (related: #1064). But then, WPML’s Twig version will be used instead of Timber’s, which is not future-proof either.

I’m all for parity. I’m not saying we shouldn’t upgrade the minimum requirements (I’m not a big fan of WPML myself 🙊), but if we do it, we probably need to mark Timber as incompatible with WPML.

jarednova:
Gahhhhhh!!!!!! Twig!!!!!

gchtr:
I guess this can be closed. We won’t release 2.x as a plugin, but only as a Composer package. See #2005.

PR #2005, opened 2019-05-19
2.x Drop plugin support

gchtr:
## Issue

I already discussed this with @jarednova and we suggest to **drop the support for installing Timber as a plugin as of 2.x**.

Supporting a plugin version of Timber has been quite a hassle lately. And when we’re honest, Timber is not really a plugin. And it’s not really a framework or a library either. It’s something in-between. It’s a tool to develop themes (and private plugins) with. And it’s evolving constantly. There will be breaking changes that require developers to update their themes – like the huge 2.x update. When Timber can be installed as a plugin, it’s really hard to introduce those breaking changes. Composer allows us to do that, because it won’t automatically install major versions. Installing a major requires more work than just pressing a button in the plugins menu. And that’s a good thing, because as a developer, installing a new major version requires you to test your project with the new version.

Related tickets:

- [#1673 – Force Timber to use Twig 2.x?](https://github.com/timber/timber/issues/1673)
- [#1688 – Provide proper upgrade guidance in admin depending on installation method](https://github.com/timber/timber/issues/1688)

## Solution

This pull request

- Removes the **timber.php** plugin file.
- Removes the script to deploy Timber to the plugin repository.
- Removes the **readme.txt** file. For release notes, I suggest we continue to use the [Releases](https://github.com/timber/timber/releases) section.
- Updates the **Usage Changes** section in the Pull Request Template. A list of changes is probably enough. We will have to fill in the details for the release notes ourselves.
- Updates the documentation.
- Updates the Upgrade Guide.

## Impact

Projects that have Timber installed as a plugin won’t see the 2.x update. The sites will stay on 1.x and continue to work as they did.

## Usage Changes

It will only be possible to install Timber as a Composer dependency.

## Considerations

Anything?

## Testing

No testing.


docs/getting-started/setup.md
```diff
@@ -8,34 +8,31 @@ menu:
...
-
-The GitHub version of Timber requires [Composer](https://getcomposer.org/download/). If you'd prefer one-click installation, you should use the [WordPress.org](https://wordpress.org/plugins/timber-library/) version.
+You can install Timber through [Composer](https://getcomposer.org/download/).
 
 ```shell
 composer require timber/timber
 ```
 
-If your theme is not setup to pull in Composer’s autoload file, you will need to add the following at the top of your `functions.php` file: 
+You can choose yourself where in your project you want to include Timber. Some developers prefer to have Timber installed as a theme dependency, so they would run this command from the theme root. But it’s also possible to use Timber as a WordPress dependency, which means you would run the command above from the WordPress root.
```
jarednova: I think this should be `**Most** developers prefer to have Timber installed as a theme dependency.` It's pretty clear that the other suggestion is more alternative `it's also possible...` I have to confess that I've only ever tested/used it as a theme or plugin dependency — never for the entire WP install

jarednova (changes requested):
😢 the end of plugin days! (honestly, good riddance). I think just one tweak to make here then we're 👍

jarednova:
@gchtr looks good! Just one small tweak and then we can say "hello" to a much simpler Composer-only future

docs/getting-started/setup.md
gchtr: Yeah, that’s probably better. I replaced "Some" with "Most".

jarednova (approved):

tomlecki:
@jarednova @gchtr 
If I can say something here I would like to vote against this change. 
I can explain it from a different view than yours. 

In last few years I've used Timber in many websites using mostly plugin version as it was just quicker. I introduced Timber to a few 'non-techy' people and they fell in love with it because of simplicity. You just install plugin, download starter theme and it works just like that. You just edit the .twig and .php files.

If the only reason is a problem with informing users about 2.0 update and changes that will brake site - maybe consider creating another plugin like "Timber 2". Current one would stay as it is and people could just install the second version for new websites that would be compatible with new features and not using deprecated stuff. If it's not technically possible for some reason just ignore this idea.

I personally can handle this change, I've used composer in a few projects with Timber but for those who wanted simplicity it's gonna be another obstacle - installing composer, using terminal etc. It's another thing to learn.

Right now Timber has 30 000+ active installs from WP repository. I don't know this for sure, but I assume a lot of those installs are just websites made by people attracted by Timber's low entry-level knowledge and that it works as I said before "out of the box". It is powerful but to start you don't need much. In fact that was the reason that attracted me 5 years ago.

Of course I don't know what is going on behind the scene with the current plugin. Maybe it would consume too much time to stay up-to-date with composer version? If yes, I understand it totally and I would also support this plugin removal change. But if the reason is not work and time but just this update issue, maybe consider leaving plugin as optional install method.

gsalkin:
@tomlecki I respect a lot of the appeal of the plugin version of Timber for fast development. But the truth is as said above: Timber is not a plugin. Shipping it as such is misleading and can cause tons of issues for site maintainers. Especially as technical debt piles up, I'm concerned about vulnerabilities. For instance, the plugin version of Timber never adopted full Twig 2.0 support. 

I think there *should* be a sunsetting plan. Can we think about a period of time for maintenance updates? And of course we'll need to work on migration documentation.

jarednova:
@tomlecki @gsalkin thanks for following-up on this. I really appreciate the passion and enthusiasm. It means that what we're doing is important and reaches people (this is good!)

The plan is to continue development into the 2.x branch as Composer-only. That doesn't mean that Timber's presence in WP.org will be going away though.

We want to maintain the best possible developer experience, so we'll leave the existing plugin on WP.org for the indefinite future. If we find any critical bugs, we'll update them too. But what we can't do is maintain two versions into the distant future. Syncing the versions is an error-prone process and I'm very wary of (and have made mistakes with, despite scripts and other tools to automate the process).

All that said, these are plans and things could change (especially if someone volunteers for a role reaching out to new devs just getting into WordPress and templating!??!??!?!) — but for now we're planning on making 2.0 Composer-only once it's released.

jarednova:
... and of course as @gsalkin notes, we'll be sure to document a migration path and on-boarding materials. I hope @tomlecki you can keep us honest on this front to ensure the great experience you have (and have introduced others too) can continue!

arun369:
Some of the points raised by @tomlecki are pretty good and I agree and I have a bit of mixed reviews about this as for now, it will definitely take away the usability for not-so-tech users of timber but considering the long term impacts and how it should be treated, I appreciate the decision. It absolutely puts it out as a more mature solution for developers and clears the confusion about where it stands as a plugin/tool/framework. 

Here is an interesting thing, a few weeks back I installed Twig using Composer directly into wordpress theme which I was developing  because timber as plugin didn't feel like the right choice, mostly because it does feel like part of a theme more than a plugin and it worked well (minus all the features that timber has but twig can not for obvious reasons). For that particular instance Timber would have been a better choice if it was available with Composer, although for most of developments I install Timber as a plugin.

I've been using Timber for about 3 years and initally I hated 😡 it but as I came to know more, I now prefer to make a whole theme on top of it because of the clean separation it provides which gives ability to a bigger team (designing & development) to work on things simultaneously. ♥😍

Composer does have a bit of learning curve but all serious developers have to learn it at some point anyway, because any complex development will also have additional libraries such as emogrifier, emailer libraries, data validation libraries, pdf generation and such which have to be installed with Composer.

I don't know all ins and outs of Timber yet, but I think this steps makes it adapt the newer twig versions easier than before. (or may be I'm incorrect and someone should correct me 😛)

Keep up the good work 👍👍

Issue #1863, opened 2018-11-30
Consider adopting type declarations


acobster:
PHP 5.1+ allows for declaring arrays and class/interface names in method signatures, e.g.

```php
class Post {
  public function my_method( Post $post ) {
    return $post->meta('whatever');
  }

  public function other_method( array $data ) {
    return $data['some_index'];
  }
}
```

This results in much clearer error messages when the caller passes something other than the declared type. For example, **Uncaught TypeError: Argument 1 passed to Post::my_method() must be an instance of Post, null given** instead of **Uncaught Error: Call to a member function meta() on null.** Whereas the latter error tells the caller "there's an error happening somewhere in my_method()", which could imply a bug in the library code, the former says definitively "you're doing it wrong."

Another benefit is much greater leverage from PHPStan, which we've already adopted.

We should consider adopting this as much as possible across the Timber codebase for 2.0. A major upgrade is really the only appropriate time to do this, as this is fundamentally a breaking change.

Here are some changes we should consider making:

* Require PHP 7.3 or 7.4 (7.2 to be EOL'd November 2020)
* Nullable types in the top-level API/Factories - requires 7.1+, returning `null` instead of `false` in "unhappy path" situations, e.g. from `::get_post()` when no post found
* Declare types of object properties
* Declare `object` param types in certain Factory methods

All of this would be a pain to do manually, but with [Rector](https://github.com/RectorPHP/Rector) we might be able to automate a lot of this as outlined in #2339.

gchtr:
I like this! Clearer error messages always help. And it’s probably better for code hinting in the various IDEs. Can we define some rules/guidance, when we would apply it?

- Would we apply it for all parameters that accept a single type only?
- Would we apply it for optional parameters?
- Would we apply it only for arrays and classes, or for [all types](https://mlocati.github.io/articles/php-type-hinting.html)?

If we use more types than arrays, classes or callables, then we will have to raise the minimum required PHP version for Timber to 7.0. I don’t know if we already defined this, but I think it would be the right step for 2.0 to drop support for PHP 5.x.

acobster:
I looked and didn't find any prior discussion of increasing the minimum PHP version. I wasn't sure how far down that road folks would want to go, so I kept my proposal to types that 5.3 already supports. But if 7.x is on the table, I'm all for that. Would be curious to hear @jarednova's thoughts.

[...]

PR #2006, opened 2019-05-19
2.x Drop support for PHP versions < 5.x

gchtr:
**Ticket: #1965**

When we drop support for Twig 1.x in #1965, we will have to drop support for PHP 5.x. The [Travis log says it](https://travis-ci.org/timber/timber/jobs/534011840):

```
twig/twig v2.9.0 requires php ^7.0 -> your PHP version (5.6.30) does not satisfy that requirement.
```

I added this pull request to handle this separately. We already discussed this a bit in #1863. WordPress updated the [minimum required PHP version to 5.6](https://make.wordpress.org/core/2018/12/08/updating-the-minimum-php-version/). WordPress has to consider a lot more backwards compatibility and consider setups that are a little more low-level than Timber. I think we can introduce a minimum PHP version of 7.0 for Timber.

## Solution

This pull request updates the Composer requirements, the Travis file and the Upgrade guide.

## Impact

Timber can only be installed in environments that support PHP 7.x.

## Usage Changes

No.

## Considerations

I’m asking myself, should we go even higher and make 7.2 the minimum required version? When we look at the [Supported Versions of PHP](https://secure.php.net/supported-versions.php), PHP 7.0 has already reached End of Life and active support for PHP 7.1 has already ended by the end of 2018.

However, I have no idea about the hosting landscape out there that Timber developers use and whether a version 7.2 would be too high. Is 7.2 already widely available on hosters? Should we use a minimum required version of 7.2 to nudge everyone into a modern stack?

What are your thoughts on this (@jarednova @acobster @pascalknecht @palmiak @chrisvanpatten)? Or anyone else who wants to chime in?

## Testing

Let Travis do the testing and see what comes up.


jarednova:
@gchtr nice to not have to worry about PHP 5 in 2.x anymore. God, that will make the tests go faster. In terms of PHP 7.0 vs. 7.2, I think we should be aware of what WP.org stats have for us there:

https://wordpress.org/about/stats/

... currently PHP 7.x is found on about 50% of installs — PHP 7.2+ is on about 20% of installs. Obviously, this number is only going to increase. My vote is to bite the bullet and go 7.2. I think the developers who are likely to use Timber 2 aren't going to be on a $5 GoDaddy server. 

WPEngine (which I consider a good barometer of a WP host here in 🇺🇸 ) is pushing customers to upgrade to 7.2. This trend is only going to continue.

gchtr:
> WPEngine (which I consider a good barometer of a WP host here in 🇺🇸 ) is pushing customers to upgrade to 7.2. This trend is only going to continue.

It’s the same here in Switzerland 🇨🇭. PHP 5.6 is not available anymore, and PHP 7.2 is the new standard.

chrisgherbert:
Requiring 7.2 feels like a step too far. While we’d all love to use hosts that stay on the cutting edge, clients don’t always make that a priority.

acobster:
I agree, we should shoot for parity with Twig.

jarednova:
"parity with Twig" feels like the perfect place to leave this: easy to explain/justify and maintain over time

gchtr:
Thank you all for your answers! 👍 We’re sure in no rush to update to 7.2. And parity with Twig indeed seems like a very good approach. Let’s do that!

PR #2009, opened 2019-05-20
 ref #1974 -- apply fix suggested by WPML team

strategio:
Hi @jarednova, we have one unrelated test failing as you can see here https://travis-ci.org/timber/timber/jobs/533954649. I didn't have time to fix my local setup to run tests, but I rebased the branch on `master` maybe it helps...

Does the error reported on Travis rings a bell to you?

### DIFF
--- a/lib/Menu.php
+++ b/lib/Menu.php
@@ -258,6 +258,9 @@ class Menu extends Core {
 					$old_menu_item = $item;
 					$item = new $this->PostClass($item);
 				}
+				if ( get_class($item) == 'WPML_LS_Menu_Item' ) {
+					$old_menu_item = $item;
+				}
 				$menu_item = $this->create_menu_item($item);
 				if ( isset($old_menu_item) ) {
 					$menu_item->import_classes($old_menu_item);
--- a/lib/Menu.php
+++ b/lib/Menu.php
@@ -258,7 +258,7 @@ class Menu extends Core {
 					$old_menu_item = $item;
 					$item = new $this->PostClass($item);
 				}
-				if ( get_class($item) == 'WPML_LS_Menu_Item' ) {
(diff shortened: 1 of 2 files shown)


jarednova:
@strategio awesome! That Travis issue was solved by #1994 so once we catch it up with `master` it should all pass

lib/Menu.php
```diff
@@ -258,6 +258,9 @@ protected function order_children( $items ) {
 					$old_menu_item = $item;
 					$item = new $this->PostClass($item);
 				}
+				if ( $item instanceof WPML_LS_Menu_Item ) {
```
jarednova:
```suggestion
				if ( $item instanceof \WPML_LS_Menu_Item ) {
```

jarednova: I think this will ensure that it's looking for the constant at the root level. See this error in Scrutinizer: 

https://scrutinizer-ci.com/g/timber/timber/inspections/b3bfce39-18a1-41db-a058-3b49d203dea7/issues/files/lib/Menu.php?status=new&orderField=path&order=asc&honorSelectedPaths=0

gchtr:
Cool that someone from WPML is looking into making Timber more compatible 👍.

@jarednova I’m a little hesitant about adding code / handling special cases specific to a plugin directly in Timber core. Usually we use custom integrations for all the required changes related to a third-party plugin. Could we instead add a filter that would allow this check to be added by an integration? Or would this be too nitpicky and pedantic?

jarednova:
Nitpicky in the right ways. I forgot we already had an `Integrations/WPML` class we could use. Now that we know this works, I'll pull this functionality into it

jarednova:
So #2013 solves this much more cleanly w/o the need for any WPML-specific code. Moving things over there to wrap up ...

PR #2012, opened 2019-05-24
Conditional eager-loading of post meta values

### LINKED ISSUE #2011: Allow timber_post_get_meta_pre filter to short circuit or alter Timber\Post::get_post_custom()

When dealing with large collections of items, sometimes the eager loading behaviour of Timber works against us and makes for heavy arrays that consume a lot of memory. We notice this particularly when the items we're querying for / working with have a lot of ACF flex content, as this really bulks out the meta.

It looks like there has already been an attempt to solve this in the get_post_custom() function, but the filter that runs has no opportunity to affect the behaviour beyond the callback, the `get_post_custom()` will be called anyway, all the time.

This is the "first" issue, and one that's much easier to deal with, then the "real" issue that I've actually had multiple times, which is the same thing on Terms, where the meta fetching and capabilities are very intertwined with ACF, rather than using the now widely available term_meta functions and using ACF if necessary.

Pull request for both inbound, this one is easy, and sets the pattern/tone for the next harder one related to terms.

Some rough pseudo code below
```php
/* Prevent eager loading of Post Meta */
add_filter( timber_post_get_meta_pre, function() { return false; } );

/* only eager load certain meta keys */

add_filter( 'timber_post_get_meta_pre', function($customs, $pid, $timberPost ) {
    $keys = [ 'key1', 'key2', 'key3' ];
    $customs = [];
    foreach( $keys as key ){
        $customs[ $key] = get_post_meta( $pid, $key );
    }
    return $customs;
}, 10 , 3 );
```

It's possible to filter the values returned to the import statment, but that's only after the data is fetched from the db, so this way aims to reduce the overhead of that on big queries / collections.

## Additional context

unsure if should be implemented as `timber_post_get_meta_pre` or `timber/post/meta/pre` ?

aj-adl:
**Ticket**: Fixes #2011 

## Issue
Allows the user to control what post meta values get eager-loaded into `\Timber\Post->custom`

## Solution
Supports both disabling, and conditional loading

## Impact
This filter currently cannot affect how the parent function of `get_post_custom` runs. There is a potential issue if people have been hooking into here to perform some action, and returned null / void etc as previously the return value had no effect. I doubt many people are getting that into the guts of Timber, but it is something to be considered

## Usage Changes

Allows the user to return `false` or `null` to stop the eager-loading, or return an array of `$key => $value` which will then be processed and imported into the `Timber\Post` or derivative object. 

## Considerations
naming? not suer how this looks in 2.0, seems like some changes there

## Testing
Unit tests included and passing in `tests-timber-post.php`

### DIFF
--- a/lib/Post.php
+++ b/lib/Post.php
@@ -531,11 +531,16 @@ class Post extends Core implements CoreInterface {
 	 * @return array
 	 */
 	protected function get_post_custom( $pid ) {
-		apply_filters('timber_post_get_meta_pre', array(), $pid, $this);
-		$customs = get_post_custom($pid);
-		if ( !is_array($customs) || empty($customs) ) {
+		$customs = apply_filters('timber_post_get_meta_pre', array(), $pid, $this);
+
+		if ( is_array($customs) && empty($customs) ) {
+			$customs = get_post_custom($pid);
+		}
+
+		if ( !is_array($customs) ) {
 			return array();
 		}
+
 		foreach ( $customs as $key => $value ) {
 			if ( is_array($value) && count($value) == 1 && isset($value[0]) ) {
 				$value = $value[0];
--- a/tests/test-timber-post.php
+++ b/tests/test-timber-post.php
@@ -319,6 +319,47 @@
 			$this->assertEquals($page2, trim(strip_tags( $post->get_paged_content() )));
 		}
 
+		function testMetaCustomPreFilterDisable(){
+
+			$callable = function(){ return false; };
+
+			add_filter( 'timber_post_get_meta_pre', $callable );
+
+			$post_id = $this->factory->post->create();
+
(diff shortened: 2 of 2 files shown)


jarednova:
@gchtr has been working on re-writing meta value handling in 2.x, so I'm gonna let him take this from here on how to best pull this (or its ideas) in

gchtr:
Hey @aj-adl, I just opened #2014 to match the behavior that I added in 2.0. Turns out that you would approach this the same way as I would do it 👍.

I think we’ll leave the name as it is. You’ll have to update the filter to the new name `timber/post/pre_get_meta_values` as soon as 2.0 is released, though.

jarednova:
@aj-adl thanks again for submitting. I see that @gchtr pulled the underlying work and tests over in #2014. Really appreciate you flagging the issue, submitting the PR AND supplying tests. Triple crown!

PR #2014, opened 2019-05-30
Ref #2012 – Update skipping get_post_meta() to match behavior in 2.0

gchtr:
**Ticket**: #2012

## Issue

In #2012 @aj-adl proposed a way to skip Timber from fetching meta values through `get_post_meta()` via the `timber_post_get_meta_pre` filter. Turns out that in 2.x, I added pretty much the same behavior, just with a new filter:

https://github.com/timber/timber/blob/35279e5fa8911c52854f2d268fc028bcf76c2313/lib/Post.php#L574-L624

I wasn’t sure whether this was over-optimization, but hey, here we have a use case already 😎.

## Solution

This pull request builds on #2012 to replicate the behavior that we’re going to add in 2.0 in the master branch so that this feature can be used now.

## Impact

None.

## Usage Changes

See #2012.

## Considerations

None.

## Testing

None.

### DIFF
--- a/lib/Post.php
+++ b/lib/Post.php
@@ -547,7 +547,7 @@ class Post extends Core implements CoreInterface {
 		 *
 		 * // Add your own meta data.
 		 * add_filter( 'timber_post_get_meta_pre', function( $post_meta, $post_id, $post ) {
-    	 *     $post_meta = array(
+		 *     $post_meta = array(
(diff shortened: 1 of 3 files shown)


jarednova:
Thanks @aj-adl and @gchtr — I added a note to the readme so this can go out with all the info in the next version

Issue #1731, opened 2018-06-19
$post->comments() function ignores WP's comment order.


kiiraklis94:
### Expected behavior
In the docs it say that the comments function takes an $order parameter.
Default is "wp" which it says is meant to use the wordpress option found in settings>discussion.

### Actual behavior
I've set this everything from 'wp' to 'wordpress' and even tried 'asc' and 'desc' but nothing seems to work. It just ignores it and displays the comments with the older comments on the top.

### What version of WordPress, PHP and Timber are you using?
WordPress 4.9.6, PHP 7.2, Timber 1.7.1

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Composer

alicam:
This FIX for this is to change CommentThread.php, line 45. From:

`return array_merge($base, $args, $overrides);`

to

`return array_merge($base, $args);`

In other words, there is currently a forced override of the comment ordering. 

Why?!

jarednova:
Thanks @alicam! I've applied your suggested fix in #2015

jarednova:
Closed by #2015 — release coming soon!

PR #2015, opened 2019-05-31
ref #1731 -- apply suggested fix for comment order override

jarednova:
**Ticket**: #1731 

## Issue
Manual ordering args were overridden by class's value

## Solution
Make class's value the default, can now be overriden by user defined value

## Impact
This will only cause an issue in the case where someone's failed attempts to override were unsuccessful — now they'll succeed

## Testing
No new tests


jarednova:
@palmiak can I get your review? I think this is an artifact of some experimental stuff I was trying many years ago, that we're in the process of stripping out

Issue #1959, opened 2019-03-29
'nav_menu_css_class' filter $args param empty array


benedict-w:
## Expected behavior
When applying the filter _nav_menu_css_class_, the third param $args should be an array of wp_nav_menu args https://developer.wordpress.org/reference/functions/wp_nav_menu/ - see: https://codex.wordpress.org/Plugin_API/Filter_Reference/nav_menu_css_class

## Actual behavior
Empty array passed instead.

See MenuItem.php, function `import_classes`, line 236 (Timber v1.9.3)

`$this->classes = apply_filters('nav_menu_css_class', $this->classes, $this, array(), 0);`

This means that when applying my own filter to nav_menu_css_class I can not identify the theme_location from the $args param to add my own CSS class.

There's an example of how you would do this in the WordPress docs on this page: https://codex.wordpress.org/Plugin_API/Filter_Reference/nav_menu_css_class

```
function my_secondary_menu_classes( $classes, $item, $args ) {
    // Only affect the menu placed in the 'secondary' wp_nav_bar() theme location
    if ( 'secondary' === $args->theme_location ) {
        // Make these items 3-columns wide in Bootstrap
        $classes[] = 'col-md-3';
    }

    return $classes;
}

add_filter( 'nav_menu_css_class', 'my_secondary_menu_classes', 10, 3 ); 
```

Adding a function like this doesn't work for TimberMenu because the $args param is set to an empty array.

gchtr:
You’re right, Timber deviates in the filter arguments for `nav_menu_css_class`. These are the default args that are used for `wp_nav_menu()`:

```php
$defaults = array(
	'menu'            => '',
	'container'       => 'div',
	'container_class' => '',
	'container_id'    => '',
	'menu_class'      => 'menu',
	'menu_id'         => '',
	'echo'            => true,
	'fallback_cb'     => 'wp_page_menu',
	'before'          => '',
	'after'           => '',
	'link_before'     => '',
	'link_after'      => '',
	'items_wrap'      => '<ul id="%1$s" class="%2$s">%3$s</ul>',
	'item_spacing'    => 'preserve',
	'depth'           => 0,
	'walker'          => '',
	'theme_location'  => '',
);
```

Out of these, only the following would be applicable to a Timber Menu, because all others relate to the menu walker, which we don’t have in Timber:

```php
$defaults = array(
	'menu'            => '',
	'depth'           => 0,
	'theme_location'  => '',
);
```

[...]

PR #2018, opened 2019-05-31
Menu optimizations

gchtr:
**Ticket**: #1959

## Issue

See #1959.

## Solution

This pull request:

- Adds a `theme_location` property to the `Timber\Menu` class that is set when the menu is registered with a theme location.
- Adds a `menu` property to the `Timber\MenuItem` so it’s possible the parent menu object from a menu item. For this, the parent menu object is passed to the constructor of `Timber\MenuItem`.
- Adds an `options` property to the `Timber\Menu` class that holds an array of default options that can be adapted through `Timber\Menu`’s constructor. It also remove the `Timber\Menu::set_options()` method and moves its content back to the constructor. I thought there’s no need to have a separate function for this.
- Passes down the new `options` property as an object for the `$args` parameter in the nav_menu_css_class filter.
- Updates the default for `depth` from `-1` to `0` to match WordPress’s defaults in [`wp_nav_menu()`](https://developer.wordpress.org/reference/functions/wp_nav_menu/).

## Impact

Better parity with WordPress menu filters and defaults.

## Usage Changes

It’s possible to access the theme location in various places, like in the `nav_menu_css_class` filter:

```php
function my_secondary_menu_classes( $classes, $item, $args ) {
    if ( 'secondary' === $item->menu->theme_location ) {
        $classes[] = 'theme-location-secondary';
    }

    return $classes;
}

add_filter( 'nav_menu_css_class', 'my_secondary_menu_classes', 10, 3 ); 
```

And the `$args` parameter in that filter contains the options passed to `Timber\Menu`.

## Considerations

Should we add the options for the args parameter in the `nav_menu_css_class` filter or is it obsolete, because the same options could be accessed through `$item->menu->options`?

I didn’t add `theme_location` to `Timber\Menu::$options`, because when `theme_location` is used as an argument for `wp_nav_menu()`, it’s a query to select only the menu from that location. But in Timber, you could pass this in the first argument of `Timber\Menu::__construct()`. So you would never use it as an option.

In `Timber\Menu::init()`, I added `$locations = get_nav_menu_locations();` to get theme the locations in order to set the `theme_location` property. We already use `get_nav_menu_locations()` in the `Timber\Menu::__construct()`, but it didn’t feel right to pass it down as a parameter in `Timber\Menu::init()`. The locations should be cached by WordPress’s internal option handling, so I hope we’re fine here.

## Testing

Yes.

### DIFF
--- a/lib/Menu.php
+++ b/lib/Menu.php
@@ -296,7 +296,7 @@ class Menu extends Core {
 	 * @return mixed an instance of the user-configured $MenuItemClass
 	 */
 	protected function create_menu_item($item) {
-		return new $this->MenuItemClass($item);
+		return new $this->MenuItemClass( $item, $this );
 	}
 
 	/**
--- a/lib/Menu.php
+++ b/lib/Menu.php
@@ -48,16 +48,24 @@ class Menu extends Core {
 
 	/**
 	 * @api
-	 * @var array The unfiltered options sent forward via the user in the __construct
-	 */
-	public $raw_options;
-
-	/**
-	 * @api
 	 * @var string The name of the menu (ex: `Main Navigation`).
 	 */
 	public $title;
 
+	/**
+	 * Menu options.
+	 *
+	 * @api
+	 * @since 1.9.6
+	 * @var array An array of menu options.
+	 */
+	public $options;
+
+	/**
+	 * @api
+	 * @var array The unfiltered options sent forward via the user in the __construct
+	 */
+	public $raw_options;
 
 	/**
 	 * Theme Location.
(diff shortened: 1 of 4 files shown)


jarednova:
> Should we add the options for the args parameter in the nav_menu_css_class filter or is it obsolete, because the same options could be accessed through $item->menu->options?
If I understand this correctly, the args should be sent as well (as the code you have currently has it). Even if it's duplicative, I think we want to play nicely with [how the filter normally works](https://codex.wordpress.org/Plugin_API/Filter_Reference/nav_menu_css_class) and where a dev is expecting to find that info

>I didn’t add theme_location to Timber\Menu::$options, because when theme_location is used as an argument for wp_nav_menu(), it’s a query to select only the menu from that location. But in Timber, you could pass this in the first argument of Timber\Menu::__construct(). So you would never use it as an option.
👍 

> In Timber\Menu::init(), I added $locations = get_nav_menu_locations(); to get theme the locations in order to set the theme_location property. We already use get_nav_menu_locations() in the Timber\Menu::__construct(), but it didn’t feel right to pass it down as a parameter in Timber\Menu::init(). The locations should be cached by WordPress’s internal option handling, so I hope we’re fine here.
My experience is that those WP functions are always cached. I think relying on that vs. a crazy chain of argument passing is better

PR #2019, opened 2019-05-31
Streamline get_meta_values functions across objects

gchtr:
**Ticket**: #1617 

## Issue

There were failing tests after merging in #2014. And it made me realize that for the meta functionality, I added way too few tests.

## Solution

This pull request first streamlines the functionality of what the `get_meta_values()` does across Post, Term, User and Comment objects. I figured it’s actually good this is coming together step by step, because when I started out with working on meta for 2.x, I was hard to get an overview.

Now, all objects import all meta values into a `custom` property. And all `get_meta_values()` functions do pretty much the same, including the same variable pattern etc.

I also added tests for the different objects. The tests that were introduced in #2012 are really good to test the `timber/{object}/pre_get_meta_values` filter (Thanks, @aj-adl!). 

## Impact

Fewer fails.

## Usage Changes

None.

## Considerations

I need to add more tests to improve coverage of changes for meta. There will be more pull requests.

## Testing

Yes.

### DIFF
--- a/lib/Comment.php
+++ b/lib/Comment.php
@@ -385,13 +385,11 @@ class Comment extends Core implements CoreInterface, MetaInterface {
 
 	/**
 	 * @internal
+	 *
 	 * @param int $comment_id
 	 * @return array
 	 */
-	protected function get_meta_values( $comment_id = null ) {
-		if ( $comment_id === null ) {
-			$comment_id = $this->ID;
-		}
+	protected function get_meta_values( $comment_id ) {
 
 		$comment_meta = array();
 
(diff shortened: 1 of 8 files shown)


gchtr:
@jarednova This pull request should fix the current build issues in `2.x`.

jarednova:
Great! Will review so we can unblock the other 2x stuff

gchtr:
@jarednova Cool, thanks! 👍 I feel like the more we iterate with small changes, the more we see the patterns that will make different object types work in the same way.

Issue #2020, opened 2019-06-03
Timber 1.9.5 breaks my site "Ce site rencontre des difficultés techniques."


AL20:
Hi,

I just updated my website with Timber 1.9.5 and I get this message '`Ce site rencontre des difficultés techniques`.' (in english : _The site is experiencing technical difficulties_. equivalent to the White Screen of Death since WP5.2) on all the pages that use Timber + Twig templates.

My setup (with a custom theme) :
- WP last version (5.2.1)
- ACF Pro (last version)
- CPT UI (last version)
- Woocommerce (last version)
- WPML (last version)

I tried to desactivate them individually but the issue remains.

In advance, thanks for your help.

palmiak:
This message is just more user friendly 500 in WP 5.2

If you would downgrade timber to earlier version will the problem still
exists?

Also try to turn on WP_DEBUG it should show you the real problem

pon., 3 cze 2019, 12:17 użytkownik AL20 <notifications@github.com> napisał:

AL20:
When I rollback to Timber 1.9.4, the issue is gone.
I set :
```
define('WP_DEBUG', true);
define('WP_DEBUG_LOG', true);
define('WP_DEBUG_DISPLAY', true);
```
but I don't get more informations.

palmiak:
I see you are using WPML. From what I remember Wpml also uses twig and there might be some conflict here. If I would have to guess I think it might be related with this update - https://github.com/timber/timber/pull/1963 . 

Is there anything in an error_log?

AL20:
Maybe, I know WPML brings a lot of issue in general…
Here's what I got from the log :

```
Warning: Cannot declare interface Twig_LoaderInterface, because the name is already in use in /Users/vinsan/Dropbox/DEV/Sites/site_skyfarms_boutique/wp-content/plugins/timber-library/vendor/twig/twig/src/Loader/LoaderInterface.php on line 61

Warning: Cannot declare interface Twig_ExistsLoaderInterface, because the name is already in use in /Users/vinsan/Dropbox/DEV/Sites/site_skyfarms_boutique/wp-content/plugins/timber-library/vendor/twig/twig/src/Loader/ExistsLoaderInterface.php on line 33

[...]

Issue #2016, opened 2019-05-31
Cannot redeclare twig_array_batch()


tlygnersjo:
## Expected behavior
I expect the website to work without errors.

## Actual behavior
Error page with the following message:
```
The site is experiencing technical difficulties.

Fatal error: Cannot redeclare twig_array_batch() (previously declared in /path/to/website/wp-content/plugins/woocommerce-multilingual/vendor/twig/twig/lib/Twig/Extension/Core.php:1557) in /path/to/website/wp-content/plugins/timber-library/vendor/twig/twig/src/Extension/CoreExtension.php on line 1667
```

## Steps to reproduce behavior
Run a WordPress website with WPML (including WooCommerce Multilingual) with Timber plugin. Update from Timber 1.9.4 to 1.9.5

## What version of WordPress, PHP and Timber are you using?
WordPress 5.2.1
PHP 7.1.29
Timber 1.9.5

## How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Upgraded to newest version via plugin updater in WordPress dashboard

tlygnersjo:
Could it be a function name (twig_array_batch) which is the same in WooCommerce Multilingual and Timber?

gio00:
Yes, same error here. I reinstalled the 1.9.4 and everything works again. I guess must be some twig version loaded twice.

jarednova:
Thanks @tlygnersjo, @gio00  — I think this is related to #2020. I'm investigating today ...

tlygnersjo:
Thanks @jarednova sounds like The sme issue. I can confirm that downgradingto 1.9.4 solves the issue

jarednova:
@tlygnersjo @gio00 I'm pretty confident this is an issue with a collision of Twig versions. Can you replace your `wp-content/plugins/timber-library` directory with this test version and see if that resolves the issue?

https://www.zipshare.com/download/eyJhcmNoaXZlSWQiOiI3YzVhMjY0NS0yNGQwLTQ3NjQtYmQyZC1mMWQxNjBkNTYzMzYiLCJlbWFpbCI6ImphcmVkQHVwc3RhdGVtZW50LmNvbSJ9

coreylafferty:
@jarednova That fixes the issue for me (tested on a new WP install with just my theme and the Timber plugin installed).

gio00:
[...]

PR #2022, opened 2019-06-04
Release of 1.10.0 and resolution of WPML errors

jarednova:
**Ticket**: #2020 #2016 

## Issue
A cascade of issues starting with us releasing 1.9.4/5 as patches instead of minor versions. Upgrading Twig collided with WPML's included Twig version wrecking havoc across sites

## Solution
Release 1.10.0 as non-WPML compatible (until we can find a permanent solution). Recommend to sites running WPML that they run 1.9.2 instead (and pull 1.9.3, 1.9.4 and 1.9.5 from Packagist and WP.org to reflect this)

## Impact
It's probably going to be annoying for WPML users (for now). But everyone else will be happy.

## Considerations
- We could do a big hunt and make a (more) stable version of Timber at 1.9.6; but there'd be few upsides versus 1.9.2 and it would be a fair amount of work. This hypothetical version would have a number of quirks that could make it very unstable or create other issues for us to chase after
- We should consider how to potentially bail loading Timber's Twig if a Twig is already loaded. This would benefit WPML and other plugins

## Next Steps
Though far from ideal, I feel strongly that this is the best option to avoid further issues with WPML setups and going on too much of a goose chase. I'd love @palmiak or @gchtr to validate this thinking before I move forward with a release though

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
   },
   "require": {
     "php": ">=5.3.0|7.*",
-    "twig/twig": "1.34|^2.4",
+    "twig/twig": "^1.34|^2.4",
     "upstatement/routes": "0.5",
     "composer/installers": "~1.0",
     "asm89/twig-cache-extension": "~1.0"
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
   },
   "require": {
     "php": ">=5.3.0|7.*",
-    "twig/twig": "^1.34|^2.4",
+    "twig/twig": "^1.38|^2.4",
     "upstatement/routes": "0.5",
     "composer/installers": "~1.0",
     "asm89/twig-cache-extension": "~1.0"
--- a/lib/Timber.php
+++ b/lib/Timber.php
@@ -35,7 +35,7 @@ use Timber\Loader;
  */
 class Timber {
 
-	public static $version = '1.9.6';
+	public static $version = '1.10.0';
 	public static $locations;
 	public static $dirname = 'views';
 	public static $twig_cache = false;
--- a/readme.txt
+++ b/readme.txt
@@ -3,7 +3,7 @@ Contributors: jarednova, connorjburton, lggorman
 Tags: template engine, templates, twig
 Requires at least: 4.7.12
 Tested up to: 5.2.1
-Stable tag: 1.9.6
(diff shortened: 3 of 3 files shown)


palmiak:
Even with the non-plugin version there was the problem with WPML and I had to load Timber as MU Plugin to be sure it was loaded faster than the WPML's one and I never had any problems with this.

Maybe this would be a temporary solution - an extra mu-plugin for WPML users that would force using our Twig.

The downside - probably if someone is using https://wpml.org/documentation/getting-started-guide/language-setup/designing-custom-language-switchers-using-template-files/ - he would run into some trouble (I never used this feature this way).

gchtr:
This seems like a reasonable approach to me.

> * We should consider how to potentially bail loading Timber's Twig if a Twig is already loaded. This would benefit WPML and other plugins

As we’ve discussed before, it’s not only a problem with WPML alone, but with any plugin that uses Twig. Plugins just [shouldn’t load public Composer packages into the global namespace](https://wptavern.com/a-narrative-of-using-composer-in-a-wordpress-plugin). This means that for the plugin version of Timber, we’d have to use put Timber’s Twig into its own namespace. That probably would require a `post-install-cmd`. We’d have to explore what options we have there.

I don’t know if it’s possible to bail loading Timber’s Twig, because as far as I know, Composer doesn’t allow fine-grained control like this.

On the other hand, when we move away from offering a plugin version, I’m not so sure anymore if we should namespace Timber’s Twig, because it would totally take away the incentive for plugin developers to fix how they include Twig. And it could be really hard to make the namespacing process work properly on all the developer machines out there, where Timber will be installed through Composer.

> Maybe this would be a temporary solution - an extra mu-plugin for WPML users that would force using our Twig.

It’s not only the language switcher. If we force Timber’s Twig, some WPML functionality in the backend might not work either, because WPML heavily uses Twig in the backend. It works as long as Timber’s Twig is compatible with how WPML uses Twig.

jarednova:
Thanks @gchtr. I did some testing on `mu-plugins` hack and it unfortunately doesn't fix the problem (basically, Timber's Twig is too new for WPML and WPML's Twig is too old for Timber). Another solution is for WPML to simply upgrade compatibility for its version of Twig to 1.38+; I'll let that team know about that resolution

PR #2023, opened 2019-06-11
2xtests

jarednova:
Update to PHPUnit 7.5, see if it makes things faster


jarednova:
So turns out, the upping the PHPUnit versions doesn't improve speed (at least, not measurably) — closing

Issue #2024, opened 2019-06-12
Timber 1.10.0 has a breaking change


joelambert:
Version `1.10.0` on Packagist includes this commit:

https://github.com/timber/timber/commit/111c70334113e3e1f6ea2121bd636a25ae93618c

This contains a breaking change to the constructor of the `MenuItem` class.

We noticed it as it causes an issue with [Lumberjack](https://lumberjack.rareloop.com) but will also break for anyone else using a custom subclass on the `MenuItem`.

jarednova:
Thanks so much @joelambert — I'll investigate a patch asap

palmiak:
This may also be the cause of menu error in other issue. I'm on mobile so
it's harder for me to check the issue ID but it has my latest response

śr., 12 cze 2019, 18:25 użytkownik Jared Novack <notifications@github.com>
napisał:

jarednova:
@palmiak I think it's #1988 ? Thanks that's helpful to know these are connected

palmiak:
I think. https://github.com/timber/timber/issues/1988#issuecomment-500921820
strange error with menu after update. Looks similar.

śr., 12 cze 2019, 18:33 użytkownik Jared Novack <notifications@github.com>
napisał:

jarednova:
This is closed by #2025. However, I don't believe that #1988 is actually connected to this issue (since that reports the issue began in 1.9.4, not 1.10.0)

joelambert:
Thanks for fixing this so quickly!

jarednova:
you bet!

Issue #1988, opened 2019-05-06
Illegal offset type in isset or empty in taxonomy.php when using Timber\Menu


yingban:
Hi,

I am using Timber(1.9.4) installed with Composer, in a Wordpress project version 5.1.1, with ACF plugin last version.

In my recent projects, I got a Wordpress warning which is:
"Warning:  Illegal offset type in isset or empty in wp-includes/taxonomy.php" in a custom post type content with a custom taxonomy.

When I comment the line in the foreach loop to retrieve all the menus, the warning disappears

   ` foreach (get_registered_nav_menus() as $key => $value) {
$context[$key] = new Timber\Menu( $key );
}`

Does somebody have the same issue? 

Thanks!

palmiak:
Hi

Does this problem also exists in new versions of Timber? There were some changes with the menu so maybe it was fixed.

Also do you have any menus created?

And finally I can't find the timber-menus file? Are you sure about the version that you are using?

yingban:
Sorry palmiak, I edited my comment.

I updated timber to version 1.9.4.
We use a foreach loop to retrieve all menus, and add each menu to the global Timber context.

yingban:
I updated Timber to version 1.10.

When I add menu to the global context like below 
`foreach (get_registered_nav_menus() as $key => $value) {
		$context[$key] = new Timber\Menu( $key );
	}`

in some page, it throws an error from the wp core taxonomy.
If I comment this line 
`$context[$key] = new Timber\Menu( $key );`
there is no more error.

palmiak:
Fist of all - could you paste the error code here?

Could you show me what will this code output:
```
var_dump( get_registered_nav_menus() );
```

It will should show you all the `$keys`. For example let's say it will output:
- menu_1
- menu_2
- menu_3

next define all those menu manually - without the loop:
```
$context['menu_1'] = new Timber\Menu( 'menu_1' );
$context['menu_2'] = new Timber\Menu( 'menu_2' );
$context['menu_3'] = new Timber\Menu( 'menu_3' );
```

next go to the page where the error appears and start commenting all the menus one after another to see which one is causing the error.

yingban:
Hi Palmiak, sorry for my late reply, just came back form holidays.

Here is the error
`<b>Warning</b>:  Illegal offset type in isset or empty in <b>/wp-includes/taxonomy.php</b> on line <b>294</b><br />`

If I dump the $taxonomy variable in WP function taxonomy_exists, I got this
[...]

PR #2025, opened 2019-06-12
ref #2024 -- patch for custom MenuItems

jarednova:
**Ticket**: #2024 #1988 

## Issue
Inadvertent breaking change introduced by new `Timber\MenuItem` constructor definition in `1.10`

## Solution
Set a default `null` value for the menu

## Impact
Filtering CSS classes (and other things that a user might do with menu info) won't work

## Usage Changes
None

## Considerations
Probably something that we don't need to bring into 2.x

## Testing
No new tests

### DIFF
--- a/lib/MenuItem.php
+++ b/lib/MenuItem.php
@@ -79,7 +79,7 @@ class MenuItem extends Core implements CoreInterface {
 	 * @param array|object $data
 	 * @param \Timber\Menu $menu The `Timber\Menu` object the menu item is associated with.
 	 */
-	public function __construct( $data, $menu ) {
+	public function __construct( $data, $menu = null ) {
 		$this->menu = $menu;
 		$data       = (object) $data;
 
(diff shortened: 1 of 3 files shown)


gchtr:
Yeah, I totally missed that the second parameter should probably be optional.

> Probably something that we don't need to bring into 2.x

I’m not sure about this. I – without thinking too much about it – assumed that a `MenuItem` always only exists in combination with a `Menu` object. But when we have a public `MenuItem` class, shouldn’t it be possible to create an instance without the knowledge of a parent `Menu` object? Maybe we need to keep this.

jarednova:
Fair point. And honestly easier to keep than selectively remove/isolate it

jarednova:
Wooohooo, tests pass! @palmiak or @gchtr could I get a quick review so I can merge?

jarednova:
@gchtr definitely. I had started a test locally that I can add to this branch

jarednova:
Test, added + merged!

PR #2026, opened 2019-06-12
Removed unused get_dimensions_* related code

jarednova:
## Issue
There seemed to be a few "dead" functions within `Timber\Image` that were non-public and unused.

## Solution
Remove them!

## Impact
If someone has extended the Image class in some way, they may be relying on those functions for a custom operation. If this is the case, the functionality is still easily replicated

## Usage Changes
None

## Considerations
We should deprecate these corresponding functions in 1.x

## Testing
No new tests. Image-related tests passed locally


lib/Image.php
```diff
@@ -108,55 +108,13 @@ public function __toString() {
 
 	/**
 	 * Get a PHP array with pathinfo() info from the file
+	 * @deprecated since 2.0 Functionality will no longer be supported in future releases.
 	 * @return array
 	 */
 	public function get_pathinfo() {
 		return pathinfo($this->file);
```
gchtr: Wouldn’t this need a call to `Helper::deprecated()`?

lib/Image.php
```diff
@@ -108,55 +108,13 @@ public function __toString() {
 
 	/**
 	 * Get a PHP array with pathinfo() info from the file
+	 * @deprecated since 2.0 Functionality will no longer be supported in future releases.
```
```suggestion
	 * @deprecated 2.0.0 Functionality will no longer be supported in future releases.
```

gchtr:
@jarednova Also see #2027 for what I mean when I say that the functions might have been accessible everywhere.

lib/Image.php
jarednova: Yep! will address in the next commit

jarednova:
@gchtr thanks for the heads-up on how these could have still been accessible. I'm good with an "abundance of caution" strategy and deprecating rather than removing. I've moved this into #2042 with that approach

PR #2027, opened 2019-06-12
[WIP] 2.x Improve magic methods in Core

gchtr:
**Ticket**: #478, #1904

## Issue

See #478. 

@acobster I realized I can create a pull request from the branch in your repository. I went ahead and did this, so we won’t forget about this and so that I can resolve the failing tests and maybe improve the documentation about this. Thanks for your work on this!

### Prevents infinite loops

Prevents infinite loops when:

- a `meta()` method wouldn’t exist.
- an inaccessible method is called.

By never calling `__get()` from `__call()`, but replicating the logic we need, we can circumvent an infinite loop.

### Removes possibility to access inaccessible properties.

The following check is removed from `__get()`:

```php
if ( property_exists( $this, $field ) ) {
    return $this->field;
}
```

With the check removed, we are not allowed anymore access properties that are normally inaccessible (protected or private properties). Inaccessible properties are inaccessible for a reason, that’s why we shouldn’t allow them to be accessible through magic methods. This will allow us to finish work in #1904.

## Impact

Possibly, there will be quite a few errors for developers who try to access inaccessible properties.

## Usage Changes

- **`{{ post.meta('my_field`) }}** is the _preferred_ way to get custom data
- **`{{ post.my_field }}`** is the _preferred_ way to get custom data that is modified by a custom method. However, if that method doesn't exist, Timber will fall back to the equivalent `{{ post.meta('my_field'} }}` call.

## Considerations

Should we add deprecation notices for anything here? I don’t even know if that would be possible.

## Testing

Still WIP.

### DIFF
--- a/lib/Core.php
+++ b/lib/Core.php
@@ -23,22 +23,59 @@ abstract class Core {
 	}
 
 	/**
-	 * This is helpful for twig to return properties and methods see: https://github.com/fabpot/Twig/issues/2
-	 * @return mixed
+	 * "Magic" method dispatcher for meta fields, for convience in Twig views.
+	 * Called when explicitly invoking non-existent methods on a Core object.
+	 * Not meant to be called directly.
+	 *
+	 * @example
+	 * ```php
+	 * $post = Timber\Post::get();
+	 * update_post_meta($post->id, 'favorite_zep4_track', 'Black Dog');
+	 * Timber::render('rock-n-roll.twig', array( 'post' => $post ));
+	 * ```
+	 * ```twig
+	 * {# Since this method does not exist explicitly on the Post class,
+	 *    it will dynamically dispatch the magic __call() method with an argument
+	 *    of "favorite_zep4_track" #}
+	 * <span>Favorite <i>Zeppelin IV</i> Track: {{ post.favorite_zep4_track() }}</span>
+	 * ```
+	 * @see https://secure.php.net/manual/en/language.oop5.overloading.php#object.call
+	 * @see: https://github.com/fabpot/Twig/issues/2
+	 * @param string the name of the method being called
+	 * @return mixed the value of the meta field named `$field`, if truthy;
+	 * false otherwise
 	 */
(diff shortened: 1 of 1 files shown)


gchtr:
@jarednova Currently, a lot of builds seem to be failing again, before tests are even starting. And not only in this pull request. Do you know why that happens?

jarednova:
Yeah, trying to figure out exactly that. Seems to affect only PRs (I restarted the current `master` build and it was successful). I checked and there's no change to the `.travis.yml`, `composer.json` or WP tests scripts. I'm gonna poke into the WordPress Slack group and see if anyone there is reporting similar

jarednova:
I posted something to the Travis community forum to see if someone from Travis could help:

https://travis-ci.community/t/prs-started-failing-on-6-12/3707

tldr: the problem is not within Timber's repo. It could be at the config level or something that changed in regards to newly submitted PRs after ~1pm EST on 6/12. I'm going to look into comparing the logs right now before succeeding and failing builds (it seems to be MYSQL related?) and then check out the configs for the GH --> Travis flow

jarednova:
Solved! I'm pretty proud of this one. It looks like Travis changed the default dist of Unbuntu from trusty to xenial. Since prior builds had already been submitted, Travis honored the prior build process with the old default. Not sure why, but it seems Xenial (Ubuntu 16.04) doesn't have the right perms for MySQL. Specifying Trusty did the trick!

https://docs.travis-ci.com/user/reference/overview/

I'll pull this change into master and the other failing PRs and we should be golden

jarednova:
@gchtr can you add `dist: trusty` to the `.travis.yml` file? That should trigger (hopefully successful) Travis tests

gchtr:
Cool thing that it’s working again 😎🎉. Thanks, Jared! Travis always seems like some kind of black box to me.

I realized I can’t add commits to this pull request, so I’ll add another one that build on this one 😉.

Issue #2039, opened 2019-06-30
Install instructions with composer do not work if followed exactly


joepetrakovich:
## Expected behavior
I expect if I use:

`composer require timber/timber`

that Timber will be added to my plugins directory.  The setup instructions suggest that if using the starter theme, that I will be able to activate Timber via wp-admin/plugins.php, but I cannot because it is not there.

## Actual behavior
composer only installs files to vendor/ but the starter-theme is not able to be activated.

## Steps to reproduce behavior
run composer require timber/timber
notice that it simply pulls everything into the vendor/ directory
download the starter-theme and put it in the /themes directory
attempt to activate the theme via the admin panel
you'll get an error that Timber is not activated but there is no way to activate it.
 
## What version of WordPress, PHP and Timber are you using?
WordPress 5.2.2, PHP 7.2.9, Timber 1.10.1

## How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
`composer require timber/timber`

jarednova:
@joepetrakovich thanks for this issue. I can see how that could be confusing. When you said you followed the "setup instructions" can you point me to the particular URL? There are a few out there, so I want to make sure we review/edit the right one

joepetrakovich:
https://timber.github.io/docs/getting-started/setup/

jarednova:
Totally hear you @joepetrakovich. That's one of those things that we can get blind to because we've done it 100x times. I think you've got the right idea in your comment. Would you be up for taking a pass at the markdown and submitting a PR?

https://github.com/timber/timber/blob/master/docs/getting-started/setup.md

That way, your name can live in infamy! (that is, the GitHub contributor statistics)

joepetrakovich:
Okay, yes let me test my theory and then I'll make the changes later today.

jarednova:
sounds good! Your basic idea there that the theme is the composer "project" is correct. Command line stuff like `composer require`... works similar to `npm install`

jarednova:
Closed via #2040

PR #2040, opened 2019-07-01
Add clarification to Getting Started docs

joepetrakovich:
Clarified which directory Composer should be used in.  Also added clarification for the scenario of using Composer with the starter theme instead of the WordPress.org install.

 

**Ticket**:  #2039 

## Issue

Trying to use both the Composer install instructions _and_ the starter theme instructions resulted in some confusion when it came to activating the Timber plugin.

## Solution

Added clarification for that specific use case.

## Impact

Possibly less headaches from new users, less support tickets.

## Considerations

It's odd that the starter theme has a composer.json file that is using an older Timber version and uses that package type that installs it as a plugin rather than the Composer autoload mechanism.  I'm not really sure the best solution.

### DIFF
--- a/docs/getting-started/setup.md
+++ b/docs/getting-started/setup.md
@@ -16,11 +16,21 @@ You can grab the all-things-included plugin at [WordPress.org](http://wordpress.
 
 The GitHub version of Timber requires [Composer](https://getcomposer.org/download/). If you'd prefer one-click installation, you should use the [WordPress.org](https://wordpress.org/plugins/timber-library/) version.
 
+Run the following Composer command from within your theme's root directory:
+
 ```shell
 composer require timber/timber
 ```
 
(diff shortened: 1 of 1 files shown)


palmiak:
I'm wondering shouldn't we just make a change in the starter theme so it would use timber as library not as a plugin.

Still the documentation here needs 3 variants plugin, lib via composer and plugin via composer. So the changes you made @joepetrakovich should still be included.

joepetrakovich:
You're welcome, and yes I also agree the starter theme should be consistent.

What about simply removing the `composer.json` from it?  Then we could remove some of the clarification I added here...  Unless it's better to have a locked Timber version to go with that theme.

jarednova:
I think you're on the right track @joepetrakovich. I still want to make the starter theme versitile to use in either setup. But the `composer.json` file should anticipate people using via Composer instead of as the WP plugin. Additionally, I added the block to load the Composer autoload so we can still reduce the number of explanation variants from 3 to 2.

Issue #2000, opened 2019-05-15
ACF custom menu fields not included in TimberMenu object


bramkorsten:
## Expected behavior
When I add custom fields to a Menu object (not the MenuItem object), I would expect to find these custom fields in `{{menu.field}}` or `menu.meta('field')`. For example, I've added a custom field that includes the logo in the menu for easy changing.

## Actual behavior
I can't find where these custom fields are stored. There seems to be no documentation on this, so researching is difficult. 

I'm using Wordpress 5.2 with ACF 5.7.9 and Timber 1.9.4

Thanks in advance for any help! If i can find the solution, I'd like to try and update the documentation for this!

EDIT: Issue #2000, what an amazing community ❤️

bobhuf:
Looking for an answer to this myself. These menu fields (not menu item fields) are stored in the wp_termmeta table. Per the docs:
> WP stores each Menu as a term object in the wp_terms table. ACF will store all custom field values in the wp_termmeta table.

[https://www.advancedcustomfields.com/resources/adding-fields-menus/](https://www.advancedcustomfields.com/resources/adding-fields-menus/)

jarednova:
@bramkorsten @bobhuf thanks for the notes on this — looking into it now!

jarednova:
@bramkorsten @bobhuf this might be a very simple fix. Check out #2043 and please let me know if that resolves the issue for you (it did in my tests)

bobhuf:
It does. Thanks, Jared!

bramkorsten:
Works like a charm! Looking good Jared 👌👌

PR #2043, opened 2019-07-04
Meta data for Menu object

jarednova:
**Ticket**: #2000

## Issue
Custom metadata stored via ACF on a WordPress menu was not accessible in templates

## Solution
Since WordPress borrows the meta table and object for its menus, we should do the same. This PR simply adds a test and has `Menu` extend `Term` rather than `Core`

## Impact
Seemingly very light. The class inheritance of the `Menu` item changes, but nothing else

## Usage Changes
None

## Considerations
A second thought/voice on whether this change in class inheritance could have larger negative consequences. This definitely feels like a very elegant solution, but a downside is that now `Menu` has some functions/properties with it that might not be meaningful or relevant to it being a `Menu` and not a `Term`

## Testing
Wrote a new test to validate the meta behavior; all other tests pass

### DIFF
--- a/lib/Menu.php
+++ b/lib/Menu.php
@@ -5,7 +5,7 @@ namespace Timber;
 use Timber\Core;
 use Timber\Post;
 
-class Menu extends Core {
+class Menu extends Term {
 
 	public $MenuItemClass = 'Timber\MenuItem';
 	public $PostClass = 'Timber\Post';
--- a/tests/test-timber-menu.php
+++ b/tests/test-timber-menu.php
@@ -415,13 +415,29 @@ class TestTimberMenu extends Timber_UnitTestCase {
 		$this->assertEquals( 'http://upstatement.com', $item->link() );
 	}
 
-	function testMenuMeta() {
+	function testMenuItemMetaAlt() {
+		$menu_info = $this->_createSimpleMenu();
+		$menu = new TimberMenu($menu_info['term_id']);
+		$item = $menu->items[0];
+		$this->assertEquals('molasses', $item->meta('flood'));
+	}
+
+	function testMenuItemMeta() {
 		self::_createTestMenu();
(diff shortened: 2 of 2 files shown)


gchtr:
It seems to me the biggest difference between `MenuItem` and `Term` is that terms can have public URLs and everything that comes with it: a link, an edit link, a description, children, assigned posts, etc.

Should we add some kind of base class that both `Term` and `MenuItem` can extend? We applied a similar pattern for the new [`Attachment` class](https://github.com/timber/timber/blob/2.x/lib/Attachment.php) that is used by `Image`. Some names I could think of:

- `CoreTerm`
- `TermBase`

What does @pascalknecht think? Does it make sense to use inheritance here? Or would we better use some other architectural pattern?

jarednova:
@gchtr just wanted to make sure you were aware that this is `Menu` and not `MenuItem`. That said, some of the things you flag (e.g. link, description) are good examples of where `Term` attributes have no relationship to `Menu`s

The idea of pulling things back into an underlying core class sound like a good approach to me. This would accomplish the goal of not duplicating functionality/code, but also not empowering `Menu` a big set of irrelevant (and therefore, likely problematic) methods/properties. That said, I'll let @pascalknecht or another contrib weigh-in before I start rewriting a big chunk

gchtr:
@jarednova Oh no, I was actually looking at `MenuItem`. But I guess my thinking still applies, even if it’s for the `Menu` class 😉.

jarednova:
I've been working on this a bit and to make it happen we'll have to pull in all the meta stuff into a new `CoreTerm` class. Since this has been extensively re-worked in 2.x I see a bunch of merge conflicts likely resulting. Rather than do a 1.x rewrite here. I think we should just pull this in as-is and then turn attention to a better `CoreTerm` plan for 2.x

Issue #399, opened 2014-11-30
Clear cache programmatically


kuus:
Is there a way to flush the timber cache programmatically?

mgmartel:
Hi @kuus, that is correct. Just a bit of background: 

## `$loader->clear_cache_twig();`
When you set `Timber::$cache` to `true`, all `twig` files will be compiled to `php` files and saved. That saves you some processor cycles on page views. However, this means that changes to your `twig` templates are not visible, unless you clear the cache. That is `$loader->clear_cache_twig();`.

This cache described [here](https://github.com/jarednova/timber/wiki/Performance#cache-the-twig-file-but-not-the-data).

## `$loader->clear_cache_timber( $cache_mode );`
This clears [this cache](https://github.com/jarednova/timber/wiki/Performance#cache-the-entire-twig-file-and-data). Timber's "cache everything" mode. Make sure you pass the cache mode you are actually using as an argument into this function!

If you use a persistent object cache like `memcached`, this will not clear your cache. I remember having some luck with [Johnny Cache](https://wordpress.org/plugins/johnny-cache/) to do this. Make sure you flush the `timberloader` group. Alternatively, purge the whole cache and you're also done!

## Partial cache
The partial cache ([this one](https://github.com/jarednova/timber/wiki/Performance#cache-parts-of-the-twig-file-and-data)) is not cleared by either method. It stores everything in the `timber` cache group by default, so to invalidate it, you have to do it manually. With Memcache see my note above about Johnny Cache. 

-- Mike

kuus:
@mgmartel thanks for the insights mike!

connorjburton:
This seems to have been covered pretty well so I am going to close this now. Feel free to comment if it's still an issue.

adskinner:
Where do you add "$loader->clear_cache_twig();" to clear the cache? in the functions.php file?

asecondwill:
@adskinner  yes, in the functions.php

```
$loader = new TimberLoader();
$loader->clear_cache_timber();
```

I've just come accross an issue where templates are not updating in staging for users not logged in as admins.  This clears that.

danburzo:
Hi everyone,

I'm going to piggyback on this issue for context. I've been reading the [Timber Docs page for Performance](https://timber.github.io/docs/guides/performance/) and tried the Twig caching as described:

```php
Timber::$cache = true;
```

[...]

PR #2046, opened 2019-07-18
Add details regarding Timber::cache to docs

danburzo:
Added some documentation around `Timber::$cache` based on [this discussion](https://github.com/timber/timber/issues/399#issuecomment-512699932).

I was not sure whether to include a more comprehensive solution to clearing the cache on a live website. I was thinking along the lines of:

```php
<?php
Routes::map('maintenance/clear-twig-cache', function() {
	if (is_user_logged_in()) {
		$loader = new Timber\Loader();
		$loader->clear_cache_twig();
		echo 'Twig cache cleared.';
		exit(0);
	}
});
```

...but since I'm not fluent in PHP / WP conventions, I'm not sure whether this is an idiomatic approach to the problem.

### DIFF
--- a/docs/guides/performance.md
+++ b/docs/guides/performance.md
@@ -103,7 +103,7 @@ if ( class_exists( 'Timber' ) ){
 }
 ```
 
-You can look in your your `/wp-content/plugins/timber/cache/twig` directory to see what these files look like.
+You can look in your `/wp-content/plugins/timber/cache/twig` directory to see what these files look like.
 
 This does not cache the _contents_ of the variables. 
 
--- a/docs/guides/performance.md
+++ b/docs/guides/performance.md
@@ -103,9 +103,21 @@ if ( class_exists( 'Timber' ) ){
 }
 ```
 
-You can look in your your `/wp-content/plugins/timber/twig-cache` directory to see what these files look like.
+You can look in your your `/wp-content/plugins/timber/cache/twig` directory to see what these files look like.
 
-This does not cache the _contents_ of the variables. This is recommended as a last-step in the production process. Once enabled, any change you make to a `.twig` file (just tweaking the HTML for example) will not go live until the cache is flushed.
+This does not cache the _contents_ of the variables. 
+
+Enabling `Timber::$cache` works best as a last step in the production process. Once enabled, any change you make to a `.twig` file (just tweaking the HTML for example ) will not go live until the cache is flushed. 
+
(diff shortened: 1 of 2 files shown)


jarednova:
@danburzo thanks so much for following up on #399 and turning it into a solution for others. 

The code you posted above feels totally workable, but doesn't really fit with other WP conventions (I feel like in general, you see an admin options page for something like that). That said, it's still totally someone could implement without any big downside

jarednova:
I'll pass the final review here to @gchtr since he's code owner on the docs. Thanks for the correction and further detail you provided

docs/guides/performance.md
```diff
@@ -103,9 +103,21 @@ if ( class_exists( 'Timber' ) ){
 }
 ```
 
-You can look in your your `/wp-content/plugins/timber/twig-cache` directory to see what these files look like.
+You can look in your your `/wp-content/plugins/timber/cache/twig` directory to see what these files look like.
 
-This does not cache the _contents_ of the variables. This is recommended as a last-step in the production process. Once enabled, any change you make to a `.twig` file (just tweaking the HTML for example) will not go live until the cache is flushed.
+This does not cache the _contents_ of the variables. 
+
+Enabling `Timber::$cache` works best as a last step in the production process. Once enabled, any change you make to a `.twig` file (just tweaking the HTML for example ) will not go live until the cache is flushed.
```
gchtr:
```suggestion
Enabling `Timber::$cache` works best as a last step in the production process. Once enabled, any change you make to a `.twig` file (just tweaking the HTML for example) will not go live until the cache is flushed. 
```

docs/guides/performance.md
```diff
@@ -103,9 +103,21 @@ if ( class_exists( 'Timber' ) ){
 }
 ```
 
-You can look in your your `/wp-content/plugins/timber/twig-cache` directory to see what these files look like.
+You can look in your your `/wp-content/plugins/timber/cache/twig` directory to see what these files look like.
```
gchtr:
```suggestion
You can look in your `/wp-content/plugins/timber/cache/twig` directory to see what these files look like.
```

danburzo:
What I ended up doing is include a meta box on the (ACF) Site Options page with an AJAX action co clear the cache.

gchtr:
@danburzo Yes, an ACF Options Page probably works just as well.

The documentation updates are now live. Thanks again for clarifying this! 😎

danburzo:
Thank you, my pleasure!

Issue #2049, opened 2019-07-24
Open theme properties to public


dtvn:
Check this dump
```
["theme"]=>
    object(Timber\Theme)#907 (10) {
...
      ["theme":"Timber\Theme":private]=>
      object(WP_Theme)#908 (12) {
        ["update"]=>
        bool(false)
        ["theme_root":"WP_Theme":private]=>
        string(52) "censored"
        ["headers":"WP_Theme":private]=>
        array(11) {
          ["Name"]=>
          string(6) "censored"
          ["ThemeURI"]=>
          string(34) "censored"
          ["Description"]=>
          string(43) "Bla bla"
          ["Author"]=>
          string(10) "This is what I need"
          ["AuthorURI"]=>
          string(20) "This is what I need"
...
```

I would like to change `["theme":"Timber\Theme":private]` to `["theme":"Timber\Theme":public]` so I can access to the Author and AuthorURI fields. Adding watermark is important so please open for public that field.

In the meantime, can you teach me how to extends the Theme class to print the Author field please?

dtvn:
A temporary solution is make one for ours, this is not optimized (in fact, it's ugly) but it can be used.

```PHP
// Put this in functions.php
function fixMetaTheme() {
	$theme = wp_get_theme();
	$meta = [
		'name' => $theme->get('Name'),
		'meta' => [
			'author' => $theme->get('Author'),
			'author_uri' => $theme->get('AuthorURI')
		]
	];
	return $meta;
}
```

After that, I override the theme array like this: 
```PHP
$context['theme'] = fixMetaTheme();
```

From there, I could use in twig template: 
```Twig
This bloody awesome {{theme.name}} made by <a href='{{theme.meta.author_uri}}'>{{theme.meta.author}}</a>
```

PR #2051, opened 2019-07-25
Add meta property for Timber\Theme

dtvn:
Hello I would like to improve `Timber\Theme` object. 

## Issue
I need to get the theme author name but the current Timber could do that.
Detail: https://github.com/timber/timber/issues/2049

## Solution
Introduce new property for the [Theme](https://github.com/dtvn/timber/blob/master/lib/Theme.php).
```PHP
/**
* @api
* @var array return an array of theme information (ex: author, author URI)
*/
public $meta;
```

Collect data for `$meta` in `init` function:
```PHP
$this->meta = [
	'theme_uri' => $this->theme->get('ThemeURI'),
	'desc' => $this->theme->get('Description'),
	'author' => $this->theme->get('Author'),
	'author_uri' => $this->theme->get('AuthorURI')
];
```

## Impact
This code does no harm things or have great impact on the core. It just simply get more data from `wp_get_theme()`. Totally safe.

## Usage Changes
See **Solution**, that's all I got.

## Considerations
Please check my issue:  https://github.com/timber/timber/issues/2049

## Testing
Not yet... I don't even know how to do that :p

### DIFF
--- a/lib/Theme.php
+++ b/lib/Theme.php
@@ -58,7 +58,7 @@ class Theme extends Core {
 
 	/**
 	 * @api
-	 * @var array the slug of the theme (ex: `my-super-theme`)
+	 * @var array return an array of theme information (ex: author, author URI)
 	 */
 	public $meta;
 
--- a/lib/Theme.php
+++ b/lib/Theme.php
@@ -57,6 +57,12 @@ class Theme extends Core {
 	public $uri;
 
 	/**
+	 * @api
+	 * @var array the slug of the theme (ex: `my-super-theme`)
+	 */
+	public $meta;
+
+	/**
 	 * @var WP_Theme the underlying WordPress native Theme object
 	 */
 	private $theme;
@@ -94,6 +100,13 @@ class Theme extends Core {
 		$this->slug = $this->theme->get_stylesheet();
 
 		$this->uri = $this->theme->get_template_directory_uri();
+
+		$this->meta = [
+			'ThemeURI' => $this->theme->get('ThemeURI'),
+			'Desc' => $this->theme->get('Description'),
+			'Author' => $this->theme->get('Author'),
+			'AuthorURI' => $this->theme->get('AuthorURI')
+		];
 
 		if ( $this->theme->parent()) {
 			$this->parent_slug = $this->theme->parent()->get_stylesheet();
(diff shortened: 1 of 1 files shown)


dtvn:
This is the first time I ever work in git stuff. Tried to understand how the pull work but I still a bit trouble. Very sorry if it troubles everyone.

gchtr:
> This is the first time I ever work in git stuff. Tried to understand how the pull work but I still a bit trouble. Very sorry if it troubles everyone.

No worries! Thank you for taking the time to sit down and make your first pull request! I know, it can be confusing first, but we’ve all been there 😉. Usually, you would want to make a pull request against the `master` branch, so you have to select `master` as the base. I already changed that for you. Now it looks a bit cleaner.

I think making all the theme info accessible is a good idea! But I wouldn’t go with the name `$meta`, because in Timber, we reserve "meta" for everything that’s save in a meta database table.

And I wonder if we would need to load in all the data, or if we want to only make the `get()` and the `display()` functions accessible:

```php
namespace Timber;

class Theme extends Core {
    /**
     * Gets a raw, unformatted theme header.
     * 
     * @api 
     * @see \WP_Theme::get()
     * @example
     * ```twig
     * {{ theme.get('Version') }}
     * ```
     *
     * @param string $header Name of the theme header. Name, Description, Author, Version,
     *                       ThemeURI, AuthorURI, Status, Tags.
     *
     * @return false|string String on success, false on failure.
     */
    public function get( $header ) {
        return $this->theme->get( $header );
    }

    /**
     * Gets a theme header, formatted and translated for display.
     *
     * @api
     * @see \WP_Theme::display()
     * @example
     * ```twig
     * {{ theme.display('Description') }}
     * ```
     *
     * @param string $header Name of the theme header. Name, Description, Author, Version,
     *                       ThemeURI, AuthorURI, Status, Tags.
     *
     * @return false|string
     */
    public function display( $header ) {
        return $this->theme->display( $header );
    }
}
```

This way, we don’t have to duplicate all the available headers into the Timber class. What do you think?

dtvn:
Ok, sorry, I didn't search carefully enough. Just found out the `meta` keywords also has something to do with `Timber\Core` method. I'll take note this and leave an image here to remind myself:

image

And I totally agree with your code. It would increase process time and saving some memory space for server. So, should I rewrite the source and make a request again?

jarednova:
Thanks @dtvn! I added tests and notes to the readme in #2065 and just merged them in. This will be in the next release of Timber!

I really hope you'll keep contributing PRs like this. These tests were pretty easy to write, here's the full info on including tests from our docs: https://timber.github.io/docs/guides/testing/

PR #2053, opened 2019-08-06
Updated VVV testing instructions based on current VVV setup

jarednova:
In setting up a new computer, I discovered that VVV no longer does some WordPress testing stuff automatically. This PR updates with info on how to setup and configure the tests manually given the new scripts


jarednova:
great, thankya — I'll get the docs re-deployed with this added

PR #2062, opened 2019-08-27
Twig\Extension\CoreExtension::setEscaper() is deprecated

rubas:
[As of Twig 2.11, the Twig\Extension\CoreExtension::setEscaper() and Twig\Extension\CoreExtension::getEscapers() are deprecated. Use the same methods on Twig\Extension\EscaperExtension instead.
](https://twig.symfony.com/doc/2.x/deprecated.html#extensions)

PR #2069, opened 2019-09-22
Twig extension coreextension set escape deprecated

romainmenke:
see https://github.com/timber/timber/pull/2062

This adds a check to see if the new class exists in Twig


romainmenke:
wrong base branch, my bad :/

PR #2070, opened 2019-09-22
Twig\Extension\CoreExtension::setEscaper() is deprecated

romainmenke:
_let's try this from the proper base branch 🤦‍♂_ 

see : https://github.com/timber/timber/pull/2062

This adds on `class_exists( 'Twig\Extension\EscaperExtension' )`

[As of Twig 2.11, the Twig\Extension\CoreExtension::setEscaper() and Twig\Extension\CoreExtension::getEscapers() are deprecated. Use the same methods on Twig\Extension\EscaperExtension instead.](https://twig.symfony.com/doc/2.x/deprecated.html#extensions)

### DIFF
--- a/lib/Twig.php
+++ b/lib/Twig.php
@@ -297,20 +297,37 @@ class Twig {
 	 */
 	public function add_timber_escapers( $twig ) {
 
-		$twig->getExtension('Twig\Extension\CoreExtension')->setEscaper('esc_url', function( \Twig\Environment $env, $string ) {
-			return esc_url($string);
-		});
-		$twig->getExtension('Twig\Extension\CoreExtension')->setEscaper('wp_kses_post', function( \Twig\Environment $env, $string ) {
-			return wp_kses_post($string);
-		});
+		if ( class_exists( 'Twig\Extension\EscaperExtension' ) ) {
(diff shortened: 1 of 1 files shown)


jarednova:
This looks good to me @romainmenke! I had thought about getting a little more cleaver with assigning the class (`'Twig\Extension\EscaperExtension'` or `'Twig\Extension\CoreExtension'`) to a variable. While it seemed more graceful at first, I can see it being a pain to manage over time (i.e. if methods start to take different args, etc.). I think your solution is actually the simplest/clearest. @gchtr, what's your take?

romainmenke:
Reusing those makes sense. I also assigned the escaper extensions to a variable as 4 calls to get the same thing wasn't optimal either.

```
$escaperExtension = $twig->getExtension('Twig\Extension\EscaperExtension');
$escaperExtension->setEscaper('esc_url', $esc_url);
$escaperExtension->setEscaper('wp_kses_post', $esc_url);
$escaperExtension->setEscaper('esc_html', $esc_url);
$escaperExtension->setEscaper('esc_js', $esc_url);
```

romainmenke:
> However, I was merely using $esc_url as an example.

Yeah still early over here. Realised it just now :)

romainmenke:
@gchtr should be better now. Thanks for the fast feedback!

gchtr:
@romainmenke Congratulations on your first contribution to Timber 🍾 !

Issue #1133, opened 2016-08-01
Timber cannot resize images with Hebrew characters


chetzof:
Name any image `איתין-נוף-נוסף.jpg` , upload to wordpress and try to resize it 

```
{{ image|resize(100, 100) }}
```

The call returns original image, no resizing is done.
Not sure if the bug is OS-dependent, but is reproducible on any Linux OS.

palmiak:
@chetzof did you tried to investigate it further? We had a similar problem with Ukrainian letters, but didn't had time to go further with it.

From what I see there is a plugin https://wordpress.org/plugins/cyrlitera/ (for russian) or https://wordpress.org/plugins/sanitize-spanish-filenames/ (many languages) that fixes those problem - maybe it would be worth a while to looke deeper into them and use their funcionality to fix the problem.

danburzo:
It seems that by the time it reaches the WP image editor, the path to the image file has been transformed / sanitized, and the WP Image Editor can't find it (I'm getting 'Error loading <sanitized-name>' when enabling Timber logging). Following the code from resize() to the WP Image Editor should reveal where this happens, but I have not had the chance yet to look into it.

danburzo:
Okay, looked into it further and the calls to the `pathinfo()` and `basename()` PHP functions mangle the source file name, since they depend on the locale set on the server. In my particular setup, these functions mostly work with the (default?) 'C' locale, except [when they start with a special char](https://bugs.php.net/bug.php?id=77239), which gets stripped :/

Any chance there's a small library that does not depend on the locale to extract path segments?

danburzo:
Okay, I edited the ImageHelper class to use `wp_basename()` and copied over the implementation of `mb_pathinfo()` and the issues went away, so it's something worth considering I think!

gchtr:
Resolved in #2072.

PR #2072, opened 2019-09-24
Add PathHelper for utf8-safe pathinfo() / basename() functions

danburzo:
**Ticket**: #1133

## Issue

The stock `pathinfo()` / `basename()` PHP functions have trouble with Unicode characters. 

## Solution

Taking a cue from changes made to WordPress core — changing instances of `basename()` with [`wp_basename()`](https://developer.wordpress.org/reference/functions/wp_basename/) — I've implemented `pathinfo()` using the same `rawurlencode` / `rawurldecode` combo.

The functions are grouped in a `PathHelper` class.

## Impact

The changes mostly impact the Image / ImageHelper classes. 

## Usage Changes

No changes in usage. The only user-facing change is in the Image's `get_pathinfo` function, but it should be backwards compatible.

## Considerations

Nothing else I can think of. 

## Testing

Unit tests are not included — I'm having a hard time bootstrapping the tests on a macOS / MAMP setup.  Some paths that should work now:

* `path/to/איתין-נוף-נוסף.jpg` (original issue report)
* `©-Dániel-.jpg` (starts with special char)

`basename()` and the `basename` and `filename` parts of `partinfo()` should leave the name in place.


gchtr:
Hey @danburzo, thanks for this pull request! I tried to add tests, but unfortunately (or maybe, actually, fortunately 😉), this error doesn’t appear on MacOS (also mentioned in https://bugs.php.net/bug.php?id=77239).

I have only one question: Is there a reason that you copied the logic over to separate functions instead of using `wp_basename()` and `mb_pathinfo()`.

@jarednova Can you add tests that confirm the issue with the default functions as well as tests that confirm the solved issue? I guess if you use your VVV environment, it might work.

danburzo:
@gchtr thanks for looking into it. Stream of consciousness thought process below:

For `wp_basename` I thought it might be a dependency on a random function in WP, and since it's so small, I copied it over. (Code Sniffer nudged me to use `rawurlencode` instead of `urlencode` as the more robust alternative.)

For `mb_pathinfo`, it's a method in PHPMailer, a separate library. Definitely too random a dependency; copying over `mb_pathinfo`, I wondered if there are licensing implications to that. At which point — 🤔 — why not use the same rawurlencode trick on `pathinfo` as well?

And now for the lesson in the value of unit tests 😅. Not being able to set them up, I coded it out of pure intuition (at midnight); but a good night's sleep helped me realize that while the approach works for `basename`, where slashes / backslashes don't matter, for `pathinfo` they definitely do. Consequently, my "solution" should break on Windows, since it turns backslashes to slashes.

In conclusion, importing `mb_pathinfo` sounds like the way to go. Would you rather reference the method in PHPMailer, or make a copy in Timber?

gchtr:
@danburzo That makes total sense to me. I also saw that `wp_basename()` doesn’t use `rawurlencode()`. Since we’re already applying a new function, it’s good to have a more robust solution.

I missed that `mb_pathinfo()` is a part of PHPMailer. So then, it also makes sense to create our own function in Timber. I’m not sure if we can just copy that function over because of the licensing, but we can surely take the idea and rewrite it or find an implementation that doesn’t have a licensing problem?

jarednova:
Thanks y'all! I'm working on those tests right now

jarednova:
I wasn't able to get the tests to fail either, but I certainly take @danburzo at his word that the current implementation contains this error (and this patch is certainly cleaner). 

Because we're on a different license (MIT versus PHPMailer's more restrictive LGPL) we can't bring over exactly that function. But if you're saying @danburzo that you can write something to accomplish the same result, presto! 🎉 — that'd work for me

danburzo:
Okay, I _think_ with the separate backslash replacement for Windows, things should work as expected there as well. Unfortunately, I don't have a Windows machine where I can run some tests, and `pathinfo()` on Mac handles backslashes differently. In case anything breaks, we can opt out of this behavior with:

```php
if (strtoupper(substr(PHP_OS, 0, 3)) === 'WIN') {
  // return plain pathinfo();
} else {
  // custom implementation
}
```

(We can do it proactively now, or as a hotfix if the need arises)

It's worth noting at this point that my experience with PHP is patchy, and limited to building some websites with WordPress. 😅

jarednova:
Thanks so much for your work on this @danburzo! I think from here, we might want to consider moving some of the methods from `Timber/URLHelper` into this snazzy new `Timber/PathHelper` class — but that's a PR/issue for another time. Thanks again for the solution here. This will be a part of the next release (sometime next week)

danburzo:
My pleasure! Thanks :)

Issue #1957, opened 2019-03-26
Should we really deprecate Timber::get_posts in 2.0?


palmiak:
While reviewing https://github.com/timber/timber/pull/1813 there is a small difference between `get_posts()` and `WP_Query()`. `get_posts()` always gets three extra params:
- ignore_sticky_posts
- no_found_rows
- suppress filters

I see that `Timber::get_posts()` is deprecated in 2.0, but shouldn't we keep `get_posts()` in some way to keep better compatibility with WP? @gchtr what do you think? Or maybe I'm missing something - I don't know 2.0 to good yet :)

gchtr:
I totally missed what was going on in that pull request. Good question!

For 2.0, @jarednova and I decided to remove a lot of functionality that practically does the same thing, like `Timber::query_posts()`. The same goes for `Timber::get_posts()`.

Keeping `Timber::get_posts()` would mean that we would have to explain the differences between that function and `Timber\PostQuery` and I fear it would lead to more confusion. We would have to answer the following questions:

- **Should you use `get_posts()` or `WP_Query` when working with WordPress?** 
- **When should you use `get_posts()` and when should you use `WP_Query`?** What’s the benefit of using `get_posts()` over `WP_Query`?

An it boils down to this one: **Do we use modern WordPress functionality or do we want to keep all the backwards compatibility that’s also baked into WordPress?**

It seems to me that `get_posts()` is a rather old function and that `WP_Query` is the preferred method to get posts. Are the defaults for the three mentioned parameters the only difference? If yes, I think it would make even more sense to drop `Timber::get_posts()` in order to prevent the confusion that could come with it. The documentation for the query parameters is way more detailed for [WP_Query](https://developer.wordpress.org/reference/classes/wp_query/) than for [get_posts()](https://developer.wordpress.org/reference/functions/get_posts/). I would prefer only a single way to fetch posts through Timber. How fetching posts works exactly can be answered better when we figure out how Object Factories work. They decide how the final API in 2.0 will look like.

palmiak:
> Should you use get_posts() or WP_Query when working with WordPress?
> When should you use get_posts() and when should you use WP_Query? What’s the benefit of using 
> get_posts() over WP_Query?

[...]

PR #1793, opened 2018-09-18
WIP: Feature/interfaced object factories

pascalknecht:
Merge of #1220. Feature will be developed further here.

PR #2073, opened 2019-09-25
2.x Documentation for API

gchtr:
**Tickets:** #1219, #1793.

@pascalknecht and I thought about what would be the most consistent way to go forward with Object Factories and the changes that come to the API. Here’s our proposal. As discussed in #1219, we’re starting with the documentation first.

## How to navigate this proposal

1. Read the new documentation for [Posts](https://github.com/timber/timber/blob/2.x-docs-api/docs/v2/guides/posts.md), [Terms](https://github.com/timber/timber/blob/2.x-docs-api/docs/v2/guides/terms.md), [Class Maps](https://github.com/timber/timber/blob/2.x-docs-api/docs/v2/guides/class-maps.md) and the updated documentation for [the Context](https://github.com/timber/timber/blob/2.x-docs-api/docs/v2/guides/context.md) (see [commit for changes](https://github.com/timber/timber/commit/d0f1740252fa8f4463ce0a16b1131d1ea609682a)).
2. Read the updates to the [Upgrade Guide](https://github.com/timber/timber/blob/2.x-docs-api/docs/v2/upgrade-guides/2.0.md) (see [commit for changes](https://github.com/timber/timber/commit/88d17edab5a2722bb993ee0bfb3c7e43a53d67ce#diff-f364d4d1c8295242b57f2ef81321d549)).
3. Continue reading this pull request.

## The API

We’ve found that the new API could evolve around **Class Maps**. Class Maps define what objects will be used for getting single objects as well as collections of objects. We already have them in Timber, but we don’t use them in all the places yet.

When we only have Class Maps, we can use them as a central hub to set what classes are used in Timber. A single source of truth, so to say. We could remove the existing possibilities that we currently have in certain functions to pass in the class that should be used, like for `Timber::get_posts()` and `Timber\PostQuery()`.

@aj-dl proposed to **always be explicit** when instantiating new posts in https://github.com/timber/timber/pull/1219#issuecomment-503400807. We think that is a good idea in general, at least better than always getting something from a `Post` class, which then might not necessarily be a `Post`, but a child class of `Post`.

However, by being explicit, we could only be explicit with objects that can be instantiated directly. But for all other functionality like `$post->thumbnail()` where the instantiation is hidden in a function, we would still have to rely on the Class Map. It still wouldn’t be clear when exactly the classes we defined in the Class Map would be used.

I myself have recently found that when I had to define the class that should be used in several places, I ran into a couple of cases where I was confused why something didn’t work, until I realized that I didn’t pass the class to be used in a function parameter. If we’d have Class Maps only, we wouldn’t have to think about instantiating objects with the proper class all the time.

Because of @aj-dl’s comments, we explored the different options we had for the API and thought about the following possibilities:

- `PostFactory::get()` and `PostQueryFactory::get()` – Like I commented in other issues before, I would find it great if we wouldn’t introduce the concept of factories into the API, but keep it hidden behind the scenes. Developers should be able to start with Timber without having to understand the concept of factories.
- `Post::get()` and `PostQuery::get()` – Here we would clearly state what we get, but there could be confusion about the naming. And we would probably have to rename the `Timber\Post` class to something like `Timber\BasePost` and make `get()` the only method in `Post`.
- `Timber::post()` or `Timber::posts()` – Here it might not be clear enough that you *get* something. It could also be understood as *posting something*.

Which leads us to the **proposed API**. We think we should use the functions we already have:

- `Timber::get_post()`  
- `Timber::get_posts()`
- `Timber::get_term()`
- `Timber::get_terms()`

This is basically what @acobster already proposed in https://github.com/timber/timber/pull/1219#issuecomment-435624322. All of these functions could rely on Class Maps. The **benefits of this API** would be:

- 👍 Developers can mostly use the API they already know. It’s really close to how WordPress works. We will make usage stricter though, with some deprecations.
- 👍 By using Class Maps, you define what custom classes are used and when in a single place. If you want to use a new method in Twig and extend Timber with a new class, you only need to update the Class Map and update your Twig file, but not every instantiation of where you need to use that class.
- 👍 By defining all public functions in the `Timber` class, developers get a nice overview of the API when looking at the code or the online documentation.
- 👍 [Automatic template contexts]() become even easier. There’s no need to manually call `setup()` when using a custom class, because the context also uses a Class Map.

The **drawbacks**:

- 😔 With Class Maps, you don’t see right in the code what class you’re dealing with and you can’t make use of navigating to that class in your IDE. Syntax highlighting might not work properly without adding custom DocBlocks with `@var` definitions.

Do you see any other benefits or drawbacks? We’d be pleased to hear them.

## Factories

For the factories implementation, I think we’re still on track in https://github.com/timber/timber/pull/1793 and can continue discussing and working out the details about the factories themselves there.

This pull request can be used for the changes to the documentation (without documentation in DocBlocks).

## Proposed Changes

When we stay with the proposed API, we can target to release Timber 2.0 sooner, because the only thing that’s holding us back is how the public API will look like. There are changes that need to be made for 2.0 and then there are changes that we can make later, in 2.x.

### Changes for 2.0

For 2.0, we could implement updates to the API for post objects only.

- [x] Add option to use callback functions in Class Maps. #2093 
- [x] Add Class Maps for Terms, Users and Comments. #2093 
- [x] Always use an array instead of a string as the default argument passed to the `timber/post/classmap` filter. #2093
- [x] <del>Deprecate</del> Update the API to _prohibit_ passing in the class to use for instantation in all post functions and use the Post Class Map as the single source of truth. #2090
- [x] <del>Deprecate</del> Update the API to _prohibit_ passing in query strings to `Timber::get_posts()`, like here: `Timber::get_posts( 'post_type=article' );` #2090
- [x] <del>Deprecate</del> Update the API to _prohibit_ passing in the class to use for instantation in all functions and use the new Class Maps as the single source of truth. #2090 
- [x] Implement `Timber::get_menu()`, `Timber::get_user()`, and `Timber::get_comment()` methods. #2089, #2197
- [x] Replace all direct instantiations in core and tests with the corresponding `Timber::get_*` calls. #2094
- [x] <del>Throw a *doing-it-wrong* exception when a `Post` [`Term`, `Menu`, `Comment` or `User` ] or `PostQuery` object is instantiated directly. (Is that even possible?)</del> Make `Post`, `Term`, `Comment`, and `User` constructors protected. #2088 ( `Menu` and `MenuItem` constructors will be made protected in #2292)
- [x] Add `to_array()` to Collection classes so that we can use something like `Timber::get_posts( $query )->to_array()`.
- [x] Add a new User Guide.
- [ ] Deprecate `Timber\PostClassMap` filter and rename it to `timber/post/classmap`. Currently, it’s called `timber/post/post_class` in the 2.x branch. #2293
- [ ] Determine what happens when a `timber/*/classmap` filter callback returns something other than a `CoreInterface`. (Probably should throw an exception rather than just letting it `TypeError`.) - also #2293 
- [ ] Finalize type declarations #1863 
- [ ] Implement the new `PagesMenu` API. #2292
- [x] Implement `::get_image()` and `::get_attachment()` top-level methods and corresponding Twig functions. #2333 

## Backwards compatibility

With the proposed API, we would replace the existing API for `Timber::get_post()` and `Timber::get_posts()`. We will have to consider what happens with the differences between `get_posts()` and `WP_Query` as discussed in https://github.com/timber/timber/issues/1957.

## Open Questions/Discussion

@jarednova, @pascalknecht, @acobster, @palmiak, @aj-adl: I’d be happy to get your opinions on this pull request. Some more questions came up while working on this pull request:

- ~~I know Inheritance is not always the best option for Object Oriented Programming. Does anybody see any problems with class conflicts?~~
- ~~Are callback functions in the Class Maps enough to have advanced control over what class is used when?~~
- For the API in Twig, we still have `Post`, `PostQuery`, etc. For consistency, should we consider renaming these functions to `get_post()`, `get_posts()`, etc?
- ~~Should we use `Timber::get_posts()` or `Timber::query_posts()`? Would it be more clear that `query_posts()` returns a query/collection of posts, and not an array of posts?~~

## Todo

- [x] Update Context Guide with new function signatures.
- [x] Document `no_found_rows`, see https://github.com/timber/timber/pull/2073#issuecomment-535720949. → Documented in https://github.com/timber/timber/blob/2.x-docs-api/docs/v2/guides/posts.md#count-rows-only-if-needed
- [x] Document `Timber::get_posts( $query )->to_array()`, see https://github.com/timber/timber/pull/2073#issuecomment-571375271.
- [x] Document difference between post arrays and PostCollections. → Documented in https://github.com/timber/timber/pull/2073/commits/508aa9efee730798288592a7d67becec20b84c0d
- [x] Optimize API for Menu Class Maps #2247.
- [x] Document JSON serialization of posts #2311. (#2335)
- [x] Update Upgrade Guide with updates added in #2311. Done in https://github.com/timber/timber/pull/2073/commits/82b6a6abb40f614e6821714f10b82d75042a6185.
- [x] Update Upgrade Guide with changes from THE PURGE in #2354, including the removal of the `timber/get_posts/mirror_wp_get_posts` filter. Done in #2367.


acobster:
@gchtr and @pascalknecht , thanks so much for putting this together! This is really helping me focus on shipping some good code.

Let me first just say that I like what I'm seeing. Keeping the API simple and small is definitely the way to go here, and I like that a Class Map can be either an array or a function. I think that'll keep things really simple for most use cases, yet endlessly extensible for more advanced usage. Just a few clarifying points before I dive in...

I think we should make the `Timber::get_posts()` usage change more explicit, and provide before/after code block examples ([Lando's upgrade guide](https://docs.lando.dev/guides/updating-to-rc2.html#overrides) does a really good job of this IMO).

> Throw a _doing-it-wrong_ exception when a Post or PostQuery object is instantiated directly. (Is that even possible?)

I don't think it's possible to detect who's calling a constructor (or any other method) without getting rrrreally hacky. :) But, we can _guarantee_ it will never be called directly by making `__construct()` protected:

```php
class Thing {
  public static function build() {
    return new static();
  }

  protected __construct() {
    $this->stuff = 'asdf';
  }

  public function get_stuff() {
    return $this->stuff;
  }
}

Thing::build()->get_stuff(); // -> "asdf"
```

This is a pretty common approach when implementing the Factory/Builder patterns in other OO languages, and I think it makes sense here. However, it does force our hand in terms of the 2.0 vs. 2.x rollout, since it would mean a _breaking_ change instead of a deprecation we can make down the line. That's unfortunate from a timing standpoint because it means we have to do more work up front, but I think it's worth doing that anyway. This will force us to be consistent across the board at 2.0, and I don't really see an alternative.

> I know Inheritance is not always the best option for Object Oriented Programming. Does anybody see any problems with class conflicts?

What do you mean by class conflicts? Are you talking about composition over inheritance? Are you worried about problems with inheritance in core Timber code, in user code, or both?

> For the API in Twig, we still have Post, PostQuery, etc. For consistency, should we consider renaming these functions to get_post(), get_posts(), etc?

I think renaming and deprecating the old Twig functions is probably a good idea. Seeing "Use `Timber::get_posts()` instead of `new Timber\PostQuery()`" and then a paragraph later seeing "You can also use PostQuery in Twig" is a little jarring, and it's one less thing for new users to have to remember.

> Should we use Timber::get_posts() or Timber::query_posts()? Would it be more clear that query_posts() returns a query/collection of posts, and not an array of posts?

I think we should keep the API as close as possible, and not rename the method. I'm definitely not dogmatic about backwards-compatibility, but in this case we're talking about one of the single most important methods in the entire library. At the very least, we should keep the name the same.

From what I can tell, the idea for the `'query' => [ ... ]` syntax got introduced in #1957 as a way to distinguish between actual `WP_Query` args and Timber-specific options like `post_class`. Do I have that right?

Assuming that's the case, I'm not a huge fan of changing the structure of this first argument, because it means users will have to update _every single call_ to `Timber::get_posts`. On the other hand, if we keep the first argument the same, only calls using a higher arity need to change. I imagine, based on my own code and my colleagues', that this is a significant portion of method calls currently in production, probably most of them.

Here's the API I propose instead:

```php
public function get_posts( array $query, array $options = [] );
```

This would allow us to extend the API in the future without changing the signature, and support whatever `$options` we want ('merge_default`, `pagination`, etc.) out of the box with 2.0. And if I'm understanding correctly, as far as `no_found_rows` and friends go, this lets us keep the spirit of [your note about performance](https://github.com/timber/timber/blob/2.x/docs/guides/context.md#performance), except `no_found_rows` etc. would just be part of the normal `$query` arg.

I'm sure I'll have more comments and questions as I move along but I feel like this is off to a good start. Thanks again.

gchtr:
Thanks for your deep dive into the proposal!

> I think we should make the `Timber::get_posts()` usage change more explicit, and provide before/after code block examples ([Lando's upgrade guide](https://docs.lando.dev/guides/updating-to-rc2.html#overrides) does a really good job of this IMO).

I agree, we could improve that in the documentation 👍. I’m adding this to the list.

> I don't think it's possible to detect who's calling a constructor (or any other method) without getting rrrreally hacky. :) But, we can _guarantee_ it will never be called directly by making `__construct()` protected:

I see, that’s the way to go. Making the constructor protected would work. And yes, we would have to do that in 2.0, because it would be a breaking change.

> > I know Inheritance is not always the best option for Object Oriented Programming. Does anybody see any problems with class conflicts?
> 
> What do you mean by class conflicts? Are you talking about composition over inheritance? Are you worried about problems with inheritance in core Timber code, in user code, or both?

Yes, exactly, I mean composition over inheritance. But probably not in Timber core, mostly in user code, when class maps are used to define which class is used when. For now, I don’t see any direct problem, but maybe there are warning signs that I don’t know of? You certainly have a better grasp of these patterns ;).

> I think we should keep the API as close as possible, and not rename the method. I'm definitely not dogmatic about backwards-compatibility, but in this case we're talking about one of the single most important methods in the entire library. At the very least, we should keep the name the same.

Well, of course! I totally forgot again about the simplest reason why it should be `Timber::get_post()`: because there’s `get_post()` 🙈. Must have been late 😄.

> From what I can tell, the idea for the `'query' => [ ... ]` syntax got introduced in #1957 as a way to distinguish between actual `WP_Query` args and Timber-specific options like `post_class`. Do I have that right?

Yes, you got that right!

> Assuming that's the case, I'm not a huge fan of changing the structure of this first argument, because it means users will have to update _every single call_ to `Timber::get_posts`.
>
> Here's the API I propose instead:
> 
> ```
> public function get_posts( array $query, array $options = [] );
> ```

Totally! Here it wouldn’t make much sense to make developers change every call to `Timber::get_posts()`. Let’s do it the way you proposed 👍. I can take care of the relevant inline documentation as well as the documentation in the guides, if you want me to.

acobster:
@gchtr sounds good! Yes, if you could update the docs that'd be great. I'll continue laying the groundwork for using Factories under the hood, everywhere. There a lot of direct instantiations in the tests...

>  For now, I don’t see any direct problem, but maybe there are warning signs that I don’t know of?

Since we're talking about "userland" code, I don't think there's a whole lot we can do in core to prevent the kinds of problems that composition-over-inheritance addresses. I say this because these kind of problems are rooted in poor software design. Aside from the premise that custom posts may be instances of different classes, Timber is pretty unopinionated about architecture, and I think that's a good thing.

I do think it would be a good idea _at some point_ to have a guide recommending using traits to accomplish composition over inheritance in tandem with class maps, since traits seem to be a somewhat underappreciated feature of PHP, especially in the WP world, in my experience. I'll write up a future issue for that.

gchtr:
@acobster Great, I’ll be updating the docs in this pull request, then.

> Since we're talking about "userland" code, I don't think there's a whole lot we can do in core to prevent the kinds of problems that composition-over-inheritance addresses. I say this because these kind of problems are rooted in poor software design. Aside from the premise that custom posts may be instances of different classes, Timber is pretty unopinionated about architecture, and I think that's a good thing.
> 
> I do think it would be a good idea _at some point_ to have a guide recommending using traits to accomplish composition over inheritance in tandem with class maps, since traits seem to be a somewhat underappreciated feature of PHP, especially in the WP world, in my experience. I'll write up a future issue for that.

Alright! I’d say we leave it at *"some point"*. I think there’s no need to add documentation for problems that might arise at some point in the future. I’m never against an additional guide that shows some advanced usage pattern, though 😉.

acobster:
@gchtr I made a bunch of updates to the description per our discussion, and created some new issues that are big enough to be scoped and tracked separately. Please review. I'll continue working on the internal Factory API for now, and circle back to tying it in with the larger API changes. Thanks!

acobster:
@pascalknecht after the discussion on this ticket, I think that our best route here is, unfortunately, to close #1219 and #1793 and start on Factories fresh, with the new API proposal as the motivating goal. There are simply more moving pieces, many of them coupled to the old API (or to a [new API](https://github.com/timber/timber/pull/2073) that we're not going to use), than I can wrap my head around right now. I could try to refactor #1219 to match the new requirements, but I think that would take me about twice as long as just starting fresh and pulling in what I need to accomplish the end goal of this API. Unless you have the bandwidth to take on this major refactor of a completely new API (which also currently lacks tests), I'm going to go ahead and close out these two PRs. Thanks.

jarednova:
@acobster @gchtr thank you for the thoughtful rundown of the issues and outcomes. Between teh limitations of PHP, WordPress and naming conventions, there's unlikely to be a _perfect_ solution, but this one seems _damn good_ (at the very least). 

Now that this seems like the appropriate path, closing #1219 and #1793 is probably the right move. Even if the exact code from @pascalknecht and @joshlevinson doesn't make it in — each of these PRs developed some of the key ideas we're now implementing. 

Happy to continue contributing my two cents — but mostly I'll stay out of the way as you two keep hammering!!!!

acobster:
@jarednova @gchtr another handful of questions for the team:

1. Should `Timber::get_posts()` always return a `PostCollection`? Will we ever want to return the more specialized `PostQuery`? Is there a distinction between the two that users will find useful going foward?
2. The same method also sometimes returns a `QueryIterator`. Can this behavior be subsumed into `PostCollection` or `PostQuery` somehow?

docs/guides/class-maps.md
```diff
@@ -0,0 +1,192 @@
...
+
+add_filter( 'timber/comment/classmap', function() {
+    return [
+        'book' => function( \WP_Post $post ) {
+            if ( 0 !== $post->post_parent ) {
+                return BookChildComment::class;
+            }
+            
+            return BookComment::class;
+        },
```
acobster: @gchtr I think this callback should still get the `WP_Comment` object, rather than the associated post. Otherwise, if someone wants logic based on comment metadata or something, there's no way to do it because the comment ID is no longer in scope. If folks want logic based on the associated post, on the other hand, they can always just call `get_post($comment->comment_post_ID)`.

jarednova:
Here's my take ...

**Should Timber::get_posts() always return a PostCollection? Will we ever want to return the more specialized PostQuery? Is there a distinction between the two that users will find useful going forward?**
The different behavior of `get_posts` vs `PostQuery` was more of a historical accident than a deliberate choice. `PostCollections` are essentially fancy arrays that are agnostic of their origin. However, `PostQuery`s are fancy `PostCollections` that _are_ aware of their origin (the query). Why does this matter? Mainly pagination. A `PostCollection` of post ids 11-20 doesn't care that it's page 2 of a query that ultimately holds 500 posts (for example). However, the `PostQuery` does care. My take is that `PostQuery` is the class that most users would want to interact with and should be what is generally returned from `get_posts` and other methods.

**The same method also sometimes returns a QueryIterator. Can this behavior be subsumed into PostCollection or PostQuery somehow? In other words, can we just always return some kind of PostCollection, and have the Factory layer make sure that that collection is initialized properly?**
This one is a little trickier. As I recall/understand `QueryIterator`s are used to handle the requesting and iterating through these `PostQuery`s. I think this _could_ be subsumed into another class. I think the reason it's separate is to contain this piece of functionality into its own class.

It strikes me now that this is really a `_Post_QueryIterator`. If this is a pattern to continue, it might  provide a model for `UserQueryIterator`, `TermQueryIterator` ... However, none of these are really meant to be "public." If there's a better organization paradigm based on factories, we should probably go that route

acobster:
@jarednova alright...buckle up. :rocket: 

> My take is that PostQuery is the class that most users would want to interact with and should be what is generally returned from get_posts and other methods.

I agree.

> I think the reason it's separate is to contain this piece of functionality into its own class.

By "this piece of functionality" do you mean translating from a `WP_Query` to an iterable collection that is aware of pagination?

> If there's a better organization paradigm based on factories, we should probably go that route

After some thinking, I believe that the best approach here is to define a `PostQueryInterface`, and return a different concrete instance depending on whether we've been passed an array or a `WP_Query`. Here's the API I propose, based on what's currently in `PostQuery`, `PostCollection`, and `QueryIterator`:

```php
interface PostQueryInterface extends Traversable, Countable, ArrayAccess {
  public function pagination( array $options = [] );
}

class PostQueryIterator implements PostQueryInterface {
  public function pagination( array $option = [] ) {
    return new Pagination($options, $this->_query);
  }
  /* all the same traversal/counting methods currently in QueryInterface, plus ArrayAccess methods... */
}

class PostArrayObject extends ArrayObject {
  public function pagination( array $options = [] ) {
    return null;
  }
}
```

A couple things to note:

* I chose to extend those interfaces specifically because they are close to the lowest common denominator of `ArrayObject` (which `PostCollection` extends) and `Iterator`/`Countable` (which `QueryIterator` implements). Also because `ArrayAccess` is nice, and easy to implement. :)
* `PostArrayObject` already implements `Traversable` etc. by virtue of extending `ArrayObject`.
* A Factory can decide which class to instantiate based on arguments to `Timber::get_posts()` (or whatever other internal code is calling the Factory).

We should document `PostQueryInterface` and the classes that implement it. Maybe a future version of Timber can expose a filter for overriding the concrete instance of `PostQueryInterface` returned from `Timber::get_posts()`, but we don't need that right now.

### The long version

I hemmed and hawed over this issue for a while before coming to the above conclusion. In case anyone's interested, here's how I got there...

In its constructor, to get the array of posts that it will internally iterate over, `PostQuery` delegates to `PostGetter::query_posts()`, which in turn returns _either_ a new instance of `PostCollection` or an instance of `QueryIterator` depending on the type of array it's been passed. Lifting that decision of which thing to give back out of `PostGetter`, the crucial lines in `PostQuery::__construct()` can effectively be translated into:

```php
		if ( is_array($args['query']) && count($args['query']) && isset($args['query'][0]) && is_object($args['query'][0]) ) {
			// We have an array of post objects that already have data
			$this->queryIterator = new PostCollection($query, $args['post_class']);
		} else {
			// We have a query (of sorts) to work with
			$this->queryIterator = new QueryIterator($query, $args['post_class']);
		}

		$posts = $this->queryIterator->get_posts();

		parent::__construct( $posts, $args['post_class'] );
```

...meaning that `PostQuery` (a `PostCollection` itself) is delegating to `QueryIterator` or to another `PostCollection`. Smelly! :)

This complicates the claim that: 

> A `PostCollection` of post ids 11-20 doesn't care that it's page 2 of a query that ultimately holds 500 posts (for example). However, the `PostQuery` does care.

Where this gets really sticky is that `PostQuery` actually only returns a `Pagination` object IFF its internal `queryIterator` is a `Timber\QueryIterator` instance (and not a `PostCollection`, as we've seen it could well be):

```php
	public function pagination( $prefs = array() ) {
		if ( !$this->pagination && is_a($this->queryIterator, 'Timber\QueryIterator') ) {
			$this->pagination = $this->queryIterator->get_pagination($prefs, $this->get_query());
		}

		return $this->pagination;
	}
```

I don't know if there's some subtlety elsewhere in the code guaranteeing that we never actually are instantiating `PostQuery` in such a way that it will never delegate to a `PostCollection`, and therefore it will _always_ return a `Pagination` instance...but in any case that's just a stronger argument for clarifying this distinction.

All that is to say: it seems like what we're doing here abstracting away two different data sources (a primitive array of posts or a `WP_Query`) within the same class/constructor. So the question then is: do we really need to worry about always returning a concrete `PostQuery`? Or is our real concern here with the _interface_? The complicated delegation going on in the constructor seems to work toward unifying pagination and traversal into a single _class_, when we could make equally strong guarantees by returning different concrete classes that implement the same interface. Hence the above proposal.

jarednova:
My questions/points all feel answered. To the point of:

> Or is our real concern here with the interface?

Yes. I don't think we'll have anyone angry/care about whether it's a `PostQuery` or not. But what _does_ matter is the ability to easily traverse, iterate, etc. (all the things you know and love about arrays). If this can be handled with fewer classes, all the better. 

The proposal feels very reasonable and a sensible organization of the classes/interfaces we already have. Interested in @gchtr's interpretation on the above details as it relates to the factories component.

docs/guides/class-maps.md
gchtr: @acobster  Thanks, that’s a very good input! I updated the guide in https://github.com/timber/timber/pull/2073/commits/4012fd625d5b951efb741ef02ba009b4b3e032cd.

When calling the filter, do you think we already have the associated post and could pass it as a second parameter, or do we only have the post type of the associated post?

Would (and should) this be possible?

```php
'book' => function( \WP_Comment $comment, \WP_Post $post ) {

}
```

Or even:

```php
'book' => function( \WP_Comment $comment, \Timber\Post $post ) {

}
```

gchtr:
> @gchtr I made a bunch of updates to the description per our discussion, and created some new issues that are big enough to be scoped and tracked separately.

Cool, thanks @acobster 👍!

> 1. Should `Timber::get_posts()` always return a `PostCollection`? Will we ever want to return the more specialized `PostQuery`? Is there a distinction between the two that users will find useful going forward?

From answering a lot of questions here in the issues and on Stack Overflow, I’ve found that in the current state, it’s not very clear what the difference between a `PostQuery` and a `PostCollection` is. It’s more like: «Sometimes you have to call `get_posts()` on the result to get all the posts that you get from a query, and sometimes you don’t.» If we could unify this so that we can always use the same functionality no matter the class of the object we get, that would be great! 👍

As far as I understand it, I think your proposal makes sense. I suppose for developers it’s important they can:

1. Loop over the collection of posts.
2. Use a pagination for the posts.
3. Access additional information, like the amount of [found posts](https://github.com/timber/timber/pull/2074).

So I agree with Jared when he says:

> Yes. I don't think we'll have anyone angry/care about whether it's a `PostQuery` or not. But what _does_ matter is the ability to easily traverse, iterate, etc. (all the things you know and love about arrays). If this can be handled with fewer classes, all the better.

---

> We should document `PostCollectionInterface` and the classes that implement it.

Yes, I think that would help developers understand what’s going on 👍. Remember to ping me if I can help here!

> Maybe a future version of Timber can expose a filter for overriding the concrete instance of `PostCollectionInterface` returned from `Timber::get_posts()` (i.e. a Class Map filter), but we don't need that right now.

Yes, let’s wait until we actually need it, because it’s probably an edge case. The only case I ever needed to overwrite a class inside the post collection logic was a [custom PostIterator](https://github.com/mindkomm/timber-integration-woocommerce/blob/master/lib/Timber/Integrations/WooCommerce/ProductsIterator.php), but even that [won’t be needed with 2.0](https://github.com/mindkomm/timber-integration-woocommerce/issues/6#issuecomment-418721525).

acobster:
Awesome. Thanks for the responses, @gchtr and @jarednova. I'll follow up with more specifics later in the week (client projects are calling me back) but I think we're in really good shape for the last big push for this API change.

gchtr:
> I think we should make the `Timber::get_posts()` usage change more explicit, and provide before/after code block examples ([Lando's upgrade guide](https://docs.lando.dev/guides/updating-to-rc2.html#overrides) does a really good job of this IMO).

@acobster I updated the documentation for the function signature for `Timber::get_posts()`. And I updated the Upgrade Guide with *Old* and *New* examples for both `Timber::get_post()` and `Timber::get_posts()`, like you proposed.

gchtr:
@acobster @jarednova While working on the function signature `Timber::get_posts()`, I realized a couple of things that I think need discussing:

## 1. Getting posts as an array instead of a collection

> I'm not a huge fan of changing the structure of this first argument, because it means users will have to update _every single call_ to `Timber::get_posts`. On the other hand, if we keep the first argument the same, only calls using a higher arity need to change. I imagine, based on my own code and my colleagues', that this is a significant portion of method calls currently in production, probably most of them.
> 
> Here's the API I propose instead:
> 
> ```
> public function get_posts( array $query, array $options = [] );
> ```

In the current state of the API, we always return a `PostCollection` when using `Timber::get_posts()`. If I wanted to have the posts array, I’d always have to do the following:

```php
$posts_array = Timber::get_posts()->get_posts();
```

Having "get_posts" twice feels weird, when it’s written in one line, right? I wonder whether we could still improve that.

In 1.x, there is the third parameter `$return_collection` for `Timber::get_posts()`. Whenever that was set to `false`, it would directly return a posts array. For 2.x, I see the following options:

**1. Leave as is**

Leave as is and live with it.

**2. Use the `$option` parameter**

We could add a `return_type` option and use `collection` as a default. Whenever I wanted to return an array, I could use:

```php
Timber::get_posts( $query, [
    'return_type' => 'array',
] );
```

**3. Add a separate function**

We could add a new function `Timber::get_posts_array()` for this.

```php
public static function get_posts_array( array $query, array $options = [] ) {
	return self::get_posts( $query, $options )->get_posts();
}
```

The benefit of a different function would be proper type hinting. `Timber::get_posts()` would always have a return type of `PostCollection` and `Timber::get_posts_array()` would always return an `array`.

## 2. Should `Timber::get_posts()` accept a range of inputs?

There are a lot of tests that use an array of post IDs as a parameter for `Timber::get_posts()`. And there are tests that check whether it’s possible to to pass an array of WordPress post objects to `Timber::get_posts()`.

For the 2.x API, should it still be possible

- to pass an array of post IDs to `Timber::get_posts()`?
- to pass an array of WordPress post object to `Timber::get_posts()`? 

Or would they first have to be converted with another function? We already have a function for converting posts: `Post::convert()`. How would that fit in? How does it fit in with the new `PostCollection`/`PostCollectionInterface` setup?

## 3. Should we add functions to get posts by slug/name?

Getting a post by slug or name through `Timber::get_post()` will be deprecated. However, I think this could still be quite helpful, because I’ve always found it a hassle to work with `get_page_by_path()` and `get_page_by_title()`. I don’t think we should make it possible to use a post slug as a parameter for `Timber::get_post()`. But:

- Should we add a `Timber::get_post_by_slug()` method?
- Should we add a `Timber::get_post_by_name()` method?

jarednova:
@gchtr @acobster you both have been deeper in the weed here, so take this as my thoughts, not my commands here. I think @gchtr has uncovered some of my "original sins" from way back in 2013 with the earliest pieces in simplifying post retrieval to be more generous with whatever it got sent. Practical, perhaps, but also a good example of Timber trying to do a bit too much.

## 1. get_posts returns an array...
Option 2 feels the most in-keeping with other stuff we've done (arguments through an arguments array instead of additional arguments). Even if it's verbose, it's consistent. If a dev is going through the upgrade pain, I find that whether I have to copy-and-paste something that's short or slightly less short isn't really more burdensome. 

## 2. Should Timber::get_posts() accept a range of inputs?
The reason I think that this behavior should be preserved is that often ACF (or other custom field tools) can be inconsistent about how they return data that Timber later processes. A site can be 100% working but if someone clicks "return as post IDs" in the ACF interface, that part will crash (and cause a mad dash to figure out why). Because it's so easy to have a non-material data error I'm a proponent of keeping its flexibility with the type of input. Perhaps this is better organized within the class through so it always hits some internal `convert` method.

## 3. Should we add functions to get posts by slug/name?
However, this is one where I don't think `get_posts()` needs to do magic in handling slugs and titles. These are distinctly diff't types of data that WP/ACF doesn't intermix (i.e. you might sometimes get a post ID with `1` instead of a `WP_Post` object corresponding to "Hello World", but you'll never just get back a `"Hello World"` or `hello-world` (make sense?)). I don't think offering interfaces into this common use case is too blurry `Timber::get_post_by_slug() ` seems straight forward and highly useful. 

Again, just thoughts. I always reserve the right to be wrong

acobster:
@jarednova @gchtr thanks for bringing this up. These are good points as always.

## 1. Getting posts as an array instead of a collection

What if instead, `PostsCollection` and friends implemented a `to_array()` method? That's simpler and shorter than options 2 and 3:

```
Timber::get_posts( $query )->to_array();
```

In addition to providing that escape hatch, I was thinking we could annotate the return type of `Timber::get_posts()` to be [`Iterable`](https://www.php.net/manual/en/language.types.iterable.php):

> Iterable is a pseudo-type introduced in PHP 7.1. It accepts any array or object implementing the Traversable interface. Both of these types are iterable using foreach and can be used with yield from within a generator. 

Both `array` and `PostsCollection` fit under this umbrella. This would leave open the possibility of providing a `return_type` option in the future.

## 2. Should Timber::get_posts() accept a range of inputs?

Yes it should, and this is already supported in all factories! Here's an example [PostFactory test](https://github.com/timber/timber/blob/2.x-factories/tests/test-post-factory.php#L166) and the [corresponding clause](https://github.com/timber/timber/blob/2.x-factories/lib/Factory/PostFactory.php#L28) for dealing with this case. Note that `is_numeric_array()` only looks at keys, meaning it doesn't care if you mix and match values/types; it handles each value independently in a recursive call. It's similar to what `convert` does, but `convert` is coupled to `PostGetter` so I avoided it.

I agree with Jared about the philosophy here and I think this implementation strikes that balance.

## 3. Should we add functions to get posts by slug/name?

I agree with Jared again here. These seem like convenience wrappers around special cases of `Timber::get_posts()`, rather than something lower-level deserving of special handing _inside_ the method.

The term "name" is overloaded though, since "slug" is confusingly called `post_name` under the hood: we should go with `Timber::get_post_by_slug` and `Timber::get_post_by_title`.

# ALSO: Progress report

I also wanted to let y'all know I have started the effort to switch core constructors to protected and switch the public API to begin using Factories under the hood. I got most of the way through Comments today; there are a few failing tests on the `2.x-factories` branch that I'm not sure about:

```
phpunit --group comments-api
```

I merged `2.x-refactor-tests` back into `2.x-factories` and I'm going to be working off of that as the canonical branch for this overarching issue.

I think lots of failing tests will have to be the norm for a while as I switch over the API. Let's try to keep the feedback loop tight and use the test failures to guide as as to where updates need to happen. Sound good?

jarednova:
Completely agree on `Timber::get_post_by_title` — good call. Lemme see if I can replicate what you're seeing on those testing errors, there's a lot of version mixing right now that's often the source ...

acobster:
@jarednova I'm getting a warning about [this test](https://github.com/timber/timber/blob/2.x-factories/tests/test-timber-post-comments.php#L112) not performing any assertions. I'm guessing it should parse the rendered HTML and assert that the content appears as expected? Is that something you can knock out?

gchtr:
> What if instead, `PostsCollection` and friends implemented a `to_array()` method? That's simpler and shorter than options 2 and 3:
> 
> ```
> Timber::get_posts( $query )->to_array();
> ```
>
> In addition to providing that escape hatch, I was thinking we could annotate the return type of `Timber::get_posts()` to be [`Iterable`](https://www.php.net/manual/en/language.types.iterable.php):

I like that. Let’s do it this way 👍. I’m going to update the description of this pull request with related todos.

> The term "name" is overloaded though, since "slug" is confusingly called `post_name` under the hood: we should go with `Timber::get_post_by_slug` and `Timber::get_post_by_title`.

"Title" is indeed better than "name". Let’s do that. I openend a separate issue for this in #2152.

> I merged `2.x-refactor-tests` back into `2.x-factories` and I'm going to be working off of that as the canonical branch for this overarching issue.

Cool, I’m going to make my pull requests against `2.x-factories`, then.

acobster:
@gchtr @jarednova @pascalknecht here's another design question: how and where should we implement "current user/post/etc." behavior? For example, in 1.x calling `new Post()` infers the current post ID from global state, but it does so in the `determine_id` method, called from the constructor. But that's too late for a Factory to determine which class to call. Similar problems occur for `User` and `Term`.

My instinct is to just lift this logic up into `Timber::get_*` methods, and keep `Timber\Core` instance methods as decoupled from global state as possible. I remember hearing about someone (@pascalknecht?) working on something related to this (something about `Timber::context()`, maybe?) but I don't recall which issue. So I'm not sure if there's existing code I can use for this, or if I'll need to just move this logic around wholesale...thoughts?

gchtr:
> My instinct is to just lift this logic up into `Timber::get_*` methods, and keep `Timber\Core` instance methods as decoupled from global state as possible.

@acobster Here we talked about going stateless in Factories: https://github.com/timber/timber/pull/1793#issuecomment-443364338.

And in the current state of the context, we still rely on passing no arguments to get the global state:

https://github.com/timber/timber/blob/7c37b162ebc3d415a9e6004fa5ba86d2f93ffd5c/lib/Timber.php#L283-L294

I agree with you that this should be in the `Timber::get_*` methods.

gchtr:
> > What if instead, `PostsCollection` and friends implemented a `to_array()` method? That's simpler and shorter than options 2 and 3:
> > ```
> > Timber::get_posts( $query )->to_array();
> > ```
> > 
> > 
> > In addition to providing that escape hatch, I was thinking we could annotate the return type of `Timber::get_posts()` to be [`Iterable`](https://www.php.net/manual/en/language.types.iterable.php):
> 
> I like that. Let’s do it this way 👍. I’m going to update the description of this pull request with related todos.

I have another question… In #1778, we’ve reworked how the context works and how we can improve compatibility with third-party through [`Timber\PostIterator`](https://github.com/timber/timber/blob/2.x/lib/PostsIterator.php) by setting up and tearing down posts, calling WordPress core hooks and calling functions like `wp_reset_postdata()`.

Now, in the Context Guide I’ve written about [when to use `Post::setup()`](https://github.com/timber/timber/blob/2.x-docs-api/docs/guides/context.md#using-a-custom-post-class). The idea was to **not call `setup()`** every time a post is instantiated, because that’s exactly what lead to errors (see https://github.com/timber/timber/issues/1639). Instead, we would only automatically call it in loops or when the default context is set up.

When we use `PostCollection::to_array()`, then we will get an array and "lose" our custom iterator, right? In the loop, the `setup()` and `teardown()` functions wouldn’t be called on posts. Usually, this isn’t a problem, because in Twig, we can use collections in for-loops. And often, we need the collection to keep pagination working.

But, I still need to loop over a `PostCollection` in PHP sometimes, preferably without turning it into an array. And I want to document how it’s done best. And maybe improve `PostCollection` to make this easy. What’s the best way to loop over an `ArrayObject` in PHP? Can we use [`iterator_apply`](https://www.php.net/manual/en/function.iterator-apply.php), or [`IteratorAggregate`](https://www.php.net/manual/en/class.iteratoraggregate.php). Or can we add a new method to a collection that allows us to loop over the collection in a callback function?

acobster:
@gchtr ah, yes, #1793 was the discussion I was looking for. Cool - I'll have `Timber::get_*` call out to `get_queried_object()` or similar for no-arg defaults.

> When we use PostCollection::to_array(), then we will get an array and "lose" our custom iterator, right?

Correct. It won't maintain the WP Loop state the way a `PostCollection` would.

> But, I still need to loop over a PostCollection in PHP sometimes, preferably without turning it into an array.

`PostCollection` extends `ArrayObject`, which itself implements `IteratorAggregate`. So unless I'm missing something, you can loop over a `PostCollection` with `foreach` directly.

gchtr:
> > When we use PostCollection::to_array(), then we will get an array and "lose" our custom iterator, right?
> 
> Correct. It won't maintain the WP Loop state the way a `PostCollection` would.

Okay, so I’ll document this better.

> > But, I still need to loop over a PostCollection in PHP sometimes, preferably without turning it into an array.
> 
> `PostCollection` extends `ArrayObject`, which itself implements `IteratorAggregate`. So unless I'm missing something, you can loop over a `PostCollection` with `foreach` directly.

@acobster I see. Then that was a misunderstanding on my side and the problem is apparently using array functions like `array_map()` or `array_filter()` with an `ArrayObject`. I’d have to do it like that:

```php
$posts->exchangeArray( array_map( function( $post ) {
    return $post;
}, $posts->get_posts() ) );

$posts->exchangeArray( array_filter( $posts->get_posts(), function( $post ) {
    return true;
} ) );
```

What would you say, would we reinvent the wheel if we would provide `$posts->map()` and `$posts->filter()` instead?

```php
$posts->map( function( $post ) {
    return $post;
} );

$posts->filter( function( $post ) {
    return true;
} );
```

acobster:
Yeah...this is one of my least favorite things about PHP. _Sigh...._

I've gone back and forth about this myself. I can live with just doing:

```php
$postsArr = array_map(fn($p) => $p, $posts->to_array); // PHP 7.4 arrow functions FTW!
```

I could see `PostCollection::filter()` being helpful, but I feel pretty strongly that it should be immutable, i.e. it shouldn't modify the original object. So it would return a new `PostCollection` built up from the result of calling `array_filter` internally. Mutable state makes everything harder, and you might want to keep the original collection.

I'm less sold on `map`. Assuming immutability for the same reason as above, what does `map` return? If it returns an array, that's fine I guess, but really doesn't give us much over `to_array -> array_map`. If it returns a `PostCollection`, well, can we really call it a `PostCollection`? What if the callback returns `null`? or `42`? Foot-guns abound.

FWIW, if you're looking for a general-purpose solution to composable collections, have a look at [functional-php](https://github.com/lstrojny/functional-php) ([docs](https://github.com/lstrojny/functional-php/blob/master/docs/functional-php.md)), which is probably a better wheel than the one we'd invent. :) I haven't used it myself, but it's pretty mature and works with any `Traversable`.

szepeviktor:
👍 for functional-php, it is highly advanced code.

gchtr:
@acobster Thanks for the feedback! I see, with these other problems, I think it’s probably best to improve the documentation about the difference between arrays and PostCollections and maybe include the link to the [functional-php](https://github.com/lstrojny/functional-php) library for advanced usage.

acobster:
FYI to @jarednova and @gchtr - I updated all Factories `2.x-factories` branch to return a fallback of `false` instead of `null`. I also added support for string IDs for those that didn't have them.

jarednova:
Excellent, thanks @acobster! Speaking of that branch, is now a good time to start a PR of that into `2.x` itself?

acobster:
@jarednova I think it might be better to wait until Factory usage is universal on `2.x-factories`. Then we can pull in the latest from `2.x` and run it against PHPStan and all that. It'd be good to have a single point in time we can point to and say "this is when we flipped the switch on Factories." Plus, that's more satisfying. :D

acobster:
@gchtr - @FlyingDR made a [great point](https://github.com/timber/timber/issues/2093#issuecomment-579194104) about the `timber/post/classmap` filter that actually applies to all Class Map filters: What happens when the object instantiated does not implement `CoreInterface`?

Currently the Factory code will just `TypeError` because the returned object does not match the return type declaration of `::build()`. An error like this bubbling up from an internal class is bound to look like a Timber bug. We should probably throw an exception here instead, to make it clear to the user what they should do.

gchtr:
> An error like this bubbling up from an internal class is bound to look like a Timber bug. We should probably throw an exception here instead, to make it clear to the user what they should do.

@acobster That makes sense to me.

But now I have a question: When do we throw exceptions and when do we call `Helper::doing_it_wrong()`, which I introduce in #2145? I think the errors where I use `Helper::doing_it_wrong()` could also be exceptions, couldn’t they? Should we even use `Helper::doing_it_wrong()` or only use exceptions? What’s better from a developer perspective? What’s easier to handle when writing tests?

FlyingDR:
@gchtr Current implementation [simply swallows](https://github.com/timber/timber/pull/2145/files#diff-ca31977bb54277782eb3af181feb80c0R294-R296) any cases covered by helper method in production. It means that in a case if some kind of issue will be generated in production environment - nobody will ever get a chance to get information about it.

Sorry, just my 2c on this topic.

acobster:
If I were designing this API from the ground up, I'd probably go with Exceptions across the board, but I think we're past that point from a pragmatic perspective. :upside_down_face: Doing so would mean the upgrade path would involve not just class maps, but also wrapping most calls to the central `Timber` methods in try/catch blocks, which is a big ask. Also, it's probably more valuable to ship something than to try to get it perfect.

Error handling in my experience is best handled as an environment-specific thing. Which is to say, I think our best move here is to support throwing an exception if the user wants to, and giving them a hook to do custom stuff whenever we catch them _doing it wrong_.

We're _kinda_ doing that here, but as @FlyingDR points out, in production (or wherever `WP_DEBUG` is falsy) this will just toss out any errors immediately, without giving the user the chance to catch anything and handle it. It would be nice if we did something like:

```php
public static function doing_it_wrong( $function, $message, $version ) {
  do_action( 'doing_it_wrong_run', $function, $message, $version );

  if ( ! WP_DEBUG ) {
    return;
  }

  // trigger_error() etc. down here
}  
```

This gives folks an escape hatch for refactoring (which I consider the main use-case for `doing_it_wrong` to begin with) where they can throw exceptions, log to an external service, or whatever else to their hearts' content.

So to zoom back in on the question at hand: should we throw an exception, or just raise a `doing_it_wrong` error and let it crash?

I'm leaning towards the second approach.

This is more consistent with how Timber (and WP in general) alerts users to their mistakes. It's unfortunate that in this case their doing it wrong also leads a `TypeError`, but I would assume that the most likely place you're going to run into this is in a dev environment where you're able to see error logs directly. So I think what would be most helpful would be a message like `You returned "<class name>" from your "timber/post/class_map" filter. This class does not implement "Timber\CoreInterface" and will therefore cause a TypeError.` Then, we don't do anything special...we just let it crash.

I guess there could still be some edge cases where your Class Map is so complex that you can't totally guarantee that what it's returning is a `CoreInterface`. In that case, you should probably just wrap all your calling code in a try/catch anyway, which is possible given the escape hatch suggested above.

```php
add_action('doing_it_wrong_run', function($fn, $msg, $version) {
  throw new \LogicException('you did it wrong!');
});

add_filter('timber/post/class_map', function($post) {
  // hella confusing class map logic here
  return $whateverIdk;
});

try {
  $post = Timber::get_post(123);
} catch (\LogicException $e) {
  FancyLoggingSystem::error($e->getMessage);
  $post = false;
}
```

gchtr:
Thanks, @FlyingDR! Your 2 cents are very much appreciated! I guess I only copied the logic we already had in `Helper::deprecated()` and didn’t think more about it.

> If I were designing this API from the ground up, I'd probably go with Exceptions across the board, but I think we're past that point from a pragmatic perspective. 🙃 Doing so would mean the upgrade path would involve not just class maps, but also wrapping most calls to the central `Timber` methods in try/catch blocks, which is a big ask. Also, it's probably more valuable to ship something than to try to get it perfect.

@acobster I agree!

I like your recommendations and added a separate pull request in #2196 to explore this further.

gchtr:
Next question! When looking at the methods in the `Timber` class, I see that most of the functions will follow the `get_*` pattern for the new API, except for `Timber::context()`. I initially changed that from `Timber::get_context()` to be more in line with functions like `Post::link()` or `Post::title()`. But now, I’m not so sure anymore if we should use `Timber::context()` or keep `Timber::get_context()` to be more in line with other `Timber::` API functions.

What do you think? @acobster @jarednova 

grafik

acobster:
@gchtr consistency makes sense to me, and the places where we omit the `get_` part are methods that are most likely going to be called from Twig, whereas `::context()` is not. On the other hand, the `get_*` methods are all about specific WP entities, whereas context is more of a grab bag. I don't have a strong opinion either way.

jarednova:
@gchtr I'd vote to leave as-is with both `Timber::get_posts()` and `Timber::context();` existing together (`get_context` is still hanging out in `Timber.php` as deprecated so it's not gone forever).

In my mind the `get_` signifies queries to WP — wheras the others are simply gathering the piece(s) of data the user needs for that piece of code

gchtr:
@jarednova Cool, thanks for the feedback 👍

gchtr:
😱 Apparently I merged in the wrong branch. I still want to merge this into `2.x`, and not `2.x-factories`. Luckily, there’s cherry-picking. All fixed now 👍.

gchtr:
As soon as #2367 is merged, I want to merge this pull request, too. We discussed and documented the new API thoroughly. The outstanding todos all have separate issues and we don’t have to wait for these to be resolved before we merge this.

acobster:
@gchtr agreed! I just did a once-over of the [2.0 tag](https://github.com/timber/timber/issues?q=is%3Aopen+is%3Aissue+label%3A2.0) and it's looking like we're in really good shape. Let's just keep an eye out for anything else that needs that tag that doesn't already have it.

Once this is merged, how about we make an actual `v2.0.0-alpha.01` release?

cc @jarednova

jarednova:
Oh my. Are we actually here????!!!!! Yes! Let's start the tagging and moving to `alpha`. #2367 looks good to me!

FlyingDR:
@acobster 
> Once this is merged, how about we make an actual 2.0.0-alpha.01 release?

Short notice: as per [p.9](https://semver.org/#spec-item-9) of semver specification:

> Numeric identifiers MUST NOT include leading zeroes

Actual version number should probably be `2.0.0-alpha.1`

gchtr:
@jarednova @acobster Yes, I think we’re getting there! For me, there’s still quite some work for the documentation, but that doesn’t stop us from releasing an Alpha version 👍🚀🔥. I’ll let you one of you review and merge this.

jarednova (approved):

jarednova:
🎉 LTGM! I'll also pull this into the docs style updates (https://github.com/timber/docs/pull/23) so we can more closely review that work with these new additions/changes

acobster:
**woohoooo!!** :tada: 

@jarednova so how do we get the 2.0 docs up to https://timber.github.io/docs/v2?

Also, I have started on a draft release - please take a look and add your notes when you get a chance.

cc @gchtr

gchtr:
@acobster @jarednova I'll have a look at everything tomorrow and will update the live documentation.

Issue #1973, opened 2019-04-22
found_posts on new PostQuery()


Grafikart:
## Is your feature request related to a problem? Please describe.

There is no way to get the total number of post from the PostQuery object which forces users to write things like this : 

```php
$query = new WP_Query([
    'cat'            => $category->id,
    'post__not_in'   => get_option('sticky_posts'),
    'posts_per_page' => $context['offset']
]);
$context['posts'] = new PostQuery($query);
$context['found'] = $query->found_posts;
```

## Describe the solution you’d like

A method (`total()` or `found()`) on `PostQuery` to get the total number of post found.

## Questions

Is there a reason why this feature is not implemented ? I may be able to submit a PR to implement this myself but I want to know if there was a reason for not including this in the PostQuery class. 

Thanks

marciojc:
@Grafikart 
```
$context['found'] = count($context['posts']);
```

MouadDB:
The count() function returns the number of elements in an array.
https://www.php.net/manual/en/function.count.php

marciojc:
oh, we want the total of posts of the query.
Don't know what is original code.
But as his already using new WP_Query he can use his solution or calling the global $query

j-masci:
> The count() function returns the number of elements in an array. https://www.php.net/manual/en/function.count.php

Thanks, but he wants the total number of posts not including pagination parameters, ie. what you get back from SQL when using SQL_CALC_FOUND_ROWS selector.

Issue #1257, opened 2016-12-01
Make $wp_query->found_posts available through Timber::get_posts()


jphilung:
On queries that do not specify arguments related to posts_per_page and which uses the default posts_per_page value, I can't find a way of getting the total number of posts found.

#425 references this issue, but the count in the array returned is at maximum the number specified by the default posts_per_page.

For the moment, I've resorted to using the global $wp_query variable and get the value from the there.

I'd like to contribute to the repo by adding this feature, how would I contribute ?

paulhuisman:
Funny you posted this yesterday, because it's exactly what i need as well.

chrisvanpatten:
I need this as well. Here's how I'm hacking it for now:

```
<?php

class MyQuery extends \Timber\PostQuery
{

	/**
	 * Uses a hack to steal access to a private property, in this
	 * case, we need to access $this->queryIterator->_query
	 *
	 * @via https://ocramius.github.io/blog/accessing-private-php-class-members-without-reflection/
	 * @return query
	 */
	public function forceGetQuery()
	{
		$query = function ($iterator) {
			return $iterator->_query;
		};

		$found = \Closure::bind($query, null, $this->queryIterator);

		return $found($this->queryIterator);
	}

	/**
	 * Get the count of results
	 *
	 * @return int
	 */
	public function getResultsCount()
	{
		$query = $this->forceGetQuery();

		// Cast to an int
		return absint($query->found_posts);
	}
}
```

In your controller/theme file:

```
$posts = new MyQuery();
$posts->getResultsCount();
```

I'm not yet super familiar with Timber internals but I'd be happy to work up a PR with some guidance… cc @jarednova (hey dude, long time no chat!)

gchtr:
I’m having this issue as well. Accessing query variables for a custom query is too hard now. And falling back to the `$wp_query` global doesn’t work when using a custom query in singular templates.

Would be great if you could work on this @chrisvanpatten.

mibu31:
Bump - same issue.

I'm using `new \Timber\PostQuery($args);` with the `paged` argument and I'm unable to access the total number of posts returned before `paged` was applied (`$wp_query->found_posts`).

mibu31:
@jarednova @chrisvanpatten 
Is there a reason this variable is set to private? (other than following best practice)
I might be able to help resolve it/PR if I'm given some insight as to why it's been done like this. 
[...]

PR #2074, opened 2019-09-26
Add found_posts property to Timber\PostQuery

gchtr:
**Ticket**: #1973, #1257

## Issue

There is no easy way to access the amount of found posts in a query.

## Solution

Add `found_posts` property to `Timber\PostQuery`. It will only be filled if `Timber\PostQuery` uses `Timber\QueryIterator` internally, but not when what you pass to `Timber\PostQuery` is already a collection of posts.

## Impact

Happyness.

## Usage Changes

You can now access the amount of found posts through the `found_posts` property in `Timber\PostQuery`:

```php
$query = new Timber\PostQuery( [
    'post_type' => 'post',
] );

$query->found_posts;
```

## Considerations

Are there other properties that should be easy to access?

## Testing

Yes. Including tests for special cases when `no_found_rows` is used in the query.


jarednova:
Let's see, we got
1) Clean code
2) Tests
3) Release notes for the readme

... yeah, I want to do something to feel useful — but I think it's all taken care of by @gchtr. Merging!

Issue #2071, opened 2019-09-23
Recursive reference in menu object


ramiroazar:
## Expected behavior

The `json_encode` function should return a JSON encoded string of the menu object.

## Actual behavior

The `json_encode` function returns a boolean with the value `false`.

The `json_last_error` function returns a constant with the value `JSON_ERROR_RECURSION`.

The `var_dump` function returns a `menu` property with the value `*RECURSION*` within every menu item.

## Steps to reproduce behavior

`json_encode(new Timber\Menu());`

## What version of WordPress, PHP and Timber are you using?

WordPress 5.2.3, PHP 7.2.9, Timber 1.11.0 

## How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)

WordPress plugin repository.

jarednova:
@ramiroazar thanks! looking into this now ...

jarednova:
I think this can be a pretty simple fix (albeit, with a few breaking changes that we should be mindful of). A fix is in on #2083

PR #2083, opened 2019-09-30
Eliminate Recursion in Menu / MenuItem

jarednova:
**Ticket**: #2071 

## Issue
There's recursion in a Menu's MenuItem's Menus

## Solution
Convert the public `MenuItem::$menu` property into a public method

## Impact
This will break any PHP-based calls to `MenuItem::$menu` however, anything in Twig should still work (since it will pass the call of `{{ menuitem.menu }}` off to the method)

## Usage Changes
In PHP, to get a MenuItem's menu, the dev should use:
```php
$menu = $menu_item->menu();
```

## Considerations
Would be great to make calls to the `$menu` property backwards compatible

## Testing
New test to validate no recursion; might need additional test to cover new method


jarednova:
Travis tests are still failing (though they pass on my local); please hold off on any review/merge until I can resolve

jarednova:
Looks like those failing tests are now resolved (good thing they failed, helped me to identify a bug that this created related to a `MenuItem`'s depth!). Would love another set of 👀  on this ...

jarednova:
@gchtr thanks for your feedback, back to you! Just a few tweaks were needed to take care of those items

gchtr:
@jarednova Thanks for the changes!

> Instead, could we add a deprecation call to a `__get()` method, that checks if the protected `$menu` property is accessed?

I’m assuming you don’t want to catch direct calls to `MenuItem::$menu` then?

jarednova:
@gchtr actually, it turns out we get that already w/o any modification to `__get()` ....

I wrote this test to validate (which passes):

```php
function testMenuItemMenuProperty() {
		$pg_1 = $this->factory->post->create( array( 'post_type' => 'page', 'post_title' => 'Foo Page', 'menu_order' => 10 ) );
		$pg_2 = $this->factory->post->create( array( 'post_type' => 'page', 'post_title' => 'Bar Page', 'menu_order' => 1 ) );
		$page_menu = new Timber\Menu();
		$items = $page_menu->get_items();
		$menu = $items[0]->menu;
		$this->assertEquals('Timber\Menu', get_class($menu));
	}
```

... the logic in `__get()` looks for methods with that name and already uses them instead of properties.

gchtr:
@jarednova Oh, I see! Very well then. Let’s merge this in!

PR #2085, opened 2019-10-01
2.x refactor tests

acobster:
**Ticket**: #1793 

## Issue

We're moving to a new API that will restrict direct instantiation of core Timber objects (`Post`, `User`, `Menu`, etc.) in favor of corresponding `Timber::get_*` calls.

## Solution

This refactors away most direct instantiations within the tests. Some refactors we'll need to do eventually are impossible right now, since the `Timber` class lacks some `get_*` methods we'll need to replace the calls with. I don't want to add any code just yet, so I'm just taking care of the low hanging fruit right now.

## Impact

None.

## Usage Changes

No changes to production code in this PR; only tests.

## Considerations

???

## Testing

You know I'm all about those tests tho.

### DIFF
--- a/tests/test-timber-comment.php
+++ b/tests/test-timber-comment.php
@@ -2,7 +2,7 @@
 
 /**
  * @group called-post-constructor
- * @todo #1793 replace direct Timber\Comment instantiations
+ * @todo #2094 replace direct Timber\Comment instantiations
  */
 class TestTimberComment extends Timber_UnitTestCase {
 
--- a/tests/test-timber-comment.php
+++ b/tests/test-timber-comment.php
@@ -2,6 +2,7 @@
 
 /**
  * @group called-post-constructor
+ * FIXME #1793 replace direct Timber\Comment instantiations
  */
 class TestTimberComment extends Timber_UnitTestCase {
 
--- a/tests/test-timber-comment.php
+++ b/tests/test-timber-comment.php
@@ -1,5 +1,8 @@
 <?php
 
+/**
+ * @group called-post-constructor
+ */
 class TestTimberComment extends Timber_UnitTestCase {
 
 	function testComment(){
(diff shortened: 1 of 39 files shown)


tests/test-timber-term.php
```diff
@@ -66,39 +70,41 @@ function testConstructor() {
...
-			$term = new Timber\Term($term_id);
+			$term = Timber::get_term($term_id);
 			$this->assertEquals('Famous Commissioners', $term->title());
 			$this->assertEquals('Timber\Term', get_class($term));
 		}
 
 		function testTermConstructWithSlug() {
 			$term_id = $this->factory->term->create(array('name' => 'New England Patriots'));
-			$term = new Timber\Term('new-england-patriots');
+			$term = Timber::get_term($term_id);
```
gchtr: Is the idea here that we don’t allow instantiation via slug, right? Shouldn’t we have to delete that test then, or make sure it’s throwing a doing-it-wrong message? Because the name of the test suggests that it’s specifically about constructing a term with a slug.

tests/test-timber-term.php
```diff
@@ -143,7 +149,7 @@ function testGetPostsWithAnyAndCustomTax() {
 
 			$term_id = $this->factory->term->create(array('name' => 'Zong', 'taxonomy' => 'arts'));
 			$posts = $this->factory->post->create_many(5, array('post_type' => 'portfolio' ));
-			$term = new Timber\Term($term_id);
+			$term = Timber::get_term($term_id, 'arts');
```
gchtr: Would this fail without `'arts'` as a second parameter? Or: is this line needed at all? I don’t see `$term` being used in the code.

tests/test-timber-comment.php
```diff
@@ -1,5 +1,9 @@
 <?php
 
+/**
```
gchtr: Are these temporary comments for working through the API or are they going to stay for 2.x, until we refactored the remaining classes, like `Comment`, `User`, etc.?

We could maybe make use of the `@todo` tag here. It’s getting picked up in a lot of IDEs, so it’s easier to work through.

acobster: The `@group` tags are temporary, to allow e.g. `phpunit --group called-post-constructor`. I've replaced the `FIXME` tags with `@todo` - good call.

tests/test-timber-term.php
acobster: You're right, `$term` isn't used in this case and that line can be safely removed. But the `category` ones fail without the `$taxonomy` param...which raises a larger question about the Term API.

`TermGetter::get_term()` currently takes a second `$taxonomy` param which defaults to `post_tag`, unlike the core API which defaults this param to the empty string. This is going to cause problems. Hence, #2087 .

For now I'm just passing empty strings explicitly everywhere; this question is out of scope here as the immediate goal is to keep the tests passing without instantiating terms directly.

tests/test-timber-term.php
acobster: Yep, I moved this into a pending test to look at as part of #2087.

acobster:
@gchtr any more feedback before I merge this in?

jarednova:
@gchtr are you sure you meant to close this issue?

gchtr:
Nope, I didn’t. For the documentation in #2073, I first created a `2.x-api` branch, that I then renamed to `2.x-docs-api`. I saw yesterday that the `2.x-api` branch still existed and assumed we didn’t use it, so I deleted it. Except this pull request is made against the `2.x-api` branch, so we still need it 🙈. It’s back now 😅. Sorry!

Issue #1817, opened 2018-10-15
`{{ post.excerpt }}` can take an array of arguments


jarednova:
A few steps here ...

### Acceptance Criteria
- [x] `{{ post.preview }}` is deprecated in favor of `{{ post.excerpt }}`. 
- [x] Allow the constructor to take an array of valid arguments so that this: `{{ post.excerpt({chars : 55, force: true}) }}` becomes possible within Twig templates
- [x] Update docs to document new behavior
- [x] Update upgrade guide to document new behavior

acobster:
@jarednova deprecating `preview` in favor of `excerpt` is just about clarifying the confusing name, correct? As opposed to moving away from the `PostPreview` class to say a `PostExcerpt` class or something...

Also, should I go ahead and remove `Post::get_preview()`?

jarednova:
@acobster yes, this is really a naming fix. Preview is used in WP to indicate post previewing (as in, viewing something from the admin before it's published).

And yes, since `{{ post.get_preview }}` has been deprecated since 1.3.1 (> 1 year) let's go ahead and remove. While we're here, we may as well rename `PostPreview` to `PostExcerpt` — lump all the pain together.

However, we should keep `{{ post.preview }}` — deprecate yes, but not fully remove until a subsequent version

gchtr:
This issue could possibly be solved in the same update: #1345.

acobster:
@jarednova here's a blast from the past for ya:

> While we're here, we may as well rename PostPreview to PostExcerpt — lump all the pain together.

Should we still do this? Just checking/revalidating.

@gchtr can you summarize what the fix would be for #1345?

gchtr:
@acobster I already summarized these outstanding changes as todos in https://github.com/timber/timber/pull/2365#issuecomment-724269263.

> @gchtr can you summarize what the fix would be for #1345?

I don’t know yet. I first want to write a test to make sure this is still an issue.

acobster:
@gchtr oh gotcha, I missed that somehow. Thanks! Seems like we can close this one out then?

gchtr:
Yes, let’s close this so we have this off our list.

PR #2098, opened 2019-10-12
Excerpt arguments and documentation 

jarednova:
@acobster had done most of the hard work here. I was going to tick off the last documentation needs in #1817 so we can merge this in!


jarednova:
I think this is ready! I ticked off the remainder of the items in #1817 with the 2.0 Upgrade guide, cleaned-up docblocks and an @example for `Timber\Post`

jarednova:
good find, thanks for the clean-up!

Issue #2099, opened 2019-10-18
composer update : Illegal byte sequence


romainmenke:
## Expected behavior
Expected `composer update timber/timber` to pass without errors or notices.

## Actual behavior
```
$ composer update timber/timber
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 0 installs, 1 update, 0 removals
  - Updating timber/timber (1.11.0 => 1.12.0): Downloading (100%)    Failed to execute unzip -qq  'wp-content/vendor/timber/timber/5f32a2f4faf825ba723993113dedf75a' -d 'wp-content/vendor/composer/00146520'

error:  cannot create wp-content/vendor/composer/00146520/timber-timber-25e91be/tests/assets/-�Robocop.jpg
        Illegal byte sequence
error:  cannot create wp-content/vendor/composer/00146520/timber-timber-25e91be/tests/assets/����ά���?-������-��������.jpg
        Illegal byte sequence

    The archive may contain identical file names with different capitalization (which fails on case insensitive filesystems)
    Unzip with unzip command failed, falling back to ZipArchive class

Writing lock file
Generating autoload files
PHP CodeSniffer Config installed_paths set to ../../wp-coding-standards/wpcs
```

## Steps to reproduce behavior
`composer update timber/timber`

When trying to update to : https://github.com/timber/timber/releases/tag/1.12.0

Composer : 

> Composer version 1.9.0 2019-08-02 20:55:32

System :

> macos 10.14.6

Filesystem :

> APFS
> Case Insensitive

## What version of WordPress, PHP and Timber are you using?
WordPress 5.2.4, PHP 7.2, Timber 1.12.0

## How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Composer

## Related commit :

https://github.com/timber/timber/commit/6c2797f958fee88f0fdb5f46de0282074e9545e9

## Related code in composer :

https://github.com/composer/composer/blob/2528654c53b88b490c3b78e111d068bcbe7fef61/src/Composer/Downloader/ZipDownloader.php#L105

## Thoughts :

Not sure if this is something the can/should be fixed in composer or if a better test case is possible in timber.

danburzo:
This might be an issue with the `unzip` command: 

```
Unzip with unzip command failed, falling back to ZipArchive class
```

[...]

PR #2100, opened 2019-10-18
Address composer errors in #2099

jarednova:
**Ticket**: #2099 

## Issue
In #2099 we were alerted to Composer errors resulting from the introduction of special characters in file names for tests.

## Solution
Run the tests, but rename files within those tests using PHP.

## Impact
None, just tests!

## Usage Changes
None, just tests!

## Considerations
These tests aren't as "pure" as the prior versions, but they theoretically cover the same scenarios. We could also consider splitting the tests out of what Composer distributes (as @palmiak suggested)

## Testing
Yup!


jarednova:
This is just some tweaks to tests, so I'm going to go ahead and merge ...

Issue #2112, opened 2019-11-18
View details link is not relative


Beee4life:
## Expected behavior
When I click the "View details" link in the Wordpress plugins screen, I expect to see the plugin details/info, but since the link doesn't take into account that some sites are run from a sub folder, this gives a 404.

## Actual behavior
I get a 404 screen, see below gif.

![LlaDMLbAAp](https://user-images.githubusercontent.com/4508731/69048050-d3eaa400-09fc-11ea-8663-ee3679b8a5af.gif)

## What version of WordPress, PHP and Timber are you using?
WP 5.3
PHP 7.2
Timber 1.12.0

## How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Composer/packagist

SOLUTION: 
Change  line 24 in /lib/Admin.php

    $links[] = '<a href="/wp-admin/plugin-..." class="thickbox" aria-label="More information about Timber" data-title="Timber">View details</a>';

to 

    $links[] = '<a href="' . get_site_url() . '/wp-admin/plugin-..." class="thickbox" aria-label="More information about Timber" data-title="Timber">View details</a>';

palmiak:
Could you try making a pull request for this? It looks ok.

pon., 18 lis 2019, 12:18 użytkownik Beee <notifications@github.com> napisał:

Beee4life:
As you can see I did...

Beee4life:
@palmiak I saw some new stuff,  like codegov and so... Do I need to do anything else ?

jarednova:
Thanks @Beee4life — just merged this in. Thanks so much for raising the issue and the fix. I'm getting things set now to release a new version later today/tonight

PR #2113, opened 2019-11-18
Fix non-relative site url

Beee4life:
**Ticket**: #2112

## Issue
The view details link on the plugins screen is 'hardcoded' and is always relative to root. If you install WPP in a subfolder it'll break.

## Solution
Solution explained [here](https://github.com/timber/timber/issues/2112#issue-524300512)

## Impact
None

## Usage Changes
None that I can see

## Testing
No

### DIFF
--- a/lib/Admin.php
+++ b/lib/Admin.php
@@ -21,7 +21,7 @@ class Admin {
 	public static function meta_links( $links, $file ) {
 		if ( strstr($file, '/timber.php') ) {
 			unset($links[2]);
-			$links[] = '<a href="' . get_site_url(). '/wp-admin/plugin-install.php?tab=plugin-information&amp;plugin=timber-library&amp;TB_iframe=true&amp;width=600&amp;height=550" class="thickbox" aria-label="More information about Timber" data-title="Timber">View details</a>';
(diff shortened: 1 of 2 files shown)


lib/Admin.php
```diff
@@ -21,7 +21,7 @@ public static function init() {
 	public static function meta_links( $links, $file ) {
 		if ( strstr($file, '/timber.php') ) {
 			unset($links[2]);
-			$links[] = '<a href="/wp-admin/plugin-install.php?tab=plugin-information&amp;plugin=timber-library&amp;TB_iframe=true&amp;width=600&amp;height=550" class="thickbox" aria-label="More information about Timber" data-title="Timber">View details</a>';
+			$links[] = '<a href="' . get_site_url(). '/wp-admin/plugin-install.php?tab=plugin-information&amp;plugin=timber-library&amp;TB_iframe=true&amp;width=600&amp;height=550" class="thickbox" aria-label="More information about Timber" data-title="Timber">View details</a>';
```
gchtr: Could we also use `admin_url( 'plugin-install.php…' )` instead of `get_site_url( '/wp-admin/plugin-install.php…' )`?

Beee4life: I think that one is even better...  Didn't think of that one... I will update my PR.

PR #2115, opened 2019-11-27
fix(ImageHelper): apply URLHelper filter to home_url

domtra:
## Issue
With WPML enabled and set to use subdirectories for languages, the relative image path is not resolved correctly. Instead it is an absolute url including the WPML language code. This leads to 404s if you rely on it.

## Solution
Apply the existing URLHelper filter `timber/URLHelper/get_content_subdir/home_url` that takes care of stripping the language code from the home url.

## Impact
This only uses existing functionality. The relative path was broken before. There should not be any other impact.

## Testing
I could not any test coverage regarding the relative path in the first place. So I did not at any new test.

Cheers.

### DIFF
--- a/lib/ImageHelper.php
+++ b/lib/ImageHelper.php
@@ -33,7 +33,8 @@ class ImageHelper {
 	static $home_url;
 
 	public static function init() {
-		self::$home_url = get_home_url();
+		$home_url = get_home_url();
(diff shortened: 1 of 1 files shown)


domtra:
sorry, there are actually some timing issues. i will have a look again and reopen when resolved.

Issue #2120, opened 2019-12-17
Investigate custom `filter` filter


jarednova:
```
@TimberWP Your `filter` filter overwrites the core Twig `filter` filter. 
Spent an hour trying to figure out why the Twig example returns an empty array:

{% set sizes = [34, 36, 38, 40, 42] %}
{{ sizes|filter(v => v > 38) }}

Any workarounds (other than editing the package code)?
```

https://twitter.com/TimberWP/status/1204121465903300608

gchtr:
Related: https://github.com/timber/timber/issues/2010, with a suggested solution in https://github.com/timber/timber/issues/2010#issuecomment-502896508.

PR #2121, opened 2019-12-17
ref #2120 #1594 -- deprecate custom filter filter

jarednova:
**Ticket**: #2120

## Issue
In #1594 we made a custom `filter` filter (confusing yet?). Not long after, Twig introduced its own `filter` filter with a similar behavior (though different syntax):

https://twig.symfony.com/doc/2.x/filters/filter.html

## Solution
Deprecate the filter in 1.x — I'm going to do another PR to remove it from 2.0 all together

## Usage Changes
No usage change. 1.x users would still use the "old" syntax to use `filter`. I can't think of a good way to let someone disable our custom `filter` filter w/o doing a whole mini-production

## Considerations
In a perfect world we would do that mini-production, but I think we should focus on ...
a) acknowledge the issue
b) truly resolve in 2.0

### DIFF
--- a/lib/Twig.php
+++ b/lib/Twig.php
@@ -207,6 +207,11 @@ class Twig {
 		$twig->addFilter(new Twig_Filter('list', array($this, 'add_list_separators')));
 
 		$twig->addFilter(new Twig_Filter('pluck', array('Timber\Helper', 'pluck')));
+
+		/** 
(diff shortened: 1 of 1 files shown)


jarednova:
I'm going to pull this one in so that the deprecated comment starts appearing in the code ASAP

Issue #2010, opened 2019-05-22
Consider removing `filter`/`pluck` filters


nlemoine:
Twig recently introduced some useful array methods like `column` (since 2.8.0), `filter`, `map` and `reduce` (since 1.41/2.10.0).

https://github.com/twigphp/Twig/blob/2.x/CHANGELOG

https://twig.symfony.com/doc/2.x/filters/column.html
https://twig.symfony.com/doc/2.x/filters/filter.html
https://twig.symfony.com/doc/2.x/filters/map.html
https://twig.symfony.com/doc/2.x/filters/reduce.html

Timber has its own filters `pluck` and `filter`. The second one makes native Twig `filter` unusable. 

This would probably introduce a BC break. But would you consider this for 2.0+?

nlemoine:
For those who want to use these new filters right now, you can still add them with another name:

```php
add_filter('timber/twig', function($twig) {
    $twig->addFilter(new \Twig_SimpleFilter('array_column', 'twig_array_column'));
    $twig->addFilter(new \Twig_SimpleFilter('array_filter', 'twig_array_filter'));
    $twig->addFilter(new \Twig_SimpleFilter('array_map', 'twig_array_map'));
    $twig->addFilter(new \Twig_SimpleFilter('array_reduce', 'twig_array_reduce'));
    return $twig;
});
```

jarednova:
@nlemoine thanks for the heads-up on those! I think it makes sense to clean those up for 2.0 to match the native abilities in Twig. Overall we're trying to reduce our "custom" layer and pass-along the features of Twig to the end-user

rubas:
> ```
> add_filter('timber/twig', function($twig) {
>     $twig->addFilter(new \Twig_SimpleFilter('array_column', 'twig_array_column'));
>     $twig->addFilter(new \Twig_SimpleFilter('array_filter', 'twig_array_filter'));
>     $twig->addFilter(new \Twig_SimpleFilter('array_map', 'twig_array_map'));
>     $twig->addFilter(new \Twig_SimpleFilter('array_reduce', 'twig_array_reduce'));
>     return $twig;
> });
> ```

FYI: `\Twig_SimpleFilter` is deprecated. You should use `\Twig\TwigFilter` instead.

nlemoine:
@rubas Thanks pointing that out, I know, i just copied/pasted code from [Timber](https://github.com/timber/timber/blob/1.9.4/lib/Twig.php#L208) and didn't pay attention 😉

[...]

PR #2124, opened 2019-12-19
Twig filter

palmiak:
**Ticket**: #2010 

## Issue
Twig and Timber have filter called `filter` and now we have a problem.

## Solution
Adds new filter called `wp_list_filter` and adds backward compatibility and warning to normal `filter`.

## Impact
User's that use the `filter` and with string or array as second param will get a warning.

## Usage Changes
The twig filter works a bit different - https://twig.symfony.com/doc/3.x/filters/filter.html.

## Testing

Not yet

### DIFF
--- a/lib/Helper.php
+++ b/lib/Helper.php
@@ -423,16 +423,38 @@ class Helper {
 
 	/**
 	 * Filters a list of objects, based on a set of key => value arguments.
+	 * Uses native Twig Filter.
+	 *
+	 * @since x.x.x
+	 * @param array                 $list to filter.
+	 * @param callback|string|array $arrow function used for filtering,
+	 *                              string or array for backward compatibility.
+	 * @param string                $operator to use (AND, NOT, OR). For backward compatibility.
+	 * @return array
+	 */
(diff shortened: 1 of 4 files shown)


lib/Helper.php
```diff
@@ -423,16 +423,38 @@ public static function pluck( $array, $key ) {
...
 	 * @ticket #1594
 	 * @param array        $list to filter.
-	 * @param string|array $filter to search for.
+	 * @param string|array $args to search for.
 	 * @param string       $operator to use (AND, NOT, OR).
 	 * @return array
 	 */
-	public static function filter_array( $list, $args, $operator = 'AND' ) {
-		if ( ! is_array($args) ) {
+	public static function wp_filter_array( $list, $args, $operator = 'AND' ) {
```
jarednova: Can we change this to `wp_list_filter` to match the Twig filter name? One less name to manage (and confuse me)

jarednova:
I'm going to close my associated PRs in favor of this one

lib/Helper.php
palmiak: good idea - pushing the fix in a moment.

Issue #2104, opened 2019-10-25
Timezone for twig date function not working as expected


columbian-chris:
I'm using Pacific timezone in daylight savings time in my example, which is UTC-7 hours.

Out of the box, both `{{ date('now')|date('F j, Y @ g:i a') }}` (the date _function_) and `{{ 'now'|date('F j, Y @ g:i a') }}` (the date _filter_) return UTC time with no timezone applied to them.

When I stick `date_default_timezone_set('America/Los_Angeles');` at the top of my `functions.php` theme file, the date _filter_ displays the correct current time (UTC-7), but the date _function_ is displaying UTC-14, almost like the equivalent timezone offset is being applied twice.

After reading [the Twig docs on the `date()` function](https://twig.symfony.com/doc/2.x/functions/date.html), I thought adding the following in my `functions.php` file might work, but it isn't affecting my date variables in twig:

```php
function add_to_twig( $twig ) {
	$twig->getExtension(\Twig\Extension\CoreExtension::class)->setTimezone('America/Los_Angeles');
	return $twig;
}
```

I couldn't find anything in the Timber docs about timezone usage or manipulation.  I have tried this on two different servers producing the same results.

jarednova:
Thanks @columbian-chris — I'll take a look!

jarednova:
@columbian-chris I can't say it makes a ton of sense, but I think this might be working as intended. Consider the following test:

```php
update_option('timezone_string', 'America/Los_Angeles');
$tz = get_option('timezone_string');
$date_format   = 'F j, Y @ g:i a';
date_default_timezone_set('America/Los_Angeles');
$pid = $this->factory->post->create(array('post_date' => '2019-10-31 12:46:41'));
$post = new TimberPost($pid);
$post_date = Timber::compile_string("{{ post.date('".$date_format."') }}", array('post' => $post));
$date_function = Timber::compile_string("{{ date('2019-10-31 12:46:41')|date('F j, Y @ g:i a', 'America/Los_Angeles') }}");
$date_filter   = Timber::compile_string("{{ '2019-10-31 12:46:41'|date('F j, Y @ g:i a') }}");
error_log('$tz = '.$tz);
error_log('post_date ='. $post_date);
error_log('date_function = '.$date_function);
error_log('date_filter = '.$date_filter);
$this->assertEquals($post_date, $date_filter);
```

You and me (normal humans) would expect that all three of these would return the same output. However, the actual output is:

```
[...]

PR #2127, opened 2019-12-24
Remove custom `date` filter

jarednova:
**Ticket**: #2104 

## Issue
As @columbian-chris points out, there's some weird stuff happening when you use the `date` filter in certain circumstances. No wonder! We're overriding Twig's default behavior here in an attempt to better integrate WP behavior

## Solution
Change the name of our custom filter from `date` ==> `intl_date`

## Impact
This could break backwards compatibility, so making this a 2.x change

## Usage Changes
Yes! Documented in the 2.x upgrade guide

## Considerations
Should the name of the filter simply be `date_i18n`? My fear is unless we're taking _exactly_ the WP behavior, we shouldn't have the same name — thus, `intl_date` is custom to Timber but something we can control even if in the future WP, Twig or PHP adopts new date/time internationalization behaviors.

## Testing
Existing tests will need to be updated to match default Twig `date` filter behavior


gchtr:
@jarednova I have some thoughts about this and I started working on a separate guide to document how we should work with dates in Timber. I will have some spare time during my holidays to finish that guide and think about a solution for when to use the new `wp_date()` function and when to still use `date_i18n()`.

jarednova:
sounds good @gchtr! I'll cool my jets on this for now then

jarednova:
Closing in favor of #2135

PR #2137, opened 2020-01-02
2.x Update post.date

gchtr:
**Ticket:** #2135

## Issue

As of WordPress 5.3, there we a lot of [updates to the Date/Time component](https://make.wordpress.org/core/2019/09/23/date-time-improvements-wp-5-3/). Using `date_i18n()` is now discouraged and it might be deprecated in the future. Also, there are new helpful functions like `get_post_timestamp()` that we can make use of.

## Solution

- Uses functions that were newly introduced in WordPress 5.3 to get post times.
- Removes usage of `date_i18n()` and tries to reduces usage of `strtotime()` for posts.
- Introduces new `$post->timestamp()` and `$post->modified_timestamp()` functions to retrieve a post timestamp as the recommended way.
- Introduces a new `DatedInterface()` class to streamline the usage of the `timestamp()`, `modified_timestamp()`, `date()`, `date_modified()`, `time()` and `time_modified()` functions. In a later stage, we can use this interface for the `Timber\Comment` class.

## Impact

Breaks backwards compatibility with WordPress version below 5.3. Discussion in #2135.

## Usage Changes

New `$post->timestamp()` and `$post->timestamp_modified()` functions.

## Considerations

- Might there be a better name for the `DatedInterface`, which says that an object has a timestamp, a publishing and a modification date?

## Testing

Ran test. Fine until now.

### DIFF
--- a/lib/DatedInterface.php
+++ b/lib/DatedInterface.php
@@ -1,0 +1,54 @@
+<?php
+
+namespace Timber;
+
+/**
+ * Interface DateTimeInterface
+ *
+ * An interface for classes that implement getting date/time values from objects.
+ *
+ * @since 2.0.0
+ */
+interface DatedInterface {
+	/**
+	 * Gets the timestamp when the object was published.
+	 *
+	 * @api
+	 * @since 2.0.0
+	 *
+	 * @return false|int Unix timestamp on success, false on failure.
+	 */
+	public function timestamp();
+
+	/**
+	 * Gets the timestamp when the object was last modified.
+	 *
+	 * @api
+	 * @since 2.0.0
+	 *
+	 * @return false|int Unix timestamp on success, false on failure.
+	 */
+	public function modified_timestamp();
+
+	/**
+	 * Gets the publishing date of the object.
+	 *
+	 * @api
+	 * @since 2.0.0
+	 *
+	 * @param string|null $date_format Optional. PHP date format. Will use the `date_format` option
+	 *                                 as a default.
+	 * @return string
+	 */
+	public function date( $date_format = null );
+
+	/**
+	 * Gets the date of the last modification of the object.
+	 *
+	 * @param string|null $date_format Optional. PHP date format. Will use the `date_format` option
+	 *                                 as a default.
+	 *
+	 * @return string
+	 */
+	public function modified_date( $date_format = null );
+}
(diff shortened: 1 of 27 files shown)


lib/DatedInterface.php
```diff
@@ -0,0 +1,54 @@
...
+namespace Timber;
+
+/**
+ * Interface DateTimeInterface
+ *
+ * An interface for classes that implement getting date/time values from objects.
+ *
+ * @since 2.0.0
+ */
+interface DatedInterface {
```
pascalknecht: Maybe name it HasDate?

jarednova:
I love this too! Pretty short road between implementing this into `Comment` as well. 

I think there are good arguments between `HasDate` vs. `DatedInterface` — though ultimately it's a pretty internal class that devs will not be interacting with directly very much. I think I'm slightly partial to `DatedInterface` b/c it's a little more explicit about its role (and could serve as a model for other interfaces, (`AuthoredInterface`?) that might be created in the future.

I'll let @pascalknecht and/or @acobster have a crack at formal review here, but everything I saw looked good!

lib/Post.php
```diff
@@ -1498,53 +1498,148 @@ public function paged_content() {
...
+	 *                                 as a default.
+	 *
 	 * @return string
 	 */
-	public function date( $date_format = '' ) {
-		$df       = $date_format ? $date_format : get_option('date_format');
-		$the_date = date_i18n($df, strtotime($this->post_date));
-		return apply_filters('get_the_date', $the_date, $df);
+	public function date( $date_format = null ) {
+		$format = $date_format ? $date_format : get_option( 'date_format' );
```
acobster: This and other instances of this pattern can all be updated to the shorthand ternary operator, `?:`:

```
$format = $date_format ?: get_option( 'date_format' );
```

lib/DatedInterface.php
acobster: I agree with @jarednova, interface names should be explicit about the fact that they're interfaces. `HasDate` sounds more like a trait to me.

gchtr: Let’s keep `DatedInterface` then 👍.

lib/Post.php
gchtr: I like your suggestion to use the more modern shorthand syntaxes. Updated in https://github.com/timber/timber/pull/2137/commits/1c84ffc1d313a686b9cbd8d1c0a73604e6ea1720.

acobster:
@gchtr hahahaha @ "Elvis operator"! I hadn't heard that term before. Love it.

gchtr:
@acobster When I learnt about it, they called it Elvis operator and I first thought that this was the official name 😆🕺🏻.

jarednova:
Took care of a few merge conflicts, all good now!

Issue #2142, opened 2020-01-03
ACF field 'content' override content 


antiseptikk:
I think i found a "problem" with the method `{{ post.content }}` and ACF fields. 
When we create ACF field named 'content', the post content output doesn't work and the field id was displayed. 

`{{ post.content }}` works for pages & posts without this ACF field.

Personally i think it's bad idea to create ACF field named `content`, but without Timber, it's possible to do without problem. 

I think it's good idea to have a special case to get the_content with filters applied with {{ post.content }} (without override). And if you want, you can use post.meta('content') to get the ACF field content.

```
<h1 class="h1">{{ post.title }}</h1>
<div class="article-body">
    {{ post.content }}
</div>
```

## Expected behavior

`post.content`
```
"real" post content
```

`post.meta('content')`
```
array:1 [▼
  0 => array:2 [▼
    "acf_fc_layout" => "bloc_file"
    "files_repeat" => false
  ]
]
```

## Actual behavior

`post.content`
```
field_5de67b6270234
```

`post.meta('content')`
```
array:1 [▼
  0 => array:2 [▼
    "acf_fc_layout" => "bloc_file"
    "files_repeat" => false
  ]
]
```
## Steps to reproduce behavior

Create ACF group and create field (in my case this is a flexible content) named `content` and fill it.

## What version of WordPress, PHP and Timber are you using?
WordPress 5.3.2, PHP 7.3, Timber 1.13.0

PR #2143, opened 2020-01-03
Avoid post content bypassing by ACF named 'content'

antiseptikk:
Proposal to avoid bypassing post content with the ACF named 'content'.

**Ticket**: #2142 

## Issue
When we create ACF field named 'content', the post content output doesn't work and the field id was displayed.

## Solution
Create special condition to get the_content with filters applied with {{ post.content }} (without override).

And if you want, you can use post.meta('content') to get the ACF field content.

I don't know if it's the best solution, but at the moment it's working very well on a project.

### DIFF
--- a/lib/Core.php
+++ b/lib/Core.php
@@ -67,7 +67,7 @@ abstract class Core {
 				}
 				if ( !empty($key) && $force ) {
 					$this->$key = $value;
-				} else if ( !empty($key) && !method_exists($this, $key) ) {
+				} else if ( !empty($key) && !method_exists($this, $key) && $key !== '_content' ) {
 					$this->$key = $value;
 				}
 			}
--- a/lib/Core.php
+++ b/lib/Core.php
@@ -57,6 +57,9 @@ abstract class Core {
 	 * @param array|object $info an object or array you want to grab data from to attach to the Timber object
 	 */
 	public function import( $info, $force = false ) {
+		$disallowed_properties = array();
+		$disallowed_properties = apply_filters('timber/post/disallowed_properties', $disallowed_properties);
+
 		if ( is_object($info) ) {
 			$info = get_object_vars($info);
 		}
@@ -67,7 +70,7 @@ abstract class Core {
 				}
 				if ( !empty($key) && $force ) {
 					$this->$key = $value;
-				} else if ( !empty($key) && !method_exists($this, $key) && $key !== '_content' ) {
+				} else if ( !empty($key) && !method_exists($this, $key) && !in_array($key, $disallowed_properties)) {
 					$this->$key = $value;
 				}
 			}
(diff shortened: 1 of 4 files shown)


jarednova:
@antiseptikk thanks so much for raising the issue in #2142 and supplying the fix here. First off, the good news. Thanks to the work on how we handle meta data in 2.x this is already "fixed" in the upcoming version of Timber (which we're furiously working on). 

But I've confirmed that this is an issue in 1.x through the test in 2bd2d0d . My question is whether to apply a quick patch like this and risk disrupting scenarios where a user has intentionally named an ACF custom field `content` and would like things to stay that way, thank you very much. Interested in @palmiak's take and anyone else from the community. Have you run into this problem too? Any other risks/concerns we should be thinking about here?

palmiak:
@jarednova I was always cautions about naming the fields, because I always had it in the back of my head that `post.meta('content')` would be the same as `post.content` which may lead to problems.

As you mentioned - this fix may lead to surprise outcome on some websites. It's potentiality "breaking". Even it's a bit of a bug (content shouldn't be overwritten).

I see two choices:
- leave as it is (and fix the docs about forbidden field names)
- do it in a progressive way - so let's add a filter first the give a possibility to change the behaviour and later change the default value with a notice.

I think I would go the second road.

jarednova:
@palmiak totally agree. @antiseptikk if you're interested in picking this up, I think maybe implementing some kind of filter that the `Integrations/ACF` class could hook into might be one way to go. That way the responsibilities of the diff't parts of the code are clear and it's something that users can override.

I realize it might seem like a bit of overkill compared to your simple patch here, but we've learned the hard way (again and again unfortunately) to be very careful around fixing certain quirks that sites might be relying upon.

antiseptikk:
hi @jarednova, i wrote a twig filter named "forcecontent" to force the real post content with "the_content" filter applied.

The approach is simple and should not break the actual websites using Timber.
Do you agree with this solution ?

jarednova:
Thanks @antiseptikk ! I realize in reading your code that it can be a little confusing with both Twig and WordPress having something called a "filter" (that means something totally diff't in both).

I was suggesting using WordPress's `apply_filters` / `add_filter` API to prevent the assignment of certain properties that are "reserved" by methods (like `content`). The way this would work is that there's an empty array of these disallowed properties that users could add to. In thinking about this, it actually isn't about ACF — it's about any property/method collision a user might want to prevent based on their particular install and theme.

I'm thinking the filter could be called something like `timber/post/disallowed_properties` (easy to change that name as we progress in this PR). In a user's theme `functions.php` they would hook into this and add whatever values they wanted (like `content` or maybe something custom like `my_method` that they have in a custom class). 

Before auto-assigning the custom field properties, `Post.php` would make sure that field name isn't in that array. If it is, it would skip the dynamic assignment so that the method is never overridden. 

Let me know if that makes sense!

antiseptikk:
I'm sorry for the misunderstanding.
Now it's working with this following line in `functions.php`.

```
add_filter('timber/post/disallowed_properties', function() {
    return array('_content');
}, 10);
```

For now, the hook is in `Integrations/ACF` class but we can change this if you want.

jarednova:
@antiseptikk goodish news on this! After I spent more time with this the other day I realized that what you got here is a straight-up BUG. The problem is that we use a field named `$_content` on the `Timber\Post` object to cache the data from `{{ post.content }}`. Because ACF names things with an underscore (so "content" becomes "_content") this was overwriting the data. 

Really appreciate the PR and you going with me on the proposed solution. This will be released in the next few days!

Issue #2094, opened 2019-10-03
Eliminate all direct instantiations of Timber\Core objects outside of Factories


acobster:
As part of the new API proposed in #2073, we're [making Core constructors protected](https://github.com/timber/timber/issues/2088). But first, we need to make sure we're not directly instantiating these objects outside of factories, or else...

✨ **Everything Will Break!** ✨

Replace all direct instantiations with calls to `Timber::get_post()` and friends.

acobster:
@jarednova @gchtr how would y'all feel about dropping the default `'hide_empty' => false` param when querying for terms, e.g.

```php
Timber::get_terms('tags');
Timber::get_terms(['taxonomy' => 'tags']);
```

I get where this is coming from, but keeping track of two levels of defaults makes things trickier from a user's perspective. If folks want to include "empty" terms they can always pass `'hide_empty' => false` explicitly, and we'd just need to make a note of that in the upgrade guide.

jarednova:
Yeah, I think I included that for Very Good Reasons™ at some point, but this feels like a great time to simplify things and use WP defaults (yeah, keeping track of "two sets of defaults" highlights that it's just adding complexity) — I say drop it!

acobster:
Alright - assuming you're on board, @gchtr, can you add that to the upgrade guide?

gchtr:
> Alright - assuming you're on board, @gchtr, can you add that to the upgrade guide?

I agree with Jared. Dropping it feels like the right thing to do. I already added a note to the Upgrade Guide.

acobster:
```
⚙ (2.x-posts-api) ~/vvv/www/timber $ grep -r --include=*.php --exclude-dir=wordpress --exclude=test-timber-post-getter.php 'new \\\?\(Timber\\\)\?Post( \?' .
⚙ (2.x-posts-api) ~/vvv/www/timber $ grep -r --include=*.php --exclude-dir=wordpress --exclude=test-timber-post-getter.php 'new \\\?\(Timber\\\)\?Attachment( \?' .
⚙ (2.x-posts-api) ~/vvv/www/timber $ grep -r --include=*.php --exclude-dir=wordpress --exclude=test-timber-post-getter.php 'new \\\?\(Timber\\\)\?Image( \?' .
⚙ (2.x-posts-api) ~/vvv/www/timber $ 
```

Done on `2.x-posts-api` branch.

Issue #2090, opened 2019-10-01
Update ::get_post() and ::get_posts() signature


acobster:
## Is your feature request related to a problem? Please describe.

Per #2073 we are disabling direct instantiation of Posts and forcing all 

### Update `Timber::get_post()`

Update the method signature to:

```php
public static function get_post( $query = false );
```

This gets rid of the second argument, `$PostClass`, whose functionality will be subsumed (much more nicely!) by [Class Maps](https://github.com/timber/timber/blob/2.x-docs-api/docs/guides/class-maps.md).

### Update `Timber::get_posts()`

Update the method signature to:

```php
public static function get_posts( array $query = [], array $options = [] );
```

This also eliminates the `$PostClass` param as above, and consolidates Timber-specific options into a single array. Note that both are optional. This method will behave more or less the same as it has within archive templates.

## Alternatives and context

* [Upgrade guide proposal](https://github.com/timber/timber/blob/2.x-docs-api/docs/upgrade-guides/2.0.md)
* [Timber 2.0 Context Guide proposal](https://github.com/timber/timber/blob/2.x-docs-api/docs/guides/context.md)
* [Class Maps](https://github.com/timber/timber/blob/2.x-docs-api/docs/guides/class-maps.md)
* [Posts API in Timber 2.0 proposal](https://github.com/timber/timber/blob/2.x-docs-api/docs/guides/posts.md)
* More discussion: #2073 #1793

gchtr:
Looking good 👍.

What we have to consider for `Timber::get_posts()` is #1957. In #1813 we added logic to mimick the default behavior for `get_posts()` in WordPress. This produced errors, so we made it optional with a filter in #1989. For 2.0, I guess we could **deprecate that filter** and add a proper description with a [link to the documentation](https://github.com/timber/timber/blob/2.x-docs-api/docs/guides/posts.md#the-difference-to-get_post) that lists the differences and the arguments that need to be passed to `Timber::get_posts()` to mimick the default behavior.

I’m pinging @palmiak here, because he started the discussion about `Timber::get_posts()`.

acobster:
Done!

PR #2145, opened 2020-01-06
2.x Update function signatures for ::get_post() and ::get_posts()

gchtr:
## Issue

**Ticket**: #2090, #2094 related to `Timber\Post` objects.

### Add `Helper::doing_it_wrong()`

~~I added a `Helper::doing_it_wrong()` function, that can be used in combination with a `@expectedIncorrectUsage` tag in the DocBlock of a test.~~

~~This is helpful for functionality that will not be deprecated, but removed, like certain function parameters in `Timber::get_post()` and `Timber::get_posts()`.~~

Added in https://github.com/timber/timber/pull/2196.

### Remove query methods that won’t work anymore

The following methods for retrieving post were removed. Yes, removed and not deprecated, because there will be no backwards compatibility.

- **Query strings** - Removed the method of getting posts via a query string (like `post_type=portfolio`) and added a doing-it-wrong message.
- **Getting a post by post slug** – Added a doing-it-wrong message to help developers in the transition. Alternatively, we could add a new function `Timber::get_post_from_slug()` that could do this (Discussion in #2073).
- **Getting a post by post name** – This case is catched with the same doing-it-wrong message as getting a post by post slug (Discussion in #2073).

### Deprecate `Timber::query_post()` and `Timber::query_posts()`

The `Timber::query_post()` and `Timber::query_posts()` functions weren’t properly deprecated yet.

### Return a `PostCollection`

As per the new API, we always return a `PostCollection` when using `Timber::get_posts()`. That’s why I changed the current implementation through `PostGetter::get_posts()` to use `true` for the `return_collection` parameter in `Timber::get_posts()`.

However, a lot of tests still break, because `Timber::get_posts()` doesn’t return a `PostCollection()` with the ability to handle pagination yet. I guess we’ll first have to make `Timber::get_posts()` use the new API before this can be merged. @acobster How would you proceed here?

## Impact

This will break backwards compatibility with how `Timber::get_post()` and `Timber::get_posts()` worked before. But I added checks and warnings to help developers in the transition.

## Usage Changes

See #2073.

## Considerations

To keep discussion in one place, I added some remarks and questions in https://github.com/timber/timber/pull/2073#issuecomment-570954735.

## Testing

Yes, but work in progress.

## Todo

- [ ] Define default `$options` for `Timber::get_posts()`
- [ ] Make `Timber::get_posts()` return a `PostCollection` that can handle pagination.
- [ ] Use `Iterable` as return type for `Timber::get_posts()`, see https://github.com/timber/timber/pull/2073#issuecomment-571814739.


gchtr:
As per https://github.com/timber/timber/pull/2073#issuecomment-571375271, I guess I should change this against the `2.x-factories` branch.

lib/Timber.php
```diff
@@ -124,67 +124,164 @@ class_alias( 'Timber\Timber', 'Timber' );
...
+		if ( func_num_args() > 1 ) {
+			// @todo Add link to Class Maps documentation.
+			Helper::doing_it_wrong(
+				'Timber::get_post()',
+				'The $PostClass parameter for passing in the post class to use in Timber::get_post() was removed in Timber 2.0. Use Class Maps (LINK HERE) instead.',
+				'2.0.0'
+			);
+		}
+
+		// @todo Will this be handled here or inside factories, or somewhere else?
```
acobster: For UserFactories we decided the analogous `get_current_user_id` would be in `Timber::get_user()`, so I think we should be consistent and do this here. Definitely not in a factory in any case.

lib/Timber.php
```diff
@@ -124,67 +124,164 @@ class_alias( 'Timber\Timber', 'Timber' );
...
 	/**
 	 * Query post.
 	 *
 	 * @api
-	 * @deprecated since 2.0.0 Use `new Timber\Post()` instead.
+	 * @deprecated since 2.0.0 Use `Timber::get_post()` instead.
 	 *
-	 * @param mixed  $query
-	 * @param string $PostClass
+	 * @param null|int|\WP_Post $post
```
acobster: this should change from `null` to `false`

lib/Timber.php
```diff
@@ -124,67 +124,164 @@ class_alias( 'Timber\Timber', 'Timber' );
...
+	 * @param null|int|\WP_Post $post Optional. Post ID or WP_Post object to get a Timber Post
+	 *                                object. Default `null`.
+	 *
+	 * @return \Timber\Post|bool `Timber\Post` object or an instance of a child class of
+	 *                           `Timber\Post` if a post was found, `false` if no post was found.
 	 */
-	public static function get_post( $query = false, $PostClass = 'Timber\Post' ) {
-		return PostGetter::get_post($query, $PostClass);
+	public static function get_post( $post = null ) {
+		if ( is_string( $post ) ) {
```
acobster: sometimes we get IDs as strings, e.g. from `$_GET`. Factories support this already; should we support it here? We could add an `!is_numeric()` check to be certain that we definitely don't have an ID. Only consideration there would be if the `post_title` were a purely string...which would be weird and edge-casey. I feel like that's an okay trade-off...?

gchtr: Good call! Resolved in e5e0a4543b85b2110f0d69552e6fe6b16537ad4e.

lib/Timber.php
gchtr: Resolved in f2d754502dd44436cf37993f3d2ff141350759a6.

lib/Timber.php
gchtr: Resolved in a2abbb38096cd5abc762c4827c37d037b1215a78.

gchtr:
@acobster I guess I should rebase this pull request against `2.x-factories`, too?

jarednova:
@gchtr are you waiting on me for a review here? I noticed the *WIP* tag so I had assumed you were still at-work on this. 

There are a number of [failing tests](https://travis-ci.com/github/timber/timber/jobs/347276809), mainly related to pagination. Do we need to resolve those here, or is that better done once the merge you and @acobster were discussing is complete?

acobster:
@jarednova I'm assuming there's going to be a lot of lifting to get these two methods working as we get them over to Factories. That work should probably still be done in `2.x-factories` IMHO before we bring it into mainline `2.x`.

acobster:
@gchtr @jarednova I just realized I may have misunderstood the question. I don't really have an opinion about whether these fixes happen in `2.x-post-functions` or `2.x-factories`, in case that's what you were asking.

Sorry, my brain has been kinda everywhere recently. Maybe you know the feeling. :upside_down_face:

jarednova:
@gchtr the PR has my 👍  on the substance you raise in the original ticket, but I think we should work out (not necessarily solve, but just we're on the same page on this todo you left ...):

> Make Timber::get_posts() return a PostCollection that can handle pagination

Right now `PostQuery`s are essentially this — a subclass of `PostCollection` that have methods to handle for pagination and the associated data/logic. I think it's crucial that pagination be attached to the returned object as opposed to a separate request. 

I'm imagining that most of this code can be brought over into a new class w/o too much hassle; but just wanted to make sure it wasn't lost.

Another thing I encountered when doing some investigation on my local is just the number of jumps from class-to-class a call to `Timber::get_posts` makes. I'm very much in favor of simplifying this as we adapt to factories even if it means compromising some backwards compatibility in places. The current chain of functions/methods is dizzying (and can't be great for performance either).

You might already be aware or have plans for these. But just wanted to call them out so they don't bite us later!

acobster:
Ported into #2316 in https://github.com/timber/timber/commit/3fabc800f33e85c3e3de15be62d0f43e2f1f2426. Calling this good.

PR #2146, opened 2020-01-06
2.x Improve Helper::deprecated()

gchtr:
## Issue

The `Timber\Helper::deprecated()` function contains some code that is unnecessary. And I saw some other optimizations that would make the function more readable.

## Solution

- Remove check for `__()` / handling of translated strings. We don’t have any translatable strings in Timber. Otherwise, we would need to have a `timber` text domain that we would have to add to strings.
- Remove unneeded check for `WP_DEBUG`.
- Use bailout instead of if-wrapping.
- Add info about using `@expectedDeprecated` in tests.

## Impact

None.

## Usage Changes

None.

## Considerations

None.

## Testing

Nope.

### DIFF
--- a/lib/Helper.php
+++ b/lib/Helper.php
@@ -273,10 +273,15 @@ class Helper {
 	/**
 	 * Triggers a deprecation warning.
 	 *
+	 * If you want to catch errors like these in tests, then add the @expectedDeprecated tag to the
+	 * DocBlock. E.g.: "@expectedDeprecated {{ TimberImage() }}".
+	 *
 	 * @api
+	 *
 	 * @param string $function    The name of the deprecated function/method.
 	 * @param string $replacement Function to use instead.
 	 * @param string $version     When we deprecated this.
+	 *
 	 * @return void
 	 */
 	public static function deprecated( $function, $replacement, $version ) {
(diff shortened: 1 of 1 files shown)


lib/Helper.php
```diff
@@ -273,45 +273,54 @@ public static function warn( $message ) {
...
+		} else {
+			$error_message = sprintf(
+				'%1$s is <strong>deprecated</strong> since Timber version %2$s with no alternative available.',
+				$function,
+				$version
+			);
+		}
 
+		// phpcs:disable WordPress.PHP.DevelopmentFunctions.error_log_trigger_error
+		// phpcs:disable WordPress.Security.EscapeOutput.OutputNotEscaped
```
jarednova: This all looks much cleaner and readable. My only Q is these two lines — are they necessary for something you're using with PHP_CodeSniffer? or just accidentally left in?

gchtr: Yes, I left them there deliberately. As described in https://github.com/timber/timber/blob/master/CONTRIBUTING.md#use-php_codesniffer-to-detect-coding-standard-violations, we are trying to stay as close possible to the standards. However, sometimes we conciously do something that wouldn’t be allowed under the standards by whitelisting certain CodeSniffer rules.

By adding these comments, CodeSniffer will not flag the issue *and* we have some kind of documentation for where we deviate from the standards.

jarednova:
Cool, just wanted to make sure — let's merge!

PR #2148, opened 2020-01-07
2.x factories test fixes

jarednova:
**Ticket**: #2073 

## Issue
Fixing some failing comment tests!

## Solution
Comments were being created w/o relationship to a post object. This meant when `CommentFactory::get_comment_class` went to try and figure out what class to use it was like, "well I can't match this up to a class b/c it doesn't have a post associated with it (thus no `post_type`)

Making this a PR to your branch so that you can double-check that I did things right and this is your intent


jarednova:
@acobster the remaining errors stem from Comment's constructor being protected vs. public. Before I went and changed this wanted to see if there was a deliberate intent here

PR #2150, opened 2020-01-07
2.x Optimize post date filters

gchtr:
**Ticket**: #2137

## Issue

In version 4.6, WordPress added additional parameters for some of the filters that we also apply.

## Solution

Let’s update the applied filters with the new parameters. And add some more documentation.

But behold. Sometimes these filters expect a `WP_Post` object as the last parameter. I think we can’t return `$this` / the `Timber\Post` object, because that could lead to compatibility issues in case any plugin uses this filter.

- `get_the_date` – Expects `int|WP_Post` The post object or ID.
- `get_the_modified_date` – Expects `WP_Post|null` WP_Post object or null if no post is found.
- `get_the_time` – Expects `int|WP_Post` The post object or ID.
- `get_the_modified_time` – Expects `WP_Post|null` WP_Post object or null if no post is found.

As you can see, `get_the_modified_date` and `get_the_modified_time` expect either a `WP_Post` or `null`. So we can either return `null` or return a `WP_Post` by using `get_post( $this->ID )`.

## Impact

See Considerations.

## Usage Changes

None. Except better compatibility with WordPress.

## Considerations

When we use `get_post()`, WordPress will get the post from the cache. However, there might be a performance impact when some additional filters are applied. I expect this to be marginal. Still, we should probably strive for best performance.

The alternative would be to return `null`.

What other alternatives do we have?

## Testing

None.

### DIFF
--- a/lib/Post.php
+++ b/lib/Post.php
@@ -1522,7 +1522,11 @@ class Post extends Core implements CoreInterface, MetaInterface, DatedInterface,
 	}
 
 	/**
-	 * Gets the publishing date of the post to use in your template.
+	 * Gets the publishing date of the post.
+	 *
+	 * This function will also apply the
+	 * [`get_the_date`](https://developer.wordpress.org/reference/hooks/get_the_date/) filter to the
+	 * output.
 	 *
 	 * @api
 	 * @example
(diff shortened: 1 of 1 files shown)


jarednova:
@gchtr looks good to me. On the Q of `get_post`, I think the only alternative would be to store the results of that directly on the `Post` object. How much does that really save us?  I ran some tests and over 10,000 attempts to call `modified_time` the difference is about 5% (440 milliseconds vs 418) — and that's *10,000* calls to that method.

When I reduce the number of calls to `1` they both report a time of `0` milliseconds. My read is that this means that WP is storing the result of `get_post` somewhere, it just takes some time to access that cache.

jarednova:
FYI — I changed the base on this to `2.x` b/c `2.x-post-date` has already been merged in

Issue #2151, opened 2020-01-07
Provide notification for WP versions no longer supported


jarednova:
Given that 2.x will be only support WordPress 5.3 and greater, we should provide notice to users _before_ they upgrade

- [x] Consider providing notice in Upgrade guide and other docs
- [x] Consider providing notice in source code itself (`Timber.php`? `composer.json`?)
- [x] Consider providing info via a WP admin notice

PR #2156, opened 2020-01-08
Provide notice for min. WordPress version

jarednova:
**Ticket**: #2151 

## Issue
As discussed in #2135, we're dropping support for WordPress versions prior to 5.3 in 2.x. To do this, we need to let people know.

## Solution
Let 'em know! Through the upgrade guide, ensuring the `require-dev` in composer is correct (we can't do comments/annotations in here, but this is one way to indicate) and an admin notice in case someone is running a WP version that's older than 5.3.

Once inside of the `Admin` class I realized we could get rid of the bulk of what's here since it modifies the plug-in entry for Timber (which we're also not supporting in 2.x, in favor of Composer/Packagist only)

## Impact
This helps let devs know to upgrade their WordPress.

## Considerations
We could copy over some of this code into 1.x (perhaps upon the last release to WP.org) to let people know that there won't be any further plugin support and to upgrade to both WP 5.3+ and Timber 2.x+ for new sites

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
     "twig/cache-extension": "dev-master"
   },
   "require-dev": {
-    "johnpbloch/wordpress": "5.*",
+    "johnpbloch/wordpress": "5.3.*",
     "phpunit/phpunit": "5.7.16|6.*",
     "squizlabs/php_codesniffer": "3.*",
     "wp-coding-standards/wpcs": "^2.0",
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
     "twig/cache-extension": "dev-master"
   },
   "require-dev": {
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -34,7 +34,7 @@
     "twig/cache-extension": "dev-master"
   },
   "require-dev": {
-    "johnpbloch/wordpress": "5.*",
+    "johnpbloch/wordpress": "5.3.*",
```
acobster: If I'm reading Composer's [semver docs](https://getcomposer.org/doc/articles/versions.md#summary) right, this is saying "any version of 5.3 _but no lower or higher_." I think we want `>=5.3.0`

jarednova: Yup! good catch

Issue #2155, opened 2020-01-08
Question: How should we treat the co-authors-plus integration wrt Factories?


acobster:
**NOTE: A first-stab PR for this is forthcoming, but as I call out below I think this deserves its own issue.**

## The main issue: which User class to instantiate?

In the `Integrations\CoAuthorsPlus`, the `authors()` method (hooked onto the `timber/post/authors` filter) determines which class to instantiate:

```php
	public function authors( $author, $post ) {
		$authors = array();
		$cauthors = get_coauthors($post->ID);
		foreach ( $cauthors as $author ) {
			$uid = $this->get_user_uid( $author );
			if ( $uid ) {
				$authors[] = new \Timber\User($uid);
			} else {
				$authors[] = new CoAuthorsPlusUser($author);
			}
		}
		return $authors;
	}

	protected function get_user_uid( $cauthor ) {
		// if is guest author
		if( is_object($cauthor) && isset($cauthor->type) && $cauthor->type == 'guest-author'){
			// if have have a linked user account
			global $coauthors_plus;
			if( !$coauthors_plus->force_guest_authors && isset($cauthor->linked_account) && !empty($cauthor->linked_account ) ){
				return $cauthor->linked_account;
			} else {
				return null;
			}
		} else {
			return $cauthor->ID;
		}
	}
```

The logic around whether `$uid` is truthy appears to depend on whether it's for a "guest author" (I gather this is just arbitrary data submitted at post creation time - not corresponding to a `wp_users` row), whether guest authors are being "forced" (no idea what this means - plugin setting?), etc. In other words, I'm not terribly certain what this code does. :)

So my main question is: how should this interface with UserFactory and the user class map? I'm taking a stab at implementing what I _think_ should happen, but I'd like to get a review specially for this.

## Handling `stdClass` instances

Another concern is that `get_coauthors()` appears to sometimes return instances of `stdClass`. I think these represent "guest authors." Trouble is, `stdClass` is not supported by `UserFactory::from()` - in this case [it calls `from_post_object`](https://github.com/timber/timber/blob/2.x-factories/lib/Factory/UserFactory.php#L21) (copy/pasta, will fix) which eventually throws an `InvalidArgumentException`. The reasoning here is that at the heart of @gchtr's design there's an assumption that the callback to the `timber/class/classmap` filter eventually gets a `WP_User`. This is a strong guarantee and the entire class map API gets weaker if we throw this away.

[...]

PR #2157, opened 2020-01-08
2.x coauthors fixes

acobster:
**Ticket**: #2155

## Issue

The Co-Authors Plus integration as it currently exists is incompatible with Factories and Class Maps. This is an attempt to make them compatible.

## Solution

I describe the proposed solution in more detail in the issue write-up, but basically what I'm doing here is treating "guest authors" as a special case of the `timber/user/classmap` filter: instead of calling a Factory, it applies the filter directly.

## Impact

Should be none.

## Usage Changes

None, other than adding support for class maps in this integration.

## Considerations

I may be wrong about everything. :D

## Testing

Updated tests where it made sense and added a test specifically for dealing with overriding the class instantiated for guest authors.


acobster:
Detaching this from `2.x-user-factories` for now. We can revisit later.

jarednova:
Oy! That wasn't any fun. But I think it's done?!?!?! Let's see how the tests go. There are a couple of things to review when it comes to `new static();` in the constructor(s)

jarednova:
The three static analysis errors are ...

```
  Line   Comment.php                                            
  123    Unsafe usage of new static().                          
  Line   Integrations/CoAuthorsPlusUser.php                     
  14     Unsafe usage of new static().                          
  Line   User.php                                               
  115    Unsafe usage of new static().                          
```

I'm a little unsure of the best way to resolve. Given that users _shouldn't_ be extending the `__constructor()` method should we just make those final @acobster @gchtr?

acobster:
Not sure I necessarily agree that users shouldn't be extending the constructor. We just don't want constructors being called _directly_, so that all instantiation happens via Class Maps and funneled through `::build()`. It's probably a rare case that you actually need a custom constructor, but I could see some default object state being useful, so if we're going to cut off that possibility completely I'd at least want to know why.

Not really sure what PHPStan's issue with this actually is. I found the commit where they [introduced](https://github.com/phpstan/phpstan/commit/7f5582e79c2cd2ed7399cc6251ec1d0c30a4b704) this check, but no justification for it.

cc @szepeviktor

szepeviktor:
Please see https://3v4l.org/On7K3

acobster:
So the issue is that type declarations on a subclass's constructor might conflict with the declarations on the parent? I don't think that's an issue here, since our constructor takes no arguments. If people want to override `__construct()`, they can do so with zero arguments. Otherwise they will find out right away that required constructor args on CPT classes will cause a Fatal Error, and we can document advanced use-case somewhere if we deem it necessary. So, to me this seems like a docs problem and not a safety issue. My proposal is to ignore the `new static()` rule in this case.

Thoughts?

szepeviktor:
> will cause a Fatal Error

PHPStan is here with us to help prevent just that :)

acobster:
Understood, and it's a great tool for that. :) All I'm saying is that at some point we need to look at the costs and benefits of following its rules to the letter. And in this _specific instance_ - which I argue is outside the scope of the problem that example lays out - I believe the costs outweigh the benefits. We'll never be able to standardize our way out of every fatal error, and even if we prevent this one from happening, without a guide for how to do it the right way, our users will just stumble onto `PHP Fatal error:  Cannot override final method User::__construct()`. Not a lot of benefit there from my perspective, so we might as well just provide a good advanced guide and keep each constructor as-is for flexilibity. So for now I'm standing by my recommendation of annotating these instances to be ignored and adding docs for when and how to override core constructors.

Would love to hear from @jarednova or @gchtr on this too!

szepeviktor:
PHPStan has a cost that may seem very high.
Although in reality PHPStan's benefit greatly exceeds its cost.

acobster:
I agree in principle, but I'm wondering what you see as being the long-term benefit in this particular case? Are you saying you disagree with my proposal? If so, why?

szepeviktor:
I run web applications for SaaS and daily see that they cannot make a sustainable product without these basic tests:

1. syntax check
1. coding standard
1. static analysis (PHPStan Level 4+)

That is my daily observation.

PR #2159, opened 2020-01-08
2.x Fix deprecated() function

gchtr:
**Ticket**: #2146, #2154

## Issue

There are many test errors in 2.x, because I got a bailout wrong in #2146. Sorry for that.

## Solution

Fix bailout.

## Impact

Tests should work again.

## Usage Changes

None.

## Considerations

None.

## Testing

None.

### DIFF
--- a/lib/Helper.php
+++ b/lib/Helper.php
@@ -298,7 +298,7 @@ class Helper {
 		 *
 		 * @param bool $trigger Whether to trigger the error for deprecated functions. Default true.
 		 */
-		if ( apply_filters( 'deprecated_function_trigger_error', true ) ) {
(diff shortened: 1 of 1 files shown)


szepeviktor:
just on ❕

gchtr:
> all it took was just one `!`, huh?

Yeah, apparently 🤪. Even if it’s only a bugfix for a mishap of mine, this has to be my smallest pull request ever.

acobster:
@jarednova  looks like I was late to the party and #2154 was the wrong approach.

Thanks for this fix @gchtr  - I merged this back into `2.x-factories` as well.

Issue #2093, opened 2019-10-02
Factories reboot


acobster:
## Is your feature request related to a problem? Please describe.

The requirements for Factories has changed a lot since the long discussion on #1793 and #1219 :

* Factories will no longer be exposed publicly. They will constitute a completely internall API to be used by the `Timber\Timber` class.
* Instead of exposing Factories externally, we're exposing a small set of filters that will be used _inside_ Factories to instantiate the correct class for a given post type, taxonomy, etc.
* The requirements for Factories is now being driven by the new [API proposal](https://github.com/timber/timber/pull/2073) for `Timber::get_posts()` and friends.

## Goals

The goals for implementing Factories are:

* To provide a centralized layer for instantiating `Timber\Core` instances (`Post`, `Term`, etc.) and their corresponding collection classes (e.g. `PostCollection`).
* To eventually replace `PostGetter` etc., to be called directly from `Timber::get_*` methods (e.g. `::get_posts`).
* To provide a centralized place to implement filters like `timber/post/classmap` as described in the [Class Map proposal](https://github.com/timber/timber/blob/2.x-docs-api/docs/guides/class-maps.md).
* To incorporate the best ideas from #1793 and #1219 in a way that matches up with the new API proposal

FlyingDR:
As promised - here are some minor things that I've noticed into factories code:

[This](https://github.com/timber/timber/blob/891f3132a7951a673c163aa3cde22a06fb8324aa/lib/Factory/PostFactory.php#L99-L101) is not completely correct because you have different type checking rules [here](https://github.com/timber/timber/blob/891f3132a7951a673c163aa3cde22a06fb8324aa/lib/Factory/PostFactory.php#L16-L18). 

Actually it looks like whole purpose of `PostFactory::is_numeric_array()` is to distinguish it from `WP_Query` arguments passed as array. If I'm correct - then it may be better to check if given array is associative or not, because `WP_Query` arguments array can be only associative and, for example `['post_id'=>123]` will pass your current check.

It may look something like this:
```php
private function is_numeric_array($arr): bool
{
    return is_array($arr) && empty(array_filter($arr, '\is_string', ARRAY_FILTER_USE_KEY));
}
```

FlyingDR:
[...]

Issue #2088, opened 2019-10-01
Make core constructors protected


acobster:
## Is your feature request related to a problem? Please describe.

Per #2073 , we are moving away from direct instantiation of `Post`, `Term`, etc. objects in favor of `Timber::get_*` methods. Using [Class Maps](https://github.com/timber/timber/blob/2.x-docs-api/docs/guides/class-maps.md), we want to make sure that the correct classes are instantiated for their respective post types, taxonomies, etc. To that end, we need to make sure these objects aren't instantiated directly, but are created in such a way that the proper hooks are applied before instantiation.

To this end, we need to disable instantiating calling these constructors.

## Describe the solution you’d like

- [ ] Define a static `Timber\Post::build()` method that takes a `WP_Post` instance and returns an instance of the called class, e.g. a `Timber\Post`.
- [ ] Make `Timber\Post::__construct()` protected.
- [ ] Evaluate whether `Comment`, `Image`, `Attachment`, and `PostQuery` need to have their own special `build` methods and protected constructors.
- [ ] Repeat for analogs `Term`, `Comment`, `Image`, `Attachment`, `PostQuery`, `Menu`, and `User` as appropriate.

## Describe alternatives you’ve considered

@gchtr threw out the question of whether it was possible to throw a _doing-it-wrong_ exception if the constructor is called directly (presumably from outside Timber core). I don't know of a way to do this without either ephemerally setting global variables or checking the call stack (i.e. `debug_backtrace()`) from within the constructor, both of which would be very hacky.

jarednova:
Whoa, @acobster is on 🔥

gchtr:
I wouldn’t go down a hacky road that involves `debug_backtrace()`. Your proposed solution makes sense to me 👍.

acobster:
TODO: also protect `Menu[Item]` constructors in #2292

PR #2149, opened 2020-01-07
Introduce PHPStan

szepeviktor:
Closes #1986

Issue #2144, opened 2020-01-05
Update Object Getter function names in Twig


gchtr:
In https://github.com/timber/timber/pull/2073#issuecomment-535720949, we discussed whether we should keep the `Post`, `Term`, … functions in Twig:

> > For the API in Twig, we still have Post, PostQuery, etc. For consistency, should we consider renaming these functions to get_post(), get_posts(), etc?
> 
> I think renaming and deprecating the old Twig functions is probably a good idea. Seeing "Use `Timber::get_posts()` instead of `new Timber\PostQuery()`" and then a paragraph later seeing "You can also use PostQuery in Twig" is a little jarring, and it's one less thing for new users to have to remember.

This is about the following functions in Twig:

https://github.com/timber/timber/blob/ca43edc0aa6c12948a71c36d7e266d559ed26ffc/lib/Twig.php#L49-L70

This leaves us with the following todo:

- [x] Deprecate `Post`, add `get_post()`
- [x] Rename `PostQuery` to `get_posts()` (no need to deprecate, added for 2.x)
- [x] Deprecate `Term`, add `get_term()`
- [x] Deprecate `User`, add `get_user()`
- [x] Deprecate `Image`, add `get_image()`
- [x] Rename `Attachment` to `get_attachment()` (no need to deprecate, added for 2.x)

PR #2160, opened 2020-01-09
2.x user factories

acobster:
**Ticket**: #2073 #2088 #2094 

## Issue

Updating the Users API per discussion in #2073.

## Solution

Extensive discussion of the solution is in #2073 and #2093. In short:

* `Timber::get_user` and `Timber::get_users()` no longer take a `$UserClass` arg; users must now rely on the `timber/user/classmap` filter to achieve polymorphic instantiation.
* The public Timber functions use Factories under the hood, which in turn apply `timber/user/classmap`.

## Impact

Discussed in docs.

## Usage Changes

Also discussed in the docs.

## Considerations

I'm not sure what the new usage of the Twig `{{ User(...) }}` fn should be. Currently it supports arrays of IDs, which seems a bit astonishing. I'd be in favor or dropping support for this, or at least deprecating that use-case and providing a new `{{ Users(...) }}` (or `{{ UserCollection(...) }}`?) function.

Speaking of which, do we need a `UserCollection` class? I kind of assumed in the back of my mind that we'd want Collections for Users and Terms (we already have `CommentThread`), but thinking more critically about it, the Loop is the main reason we have `PostCollection`, and there's not really an analog to that with Users or Terms. I could see there being a use-case for pagination or some other advanced query-related stuff, but I'm not sure. Thoughts?

**Edit:** oh, and I'm hoping one of you can look at #2157 before we merge this in. :)

## Testing

Tests have been refactored and removed where appropriate. The `UserFactory` class has some prety thorough tests as well. It would be great add to tests for any edge cases y'all can think of.


gchtr:
> I'm not sure what the new usage of the Twig `{{ User(...) }}` fn should be. Currently it supports arrays of IDs, which seems a bit astonishing. I'd be in favor or dropping support for this, or at least deprecating that use-case and providing a new `{{ Users(...) }}` (or `{{ UserCollection(...) }}`?) function.

I think this was applied to match the behavior of other other types (post, terms, ...), without thinking about a use case.

But we will deprecate the Twig functions anyway and use the same function names that we use in PHP, see https://github.com/timber/timber/issues/2144. That would mean we should have a `{{ get_users(...) }}` function as well. The same goes for the other objects.

> Speaking of which, do we need a `UserCollection` class? I kind of assumed in the back of my mind that we'd want Collections for Users and Terms (we already have `CommentThread`), but thinking more critically about it, the Loop is the main reason we have `PostCollection`, and there's not really an analog to that with Users or Terms. I could see there being a use-case for pagination or some other advanced query-related stuff, but I'm not sure. Thoughts?

We discussed this quite some time ago in https://github.com/timber/timber/issues/1689. My take is that we don’t need UserCollections, because the loop isn’t involved and usually we don’t need pagination. The same goes for Terms. In case we ever need that in the future, we can add it later.

I will have a thorough look at this later this week.

acobster:
@gchtr ok, I remember that discussion now. I guess we can always add a `::get_users()` option to return a `UserCollection` instead of an array if we decide to implement it. It also occurs to me that we should probably have a `users.md` among the [Guides](https://github.com/timber/timber/tree/2.x-docs-api/docs/guides). Do you have the bandwidth to write one up?

gchtr:
👍

> It also occurs to me that we should probably have a `users.md` among the [Guides](https://github.com/timber/timber/tree/2.x-docs-api/docs/guides) [](/timber/timber/tree/HEAD@{2020-01-09T20:20:57Z}/docs/guides). Do you have the bandwidth to write one up?

Yes, I will write one up 👍

lib/Timber.php
```diff
@@ -215,6 +216,65 @@ public static function get_term( $term, $taxonomy = 'post_tag', $TermClass = 'Ti
...
+	================================ */
+
+	/**
+	 * Get one or more users as a collection.
+	 * @api
+	 * @param array   $query The params for the WP_User_Query to performed
+	 * @param array   $options optional; none are currently supported
+	 * @return mixed
+	 */
+	public static function get_users( array $query = [], array $options = [] ) : Iterable {
```
gchtr: Should we really add an `$options` parameter if there are no options supported yet?

lib/Factory/UserFactory.php
```diff
@@ -35,10 +35,12 @@ public function from($params) {
...
 
 	protected function from_id(int $id) {
-		return $this->build(get_user_by('id', $id));
+		$wp_user = get_user_by('id', $id);
+
+		return $wp_user ? $this->build($wp_user) : null;
 	}
 
-	protected function from_post_obj(object $obj) : CoreInterface {
+	protected function from_user_object(object $obj) : CoreInterface {
```
gchtr: The `object` type hint is only supported in PHP 7.2+. So I guess we can’t use it yet.

lib/Factory/UserFactory.php
```diff
@@ -35,10 +35,12 @@ public function from($params) {
 	}
```
gchtr: I’m getting a "Missing return statement" warning here.

acobster: I added an explicit `return null` at the end of this method, but that's only a guess as to what this means. I'm not seeing that warning on my end. Are you getting it running `phpunit`?

lib/Factory/UserFactory.php
acobster: Removed.

lib/Timber.php
acobster: Not sure about this one. This was back when I thought we'd return a `UserCollection`, so adding such an option might be imminent. So not as necessary. However, it does still come down to what constitutes a "breaking change" and the kinds of use-cases we explicitly support.

If don't ship with `$options` but then add it later, then someone could literally do `MyClass extends Timber` and override this method with single arity on 2.0 and then encounter a breakage just by updating to a minor version when we add the param. So that's technically a breaking change, even if we can't imagine why someone would do that instead of just using the filters we provide. But as long as there's a possibility, I think it might be good to signal that intention and leave the door open.

Curious to hear from @jarednova here.

jarednova: So it's basically like no options yet, but (perhaps) options to come? A good case could be made either way, but it seems like it could be helpful in remembering for when options get added this is where they go in order to match patterns seen elsewhere in the code base. Do I have that right?

acobster: Yes.

lib/Factory/UserFactory.php
gchtr: I’m seeing this in PhpStorm as an [Inconsistent return points](https://www.jetbrains.com/help/phpstorm/php-inconsistent-return-points.html) warning.

I’m not sure what the best solution is here. A `@return mixed|void` would also make the warning go away, but should we ever return `void` if we have other returns in a method?

If other methods like `UserFactory::from_id()` return `null` when no `Timber\User` could be returned, wouldn’t it make sense to be explicit and return `null` here, too?

lib/Timber.php
gchtr: I didn’t think about the breaking change if someone extends `Timber\Timber`. Under these circumstances, it makes total sense. And even if we don’t have a use case now, there can very well be a use case in the future, as we have learned time and time again while developing and maintaining Timber.

Let’s leave it, then! 👍

lib/Factory/UserFactory.php
acobster: `::from_id` already returns explicitly...right? It just returns the result of the ternary expression.

I added some doc comments that I feel are appropriately specific. I don't care too much about appeasing PHPStorm since we're already [adding PHPStan](https://github.com/timber/timber/pull/2149) in parallel. Can't please everybody. :)

acobster:
@gchtr I just got all unit tests passing for this locally.

Per discussion on #2157 I'm dropping support for Co-Authors Plus for now, as I just don't have time to continue supporting it. Discussion can continue on that PR, where the `2.x-coauthors-fixes` branch contains all the code needed to re-add support for it.

How are you feeling about this PR?

lib/Timber.php
```diff
@@ -215,6 +216,65 @@ public static function get_term( $term, $taxonomy = 'post_tag', $TermClass = 'Ti
...
+	 * - email
+	 * - login
+	 * @param int|string $value The value for $field
+	 * @return \Timber\User|null
+	 */
+	public static function get_user_by( string $field, $value ) {
+		$wp_user = get_user_by($field, $value);
+
+		if ($wp_user === false) {
+			return false;
```
jarednova: Just wanted to note that I changed this from returning `null` to `false`. It'd seem to me that if no user is found `false` is a more logical representation

gchtr: @jarednova To me, this is a question of: What do we return in any of the `Timber::get_*` functions when no object could be found based on the input. Is it `null` or `false`? And does it matter?

jarednova:
Extensive work here @acobster! I found a few tiny things, but overall this is looking super. Take a look and let me know if you agree with my one substantive change (returning `false` instead of `null` when no user found via `Timber::get_user_by` and other similar methods). This is the pattern I saw in `CommentFactory` and what would make the most logical sense to me*

Merge if you agree! (or we can discuss what makes the most sense here and close that out)

`*` I realize there are a few other places in the 2.x codebase not related to factories where things are returning `null` instead of `false`. I'll take a separate PR/issue to investigate and resolve these

jarednova:
... and also to close the loop on CoAuthorsPlus. I think #2157 is the perfect place to either resolve or abandon attempting to make it compatible. Let's not let that block us here

acobster:
@jarednova yeah, I'm assuming the magic Timber getters will still return `false` when they don't find anything? In keeping with that I think it makes sense to return `false` here too.

Thanks for adding those tests, looking good :+1:

jarednova:
Awesome, feels great to get this merged!

jarednova:
Awesome, feels great to get this merged!

PR #2162, opened 2020-01-09
Clarify PostCollection semantics (docs only)

acobster:
* clarify that you don't need to call `->to_array()` to loop
* clarify distinctions from core `get_posts` and `WP_Query`
* edit style for clarity, conciseness :)


acobster:
Thanks for adding that, even better now. :)

PR #2164, opened 2020-01-10
Fix/Ignore Level 1 problems

jarednova:
Picks up on #2149 to fix the first level of PHPStan issues


jarednova:
Closed in favor of #2165

PR #2165, opened 2020-01-10
Szepeviktor phpstan level 1

jarednova:
Resolves merge conflict from #2164

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ before_script:
 
 script:
     - if [ "$COVERAGE" == 1 ]; then mkdir -p build/logs; vendor/phpunit/phpunit/phpunit --coverage-clover build/logs/clover.xml -c phpunit.xml --verbose; else vendor/phpunit/phpunit/phpunit -c phpunit-nocover.xml --verbose; fi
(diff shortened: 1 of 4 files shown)


szepeviktor:
Please be aware that 2 of the 4 problems on Level 1 have been ignored.

jarednova:
@szepeviktor gotcha, I didn't realize the implications of ignoring those. In that case, I changed this to be just a level "0" evaluation. The thing we're ignoring there:

```
  Line   Term.php                                               
  99     Unsafe usage of new static().                          
  102    Unsafe usage of new static().                          
```

... I believe to be appropriate (and maybe negated by the work on the Factories branch). To get to level 1, I don't know if there's anything we can do about the `apply_filters` argument counts, but I'd like to see if we can resolve the undefined properties errors as we merge that in.

I updated this PR with those in mind. I'm good to pull this in then once tests pass and follow up on the level 1 issues and beyond in a fresh PR

szepeviktor:
All right. The ignoring "apply_filters" comes from WordPress core, there is no `...$arg` there _yet_. So this is a "stock" line in PHPStan config.

PR #2167, opened 2020-01-12
2.x Add a Users Guide

gchtr:
**Ticket**: https://github.com/timber/timber/pull/2160#issuecomment-572745383

## Issue

We already have new guides for Posts and Terms in #2073. We should have one for Users as well.

## Solution

Add Users Guide and clarify working with User Class Maps.

## Impact

None.

## Usage Changes

None.

## Considerations

None?

## Testing

None needed.

### DIFF
--- a/docs/guides/users.md
+++ b/docs/guides/users.md
@@ -1,0 +1,142 @@
+---
+title: "Users"
+---
+
+To get a user object in Timber, you use `Timber::get_user()` and pass the WordPress user ID as an argument.
+
+```php
+$post = Timber::get_user( $user_id );
+```
+
+This function is similar to [`get_userdata()`](https://developer.wordpress.org/reference/functions/get_userdata/) and accepts one argument: a user ID. If you don’t pass in any argument, Timber will use `get_current_user_id()` to work with the currently logged in user.
+
+```php
+$post = Timber::get_user();
+
+// Is the same as…
+
+$post = Timber::get_user( get_current_user_id() );
+```
+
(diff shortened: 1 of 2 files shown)


docs/guides/users.md
```diff
@@ -0,0 +1,148 @@
...
+
+```twig
+{% for user in get_users(user_ids) %}
+
+{% endfor %}
+```
+
+## Invalid user
+
+If no user can be found with the user ID you provided, the `Timber::get_user()` function will return `null`. With this, you can always check for valid users in a template a simple if statement.
```
jarednova: If no user can be found with the user ID you provided, the `Timber::get_user()` function will return `false`. With this, you can always check for valid users in a template a simple if statement.
```suggestion
If no user can be found with the user ID you provided, the `Timber::get_user()` function will return `false`. With this, you can always check for valid users in a template a simple if statement.
```

docs/guides/users.md
```diff
@@ -0,0 +1,148 @@
...
+
+```twig
+{% if user %}
+    {{ user.name }}
+{% endif %}
+```
+
+## Login state
+
+Similar to checking for valid users, you can also check whether a user is currently logged in to WordPress. When you don’t provide an ID for `Timber::get_user()`, it will return `null` if no user is currently logged in.
```
jarednova:
```suggestion
Similar to checking for valid users, you can also check whether a user is currently logged in to WordPress. When you don’t provide an ID for `Timber::get_user()`, it will return `false` if no user is currently logged in.
```

acobster:
@jarednova I committed your suggestions, merging this in now.

PR #2168, opened 2020-01-12
2.x User factories DocBlocks

gchtr:
**Ticket**: #2160

This pull request improves the DocBlocks for the new User API and the User Class Map, so that they will properly appear in the online documentation.

### DIFF
--- a/lib/Timber.php
+++ b/lib/Timber.php
@@ -229,6 +229,9 @@ class Timber {
 	 * @since 2.0.0
 	 * @example
 	 * ```php
+	 * // Get users with on an array of user IDs.
+	 * $users = Timber::get_users( [ 24, 81, 325 ] );
+	 *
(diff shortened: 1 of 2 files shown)


lib/Timber.php
```diff
@@ -220,11 +220,44 @@ public static function get_term( $term, $taxonomy = 'post_tag', $TermClass = 'Ti
 	================================ */
 
 	/**
-	 * Get one or more users as a collection.
+	 * Gets one or more users as an array.
+	 *
+	 * By default, Timber will use the `Timber\User` class to create a your post objects. To
+	 * control which class is used for your post objects, use [Class Maps]().
```
jarednova: This is the one thing I was like, "oh, we should remember to make a guide/link for that" and then I saw you were one step ahead of me with yr `@todo`

Issue #2152, opened 2020-01-07
Add Timber::get_post_by_slug() and Timber::get_post_by_title() functions


gchtr:
## Is your feature request related to a problem? Please describe.

Per the discussion in https://github.com/timber/timber/pull/2073#issuecomment-570954735, we should add an alternative to passing a post slug or title to `Timber::get_post()`, because that won’t be supported anymore in Timber 2.0.

## Describe the solution you’d like

- [x] Add `Timber::get_post_by_slug()`.
- [x] Add `Timber::get_post_by_title()`.
- [x] Update Upgrade Guide.

## Describe alternatives you’ve considered

See https://github.com/timber/timber/pull/2073#issuecomment-570954735.

gchtr:
Merged in #2169.

PR #2169, opened 2020-01-12
2.x Add Timber::get_post_by()

gchtr:
**Ticket**: #2152

## Issue

See #2152.

## Solution

~~Add `Timber::get_post_by_slug()` and `Timber::get_post_by_title()`.~~
Add `Timber::get_post_by()`.

I chose `post_exists()` when searching for any post type and `get_page_by_title()` when searching for specific post types as the functions to retrieve a post by post title, because they seem to be the most direct and performant ones.

## Impact

Closes a gap in what `Timber::get_post()` can’t (and shouldn’t be able) to do in Timber 2.0.

## Usage Changes

Instead of `Timber::get_post()` with a post title or post slug, developers should use one of the two new functions.

## Considerations

The two functions always run database queries to get post IDs, which got me thinking about performance. When we use `::build()` to create an object, could we eventually use `wp_cache_get()` and `wp_cache_add()` in the future to create an internal object cache for Timber objects?

## Testing

Yes.


jarednova:
@gchtr I know you're going to scream, but having just spent time with #2160 and #2167, I'm thinking of `Timber::get_user_by` — does it make sense for us to follow the same pattern? I realize that's piggy-backing on the corresponding [WordPress function](https://developer.wordpress.org/reference/functions/get_user_by/) that isn't matched by Posts. Had you given that any thought?

gchtr:
> I know you're going to scream

@jarednova Naaah, that’s fine 😄.

To be honest, I didn’t think of this at all yet. But I like it! Let’s get @acobster’s opinion, before I’m going to change this.

acobster:
@jarednova @gchtr I like it! It's more consistent and keeps the surface area smaller.

lib/Timber.php
```diff
@@ -157,6 +157,75 @@ public static function get_posts( $query = false, $PostClass = 'Timber\Post', $r
...
+	 * ```
+	 * $post = Timber::get_post_by_slug( 'about-us' );
+	 * ```
+	 *
+	 * @param string       $slug      The post slug to search for.
+	 * @param string|array $post_type Optional. The post type or an array of post types to look
+	 *                                for. Default `any`, which will look for posts in any post
+	 *                                type.
+	 *
+	 * @return \Timber\Post|null A Timber post or null if no post could be found.
```
acobster: `null` ➡️ `false`

gchtr: Resolved in https://github.com/timber/timber/pull/2169/commits/5ae15e53ce9d194199a63f2d328a41f37fd5b6a4.

tests/test-timber.php
```diff
@@ -28,12 +28,92 @@ function testGetPostString(){
...
+		] );
+
+		$post_movie = Timber\Timber::get_post_by( 'slug', 'kill-bill', 'movie' );
+		$post_page  = Timber\Timber::get_post_by( 'slug', 'kill-bill', 'page' );
+
+		$this->assertEquals( $post_id_movie, $post_movie->ID );
+		$this->assertEquals( $post_id_page, $post_page->ID );
+
+		$post_any = Timber\Timber::get_post_by( 'slug', 'kill-bill' );
+		$this->assertEquals( $post_id_movie, $post_any->ID );
```
jarednova: This actually returns the value of `$post_id_page` because it's the newest post (and WP's default sort is `post_date DESC`

tests/test-timber.php
```diff
@@ -28,12 +28,92 @@ function testGetPostString(){
...
+
+		$post_movie    = Timber\Timber::get_post_by( 'title', $post_title, 'movie' );
+		$post_page     = Timber\Timber::get_post_by( 'title', $post_title, 'page' );
+		$post_multiple = Timber\Timber::get_post_by( 'title', $post_title, [ 'page', 'book' ] );
+
+		$this->assertEquals( $post_id_movie, $post_movie->ID );
+		$this->assertEquals( $post_id_page, $post_page->ID );
+
+		// Multiple post types should return the post with the lower ID.
+		$this->assertEquals( $post_id_page, $post_multiple->ID );
```
jarednova: This returns the oldest by post date which SHOULD be the lower ID but isn't guaranteed

jarednova: see this for more info: https://developer.wordpress.org/reference/functions/get_page_by_title/ `Be careful: in case of more than one post having the same title, it will check the oldest publication date, not the smallest ID.`

gchtr:
@jarednova Yeah, I remember reading that in the function description, but then the tests ran successfully, so I thought it’s fine like it is. Apparently it isn’t and I must have missed something. Thank you for catching this! I’d be happy if you could take over to solve this 👍.

jarednova:
np! will take care of this in the next couple days

jarednova:
@gchtr this is now ready for your final review!

gchtr:
Thanks @jarednova for working on this! Your changes actually pointed me in the right direction and made me rethink how we should handle cases when multiple posts are found. I created another pull request in #2195 that we need to discuss first.

jarednova:
@gchtr just wanted to make sure this didn't get lost in the cracks. With #2195 merged, is this also ready?

gchtr:
> @gchtr just wanted to make sure this didn't get lost in the cracks. With #2195 merged, is this also ready?

Yes, this one’s ready!

jarednova:
We're good here! The test failures are related to static analysis and a failure in an unrelated issue in the 2.x branch

PR #2170, opened 2020-01-12
2.x Update Timber Object Getter function names in Twig

gchtr:
**Ticket**: #2144 

## Issue

See #2144.

## Solution

- Added new Twig functions for the relevant `Timber::get_*` functions.
- Updated the deprecation warnings. There’s now always two possibilities. Instead of `Post()`, it can either be `get_post()` or `get_posts()`, depending on what’s at hand. The same goes for `Term` and `User`.
- I deliberately removed the `{{ Attachment() }}` function as well as deprecated the `{{ Image() }}` function without replacement, because we can use `{{ get_post() }}` and `{{ get_posts() }}`. The Class Maps will handle returning either a `Timber\Image` or `Timber\Attachment`.

## Impact

A more unified way of interacting with Timber’s API throughout PHP and Twig.

## Usage Changes

New function names. Pretty much the same functionality, except for handling singular and multiple objects.

## Considerations

Should we add Twig functions for the special-case `Timber::get_*` functions, like `Timber::get_user_by()` or `Timber::get_post_by_slug()`. I think, currently we don’t have to do that.

## Testing

Tests are currently breaking. They will only run without errors as soon as we merge object factories into the `2.x` branch.

## Todo

- [ ] Add testing for comment functions in Twig.
- [ ] Resolve todos.


jarednova:
@gchtr should the base branch of this be set as `2.x-factories` so we can checkout which tests pass/fail? I want to hop in and help but want to make sure I actually help in the right places

gchtr:
Added as a new pull request in #2178.

PR #2172, opened 2020-01-12
phpstan: Level 1 compliance

jarednova:
This branch / PR is meant to handle compliance with PHPStan Level 1 issues. Per @szepeviktor's [comment](https://github.com/timber/timber/pull/2165#issuecomment-573457549) in #2165, we won't be able to do anything about the `apply_filters` argument count errors (due to how WP declares that function) but we can get our house in order with declaring properties on various Timber classes.

### DIFF
--- a/lib/Comment.php
+++ b/lib/Comment.php
@@ -47,6 +47,11 @@ class Comment extends Core implements CoreInterface, MetaInterface {
 	public $id;
 
 	/**
+	 * @var int
+	 */
+	public $comment_approved;
+
+	/**
 	 * @api
 	 * @var string
 	 */
@@ -69,6 +74,11 @@ class Comment extends Core implements CoreInterface, MetaInterface {
 	 * @var int
 	 */
 	public $comment_ID;
+
+	/**
+	 * @var int
+	 */
+	public $comment_parent;
 
 	/**
 	 * @api
--- a/lib/Image.php
+++ b/lib/Image.php
@@ -133,8 +133,7 @@ class Image extends Attachment {
 			'Image::get_dimension',
 			'2.0.0'
 		);
-		$ds = array($this->width(), $this->height());
-		return $ds;
+		return array($this->width(), $this->height());
 	}
 
 	/**
--- a/lib/Image.php
+++ b/lib/Image.php
@@ -133,16 +133,8 @@ class Image extends Attachment {
 			'Image::get_dimension',
 			'2.0.0'
 		);
-		if ( isset($this->_dimensions) ) {
-			return $this->get_dimensions_loaded($dim);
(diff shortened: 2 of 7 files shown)


jarednova:
This did what it was supposed to! This PR now applies PHPStan level 1 standards. In doing so, it also caught some deprecated code/properties that could be easily simplified. Turning to @gchtr, @acobster or @pascalknecht to review before merge

lib/Image.php
```diff
@@ -133,16 +133,8 @@ protected function get_dimensions( $dim ) {
...
-			return $this->get_dimensions_loaded($dim);
-		}
-		if ( file_exists($this->file_loc) && filesize($this->file_loc) ) {
-			list($width, $height) = getimagesize($this->file_loc);
-			$this->_dimensions = array();
-			$this->_dimensions[0] = $width;
-			$this->_dimensions[1] = $height;
-			return $this->get_dimensions_loaded($dim);
-		}
+		$ds = array($this->width(), $this->height());
```
gchtr: Call me picky, but can we cut down the use of abbreviations in variable names? I will only know that `$ds` is supposed to mean if look at what kind of array it is going to build. I’d rather have `$dimensions` as a variable name.

lib/Comment.php
```diff
@@ -46,6 +46,11 @@ class Comment extends Core implements CoreInterface, MetaInterface {
 	 */
 	public $id;
 
+	/**
+	 * @var int
```
gchtr: I think in the future we should try to get properties fully documented, including summary and descriptions, because these will also appear in the documentation. But it’s totally fine for the scope of this pull request.

lib/Image.php
szepeviktor: This should be
`return array($this->width(), $this->height());`

szepeviktor:
Releasable code should pass Level 4 or higher.

I could help in that.

lib/Image.php
gchtr: > This should be
> `return array($this->width(), $this->height());`

Even better, yes!

jarednova: done!

lib/Comment.php
jarednova: Agreed. There are LOTS of them so good for a future PR/task

jarednova:
Thanks for the suggestions all! Incorporated here.

@szepeviktor that would be FANTASTIC — let's go to level 2 next. Just be aware of the active [`2.x-factories`](https://github.com/timber/timber/tree/2.x-factories) branch. I'd hate to have your work or @acobster's get into conflict

szepeviktor:
Okay.

PR #2174, opened 2020-01-14
Removing one of the two before_script-s

szepeviktor:
as Ubuntu pkg https://packages.ubuntu.com/xenial/php-imagick


szepeviktor:
How could we had two `before_script`-s?!

szepeviktor:
Oh no!!! xenial contains the extension only for PHP 7.0
Reverting...

jarednova:
Oh man, great catch! I think this is a result of a bad git auto-merge between 1.x and 2.x.

PR #2181, opened 2020-01-22
[Doc] Use apply instead of filter

rubas:
> As of Twig 2.9, you should use the apply tag instead which does the same thing except that the wrapped template data is not scoped.

https://twig.symfony.com/doc/2.x/tags/filter.html


jarednova:
Thanks @rubas! I accidentally screwed something up when I was monkeying around with the base branch, so I pulled this in via this commit:

https://github.com/timber/timber/commit/901e8400802598dc0cb5a5fe732100d75294aec9

Issue #2182, opened 2020-01-22
Variables naming in docs.


bartnovak:
In https://timber.github.io/docs/guides/routing/ 
load method you are using as an example
Routes::load($php_file, $args, $query = null, $status_code = 200)
but in description there is $template_params instead of $args.

palmiak:
@nowakosom could you create a PR for this?

jarednova:
Closed by #2183

PR #2183, opened 2020-01-23
2182 DOCfix

bartnovak:
**Ticket**: #2182

## Issue

Changing name of variable in routing docs.

## Solution

Changed name of variable in routing docs.


jarednova:
Good call, thanks @bartnovak !

PR #2195, opened 2020-02-16
2.x Improve Timber::get_post_by()

gchtr:
**Ticket**: #2169

This pull request adds on top of #2169. But because I had to change more than I thought, this comes in a separate pull request so we can discuss it first.

In https://github.com/timber/timber/pull/2169/commits/302e08b78937fcc7ec0505d1e1592b78a38e4e99, you changed the order the posts were created. This go me thinking. If we want to check for the oldest post and make sure we ignore the post ID, which is an indicator of the order the posts were created in, then the post expect to be returned [shouldn’t be the first to be created](https://github.com/timber/timber/commit/8705f311e6fcdd854673d001bcec2d1a8f02a2b6).

In https://github.com/timber/timber/pull/2169/commits/58d24dd9657674f277fd6e7c6b155a580321f2c6, you changed the parameter in the test to `any`, which is not the same as an array of post types. Because of this, a different function to search for posts was used, making the tests pass. I think we [should check for both](https://github.com/timber/timber/commit/e599babc75636abc1e6e23fae302deb45dd278b7).

I figured out that `post_exists()` and `get_page_by_title()` won’t work if we go for the **oldest post** and don’t want to sort by ID. That’s why I replaced `post_exists()` and `get_page_by_title()` with our own query in https://github.com/timber/timber/commit/d6f95d303dada9126be2a775a5fbdbd3a4756e63.  I usually try to avoid creating direct queries, but I think there’s no easier way to search for a post title.

But by doing that, we open a new possibility to make the function more useful. We could add order arguments:

```php
public static function get_post_by( $type, $search_value, $args = [] ) {
    $args = wp_parse_args( $args, [
        'post_type' => 'any',
        'order_by'  => 'post_date',
        'order'     => 'ASC'
    ]

    // ...
}
```

This way, developers could decide on their own to change the logic a post is selected when there are multiple posts available.

### DIFF
--- a/lib/Timber.php
+++ b/lib/Timber.php
@@ -175,9 +175,16 @@ class Timber {
 	 *                                   for `title`, this parameter doesn’t need to be
 	 *                                   case-sensitive, because the `=` comparison is used in
 	 *                                   MySQL.
-	 * @param string|array $post_type    Optional. The post type or an array of post types to look
-	 *                                   for. Default `any`, which will look for posts in any post
-	 *                                   type.
+	 * @param array        $args {
+	 *     Optional. An array of arguments to configure what is returned.
+	 *
+	 * 	   @type string|array     $post_type   Optional. What WordPress post type to limit the 
+	 *                                         results to. Defaults to 'any'
+	 *     @type string           $order_by    Optional. The field to sort by. Defaults to 
+	 *                                         'post_date'
+	 *     @type string           $order       Optional. The sort to apply. Defaults to ASC
+	 *
+	 * }
 	 *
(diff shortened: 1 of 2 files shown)


jarednova:
Looks good! You sure caught my play-by-play as I was trying to get this to a good place last week. Seems like a crazy little bit of WordPress weirdness left in by the WordPress gods that d6f95d303dada9126be2a775a5fbdbd3a4756e63 solves. 

I totally agree with the last option (I was already feeling a bit unsure about "assuming" that a user would want the oldest) — so this makes it a pure default/option. I'm pulling down now to add that last piece ...

PR #2196, opened 2020-02-16
2.x Improve error handling

gchtr:
**Ticket**: https://github.com/timber/timber/pull/2073#issuecomment-586538864, #2145

## Issue

In https://github.com/timber/timber/pull/2073#issuecomment-580343556 and the following comments, we discussed whether we should use Exceptions or work with calls to `Helper::doing_it_wrong()`.

## Solution

This pull request pulls out the `Helper::doing_it_wrong()` function I added in #2145 and adds improvements.

I would like to go with the solution that @acobster suggested in https://github.com/timber/timber/pull/2073#issuecomment-586538864 and move the calls to `doing_it_wrong_run` and `deprecated_function_run` in front of the bailout check with `WP_DEBUG`.

## Impact

To say it in @acobster’s words:

> This gives folks an escape hatch for refactoring (which I consider the main use-case for `doing_it_wrong` to begin with) where they can throw exceptions, log to an external service, or whatever else to their hearts' content.

When we look at `_doing_it_wrong()` and `_deprecated_function()` in WordPress Core, the two action hooks `doing_it_wrong_run` and `deprecated_function_run` also run before a check to `WP_DEBUG`, so this change will make error handling work the same as in WordPress, which is a good thing.

## Usage Changes

~~This will make it possible for developers to hook into their favorite logging service, be it with PHP Exceptions itself:~~

```php
add_action( 'doing_it_wrong_run', function( $function, $message, $version ) {
    throw new \Exception( $message );
}, 10, 3 );
```

~~Or, for example, with a service like [Bugsnag](https://www.bugsnag.com/):~~

```php
add_action( 'doing_it_wrong_run', function( $function, $message, $version ) {
    if ( ! class_exists( 'Bugsnag_Wordpress' ) ) {
        return;
    }

    /**
     * Bugsnag Instance.
     *
     * @var \Bugsnag_WordPress|\Bugsnag_Client $bugsnagWordpress
     */
    global $bugsnagWordpress;

    $bugsnagWordpress->notifyError(
        'Doing it wrong!',
        $message,
        null,
        'warning'
    );
}, 10, 3 );
```

Ignore the code examples above. It’s not clear yet what the best way is for developers to hook into this.

## Considerations

We should probably add documentation about this. I wonder whether the [Debugging Guide](https://timber.github.io/docs/guides/debugging/) is the right place for this or whether

- We should add a separate Error Handling / Logging Guide?
- We should rename and extend the Debugging Guide with a section about Error Handling / Logging?

## Testing

Not directly. But if something won’t work, it will show up in testing, because a lot of tests that run functions including calls to `Helper::doing_it_wrong()` and `Helper::deprecated()` will require these two functions to work properly.

### DIFF
--- a/lib/Helper.php
+++ b/lib/Helper.php
@@ -337,7 +337,7 @@ class Helper {
 			}
 
 			$message .= sprintf(
-				' Please see <a href="%s">Debugging in WordPress</a> as well as <a href="%2$s">Debugging in Timber</a> for more information.',
+				' Please see Debugging in WordPress (%1$s) as well as Debugging in Timber (%2$s) for more information.',
 				'https://wordpress.org/support/article/debugging-in-wordpress/',
 				'https://timber.github.io/docs/guides/debugging/'
 			);
--- a/lib/Helper.php
+++ b/lib/Helper.php
@@ -351,7 +351,7 @@ class Helper {
 
 			// phpcs:disable WordPress.PHP.DevelopmentFunctions.error_log_trigger_error
 			// phpcs:disable WordPress.Security.EscapeOutput.OutputNotEscaped
-			trigger_error( $error_message );
+			trigger_error( '[ Timber ] ' . $error_message );
 		}
 	}
 
@@ -412,7 +412,7 @@ class Helper {
 
 		// phpcs:disable WordPress.PHP.DevelopmentFunctions.error_log_trigger_error
 		// phpcs:disable WordPress.Security.EscapeOutput.OutputNotEscaped
-		trigger_error( $error_message );
+		trigger_error( '[ Timber ] ' . $error_message );
 	}
 
 	/**
--- a/lib/Helper.php
+++ b/lib/Helper.php
@@ -285,17 +285,25 @@ class Helper {
 	 * @api
 	 * @since 2.0.0
 	 * @since WordPress 3.1.0
+	 * @see \_doing_it_wrong()
 	 *
 	 * @param string $function The function that was called.
 	 * @param string $message  A message explaining what has been done incorrectly.
 	 * @param string $version  The version of Timber where the message was added.
 	 */
 	public static function doing_it_wrong( $function, $message, $version ) {
+		/**
+		 * Fires when the given function is being used incorrectly.
+		 *
+		 * @param string $function The function that was called.
+		 * @param string $message  A message explaining what has been done incorrectly.
+		 * @param string $version  The version of WordPress where the message was added.
+		 */
+		do_action( 'doing_it_wrong_run', $function, $message, $version );
+
 		if ( ! WP_DEBUG ) {
 			return;
 		}
-
-		do_action( 'doing_it_wrong_run', $function, $message, $version );
 
 		/**
 		 * Filters whether to trigger an error for _doing_it_wrong() calls.
@@ -354,19 +362,27 @@ class Helper {
 	 * DocBlock. E.g.: "@expectedDeprecated {{ TimberImage() }}".
 	 *
 	 * @api
+	 * @see \_deprecated_function()
 	 *
 	 * @param string $function    The name of the deprecated function/method.
-	 * @param string $replacement Function to use instead.
-	 * @param string $version     When we deprecated this.
+	 * @param string $replacement The name of the function/method to use instead.
+	 * @param string $version     The version of Timber when the function was deprecated.
 	 *
 	 * @return void
 	 */
 	public static function deprecated( $function, $replacement, $version ) {
+		/**
+		 * Fires when a deprecated function is being used.
+		 *
+		 * @param string $function    The function that was called.
+		 * @param string $replacement The name of the function/method to use instead.
+		 * @param string $version     The version of Timber where the message was added.
+		 */
+		do_action( 'deprecated_function_run', $function, $replacement, $version );
+
 		if ( ! WP_DEBUG ) {
 			return;
 		}
-
-		do_action( 'deprecated_function_run', $function, $replacement, $version );
 
 		/**
 		 * Filters whether to trigger an error for deprecated functions.
--- a/lib/Helper.php
+++ b/lib/Helper.php
@@ -271,6 +271,83 @@ class Helper {
 	}
 
 	/**
+	 * Marks something as being incorrectly called.
+	 *
+	 * There is a hook {@see 'doing_it_wrong_run'} that will be called that can be used
+	 * to get the backtrace up to what file and function called the deprecated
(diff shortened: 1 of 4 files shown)


jarednova:
@gchtr Looks good! I pulled it down to add an application and test coverage. Let me know if I got all that right.

I think while this is open and fresh we should go ahead and add the info to the docs. Of these, I think this is the way to go:

> We should rename and extend the Debugging Guide with a section about Error Handling / Logging?

Perhaps "Debugging & Error Logging"? This is something I do with internal documentation a lot at Upstatement. Rather than a section/folder/document be one thing, making something "This & That" communicates that it's both "this", "that', and everything in between.

gchtr:
> Perhaps "Debugging & Error Logging"? This is something I do with internal documentation a lot at Upstatement. Rather than a section/folder/document be one thing, making something "This & That" communicates that it's both "this", "that', and everything in between.

@jarednova Sounds good! I’ll keep working on this.

lib/Helper.php
```diff
@@ -270,27 +270,120 @@ public static function warn( $message ) {
...
+			$error_message = sprintf(
+				'%1$s was called <strong>incorrectly</strong>. %2$s %3$s',
+				$function,
+				$message,
+				$version
+			);
+
+			// phpcs:disable WordPress.PHP.DevelopmentFunctions.error_log_trigger_error
+			// phpcs:disable WordPress.Security.EscapeOutput.OutputNotEscaped
+			trigger_error( $error_message );
```
gchtr: Should we use the `[ Timber ]` prefix here, too?

lib/Helper.php
```diff
@@ -270,27 +270,120 @@ public static function warn( $message ) {
...
+				$version = '';
+			} else {
+				$version = sprintf(
+					'(This message was added in Timber version %s.)',
+					$version
+				);
+			}
+
+			$message .= sprintf(
+				' Please see <a href="%s">Debugging in WordPress</a> as well as <a href="%2$s">Debugging in Timber</a> for more information.',
```
gchtr: Using HTML in an error message will only be useful if the error message is output directly into HTML. However, when we have xDebug enabled, we’ll see raw HTML in the HTML, which is harder to read. Also, when you look at the error message in the console, HTML won’t be parsed either.

That’s why I’d rather have something like this:

```suggestion
				' Please see Debugging in WordPress (%1$s) as well as Debugging in Timber (%2$s) for more information.',
```

acobster: Agreed.

acobster:
> `do_action( 'timber/trigger_error_run', $message, $type );`

This is a confusing hook name IMO. If we're going to introduce a new hook, why not simplify to, say, `timber/error`?

> Deprecate the `Helper::error_log()` function and introduce a `Helper::trigger_error()` function.

The difference being that the former is not affected by `set_error_handler`, while the latter is? Wouldn't that mean that if you have $LoggingService hooked into `timber/trigger_error_run` (or whatever) _and_ `set_error_handlers`, you'd get log $LoggingService entries for both? I don't think that's what I'd want. :)

Or is that the only way that `WP_DEBUG` can have the intended effect?? PHP is weird.

lib/Helper.php
jarednova: yes! Much nicer

lib/Helper.php
jarednova: Yeah, may as well apply that here too. I was always a bit blind to this need (I'm always working in Timber, therefore the errors are always coming from Timber!) — but can see how it's a big help in isolating where errors originate from

jarednova:
> Or is that the only way that WP_DEBUG can have the intended effect?? PHP is weird.

... and WordPress makes it weirder. My take is two-fold. First, I'm persuaded that `trigger_error` is the proper place to send errors as `error_log` is sending errors to a specific place (thus, perhaps bypassing potential error/QA/monitoring workflows). 

My only pushback is in the scope of this ticket and PR. Does the current code satisfy what was needed by https://github.com/timber/timber/pull/2073#issuecomment-586538864, #2145 with the `doing_it_wrong` stuff applied? My impulse is to merge the work and then isolate the `error_log` vs. `trigger_error` solve in its own issue/PR so we don't go too deep down a rabbit hole (esp. b/c the 2.x code isn't deployed yet, thus no backwards compatibility concerns). 

Of course, if you really see it as just as a few tweaks away from resolving, go for it! But if there's a risk that it opens up additional rounds of Qs/considerations, I think it better to isolate.

gchtr:
> The difference being that the former is not affected by `set_error_handler`, while the latter is? Wouldn't that mean that if you have $LoggingService hooked into `timber/trigger_error_run` (or whatever) _and_ `set_error_handlers`, you'd get log $LoggingService entries for both? I don't think that's what I'd want. :)

Hm... 🤔 yes. That’s indeed not good. Damn …

The question then is, do we advise developers to use the `doing_it_wrong_run` action to add their custom error handling, or do we advise to use `set_error_handler()` to add a custom error handler?

From what I saw `doing_it_wrong_run` is usually only used by testing suites to make the special tags like `@expectedIncorrectUsage` or `@exptectedDeprecated` work.

But then, if we advise to use `set_error_handler()` and also advise to not set `WP_DEBUG` to `true` in production, these errors would still never be logged.

> My only pushback is in the scope of this ticket and PR. Does the current code satisfy what was needed by [#2073 (comment)](https://github.com/timber/timber/pull/2073#issuecomment-586538864), #2145 with the `doing_it_wrong` stuff applied? My impulse is to merge the work and then isolate the `error_log` vs. `trigger_error` solve in its own issue/PR so we don't go too deep down a rabbit hole (esp. b/c the 2.x code isn't deployed yet, thus no backwards compatibility concerns).
> 
> Of course, if you really see it as just as a few tweaks away from resolving, go for it! But if there's a risk that it opens up additional rounds of Qs/considerations, I think it better to isolate.

Well, if we leave this pull request at the current state, we have consistency with how WordPress does it. But we **can’t add documentation** for *Advanced Error Logging* then, because we would first need to decide how to do it right.

I guess it is a rabbit hole. We should take the time to explore how it’s properly done in the WordPress context.

I’ll clean this up next week for a final review and open a separate issue to work on in the future.

lib/Helper.php
gchtr: Updated in https://github.com/timber/timber/pull/2196/commits/83a06163dd157bb0f7211fc49fdc18655ba72489.

gchtr:
@jarednova @acobster Back to you for review.

acobster:
If @jarednova is happy, I'm happy!

PR #2202, opened 2020-02-21
Remove useless error log for Class Maps

gchtr:
## Issue

When a Class Map filter is used and Timber doesn’t find a post type in the Class Map, it will output a notice:

```
[ Timber ]posttype not found in Array
```

I think that the notice is unnecessary, because it’s possible to only add certain post types to the Class Map and let Timber fall back to the default `Timber\Post` for the ones that are not defined. I think the notice might even be confusing.

## Solution

Remove the error log. And fix a small spelling error.

## Impact

None.

## Usage Changes

None.

## Considerations

I wonder if there’s some other thinking to this, that I didn’t consider yet?

## Testing

No.


jarednova:
Useless? Ouch! But I agree.

PR #2203, opened 2020-02-22
2.x term factories

acobster:
Implements the Terms API.

Per my recommendation on #2094, this drops support for automatically including `'hide_empty' => false` in Term query params.

I did manage to keep the "corrections" e.g. fixing `taxonomies => taxonomy` and `tags => post_tag`, etc. without too much fanfare. So we don't have to break code that relies on that.


jarednova:
Awesome, thanks @acobster! Things are looking good on the testing side. There are some failures (not in PHPUnit, but static analysis). Most of them are in code not connected to this PR — but there's one we should def. take a look at ...

```
 ------ -------------------------------------------------------------------- 
  Line   Term.php                                                            
 ------ -------------------------------------------------------------------- 
  229    Static method Timber\Timber::get_term() invoked with 2 parameters,  
         0-1 required.        ```

**Comment by acobster** [comment]
@jarednova thanks for that - fixed. I'll make a note to start running the analyzer before making PRs, too.

@szepeviktor not directly related to this PR, but some of my code is getting flagged:

```
  Line   Comment.php                                           
  123    Unsafe usage of new static().

  Line   User.php                                              
  115    Unsafe usage of new static().
```

what about `new static()` is unsafe? We need to use late static binding here, or else class maps simply won't work.

**Comment by szepeviktor** [comment]
new static() is unsafe when the class is not final.
@ondrejmirtes is very good at explaining this.

**Comment by ondrejmirtes** [comment]
Someone can extend from your class and override the constructor with different arguments, so that your `new static` call will crash. See: https://3v4l.org/7JlG4

You have these options:

1) Use `new self()` - you might realize you don't need `new static()`.
2) Make your constructor final.
3) Make your class final at which point you can also use `new self()` :)
4) Create an abstract method `createX()` that each child class needs to implement with their own `new self` implementation.

**Comment by ondrejmirtes** [comment]
5) Make your constructor part of an implemented interface at which point its signature cannot change.

**Comment by acobster** [comment]
Thanks @ondrejmirtes and @szepeviktor - making the constructors final seems heavy-handed to me for this case but sound like we have some other good options.

@jarednova this discussion made me realize that I never actually implemented `Term::build()`...d'oh! So standby for that.

**Comment by acobster** [comment]
OK @jarednova  I'll have an update soon but in the meantime I wanted to get back to this discussion about the `new static` warning. In short, I believe each of those options _in this specific case_ is untenable for our users and I'm still recommending that we annotate these instances with ignores.

@szepeviktor - please note that this is not about the cost/benefit of syntax checks, coding standards, or static analysis in general. The net benefit of all those things is clear and is taken as a given (tip of the hat to @ondrejmirtes - I'm already liking this tool a lot more that phpcs! :smiley:). This is about answering the question: what is the cost vs. the benefit _for our users_ of appeasing or ignoring this specific warning?

Here is why I think each option won't work in our particular case:

1. We need late static binding here or else [Class Maps](https://github.com/timber/timber/blob/2.x-docs-api/docs/guides/class-maps.md) simply will not work.
2. Making the constructor `final` precludes any user in the future doing anything involving a custom constructor. For example, I might want to do this:

```php
class MyPost extends Post {
  protected function __construct($stuff, $things) {
    // do something special with $stuff and $things
  }

  public static function build( WP_Post $wp_post ) {
    [ $stuff, $things ] = SomeApi::special_sauce();
    return new static($stuff, $things);
  }  
}
```

3. Making any of these classes final would essentially make most of Timber's API useless.
4. I'd be fine with each subclass implementing its own `build()` method to call `new self()` _if_ these were purely internal classes. Unfortunately these are classes that are designed to be extended, which means we'd be putting the burden on users to implement something that Factories are already capable of doing, and defeats the purpose of the Factory pattern to begin with.
5. Protected methods, including constructors, cannot be part of an interface by definition. I tried this on the abstract `Timber\Core` class but abstract classes do not make the same guarantees about the signature.

To sum up, following the letter of this specific error just swaps out a (already very rare) Fatal Error about required arguments with one about final constructors, and moreover limits our users' movement for very little benefit to them.

ondrejmirtes:
The most non-disrupting way would probably be to introduce a new interface implemented by the class where the `new static` code is. The interface would contain `public function __construct` with the arguments to enforce. This would only break code that already has the bad constructor and already represents broken code today.

szepeviktor:
> I believe each of those options _in this specific case_

I think a developer has to have some ground rules
which are not evaluated on per project basis.

Many businessmen do this.
I am looking at things from a 2 years perspective.

acobster:
> a developer has to have some ground rules

Agreed. It seems we only differ on what those ground rules are. Here are some of mine:

* Think from your users' perspective
* Use the right tool for the job
* Identify reasonable exceptions

szepeviktor:
> Think from your users' perspective

You're right.
I'm not able to manufacture a WordPress developer compatible software
as ~30-50% of the source code will be highly abnormal from modern PHP perspective.

acobster:
Hahaha. The WP core is a rat's nest, no two ways about it!

Sometimes WP is the right tool for the job, and sometimes it isn't.

jarednova:
Thanks everyone for the thoughts and expertise on this issue. I'm really glad we've been able to work out both the guidelines and the exceptions we need to make given the circumstances. @acobster do you have any more commits coming to this branch, or is ready for final review?

acobster:
One more commit to finalize and then I think we'll be good.

acobster:
@jarednova @gchtr turned out there was quite a bit more work that I'd missed once I made the constructor protected. Should be ready for review now.

lib/Factory/TermFactory.php
```diff
@@ -66,6 +76,14 @@ protected function from_term_obj(object $obj) : CoreInterface {
...
+		return $this->from_wp_term_query(new WP_Term_Query(
+			$this->filter_query_params([
+				'taxonomy' => $names,
+			])
+		));
+	}
+
 	protected function get_term_class(WP_Term $term) : string {
 		// Get the user-configured Class Map
 		$map = apply_filters( 'timber/term/classmap', [
```
gchtr: I see that in the documentation for [Term Class Maps](https://github.com/timber/timber/blob/2.x-docs-api/docs/guides/class-maps.md#the-term-class-map), we currently overwrite the passed Class Maps in the examples. I should probably update the documentation with `array_merge()`, like this:

```php
use Genre;

add_filter( 'timber/term/classmap', function( $classmap ) {
    return array_merge( $classmap, [
        'genre' => Genre::class,
    ] );
} );

```

acobster:
@gchtr I was definitely coming at it with PHPUnit-induced tunnel vision. :laughing: You're right on with all of this; I'll circle back this week. Thanks for the detailed review!

cc @jarednova

lib/Timber.php
```diff
@@ -196,24 +197,76 @@ public static function query_posts( $query = false, $PostClass = 'Timber\Post' )
...
+	 * ]);
+	 * ```
 	 */
-	public static function get_terms( $args = null, $maybe_args = array(), $TermClass = 'Timber\Term' ) {
-		return TermGetter::get_terms($args, $maybe_args, $TermClass);
+	public static function get_terms( $args = null, array $options = [] ) : Iterable {
+		// default to all queryable taxonomies
+		$args = $args ?? [
+			'taxonomy'   => get_taxonomies(),
+			'hide_empty' => false,
```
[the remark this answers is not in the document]
jarednova: Yup! this could just be eliminated since we're using WP's defaults

jarednova:
@acobster @gchtr looking good! I second the notes/comments made above. As for ...

```php
public function terms( $query_args = [], $options = [] ) {}  
```
> @jarednova Would this be something you could work on?

Yup! I'll get to work on a PR for that

lib/Timber.php
```diff
@@ -196,24 +197,76 @@ public static function query_posts( $query = false, $PostClass = 'Timber\Post' )
...
+	 * $cats = Timber::get_terms([
+	 *   'taxonomy' => 'my_taxonomy',
+	 *   'orderby'  => 'slug',
+	 *   'order'    => 'DESC',
+	 * ]);
+	 * ```
 	 */
-	public static function get_terms( $args = null, $maybe_args = array(), $TermClass = 'Timber\Term' ) {
-		return TermGetter::get_terms($args, $maybe_args, $TermClass);
+	public static function get_terms( $args = null, array $options = [] ) : Iterable {
```
gchtr: For `$options`, we get a hint in #2208 as to what should be supported in the future: We could make the `merge` option work here. And we could even move the logic for that option from `Post::terms()` to this function.

acobster: Great idea. #2211

lib/Post.php
```diff
@@ -686,7 +680,6 @@ public function comment_form( $args = array() ) {
 	 *     @type bool $merge        Whether the resulting array should be one big one (`true`) or
 	 *                              whether it should be an array of sub-arrays for each taxonomy
 	 *                              (`false`). Default `true`.
-	 *     @type string $term_class The Timber term class to use for the term objects.
 	 * }
 	 * @return array An array of taxonomies.
 	 */
```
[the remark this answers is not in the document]
acobster: Done in #2208

lib/Factory/TermFactory.php
acobster: Agreed, I'll let you do that.

lib/Timber.php
```diff
@@ -196,24 +197,76 @@ public static function query_posts( $query = false, $PostClass = 'Timber\Post' )
...
-	 * @return \Timber\Term|\WP_Error|null
+	 * @param int|\WP_Term $term a WP_Term or term_id
+	 * @return \Timber\Term|false
+	 * @example
+	 * ```php
+	 * // Get a Term.
+	 * $tag = Timber::get_term(123);
+	 *
+	 * // Get a tag specifically.
+	 * $tag = Timber::get_term(123, 'post_tag');
```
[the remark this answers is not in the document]
acobster: yep, fixed.

acobster:
@gchtr @jarednova I think I've addressed all the review feedback except the docs update. Take another look at let me know?

lib/Factory/TermFactory.php
gchtr: Updated in https://github.com/timber/timber/pull/2073/commits/6db9964fd10164d65f5f7eb136d647d7d85fdcdd

jarednova:
Copy that, added one last test — and presto!

PR #2205, opened 2020-02-25
Add Timber\Post class as a return type to query_post

mweimerskirch:
##  Issue
Autocompletion fails with phpStorm because it can't derive the correct class.

## Solution
This helps with autocompletion when working with phpStorm.


jarednova:
Thanks for the PR @mweimerskirch. I just merged it and will release later as a new version bump

Issue #1504, opened 2017-08-02
Document hooks used in Timber


gchtr:
As @dseidl mentioned in [#1481](https://github.com/timber/timber/issues/1481#issuecomment-319700603), it would be nice if filters used in Timber were documented. I agree with that. And I think that we should also document actions.

Here’s a crazy idea: What if we could document filters the same way as they are documented in WordPress and follow the [Inline Documentation Standards for Hooks](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#4-hooks-actions-and-filters) and then auto-generate a Filter Reference in the documentation?

It would mean that we could add DocBlocks right in front of where filters and actions are used in the code base. Here’s an example for a filter:

```php
/**
 * Filters the Twig environment.
 *
 * Makes it possible to add custom functionality to Twig. Examples are exposing functions
 * to Twig or adding custom filters.
 *
 * @since 1.x.x
 *
 * @api
 * @example
 * ```php
 * // Possibility to add an example
 * ```
 *
 * @param \Twig\Environment $twig The Twig Environment.
 */
$twig = apply_filters( 'timber/twig', $twig );
```

It doesn’t seem to be possible to read out DocBlocks *inside* functions with the [ReflectionClass](http://php.net/manual/en/class.reflectionclass.php) that is already used when generating the current documentation. However, it would be possible with [token_get_all](http://php.net/token_get_all). We could check if the DocBlock starts with `Filters …` for filters and `Fires …` to detect actions. These are the same keywords that you see for hooks in WordPress core.

In the same way as tags like **@api** (for public methods) or **@example** (to provide example code) are already used for class methods for the [Reference](https://timber.github.io/docs/reference/), we could also use these for hooks.

The benefit of having the documentation right in the code is that it’s way easier to keep track of changes. For example: when a filter is deleted, it will be removed from the filter reference automatically. However, I guess we can’t document every aspect of hooks inside DocBlocks. [Guides](https://timber.github.io/docs/guides/) are still a good place to show how filters can help *in a workflow* across different functionalities that Timber provides.

I can start working on this when I finished deprecating the wiki.

Opinions?  
Who wants to help?

dseidl:
@gchtr great idea! auto generated docs all the way.
if you start here let me know. im in. 👍

gchtr:
Hey @dseidl 

[...]

PR #2206, opened 2020-02-27
Updates to documentation for new documentation builder

gchtr:
I’ve been working on a new version for [Timber’s Documentation](https://timber.github.io/docs/). Here’s a sneak peek:

timber-docs-v2

In short we’ll have

- a version switcher
- a reworked navigation

The new documentation will be built with [Eleventy](https://www.11ty.dev/), which is a lot closer to what we’re used to when working with Timber. The documentation templates will be based on [Nunjucks](https://mozilla.github.io/nunjucks/), which has the same syntax as Twig. Work for this continues in https://github.com/timber/docs/pull/12.

### Different docs for v1 and v2

When we release version 2 of Timber, we need to have separate docs for 1 and 2. This pull request moves documentation for v1 into a separate `docs/v1` folder. For the documentation for version 2, we’ll use the `docs/v2` folder.

In the documentation, there will be a version switcher.

### A better navigation

The navigation will build itself from all the pages it can find in subfolders. This way, a page doesn’t need to contain data about where it should appear in the navigation.

For this, we’ll use .json files to describe sections. Here’s an example for the *Getting Started* section:

```
{
    "title": "Getting Started",
    "layout": "page"
}
```

### Different order for doc pages

I added an **order** argument to the FrontMatter block in guides to support a custom order in the documentation. The guides will not be ordered by name anymore, but by sort order. This will allow us to better guide users through the documentation.

The new order will be as follows:

- Context
- Template Locations
- WordPress Integration
- Posts
- Terms
- Users
- Comments
- Menus
- Twig Cookbook
- Text Cookbook
- Escaping
- Custom Page Templates
- Custom Fields
- Functions
- Filters
- Image Cookbook
- Pagination
- Sidebars
- Performance/Caching
- Date/Time
- Internationalization
- Cheatsheet
- Extending Timber
- Debugging
- Unit Testing
- Routing

## An Installation section 

There will be a new **Installation** section before the **Getting Started** section with some new Guides for installing, managing and updating Timber through Composer (added in #2240).

I’m thinking that the **Getting Started** section would mostly be used by developers who are starting out with Timber. And the **Installation** section as well as all the other sections will probably be visited frequently by everyone.

## Getting Started

I’m working on updating the **Getting Started** section with a slightly updated structure and code examples. I will do a separate pull request for that.

## A hooks section

I already started working on this in #1504. We will have section that is all about hooks. Using [teak](https://github.com/timber/teak) (yes, another side project 🙊), we’ll extract documentation for hooks starting with `timber/` right out of PHP files. This way, we’ll have a documentation page that lists all **action hooks** and another page that lists all **filter hooks**, including example code blocks. Here’s what this will look like:

timber-docs-hooks

That’s another reason why I’m always so pedantic about getting the all DocBlocks right 😜.

## An Integrations section

While Timber grew, support for several third party plugins was added. But not everyone might use all available integrations in all the projects. In the documentation, there will be a separate section for integrations. Though it probably won’t be active right after the 2.x release. We can still activate that later.

Some integrations might have to be loaded into Timber separately in the future, like a [WooCommerce integration](https://github.com/mindkomm/timber-integration-woocommerce). And the documentation for them can grow so big that one documentation page or each integration wouldn’t be enough. That’s why the integration section will support subpages.

## Redirects work differently

To make redirects work with Eleventy, there’s now a central [**redirects.njk**](https://github.com/timber/timber/blob/2.x-docs-11ty/docs/v2/redirects.njk) file in the docs folder. We won’t list aliases in the Front Matter of documentation files directly anymore. This is not necessarily a better pattern, but I didn’t find another way to do it.

## Other goodies

- Anchor links will be added to the headings themselves. This will make it easier to find a link to a section.
- The [Contributing](https://github.com/timber/timber/blob/master/CONTRIBUTING.md) file will be included as a separate page in the documentation as well.

## Testing

To test out this pull request, you will have to check out the `2.x-docs-11ty` branch and use the `2.x` branch in the [timber/docs](https://github.com/timber/docs) repository, (see https://github.com/timber/docs/pull/12).


gchtr:
@jarednova I think this is now ready for you to have a look at 🙂.

jarednova:
@gchtr wow, lots here! Overall this is beautiful and the info/navigation all works great. I obviously didn't read through everything, but did some spot checks and then spent some time in the front sections.

In b57b3e7 I removed some duplicate info (we were showing how to install w Composer three times) and a few old references to installing/managing as a plugin. I _think_ I caught them all, but there could be more things hiding in there.

I think the *content* here is all good and approved. I'll let you manage whether this can/should be merged in ahead of the [matching PR in docs](https://github.com/timber/docs/pull/12)

jarednova:
Excellent, looks good! Travis is failing, but only due to static analysis which we should address in its own issue/PR from the 2.x branch. I'll finish up my comments on https://github.com/timber/docs/pull/12

The merge order makes sense, but I'm gonna put that in your hands to execute since you know the details (and what to watch for in case there are issues)

jarednova:
🎩

PR #2213, opened 2020-03-08
Fix s3 url

hoangnd25:
## Issue
I'm using https://github.com/humanmade/S3-Uploads to handle file upload.  `Timber\URLHelper::remove_double_slashes` is breaking `resize` filter for any image with `s3://` prefix

## Solution
I'm just adding s3 to the schemes whitelist

## Impact
Shouldn't break anything

## Usage Changes
None

## Considerations
Maybe build url without scheme first, remove double slashes then prepend url scheme.

## Testing
Yes

### DIFF
--- a/lib/URLHelper.php
+++ b/lib/URLHelper.php
@@ -207,13 +207,14 @@ class URLHelper {
 	 */
 	public static function remove_double_slashes( $url ) {
 		$url = str_replace('//', '/', $url);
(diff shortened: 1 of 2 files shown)


jarednova:
- [x] Clear problem
- [x] Small fix
- [x] Cleaner code 
- [x] Tests included

You make my job easy @hoandang! Thanks so much for the contribution. I'll release it as the new version sometime next week

Issue #2214, opened 2020-03-09
[filter:time_ago] isn't translated correctly


muuvmuuv:
The twig filter `time_ago` isn't properly translated. Both past and future variables are hardcoded in english.

> /wp-content/plugins/timber-library/lib/Twig.php:367

Change this:

```php
		if ( $from < $to ) {
			return sprintf($format_past, human_time_diff($from, $to));
		} else {
			return sprintf($format_future, human_time_diff($to, $from));
		}
```

to this:

```php
		if ( $from < $to ) {
			return sprintf(__($format_past), human_time_diff($from, $to));
		} else {
			return sprintf(__($format_future), human_time_diff($to, $from));
		}
```

## Expected behavior

Should print:

English: 2 days ago
German: Vor 2 Tagen

## Actual behavior

English: 2 days ago
German: 2 Tage ago

## Steps to reproduce behavior

```twig
{{ <date> | time_ago }}
```

## What version of WordPress, PHP and Timber are you using?

Latest

## How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)

Wordpress plugin

gchtr:
Thanks @muuvmuuv to take the time to create an issue for this! While working on the documentation for dates and times in Timber, I stumbled over the same problem with the `time_ago()` function. And I think there’s actually a bigger problem.

But first, a quick solution for you. You can pass the translated string that should be used to `time_ago()`.

```twig
{{ date|time_ago(null, 'Vor %s') }}
```

## The problem

Here’s the function that is being called by the `|time_ago` filter:

https://github.com/timber/timber/blob/3eaddbc87a56aea44a37cb3ffda24e4b1bd0041b/lib/Twig.php#L353-L370

That function accepts two parameters with English strings: `%s ago` and `%s from now`.

If I want to translate these strings, I have to pass the translations in every instance where I want to use `time_ago()`. Also, if I only want to change the strings, I’d have to pass in the first two arguments `$from` and `$to` as well.

Timber uses very few instances of translatable strings. Mostly translatable strings are used in helper functions like `time_ago()` mentioned above or [`Timber\Twig::add_list_separators()`](https://github.com/timber/timber/blob/3eaddbc87a56aea44a37cb3ffda24e4b1bd0041b/lib/Twig.php#L378) as well as in debug output.

I wondered whether we could improve this. I see a couple of ways to approach this.

### Rely on Core Translations

As proposed by @muuvmuuv here, we could rely on Core Translations.

```php
[...]

PR #2215, opened 2020-03-09
Make strings in time_ago() translatable

gchtr:
**Ticket**: #2214

## Issue

The `time_ago()` method has two parameters that aren’t translated yet.

## Solution

Use WordPress Core translations when default parameters are used.

I’ve thought about whether relying on the Core Translations is a good idea, because it’s actually not advised by WordPress. WordPress suggests introducing your own text domain for your translations, because WordPress translations might change.

I then checked what translations Timber actually uses, and it’s really only a couple of Core Translations:

- `%1$s %2$d`
- `%1$s`
- `&hellip;`

And then there are translations that are never really used, because they are only used as defaults:

- `&laquo; Previous`
- `Next &raquo;`

https://github.com/timber/timber/blob/a882871a2174e4d17fbccdb7448a9df9366f72d8/lib/Pagination.php#L115-L116

I think these translations are very unlikely to change. But still, we should be cautious and still add tests to check whether the translations work correctly.

## Impact

The proper translation will be returned when parameters aren’t passed explicitly.

## Usage Changes

None.

## Considerations

Discussed under **Solution**.

## Testing

Yes! I thought this would be going to be very complicated, but there’s actually a `wp_download_language_pack()` function. Let’s see how it runs in the automated tests.

### DIFF
--- a/lib/Twig.php
+++ b/lib/Twig.php
@@ -208,7 +208,7 @@ class Twig {
 
 		$twig->addFilter(new Twig_Filter('pluck', array('Timber\Helper', 'pluck')));
 
-		/** 
+		/**
 		 * @deprecated since 1.13 (to be removed in 2.0). Use Twig's native filter filter instead
      *  @todo remove this in 2.x so that filter merely passes to Twig's filter without any modification
 		 * @ticket #1594 #2120
@@ -353,13 +353,33 @@ class Twig {
 	}
 
 	/**
-	 * @param int|string $from
-	 * @param int|string $to
-	 * @param string $format_past
-	 * @param string $format_future
+	 * Returns the difference between two times in a human readable format.
+	 *
+	 * Differentiates between past and future dates.
+	 *
+	 * @see \human_time_diff()
+	 *
+	 * @param int|string $from          Base date as a timestamp or a date string.
+	 * @param int|string $to            Optional. Date to calculate difference to as a timestamp or
+	 *                                  a date string. Default to current time.
+	 * @param string     $format_past   Optional. String to use for past dates. To be used with
+	 *                                  `sprintf()`. Default `%s ago`.
+	 * @param string     $format_future Optional. String to use for future dates. To be used with
+	 *                                  `sprintf()`. Default `%s from now`.
+	 *
 	 * @return string
 	 */
-	public static function time_ago( $from, $to = null, $format_past = '%s ago', $format_future = '%s from now' ) {
+	public static function time_ago( $from, $to = null, $format_past = null, $format_future = null ) {
+		if ( null === $format_past ) {
+			/* translators: %s: Human-readable time difference. */
+			$format_past = __( '%s ago' );
+		}
+
+		if ( null === $format_future ) {
+			/* translators: %s: Human-readable time difference. */
+			$format_future = __( '%s from now' );
+		}
+
 		$to = $to === null ? time() : $to;
 		$to = is_int($to) ? $to : strtotime($to);
 		$from = is_int($from) ? $from : strtotime($from);
--- a/readme.txt
+++ b/readme.txt
@@ -31,7 +31,7 @@ _Twig is the template language powering Timber; if you need a little background 
 = Develop (next release) =
 
 **Fixes and improvements**
-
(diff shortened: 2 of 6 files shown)


jarednova:
I have been banging my head against this for a week. Does anyone have experience integrating language switching into WordPress's PHPUnit tests? Here's where I'm stuck ...

```php
function testTimeAgoFutureTranslated() {
	wp_download_language_pack( 'de_DE' ); // also doing this in bootstrap.php, adding here for demonstration
	$locale = $this->change_locale('de_DE'); // returns "1" for success
	$locale = get_locale(); // returns de_DE
	$test = __( '%s from now'); // doesn't translate, outputs "%s from now", should output "%s Tag ab jetzt"
	$str = Timber\Twig::time_ago( '2016-12-01 20:00:00', '2016-11-30, 20:00:00' );
	$this->assertEquals( '1 Tag ab jetzt', $str );

	$this->restore_locale();
}
```

Right now this test (and a similar one for time in the past) fails:

```php
1) TestTimberDates::testTimeAgoFutureTranslated
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'1 Tag ab jetzt'
+'1 day from now'
```

Would love any help/suggestions!

gchtr:
Language switching really during runtime really is not easy in WordPress. I can take a lot at this again this week.

jarednova:
Awesome, that worked!!!! There are still a few failing tests, but that's due to WP version issues. I should be able to solve those from my end so you can move on to more exciting stuff

gchtr:
I love it when these tiny tiny issues have consequences. I thought I found the issue. When translation files already exist, WordPress will not re-download them. I had some German translation files there, that didn’t make much sense and didn’t contain translations we needed – so I thought that’s why the strings weren’t translated.

So then apparently the solution is to delete the files before downloading a language pack. I didn’t find a function in WordPress that does this. But [WP CLI does](https://github.com/wp-cli/language-command/blob/24858c4c87e350f5587944da2e62a814c0cd6be1/src/Core_Language_Command.php#L246-L294), so I borrowed it from there.

I didn’t know where to put that function, though, so I added it as a static function to `Timber_UnitTestCase`. Does this make sense to be there?

But now, the tests still fail. So, why do they fail in WordPress 4.9.8 and not in 5.x? Apparently, the translation for `%s ago` [exists in 5.x](https://translate.wordpress.org/projects/wp/5.3.x/de/default/?filters%5Bterm%5D=%25s+ago&filters%5Buser_login%5D=&filters%5Bstatus%5D=current_or_waiting_or_fuzzy_or_untranslated&filter=Filter&sort%5Bby%5D=priority&sort%5Bhow%5D=desc), but [not in 4.9.8](https://translate.wordpress.org/projects/wp/4.9.x/de/default/?filters%5Bterm%5D=%25s+ago&filters%5Buser_login%5D=&filters%5Bstatus%5D=current_or_waiting_or_fuzzy_or_untranslated&filter=Filter&sort%5Bby%5D=priority&sort%5Bhow%5D=desc) 🤷‍♂️.

I warned about this https://github.com/timber/timber/issues/2214#issuecomment-596415737, but then also thought that these are such basic translations, they surely won’t change and always be available. Well, guess what! 😂

Back to the drawing table. What’s the best approach here?

Maybe I wouldn’t have had to redownload translations, but on the other hand, for tests, this probably is quite a good approach.

jarednova:
@gchtr looks like I responded before your comment. The 4.9.8 to 5.x fix can be as simple as skipping those tests in WP < 5.x (I think those were introduced with the new DateTime stuff in 5.2). The PHPUnit for that is `$this->markTestSkipped('%s ago not available in WP 4.9.8');` with a conditional to check the WP version.

Again, happy to do that part but if you're on a roll and want to bring it home — all yours!

gchtr:
Okay, let’s try to make it work like that. Maybe in the future we will run into an error for a missing translation again.

Thanks for the hint to `markTestSkipped()`. I just added that. Let’s see if that works now.

gchtr:
Cool, it worked 🙌! Back to you for merging this in.

jarednova:
Nice, thanks @gchtr !!!!

PR #2220, opened 2020-03-28
Extending for Google Storage

hatto:
## Issue
Handling image resizing for images with gs:// url

## Solution
Extending url whitelist with the 'gs'.

## Impact
none

## Usage Changes
Changes are same as it was in the case of s3 urls. Basically just copy-paste.

## Testing
The very same test as in the case of testDoubleSlashesWithS3.

### DIFF
--- a/lib/URLHelper.php
+++ b/lib/URLHelper.php
@@ -207,7 +207,7 @@ class URLHelper {
 	 */
 	public static function remove_double_slashes( $url ) {
 		$url = str_replace('//', '/', $url);
-		$schemes_whitelist = array( 'http', 'https', 's3' );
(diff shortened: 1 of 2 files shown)


palmiak:
@hatto @jarednova it looks ok for me but I was wondering maybe it would be a good idea to add:
```php
$schemes_whitelist = apply_filters( 'timber/url/schemes-whitelist', array( 'http', 'https', 's3', 'gs' )  );
```

palmiak:
I added the filter.

jarednova:
Perfect. Thanks for adding the filter @palmiak — should make future tweaks that people need to do for specific protocols much easier.

Issue #2185, opened 2020-01-26
PostType instantiation may produce warning


FlyingDR:
Creating instance of `\Timber\PostType` object with unavailable post type currently produces warning in a case of passing unavailable post type in constructor.  

Warning is produced by [call](https://github.com/timber/timber/blob/828bdcc8bfaeb4b76b49d4dba8baec3ea6950a8e/lib/PostType.php#L24-L25) of `get_object_vars(get_post_type_object($post_type))` which implies that `object` is always passed to `get_object_vars()`. But return type of `get_post_type_object()` is [actually](https://developer.wordpress.org/reference/functions/get_post_type_object/#return) `object|false` and this case does not handled by the code anyhow.

2.x branch is also currently [affected](https://github.com/timber/timber/blob/03bf4bf20dc763895b776b7f51ef856050821da8/lib/PostType.php#L29-L30).

FlyingDR:
One of practical scenarios is creation of `\Timber\Post` with no explicit `$pid` in a case if there is no post at all, e.g. on `404.php`. Hence this issue is one of possible consequences of #2184

acobster:
> return type of `get_post_type_object()` is actually `object|false`

Looks like it's `object|null`?

In any case, it seems like what we need here is just a simple guard in `init()` against `get_post_type_object()` returning `null`/empty. Can you make a PR?

Thanks!

jarednova:
Closed by #2222

FlyingDR:
@jarednova thank you! Sorry for not preparing patch by myself, had a lot to do over last months, hope to contribute more into project in the future.

jarednova:
all good @FlyingDR! You're doing your part by raising this stuff in a clear and helpful fashion. From there, PRs are a lot easier — stay safe out there!

PR #2221, opened 2020-03-30
#2185 post type extra check

palmiak:
**Ticket**: #2185 

## Issue
The possible bug was described in the issue. All in all there was one check missing for Timber/PostType.

## Solution
Just added the check is `! empty()`.

## Impact
It shouldn't have any impact.

## Usage Changes
No changes.


jarednova:
Thanks @palmiak ! Just added a quick test in #2222 to confirm everything

Issue #2224, opened 2020-04-09
no error on invalid template


flip111:
When i supply a non-existing template i don't get an error. Instead it fails silently with no html output

Recreate:
```php
$context          = Timber::context();
$bla = Timber::render(['non-existing.twig'], $context);
```

tested with `define( 'WP_DEBUG', true );`

palmiak:
@jarednova what do you think? I got a bit used to this, but it might be a good idea to add some warning if WP_DEBUG is true.

In compile method we could add this extra check
```php
if ( $file === false ){
Helper::error_log('error loading your templates, check to make sure the file exists');
}
```

jarednova:
Yeah, I agree. The reason to fail silently is so that more specific template preferences don't trigger errors (e.g.: 'single-my-custom-post-type.twig' > 'single.twig'), but I guess this blows away the expected error when _all_ the templates are missing. A debug statement like you have there would be grand. Wanna PR?

palmiak:
Yup 👍 I'll do it.

pt., 10 kwi 2020, 14:43 użytkownik Jared Novack <notifications@github.com>
napisał:

jarednova:
Closed by #2230

PR #2230, opened 2020-04-13
Throws warning on invalid template

palmiak:
**Ticket**: #2224 

## Issue
If the twig file doesn't exist than it fails silently.

## Solution
Throwing a warning if a template doesn't exist.

## Impact
Shouldn't have any. `compile()` still return false


jarednova:
Thanks again @palmiak — glad to get this in there

PR #2232, opened 2020-04-18
Update User.php

flip111:
prevent template from crashing
```
Recoverable fatal error: Method Timber\User::__toString() must return a string value in /var/www/html/wp-content/plugins/timber-library/vendor/twig/twig/src/Environment.php(497) : eval()'d code on line 56 
```

This happened because i programmatically inserted some posts which had no user attached. Other parts of wordpress have no problem with this scenario.

### DIFF
--- a/lib/User.php
+++ b/lib/User.php
@@ -148,6 +148,8 @@ class User extends Core implements CoreInterface {
 			if ( isset($data->roles) ) {
 				$this->roles = $this->get_roles($data->roles);
 			}
+		} else {
+			$this->display_user = '';
 		}
 		unset($this->user_pass);
 		$this->id = $this->ID = (int) $this->ID;


jarednova:
Thanks @flip111 — could you post a snippet of code showing how you're programmatically creating those posts? I want to make sure I write a test that properly matches your situation (I wasn't able replicate the error thus far in my tests)

flip111:
https://developer.wordpress.org/reference/functions/wp_insert_user/

`$posts` variable below is from some data i was loading earlier.

```php
foreach ($posts as $post) {
  $post_id = wp_insert_post(
    [ 'post_date' => date('Y-m-d H:i:s', $post['date'])
    , 'post_title' => $post['topic']
    , 'post_content' => $post['text']
    , 'post_status' => 'publish'
    ]);

  wp_set_post_categories($post_id, [$some_category_id]);
}
```

PR #2234, opened 2020-04-22
Flip111 patch 1

jarednova:
Follow-up on #2232

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ env:
 
 matrix:
     include:
-    - php: 5.6.30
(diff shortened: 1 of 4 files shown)


jarednova:
@flip111 thanks so much for your first PR!!! I added some tests on #2234 (and dealt with some random Travis/Composer bugs). I was able to simplify things a bit by moving the empty string to the top in the property declarations.

Thanks as well for the code snippet. Using that I was able to replicate the issue. The way that I was creating fake posts/users before (using WordPress's testing framework) somehow suppressed that error.

PR #2235, opened 2020-04-22
2x/no factories

jarednova:
Working to resolve errors in #1617 merge ...

PR #2236, opened 2020-04-23
One static analysis run is enough

szepeviktor:
and protect an asterisk!

**Ticket**: #2235 

## Issue

@jarednova summoned me


szepeviktor:
I wonder whether `$PREFER_LOWEST` is a stray variable...

PR #2237, opened 2020-04-23
Fix autoloading causing unit test errors

szepeviktor:
**Ticket**: #2235

## Issue

@jarednova summoned me


jarednova:
Thank you @szepeviktor !!!! that is some crazy summoning service

Issue #1754, opened 2018-07-20
Implement gettext extraction within Timber


drzraf:
### Expected behavior
`wp-cli timber twig-pot`
Generate a Gettext pot file for the Timber-back theme.

### Actual behavior
POT for Twig is restricted to Poedit2 without any practical alternative.
That means it's unsuitable for collaborative i18n platforms which usually rely on continuous/automated POT upload/updates.

### But Extracting translatable strings needs the combined knowledge of several things:
1. How to parse a Twig template to extract specific node.
1. Know all the custom function/filters and register them
1. Where are the templates files located
1. How are called gettext function wrappers (`__;_e;_n;_x;_ex;...`)

### So which software can do that:
* Point 1 Can be done by any codebase relying on Twig parser + a couple of calls to `loadTemplate()`
* Point 2: A part can only be done by Timber (which contains several additional extensions). But a theme may declare other ones. So `functions.php` needs to be loaded too.
* Point 3 could be specified by hand/parameters to external tool, but is already known by Timber
* Point 4 could be specified by hand/parameters to external tool, but are already known by Timber

### Narrowing:
Point 2 exclude defacto external tools like `Twig-Gettext-Extractor` and imply WP bootstrap (`wp-cli`).
`Twig-Gettext-Extractor` compiles one user-provided Twig templates and run `xgettext()` on resulting PHP code. This is suboptimal and trying to give it the knowledge of WordPress Twig customizations not doable.
Moreover, passing through the `xgettext` `system()` call is probably not needed.
Twig parser is (_probably_) able to provide us every `callable` node whose `name` attribute is `_x` or `__`, ...
Provided such a list it should be fairly easy to generate to corresponding PO file without the need for `xgettext()`. See for example [how wp-cli i18n](https://github.com/wp-cli/i18n-command/blob/master/src/PhpCodeExtractor.php#L80) is creating [new Translation](https://github.com/oscarotero/Gettext/blob/master/src/Translation.php) from code.

### Conclusion
It's either the responsibility of `Timber` or `wp-cli i18n` to provide the feature (or a combination of both: `wp-cli` for bootstrapping, `wp-cli i18n` for Po-file format and `Timber` for Twig-related)

### Blocked by
- [x] [Avoid overly nested function call](https://github.com/timber/timber/pull/1753)
- [x] [AST-based php-gettext Twig parser (supporting the `Timber` loader)](https://github.com/php-gettext/Twig-Scanner)
[...]

PR #2238, opened 2020-04-27
Gettext extraction

gchtr:
**Ticket**: #1754

## Issue

To make gettext extraction work, the `Timber\Twig` class needs to be instantiated without a WordPress environment. For this, we need to remove / work around WordPress functions in the `Timber\Twig::add_timber_functions()` method.

## Solution

This pull request

- Moves the hooks out of the constructor.
- Moves the `timber/twig` filter to the `Timber\Loader` class.

## Impact

There should be no impact.

By moving the `timber/twig` filter into the `Timber\Loader` class, we make sure that custom functionality is added to the Twig Environment as a last step, which is probably what most would expect.

If a duplicate filter or function is added to Twig, the order doesn’t matter, there will always be a `LogicException`. However, if your custom function or filter causes the error, the error log would probably make more sense, since it’s your code that will be flagged, and not Timber’s.

## Usage Changes

None.

## Considerations

This pull request is not about proper handling of the Twig Environment. This could probably be improved by the ideas proposed in #1493.

Using an `init()` function for adding hooks is a pattern that we often see in WordPress development. Also see [this blog post](https://carlalexander.ca/designing-class-wordpress-hooks/). I don’t know however whether it’s better to use a static pattern or a non-static pattern.

```php
// Static

Twig::init();

class Twig {
    public static function init() {
        $self = new self();

        add_action( 'example', [ $self, 'example' ] );        
    }
}

// Static with __CLASS__

Twig::init();

class Twig {
    public static function init() {
        add_action( 'example', [ __CLASS__, 'example' ] );        
    }
}

// Non-static

$twig = new Twig();
$twig->init();

class Twig {
    public function init() {
        add_action( 'example', [ $this, 'example' ] );
	}
}
```

Of course these are not all popular patterns that exist out there. The pattern using `__CLASS__` is common in order to make it easier to unhook any hooks. But then, we would have to change the functions that are being called by the hooks to be static. Maybe @pascalknecht has some insight and experience with this? I’m not sure whether it’s better to let people unhook some hooks (and possibly run into side effects) or whether there should be a different way to control functionality in case something needs to be disabled.

## Testing

No additional tests added.

@drzraf Can you check if the pull requests adds the changes we need with the5 [`gettext-extraction`](https://github.com/timber/timber/tree/gettext-extraction) branch.

### DIFF
--- a/lib/Loader.php
+++ b/lib/Loader.php
@@ -178,6 +178,14 @@ class Loader {
 		$twig = apply_filters('timber/twig/functions', $twig);
 		$twig = apply_filters('timber/twig/escapers', $twig);
 		$twig = apply_filters('timber/loader/twig', $twig);
+
+		$twig = apply_filters('timber/twig', $twig);
+
+		/**
+		 * get_twig is deprecated, use timber/twig
+		 */
+		$twig = apply_filters('get_twig', $twig);
+
 		return $twig;
 	}
 
--- a/lib/Timber.php
+++ b/lib/Timber.php
@@ -108,7 +108,9 @@ class Timber {
 	 */
 	protected static function init() {
 		if ( class_exists('\WP') && !defined('TIMBER_LOADED') ) {
-			Twig::init();
+			$twig = new Twig();
+			$twig->init();
+
 			ImageHelper::init();
 			Admin::init();
 			new Integrations();
--- a/lib/Timber.php
+++ b/lib/Timber.php
@@ -108,9 +108,7 @@ class Timber {
 	 */
 	protected static function init() {
 		if ( class_exists('\WP') && !defined('TIMBER_LOADED') ) {
-			$twig = new Twig();
-			$twig->init();
-
+			Twig::init();
 			ImageHelper::init();
 			Admin::init();
 			new Integrations();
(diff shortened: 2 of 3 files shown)


jarednova:
Ooof, this sounds quite a rabbit trail — thanks for the code and the notes. Good news is that all tests pass! 

I wish I had stronger feelings on the static vs. non-static Q. My thought is that we should default to having the lightest touch possible at this point in the 1.x lifecycle. Thus, while changing it from static to non-static might be the right move in the future, I think we'd want to know with certainty (and perhaps, hold that change until a later 2.x release).

Other than that, I just want to be super-careful once we merge this to immediately catch it up with the 2.x branch. Handling the merge conflicts there has started to become a real minefield (esp. as we're doing more code rearrangement vs. new files).

gchtr:
> Thus, while changing it from static to non-static might be the right move in the future, I think we'd want to know with certainty (and perhaps, hold that change until a later 2.x release).

Yes! And oh damn, I realized I actually got it mixed up in the code! Let’s go with the slightest touch for now.

If can also take care of merging this into 2.x. Especially with the filter hook refactors, there will probably be merge conflicts.

jarednova:
Looks good! I'll see how the 2.x merge goes ...

PR #2239, opened 2020-04-28
2.x Fix gettext-extraction merge bugs

gchtr:
**Ticket**: #2238

## Issue

Just as I expected, the filters weren’t merged properly, probably because of earlier filter refactors involving big DocBlocks.

## Solution

Move the filters to their proper place.

## Impact

None.

## Usage Changes

None.

## Considerations

None.

## Testing

None.

### DIFF
--- a/lib/Loader.php
+++ b/lib/Loader.php
@@ -154,7 +154,7 @@ class Loader {
 				'2.0.0',
 				'timber/loader/render_data'
 			);
-			
+
 			$template = $twig->load($file);
 			$output = $template->render($data);
 		}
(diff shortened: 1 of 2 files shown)


jarednova:
@gchtr I had noticed that as well, but then saw they were located in `Twig.php`:

https://github.com/timber/timber/blob/2.x/lib/Twig.php#L257-L297

— should they be moved to `Loader` as in this PR?

gchtr:
@jarednova In #2238, I moved these filters to the `Loader` class. In `2.x`, we have refactored versions of these filters, including documentation and the deprecation for `get_twig`. According to the changes in #2238, we also need to move the filters to `Loader` in `2.x`.

So

> — should they be moved to `Loader` as in this PR?

yes.

I see that in #2238, I talked about `Twig::add_timber_functions()`, but it’s actually `Twig::add_timber_filters()` that’s affected.

jarednova:
Cool, just wanted to make sure!

Issue #2095, opened 2019-10-06
Document migration from plugin version to Composer version


gchtr:
As mentioned in https://github.com/timber/timber/pull/2005#issuecomment-518356963, we should a better migration docs for migrating from the plugin version to the Composer version of Timber.

> ... and of course as @gsalkin notes, we'll be sure to document a migration path and on-boarding materials.

@jarednova What exactly do you mean with "on-boarding materials"?

jarednova:
"on-boarding materials" = the things that intro people to Timber: the readme, getting started guide, docs main page (and even articles / tutorials we might promote or create)

gchtr:
Solved by #2240.

PR #2240, opened 2020-04-30
2.x Add guides for working with Composer

gchtr:
**Ticket**: #2095

## Issue

In #2005, we dropped plugin support for Timber. When developers want to use Timber 2.x, it might be the first time that they will use Composer. This can be a hard entry.

## Solution

This pull request adds three new Guides:

- [Installation](https://github.com/timber/timber/blob/2.x-docs-composer/docs/v2/getting-started/installation.md) – All about how and where to install Timber with Composer.
- [A short intro to Composer](https://github.com/timber/timber/blob/2.x-docs-composer/docs/v2/getting-started/intro-to-composer.md) – A complementary guide that can be read before reading the *Installation* guid.
- [Managing Timber with Composer](https://github.com/timber/timber/blob/2.x-docs-composer/docs/v2/getting-started/manage-timber-with-composer.md) – All about updating Timber and installing specific versions.

I’m also working on some new Getting Started Guides and will change the structure of the onboarding documentation a bit. That’s why in this pull request, I didn’t delete any notes about installation yet.

## Impact

Hopefully helps with the transition from the plugin version to Composer.

## Usage Changes

Obviously, use Composer instead of plugin installation.

## Considerations

- Is there more we should document?
- Where should we add notes about system requirements? In the installation guide? Or in the Readme?
- Should we explicitly mention that the plugin version would have to be uninstalled when upgrading from v1 to v2?

## Testing

No.


jarednova:
Looking good @gchtr. I wasn't able to fully test how these come out when compiled with Eleventy (looks like #2206 needs to be taken care of first with https://github.com/timber/docs/pull/12). I adjusted just a few wording things and such.

Overall, I think this reads great. Includes a pretty solid explanation of why Composer is the option to go with for devs and some of the common Qs people are likely to have!

gchtr:
@jarednova Thanks man for looking over this!

> I wasn't able to fully test how these come out when compiled with Eleventy (looks like #2206 needs to be taken care of first with [timber/docs#12](https://github.com/timber/docs/pull/12)).

Yep, I will have to push some more changes to the docs repository so that you can check it out. I made good progress this weekend and I’ll ping you if more reviews are required.

Issue #1834, opened 2018-10-24
Timber sends an excessive amound of DELETES to the database


dennisjac:
### Expected behavior
Timber should not send an excessive number of requests to the database.

### Actual behavior
Timber sends an excessive amount of DELETE queries to the database server in an attempt to purge transients.

### Steps to reproduce behavior
Call the render() function often with a non-false value for the $expires parameter.

### What version of WordPress, PHP and Timber are you using?
Timber 1.7.1, Wordpress 4.9.8, PHP 7.0

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Packagist

### Notes
In `lib/Loader.php::render()` the `delete_cache()` function is called every time the function is called with a non-false $expires parameter which in our setup causes an excessive number of DELETE queries that make up a non-trivial amount of total queries sent to the database (this DELETE query is sent several times each second).

The core Wordpress code already does an expiration and it uses a scheduled event on order to prevent this kind of flood. See `wp-admin/admin.php`:
```
if ( ! wp_next_scheduled( 'delete_expired_transients' ) && ! wp_installing() ) {
        wp_schedule_event( time(), 'daily', 'delete_expired_transients' );
}
```
Timber should adopt a similar approach to cap the amount of cache purges to a more reasonable value.

chads2000:
Just hit this issue, it basically took down website down for a few hours while it was clearing.

I have also noticed that this system does a hard delete of the cache, essentially removing and inserting a new.  For a site that is ~5 Years old, my wp_options primary index is getting over 160mil.  Wondering if would be better to instead of removing the record, but to null the value until the next write.

chads2000:
Can we get an update or movement on this. Here is my DB connection stack right now. Over 126 connections!!!!!

<img width="605" alt="Screen Shot 2020-04-17 at 11 17 43 AM" src="https://user-images.githubusercontent.com/1414604/79596196-5efbc300-809d-11ea-92bc-bfa901c7a0eb.png">

jarednova:
hi @chads2000 — I'm turning my attention to this now. Thanks for your follow-ups; this is helpful to see if I can suss out what's going on here

chads2000:
This may help

calling `Timber::render('index.twig', $data, 600);` with an expiration time.

[...]

PR #2243, opened 2020-05-04
Add hack to set expires to false vs. 0

jarednova:
**Ticket**: #1834 

## Issue
A bunch of `DELETE` commands are sent to the database in certain situations where Timber is trying to clear its own cache.

## Solution
@chads2000 noticed a situation where an `$expires` value of `0` is being strictly compared to `false` which might be throwing off the logic of the cache inside of `Loader::render`

## Impact
Need to watch out to see how this works inside other caching and compile/render tests

## Usage Changes
None

## Considerations
The WP VIP team has asked me to removing caching entirely in order to be truly compatible with their platform. Ideally this can patch the problem behavior. But I think the future for Timber caching is relying on external caching tools rather than some secondary caching internal to Timber.

## Testing
No new tests, let's see what happens to the old ones!

### DIFF
--- a/lib/Loader.php
+++ b/lib/Loader.php
@@ -54,6 +54,10 @@ class Loader {
 			}
 		}
 
+		if ( $expires === 0 ) {
+			$expires = false;
+		}
+
 		$key = null;
 		$output = false;
 		if ( false !== $expires ) {


gchtr:
> But I think the future for Timber caching is relying on external caching tools rather than some secondary caching internal to Timber.

I second that. Caching is complicated. But to be clear, would you rely only on tools like the [Twig cache extension](https://github.com/asm89/twig-cache-extension) or Twig’s template file caching and leave out transient caching?

Also, would we need documentation specific to platforms like WordPress VIP?

jarednova:
> But to be clear, would you rely only on tools like the Twig cache extension or Twig’s template file caching and leave out transient caching?

I think in the future we should assume "no caching" as a default and ...
- (At least) provide easy guidance for enabling Twig's cache extension, file caching and/or more
- (At most) provide easy hooks to activate the above

jarednova:
@gchtr ... but I think this should be a future thing we plan out and execute following 2.x's release and this ticket

PR #2245, opened 2020-05-04
WordPress VIP Requests

jarednova:
In talking with the WordPress VIP Team, they asked for some top-level controls to disable file writing and caching. This adds/confirms those abilities as related to ...

1. Image file writing
2. Cache writing to the filesystem
3. Cache writing to the DB

... with these off, WP VIP said that Timber will be able to play with their infrastructure much easier.

 

**Ticket**: #


jarednova:
Replaced by #2250

PR #2194, opened 2020-02-09
2.x factories

szepeviktor:
Let's have a PR for Factories.

PR #2246, opened 2020-05-06
Fix problems reported by PHPStan

szepeviktor:
Please see separate commits


lib/Comment.php
```diff
@@ -29,7 +29,7 @@
  * <p class="comment-attribution">- Sullivan Ballou</p>
  * ```
  */
-class Comment extends Core implements CoreInterface, MetaInterface {
+final class Comment extends Core implements CoreInterface, MetaInterface {
```
gchtr, line 32: We won’t make classes `final`. Extending Timber’s classes for WordPress objects is one of the [basic concepts of working with Timber](https://github.com/timber/timber/blob/2.x/docs/guides/class-maps.md).

szepeviktor: Oh! I see!

szepeviktor: Has been force-removed.

szepeviktor:
This PR has been reduces to two basic things.

jarednova:
Agree with @gchtr. This PR is now totally fine, but more from a time perspective I want to wait on review and merge so we can remain focused on #2194 (Factories) which is our big blocker to getting 2.x complete (and then being able to really devote focus here and other clean-up issues). 

Note to future self: the correct `CoAuthorsPlus` we should use once Factories is merged is this one:

https://github.com/timber/timber/blob/2.x-factories/lib/Integrations/CoAuthorsPlusUser.php
https://github.com/timber/timber/blob/507db4c3f993468a6391deed98d1852791948d3d/lib/Integrations/CoAuthorsPlusUser.php

szepeviktor:
> Note to future self:

:-D

szepeviktor:
Resolved in https://github.com/timber/timber/commit/ff7a2171e83f79ddb9670f5c0cf26dfd73a95579

Issue #2089, opened 2019-10-01
Implement Timber::get_menu() and Timber::get_user() methods.


acobster:
## Is your feature request related to a problem? Please describe.

Per #2073 and #2088, we're making `Menu`, `User`, and `Comment` constructors protected.

For more context, see the linked issues.

## Acceptance Criteria

- [x] Implement static `Timber::get_menu()` method which in turn call the appropriate factory methods (factory implementations themselves are out of scope here)
- [x] Ditto `Timber::get_user()`
- [x] Ditto `Timber::get_comment()`
- [x] Implement `Timber::get_users()` methods for getting collections. 
- [x] Ditto `Timber::get_comments()`
- [x] Determine whether we need a `::get_menus()` method? (No - see discussion.)
- [x] Determine which hooks, if any, we need to apply before instantiating `Menu` or `User` objects (should they get their own class maps?)
- [x] Implement Class Maps for `MenuItem`
- [x] Implement tests for these methods.

## Describe alternatives you’ve considered

The alternative is having an inconsistent API, which I personally don't find acceptable.

## Additional context

See #1793 for how this all got started.

gchtr:
> * Do we need a `::get_menus()` method?

If we want to strive for completeness and consistency, we should add it. Currently, I can’t think of a use case, through. It’s going to be hard to come up with meaningful examples. Any idea?

> * Determine which hooks, if any, we need to apply before instantiating `Menu` or `User` objects (should they get their own class maps?)

In #2073, I’ve already considered a Class Maps for `User` objects, but not yet for `Menu` objects. In the same way as user objects are not associated with anything in WordPress default, menu objects are agnostic as well. The Class Map could then look like this:

```php
use MenuExtended;

add_filter( 'timber/menu/classmap', function( $class, \WP_Menu $menu ) {
    return MenuExtended::class;
}, 10, 2 );
```

More logic could also be included through Conditional Tags.

What do you think, @acobster, should I add this to #2073?

gchtr:
I added the relevant documentation for menus in #2197. But let’s keep the main discussion and tracking of todos in here.

## Function signature for `Timber::get_menu()`

[...]

PR #2247, opened 2020-05-10
2.x Menu docs optimizations

gchtr:
**Ticket**: #2089

## Issue

There are some improvements that we can make to the Menu Class Maps.

## Solution

According to what was proposed in #2089 

- The menu class map should receive a menu object (a `WP_Term`) instead of a `$location` string.
- Removes the `$location` parameter from the Menu Item Class Map, because that is already present through `$menu->theme_location`.
- Adds more examples for how to use the Menu Class Maps.

## Impact

None.

## Usage Changes

None.

## Considerations

\-

## Testing

No.

### DIFF
--- a/docs/guides/class-maps.md
+++ b/docs/guides/class-maps.md
@@ -2,7 +2,7 @@
 title: "Class Maps"
 ---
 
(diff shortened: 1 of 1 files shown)


jarednova:
Made a few small changes — back to you for merging!

gchtr:
@jarednova Cool, thanks for your improvements! I’ll merge this in then and then work on the documentation about the pages menu.

PR #2248, opened 2020-05-10
Add example to apply_filters in doc

19h47:
## Issue

Missing examples to how to use **Twig** `filter` tag in template file. 

## Solution

Examples have been added to doc.

## Considerations

This examples are useful with **WooCommerce** integration when you need to pass a large amount of code inside a filter. Better for readability.

### DIFF
--- a/docs/guides/wp-integration.md
+++ b/docs/guides/wp-integration.md
@@ -80,16 +80,16 @@ Timber already comes with a [set of useful filters](https://timber.github.io/doc
 {{ "my custom string"|apply_filters('my_filter', param1, param2, ...) }}
 ```
 
-Since __Twig 3__ you can use your filter with a [apply](https://twig.symfony.com/doc/3.x/tags/apply.html) tag.
+You can use your filter with a [filter](https://twig.symfony.com/doc/2.x/tags/filter.html) tag.
 
 ```twig
-{% apply apply_filters( 'my_filter') %}
+{% filter apply_filters( 'my_filter') %}
         {{ post.content }}
-{% endapply %}
+{% endfilter %}
 
-{% apply apply_filters('my_filter', param1, param2, ...) %}
+{% filter apply_filters('my_filter', param1, param2, ...) %}
         my custom string
-{% endapply %}
+{% endfilter %}
 ```
 
 ## Widgets
--- a/docs/guides/wp-integration.md
+++ b/docs/guides/wp-integration.md
@@ -106,6 +106,36 @@ function filter_my_filter( $tag, $param1, $param2, $param3 ) {
 }
 ```
 
+### Real world example with WooCommerce
+
+Sometimes in __WooCommerce__ we found very long line of code:
+
+```php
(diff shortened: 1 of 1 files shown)


19h47:
Soory, didn't see Timber use Twig 2.10

jarednova:
This is very cool @19h47 — but I think it might need a matching example to show what would match an example of `add_filter` in PHP?

19h47:
@jarednova You totally right, it might be confusing since a [first argument](https://github.com/timber/timber/blob/master/lib/Twig.php#L223) is added to get the content of the block.

I update my example, what do you think about i?

jarednova:
@19h47 this definitely helps me to better understand. Is there a slightly more real-world example you could provide — perhaps something from your WooCommerce work (maybe abbreviated if it's lenghty)?

19h47:
Sometimes in __WooCommerce__ templates we found very long line of [code](https://github.com/woocommerce/woocommerce/blob/master/templates/checkout/payment.php#L33):

```php
echo '<li class="woocommerce-notice woocommerce-notice--info woocommerce-info">' . apply_filters( 'woocommerce_no_available_payment_methods_message', WC()->customer->get_billing_country() ? esc_html__( 'Sorry, it seems that there are no available payment methods for your state. Please contact us if you require assistance or wish to make alternate arrangements.', 'woocommerce' ) : esc_html__( 'Please fill in your details above to see available payment methods.', 'woocommerce' ) ) . '</li>';
```

And I want to preserve **WooCommerce** filters to not losing compatibility.

In __Twig__ we can write this piece of code like this (I store `WC()->customer` in context):

```twig
<li class="woocommerce-notice woocommerce-notice--info woocommerce-info">
    {{ customer.get_billing_country() ? __( 'Sorry, it seems that there are no available payment methods for your state. Please contact us if you require assistance or wish to make alternate arrangements.', 'woocommerce' ) : __( 'Please fill in your details above to see available payment methods.', 'woocommerce' ) | apply_filters( 'woocommerce_no_available_payment_methods_message' ) }}
</li>
```

Better, but it's not very readable and scalable IMHO.

So my idea is:

```twig
<li class="woocommerce-notice woocommerce-notice--info woocommerce-info">
    {% filter apply_filters( 'woocommerce_no_available_payment_methods_message' ) %}
        {% if customer.get_billing_country() %}
            {{ __( 'Sorry, it seems that there are no available payment methods for your state. Please contact us if you require assistance or wish to make alternate arrangements.', 'woocommerce' ) }}
        {% else %}
            {{ __( 'Please fill in your details above to see available payment methods.', 'woocommerce' )  }}
        {% endif %}
    {% endfilter %}
</li>
```

I think it's more clear.

We can go further and pass `customer.get_billing_country()`, a boolean, to the filter, make the logic inside the controller, maybe change some texts, add HTML, and return everything.

```twig
<li class="woocommerce-notice woocommerce-notice--info woocommerce-info">
    {% filter apply_filters( 'woocommerce_no_available_payment_methods_message', customer.get_billing_country() ) %}
            {{ __( 'Please fill in your details above to see available payment methods.', 'woocommerce' )  }}
    {% endfilter %}
</li>
```

```php
add_filter( 'woocommerce_no_available_payment_methods_message', 'no_available_payment_methods_message', 10, 2  );

function no_available_payment_methods_message( $tag, $billing_country ) {
    if ( $billing_country ) {
        $tag =  __( 'Sorry, it seems that there are no available payment methods for your state. Please contact us if you require assistance or wish to make alternate arrangements.', 'woocommerce' );
    }

    echo $tag;
}
``` 

(A lot for a simple message, I know, but it's for demonstrate how we could use the filter.)

My idea is really to be clearer and to be able to reuse the templates with the same logic as **WooCommerce**.

What do you think?

jarednova:
Thanks again @19h47 — I made one last pass to clarify as much as I could. Will be in the next deployment!

PR #2250, opened 2020-05-12
WordPress VIP Compatibility Requests

jarednova:
## Issue
The WordPress VIP team requested the ability to disable file writes in order to make Timber play nice with their tech stack.

## Solution
Add a `timber/allow_fs_write` filter that any dev can control. Defaults to true.

## Impact
Should have no backwards compatibility effect. This defaults to true, so it will have the same behavior in all existing installs.

## Usage Changes
Rather than create a guide _just_ for WP VIP, I added a new guide for "Hosts & Servers" that describes the steps WP VIP users will need to consider. This can be expanded as people contribute any platform-specific notes (both hosts like Pantheon, but also services like AWS/Azure).

## Considerations

1. For WP VIP: we should investigate how `resize` can play with the [Photon API](https://developer.wordpress.com/docs/photon/api/). Shouldn't be too hard, but I think it's for a future PR.
2. I already have experience with a few hosts that I can add to the Hosts & Servers guides, but holding back for now (all are "works great w/o modification") — Kinsta, Pantheon and WP Engine

## Testing
Yes, new test class for WP VIP

### DIFF
--- a/docs/guides/hosts-servers.md
+++ b/docs/guides/hosts-servers.md
@@ -1,0 +1,41 @@
+---
+title: "Hosts & Servers"
+menu:
+  main:
+    parent: "guides"
+---
+
+This guide serves as reference for any host or server-specific information we gather. If you have experience hosting WordPress with a particular host, stack, or service (AWS, Azure, etc.) please add that information here so it can be shared.
+
+
+## WordPress VIP
+
+Automattic offers a paid service called [WordPress VIP](https://wpvip.com/) for enterprise customers. To get Timber to play nice with their stack, we need to disable functionality related to caching and writes to the filesystem:
+
+**functions.php**
+
+```php
+add_filter('timber/cache/mode', function() {
+	return 'none';
+});
+```
+
+```php
+add_filter('timber/allow_fs_write', function() {
+	return false;
+});
+```
+
+This means you will not be able to use on-the-fly image resizing through Timber. Don't despair! You can set custom image sizes for WordPress to use:
+
+**functions.php**
+```php
+add_image_size( 'my_custom_size', 220, 220, array( 'left', 'top' ) );
+```
+
+**single.twig**
+```twig
+<img src="{{ post.thumbnail.src('my_custom_size') }}" alt="{{ post.thumbnail.alt() }}">
+```
+
+WordPress VIP has its own caching mechanisms. So when we disable caching, we're only disabling Timber and Twig's caching — not other layers that WP VIP applies.
--- a/docs/guides/hosts-servers.md
+++ b/docs/guides/hosts-servers.md
@@ -21,9 +21,7 @@ add_filter('timber/cache/mode', function() {
 ```
 
 ```php
-add_filter('timber/allow_fs_write', function() {
-	return false;
-});
+add_filter( 'timber/allow_fs_write', '__return_false' );
 ```
 
 This means you will not be able to use on-the-fly image resizing through Timber. Don't despair! You can set custom image sizes for WordPress to use:
--- a/tests/test-timber-wp-vip.php
+++ b/tests/test-timber-wp-vip.php
@@ -3,20 +3,19 @@
 	class TestTimberWPVIP extends TimberImage_UnitTestCase {
 
 		function testDisableCache() {
-			add_filter('timber/cache/mode', function() {
-				return 'none';
-			});
-
+			$filter = function() {
+			    return 'none';
+			};
+			add_filter( 'timber/cache/mode', $filter );
 			$loader = new Timber\Loader();
 			$cache = $loader->set_cache('test', 'foobar');
 			$cache = $loader->get_cache('test');
 			$this->assertFalse($cache);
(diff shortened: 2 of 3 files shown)


jarednova:
I changed this to target the 1.x branch so that VIP can start using this stuff ASAP (it's also a very small change, so the port should be easy — even if it means using #2245 instead). 

@gchtr I think you had mentioned the lack of docs somewhere in a comment. I added those amongst the start of a larger "Hosts & Servers" guide here.

docs/guides/hosts-servers.md
```diff
@@ -0,0 +1,41 @@
...
+
+```php
+add_filter('timber/cache/mode', function() {
+	return 'none';
+});
+```
+
+```php
+add_filter('timber/allow_fs_write', function() {
+	return false;
```
gchtr, line 25: We could also use 

```php
add_filter( 'timber/allow_fs_write', '__return_false' );
```

tests/test-timber-wp-vip.php
```diff
@@ -0,0 +1,69 @@
+<?php
+
+	class TestTimberWPVIP extends TimberImage_UnitTestCase {
+
+		function testDisableCache() {
+			add_filter('timber/cache/mode', function() {
```
gchtr, line 6: This will set the filter for all subsequent tests. You should also remove the filter again.

```php
$filter = function() {
    return 'none';
};

add_filter( 'timber/cache/mode', $filter );

// Test aaaaall the things.

remove_filter( 'timber/cache/mode', $filter );

**Comment by gchtr** [comment]
Good idea to create a general guide for hosts and servers 👍. We could still split it up in the future if needed.

> I changed this to target the 1.x branch so that VIP can start using this stuff ASAP (it's also a very small change, so the port should be easy — even if it means using #2245 instead).

Yeah, let’s add this to 1.x!

> @gchtr I think you had mentioned the lack of docs somewhere in a comment. I added those amongst the start of a larger "Hosts & Servers" guide here.

We talked about this when we realized there’s differences in how CDN work in https://github.com/timber/timber/pull/1944. That was specific to CDN networks regarding images. It was you who mentioned:

> A list of which WP hosts do / don't have a push vs. pull CDNs would be a valuable add to this.

That’s still a future todo. But I think this new guide would be a good place for it. We could link to the Hosts & Servers guide from the Image Cookbook.

**Comment by jarednova** [comment]
Thanks @gchtr! Integrated those suggestions in the latest commit

**Comment by jarednova** [comment]
awesome, just waiting on word from the VIP team — want to make sure they have a look at what they're getting

**Comment by szepeviktor** [comment]
Doesn't disabling Twig template caching have detrimental effects?

**Comment by jarednova** [comment]
@szepeviktor you're correct, but given VIP's architecture they (and perhaps other platforms) would prefer to handle all performance using their existing layers. The file write specifically causes problems with their infrastructure

**Comment by szepeviktor** [comment]
> you can use the system temp directory, which is the only local, writeable path on web servers on the VIP Platform

https://wpvip.com/documentation/vip-go/writing-files-on-vip-go/#local-file-operations

@jarednova We may use the temp directory.

**Comment by jarednova** [comment]
@szepeviktor that is a good point. @mjangda how does this square with what we talked about a few weeks back? When it comes to using this to store the transpiled twig => PHP I'm wondering if it's rendered moot by: `Files and directories can only be relied on for the duration of the current request.`

**Comment by szepeviktor** [comment]
> for the duration of the current request.

I think it is **much** better to cache compiled twigs on the current instance than recompile them on each request.

**Comment by mjangda** [comment]
> We may use the temp directory.

You technically could but it's not really recommended.

Our general suggestion for compile operations like this is to do them in a build process prior to deploy (e.g. compile and cache the Twig templates in a CI process). This way, you get the benefits of Twig caching and avoid the potential negative impact of runtime-compilation.

**Comment by mjangda** [comment]
@jarednova the docs look good and line up with what we talked about. I think there's definitely opportunity to optimize how well Timber works on VIP (and other similar platforms) like support for pre-deploy compilation and integrating with the Photon API (which you noted; this would benefit VIP sites and anyone using Jetpack as well).

An additional option to consider: auto-load these tweaks by detecting the environment.

For VIP, you could do something like this:

```
if ( defined( 'VIP_GO_ENV' ) && false !== VIP_GO_ENV ) {
    // enable tweaks
}
```

**Comment by szepeviktor** [comment]
Be expressive!

```php
if ( defined( 'VIP_GO_ENV' ) && true === VIP_GO_ENV ) {
```

mjangda:
The value of that constant is normally a string so the `true` check wouldn't work.

jarednova:
Thanks for the feedback @mjangda. I'm going to go ahead and merge this in with a future todos...

- [ ] Develop a build process to precompile Twig templates
- [ ] Integrate with Photon API
- [ ] Consider detection of the VIP/VIPGO environments to activate variables

... let's wait to make these optimizations until we actually get Timber/VIP going on the next project. An actual implementation will hopefully guide us on the specifics and whether some of the convenience features are helpful/necessary

PR #2254, opened 2020-05-19
2.x Add documentation for new Pages Menu API

gchtr:
**Ticket**: #2089

## Issue

As discussed in https://github.com/timber/timber/issues/2089#issuecomment-629852632, we need documentation for the new Pages Menu.

## Solution

Add documentation in the Menus Guide, the Class Maps Guide and the Upgrade Guide.

## Usage Changes

A new API to create a menu from the existing pages dynamically:

* A `Timber\PagesMenu` class. 
* A `timber/pages_menu/classmap` filter.
* A `Timber::get_pages_menu()` function.

## Considerations

I didn’t come up with big examples, because I think this will feature will be mainly for simpler sites, where custom classes might often not be needed anyways. Or may I just felt like a lazy dog.

## Testing

No.

### DIFF
--- a/docs/guides/class-maps.md
+++ b/docs/guides/class-maps.md
@@ -285,7 +285,7 @@ add_filter( 'timber/pages_menu/classmap', function( $class ) {
 } );
 ```
 
-The Menu Class Map receives the default `Timber\PagesMenu` class name as an argument.
+The Pages Menu Class Map receives the default `Timber\PagesMenu` class name as an argument.
 
 ## The User Class Map
 
--- a/docs/guides/class-maps.md
+++ b/docs/guides/class-maps.md
@@ -265,6 +265,28 @@ add_filter( 'timber/menuitem/classmap', function( $class, $menu ) {
 }, 10, 2 );
 ```
 
+## The Pages Menu Class Map
+
+With the `timber/pages_menu/classmap` filter, you can tell Timber which class it should use for pages menu object.
+
+The Pages Menu Class Map is used:
+
+- When you get a menu through `Timber::get_pages_menu()`.
+
+Here’s an a example for a basic filter where you would always return your custom `ExtendedMenu` class.
+
+**functions.php**
+
+```php
+use ExtendedPagesMenu;
+
+add_filter( 'timber/pages_menu/classmap', function( $class ) {
+    return ExtendedPagesMenu::class;
+} );
+```
+
+The Menu Class Map receives the default `Timber\PagesMenu` class name as an argument.
+
 ## The User Class Map
 
 With the `timber/user/classmap` filter, you can tell Timber which class it should use for user objects.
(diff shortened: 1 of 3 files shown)


docs/guides/class-maps.md
```diff
@@ -265,6 +265,28 @@ add_filter( 'timber/menuitem/classmap', function( $class, $menu ) {
...
+
+```php
+use ExtendedPagesMenu;
+
+add_filter( 'timber/pages_menu/classmap', function( $class ) {
+    return ExtendedPagesMenu::class;
+} );
+```
+
+The Menu Class Map receives the default `Timber\PagesMenu` class name as an argument.
```
acobster, line 288: should we be consistent and say "The Pages Menu Class Map..."?

docs/guides/menus.md
```diff
@@ -95,6 +95,27 @@ $menu = Timber::get_menu( 'primary' );
...
+$menu = Timber::get_pages_menu();
+```
+
+This function will return an instance of `Timber\PagesMenu`, which is not quite the same as `Timber\Menu`, but it contains the same `Timber\MenuItem` objects as you know them.
+
+If you want to extend a pages menu, you would do it like this:
+
+
+```php
+class CustomPagesMenu extends Timber\PagesMenu {
```
acobster, line 112: Using the fully qualified `\Timber\PagesMenu` is safer - it'll still work inside a custom namespace. Maybe we should also name this custom class `ExtendedPagesMenu` to reinforce the hook example?

docs/upgrade-guides/2.0.md
```diff
@@ -331,7 +331,21 @@ Always pass a parameter.
...
+
+```php
+$menu = new Timber\Menu();
+```
+
+**New**
+
+```php
+$menu = Timber::get_pages_menu();
+```
```
acobster, line 348: It's probably worth noting specifically that calling `Timber::get_menu()` without arguments will return `false` if there are no menus set up.

docs/upgrade-guides/2.0.md
```diff
@@ -331,7 +331,21 @@ Always pass a parameter.
 $menu = Timber::get_menu( 'primary' );
 ```
 
-@TODO
+#### The Pages Menu
+
+If you didn’t provide a parameter and didn’t have any menus registered, Timber would build **a menu from your existing pages**. This still works, but you have to use the new `Timber::get_pages_menu()` function.
```
jarednova, line 336:
```suggestion
Previously, if you didn’t provide a parameter to `Timber\Menu()` and didn’t have any menus registered, Timber would build **a menu from your existing pages**. To achieve this same functionality, you must now use the new `Timber::get_pages_menu()` function.
```

docs/guides/class-maps.md
gchtr: Added in bb2fafe795b81826ff87d2b4bdc1512e567aeda3 👍

docs/guides/menus.md
gchtr: Added in bb2fafe795b81826ff87d2b4bdc1512e567aeda3 👍 I should probably update to fully qualified namespaces in other places, too.

docs/upgrade-guides/2.0.md
gchtr: Added in bb2fafe795b81826ff87d2b4bdc1512e567aeda3 👍

PR #2260, opened 2020-06-07
2.x Update Extending Timber Guide

gchtr:
## Issue

With 2.x, the concept of extending Timber with your own classes is becoming more prominent and relevant, especially because everything will evolve around Class Maps. While the new [Class Maps Guide](https://github.com/timber/timber/blob/2.x-docs-api/docs/v2/guides/class-maps.md) is good for showing the different Class Maps and how to work with them, we could still show some more real-world examples for how to use your custom methods.

## Solution

You can look at the updated [Extending Timber Guide](https://github.com/timber/timber/blob/2.x-docs-extending-timber/docs/v2/guides/extending-timber.md).

What’s new:

- I started out with an example for a more basic custom method in a custom class and how to add it to the Class Maps.
- Added an example for how to move from adding to `$context` to using a custom method. In this case, I used an example for fetching related posts, because that’s one of the more typical use cases I have seen in the wild.
- Refined the Magazine example and removed the special mention of the mini cache with a property, because I moved that explanation to the Related Posts example
- Added two examples for working with meta fields. The first is exploring how to working with date fields and the second one is exploring how to extend an array of post IDs to Timber objects.

## Impact

More developers using their custom object classes.

## Usage Changes

None.

## Considerations

For the future: I figured that when adding custom methods that fetch posts, what we recommend now is to add a small caching mechanism with properties. We could think about how we could automate this. Caching shouldn’t happen automatically, though. For example:

```php
Timber::get_posts( $this->meta( 'sponsors' ), [
  'cache' => true,
] );
```

## Testing

No unit tests.

### DIFF
--- a/docs/v2/guides/extending-timber.md
+++ b/docs/v2/guides/extending-timber.md
@@ -138,7 +138,7 @@ add_filter( 'timber/post/classmap', function( $classmap ) {
 } );
 ```
 
-### Move your logic from your template files to your classes
+## Move your logic from your template files to your classes
 
 In the Getting Started Guide for [Archive Pages](/docs/v2/getting-started/a-post-archive/), we looked at how you can load related posts and add them to the context.
 
@@ -252,7 +252,7 @@ class BlogPost extends \Timber\Post {
 }
 ```
 
-### Magazine example
+## Magazine example
 
 Let’s look at an example where you work with a post’s relationship to other WordPress objects. In the following example, each post is a part of an "issue" of a magazine. We want an easy way to reference the issue in the Twig file:
 
@@ -303,13 +303,13 @@ class MagazinePost extends \Timber\Post {
 
 So now we’ve got an easy way to refer to `{{ post.issue }}` in our Twig templates.
 
-### Working with meta fields
+## Working with meta fields
 
 A very common use case is to use custom methods for handling meta values. Also refer to the [Custom Fields Guide](/docs/v2/guides/custom-fields/) for more information.
 
 Here, we take an custom post type `event` an example.
 
-#### Keep your logic in PHP
+### Keep your logic in PHP
 
 An event probably has a date that you define in the meta field `date_start`. You could get the date and manipulate the output format through Twig’s `date` filter.
 
@@ -399,7 +399,7 @@ class Event extends \Timber\Post {
 ```
 
 
-#### Extending IDs to Timber objects
+### Extending IDs to Timber objects
 
 Imagine that each `event` has a meta field that contains an array of post IDs of posts of the post type `sponsor`, with which you can select all the organizations that make your event possible with their donations.
 
--- a/docs/v2/guides/extending-timber.md
+++ b/docs/v2/guides/extending-timber.md
@@ -329,14 +329,14 @@ class Event extends \Timber\Post {
 	 *
 	 * @return string
 	 */
-    public function date_display() {
+    public function date_display( $date_format = 'F j Y' ) {
         $date_start = DateTimeImmutable::createFromFormat( 'Ymd', $this->meta( 'date_start' ) );
 
         if ( empty( $date_start ) ) {
 			return '';
 		}
 
-        return wp_date( 'F j Y', $date_start->getTimestamp() );
+        return wp_date( $date_format, $date_start->getTimestamp() );
     }
 }
 ```
@@ -344,7 +344,11 @@ class Event extends \Timber\Post {
 And use it in Twig:
 
 ```twig
+{# With your default date format #}
 {{ post.date_display }}
+
+{# With a different format #}
+{{ post.date_display('F j') }}
 ```
 
 This is better, because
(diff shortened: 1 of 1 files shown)


docs/v2/guides/extending-timber.md
```diff
@@ -3,107 +3,471 @@ title: "Extending Timber"
...
+/**
+ * Class Event
+ */
+class Event extends \Timber\Post {
+    /**
+	 * Gets display date.
+	 *
+	 * @return string
+	 */
+    public function date_display() {
```
acobster, line 333: I've had to implement this exact use-case a few times now. What I've found to be most powerful is to decouple the date format from the post type completely. Taking an optional parameter here makes that very easy to do:

```
public function date_display( $format = 'F j Y' ) {
  // ...
  return wp_date( $format, $date_start->getTimestamp() );
}
```

This could even apply to the other more complicated start/end date example below, where I imagine you could take `$format_long` and `$format_short` params in place of the hard-coded `F j` and `F j Y` strings, respectively.

I think then providing examples of calling `date_display()` with various formats would illustrate the power of this approach.

docs/v2/guides/extending-timber.md
```diff
@@ -3,107 +3,471 @@ title: "Extending Timber"
 order: "1600"
 ---
 
-Myth: Timber is for making simple themes. Fact: It's for making incredibly complex themes _look_ easy. But yes, you can also make simple sites from it.
+There’s a myth that Timber is for making simple themes, but in fact it’s for making incredibly complex themes _look_ easy. But yes, you can also make simple sites from it.
 
 The beauty of Timber is that the object-oriented nature lets you extend it to match the exact requirements of your theme.
 
-## An example that extends Timber\Post
+## Extending Timber objects
```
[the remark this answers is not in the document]
gchtr: Yeah, I probably wasn’t careful enough to use the correct terms. I’m glad if you call me out on those! Will update!

docs/v2/guides/extending-timber.md
gchtr: I make the date formats filterable in my own implementation. Having dynamic date formats is definitely a benefit.

> This could even apply to the other more complicated start/end date example below, where I imagine you could take `$format_long` and `$format_short` params in place of the hard-coded `F j` and `F j Y` strings, respectively.

Yes! Are `F j` and `F j Y` even the right date formats in English? I’m way more used to German date formats.

> I think then providing examples of calling `date_display()` with various formats would illustrate the power of this approach.

Sure! I’m going to add those.

docs/v2/guides/extending-timber.md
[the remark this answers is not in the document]
gchtr: Updated in https://github.com/timber/timber/pull/2260/commits/e7ac602c891b17d2f338296d34073694c5cd03e6

docs/v2/guides/extending-timber.md
gchtr: @acobster I made some updates in [https://github.com/timber/timber/pull/2260/commits/865d2ddd84ac2eeb32210392880aa4d112b0cf3a ](https://github.com/timber/timber/pull/2260/commits/865d2ddd84ac2eeb32210392880aa4d112b0cf3a). Would you mind checking it again?

acobster: This looks great! I didn't go over your full date example in great detail (i.e. for logic errors in the example code), but it looks like the right idea. Nice work!

> Are F j and F j Y even the right date formats in English?

Depends on the locale. I think Americans are more used to "January 3rd" whereas a Brit would be more likely to say "3 January"...but probably any English speaker will know what either one means. :)

jarednova:
Thanks @gchtr. This looks good! I just made a few tiny clarity tweaks (and learned something, I didn't know I could use Composer to set the namespaces so easily and not do a bunch of `require_once`s)

docs/v2/guides/extending-timber.md
```diff
@@ -573,7 +573,7 @@ function whateverify( $text ) {
 This can now be called in your twig files with:
 
 ```twig
-<h2 id="{{ post.title|slugify }}">{{ post.title|whateverify }}</h2>
```
gchtr, line 142: Are we using spaces for filters officially or do we follow the Twig syntax that is used in the Twig documentation?

gchtr:
@jarednova Thanks for looking over it. I’ve always used Composer’s built-in autoloader. It’s great I don’t have to care about autoloading, too :).

I only have [one question](https://github.com/timber/timber/pull/2260/commits/da47161f3dd0668bd52b1762ebce4b1981965e1a#r441300487) about Twig Coding Standards before merging this in.

docs/v2/guides/extending-timber.md
jarednova: My apologies. I thought spaces were Twig's syntax, but I looked this morning and you're correct — no spaces. I was up too late!

gchtr: I feel you, especially in these times 😴.

PR #2261, opened 2020-06-09
Add missing import to Factory

szepeviktor:
Discovered by @phpstan


gchtr:
Duplicate of #2246.

We’re going to fix the errors that PHPStan finds, Viktor, no worries! We just need to do some other things first 👍😉

szepeviktor:
# 🤖 

I may be a copy of Viktor

PR #2263, opened 2020-06-10
Whoops! Use the suggested syntax

jarednova:


szepeviktor:
Could it be that I'm wrong for years? I use `phpenv version-name` in many repos!

https://github.com/phpenv/phpenv/blob/master/libexec/phpenv-version-name

szepeviktor:
Found!

`- php: 5.6.40` -> `- php: 5.6`
then you'll get the latest version

kép

jarednova:
Nice! That makes a lot of sense given how `php: 7.4`, etc. operates

szepeviktor:
All green 🍏

Issue #741, opened 2015-11-11
Using theme.link in parent/child-theme on multisites-setup


Dodsaren:
Please see: https://github.com/jarednova/timber/wiki/Cheatsheet

Specifically:

> get_template_directory_uri() => {{ theme.link }} (Parent Themes)
> get_template_directory_uri() => {{ theme.parent.link }} (Child Themes)
> get_stylesheet_directory_uri() => {{ theme.link }}

If I use **theme.link** Timber will use **get_template_directory_uri()** if we're in the parent theme and **get_stylesheet_directory_uri()** if we're in the child theme.

However, if I want to inherit a complete template to my child, the child won't find whatever asset I'm looking for since it resides in the parent theme and we get the uri for the child theme.

Now if I use **theme.parent.link** instead my asset will load correctly from my child but the parent itself won't output anything.

This means that I have to overwrite the template file in my child theme and use _theme.parent.link_ there and _theme.link_ in my parent theme. Alternatively I have to copy my assets folder into my child OR create my uris with _get_template_directory_uri()_ in my backend with PHP/WP and pass them with my timbercontext.

Is this the intended behaviour? Shouldn't I be able to inherit this kind of functionality without having to overwrite the template file in my child theme or duplicate my assets?

Dodsaren:
For now a simple solution for my specific enquiry is to add my own variable to my global timbercontext that always gives me the uri to my parent theme root:

> $context["parent_link"] = get_template_directory_uri();

palmiak:
@Dodsaren Is this still something you're actively working on or need addressed? Please respond in two weeks or we'll close this thread

Rocketpilot:
This appears to still be broken in Timber 1.15. Seems to clearly be a bug in Timber itself and the workarounds @Dodsaren suggested aren't going to be practical for all users.

jarednova:
@Rocketpilot I'd love your insight to make sure we can solve this issue correctly. I realize it's an old issue, but it seems like a live one and meaningful to solve for the future.

The potential solve would be for `{{ theme.parent }}` to always return the parent theme (even if that's the SELF). The issue I see with this is any existing logic that tests for a parent theme would break. Example:

```twig
{{ theme.parent ? theme.parent.link : theme.link }}
```

This feels like an edge case though. I'm left with two potential solves:

A) In 2.x, break backwards compatibility (which I think is pretty edge-casey anyway) to make `{{ theme.parent }}` always give you a theme, even if it's the same theme as the one that's active
[...]

PR #2264, opened 2020-06-10
Resolve tricky {{ theme.parent }} behavior

jarednova:
**Ticket**: #741

## Issue
`theme.parent` has been tracking a theme's relative parent as opposed to the WP install's objective parent. This can cause issues when trying to link assets and other common use cases.

## Solution
Make `theme.parent` return the top-most theme even if it's `$this`

## Impact
Could break some theoretical edge-cases; thus applying this patch straight to 2.x. There was a cleanliness to the idea of relative parents; but given that WP's theme architecture doesn't really work that way, I don't think it's worth dwelling on the platonic ideal.

## Usage Changes
None

## Considerations
@gchtr suggests some [additional methods](https://github.com/timber/timber/issues/741#issuecomment-6410508950):

`{{ theme.file_uri('relative/path/to/asset') }}` which could use get_theme_file_uri().
`{{ theme.file_path('relative/path/to/asset') }}` which could use get_theme_file_path()

which I think are @todos subsequent to this PR which solves the issue

## Testing
New test to validate


lib/Theme.php
```diff
@@ -43,12 +43,12 @@ class Theme extends Core {
 	public $version;
 
 	/**
-	 * Timber\Theme object for the parent theme (if it exists), false otherwise
+	 * Timber\Theme object for the parent theme (even if the current theme is the parent theme)
```
gchtr, line 46:
```suggestion
	 * Timber\Theme object for the parent theme.
	 *
	 * Always returns the top-most theme. If the current theme is also the parent theme, it will
	 * return itself.
```

PR #2265, opened 2020-06-10
codecov should make tests fail

szepeviktor:
See https://github.com/codecov/codecov-bash/blob/master/codecov#L145

That stray "coveralls" was there for 5 months!

"after_success" executes first.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,7 @@ before_script:
     - if [ "$IMG_LIBS" == 1 ]; then pear config-set preferred_state beta; pecl channel-update pecl.php.net; yes '' | pecl install imagick; fi
     - bash bin/install-wp-tests.sh wordpress_test root '' localhost $WP_VERSION
     - composer update $PREFER_LOWEST --dev --prefer-source
+    - vendor/bin/pcov clobber
     - if [ "$COVERAGE" == 1 ]; then composer require satooshi/php-coveralls:1.0.*; fi
 
 script:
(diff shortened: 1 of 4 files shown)


szepeviktor:
Why doesn't PR-s got built?

szepeviktor:
My CI https://travis-ci.org/github/szepeviktor/timber/builds/696707899

szepeviktor:
> No code coverage driver available

phpunit depends on `phpunit/php-code-coverage:^6` which does not support pcov!

https://github.com/sebastianbergmann/php-code-coverage/blob/master/ChangeLog.md#700---2019-02-01

Could we upgrade to phpunit v8?

jarednova:
thanks @szepeviktor! Now that jobs run so much quicker w/ speedy 😉  I've enabled the PR builds. 

As for PHP Unit — WordPress unit tests are limited to 7.x:
https://make.wordpress.org/core/handbook/testing/automated-testing/phpunit/

Why? That'd be a question for the WordPress Gods.

jarednova:
@szepeviktor I think I need your help understanding this PR better. I think what you're showing is we had an unrecognized test failure with `php-code-coverage` and PCOV; but that was masked based on the travis build order.

We want to use PCov b/c it's way faster, but that's limited by the PHPUnit version. Which (if I'm following you here) means we need to look at following this article:

https://dev.to/swashata/setup-php-pcov-for-5-times-faster-phpunit-code-coverage-3d9c

... to properly run PHPUnit 7 w PCOV?

szepeviktor:
# 🎉 
@jarednova We've got some phpunit activity :) :) :)

jarednova:
awesome, we got this!

PR #2286, opened 2020-07-19
Fix coverage chaos in README

szepeviktor:
We have **two** coverage status providers.

@jarednova Looks very nice.

kép


jarednova:
Thanks @szepeviktor I think this is an opportunity to bring some sanity here (and cut down on all these automated thread comments too) by sticking to a single provider. I set both of these up circa-2015. I find the interface in Coveralls to be slightly easier to follow, but it's a light preference. Is there a coverage platform that's recognized as the standard these days? Either of these? Or something else?

szepeviktor:
My 2¢

1. Codecov's comment fills my screen (it is a disadvantage)
1. Coveralls has a fine-grained interface

jarednova:
In full agreement @szepeviktor. I prefer Coveralls as well. Merging and will remove the last remnants of the Codecov integration (and those annoying giant comments on PRs)

szepeviktor:
Thank you.

PR #2289, opened 2020-07-23
Get rid of IMG_LIBS variable in Travis config

szepeviktor:
Basically conditionals should be avoided in all CI workflows.
CI jobs should be listed explicitly.

I plan to remove `$COVERAGE` after this one gets merged.


jarednova:
I've got a ticket into @codecov-commenter to disconnect the repo so we stop getting those comments

jarednova:
👍  on to `COVERAGE`!

PR #2294, opened 2020-07-25
2.x Docs – Update section data to fix doc search

gchtr:
## Issue

Currently, when searchin the Docs, the sections are messed up, because I obviously used a selector that isnt’t unique on a page.

In the following example, you can see that a search for "Comments" is marked under the sections "Upgrade Guides" and "Contributing".

Bildschirmfoto 2020-07-25 um 13 25 00

This is because the section title is determined by the `.navMain-sectionTitle` selector, which appears multiple times. Algolia obviously has no way to know which section we’re in.

## Solution

This pull request changes the the title in the data from `title` to `sectionTitle` so we can reference it in the sidebar and use a unique CSS class that hopefully solves this issue. If we’d continue using `title`, it would always be overriden by a page’s title. So I had to use another name.

## Impact

Hopefully fixes search.

## Usage Changes

None.

## Considerations

None yet.

## Tests

I already put the changes live in https://github.com/timber/docs/commit/d593e247acb761ae951e63673dba0d7600eb2b88. I can check tomorrow if Algolia picked up the changes.

### DIFF
--- a/docs/v1/getting-started/getting-started.json
+++ b/docs/v1/getting-started/getting-started.json
@@ -1,4 +1,4 @@
 {
-  "title": "Getting Started",
+  "sectionTitle": "Getting Started",
   "layout": "page"
 }
--- a/docs/v1/guides/guides.json
+++ b/docs/v1/guides/guides.json
@@ -1,4 +1,4 @@
 {
-  "title": "Guides",
+  "sectionTitle": "Guides",
   "layout": "page"
 }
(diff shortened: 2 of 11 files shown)


gchtr:
Aaaand it seems to be fixed.



The remaining issue is that faceting doesn’t work properly yet. The search shouldn’t show results both for v1 and v2. Results vor v2 should only be shown when on https://timber.github.io/docs/v2, and vice versa. Will have to investigate, but in a separate issue.

jarednova:
👌

gchtr:
I fixed the remaining issue in https://github.com/timber/docs/commit/5f4ebdd86fb8d04206c84ce2a3d33004efb66aaf. Now, everything seems to work as expected 😎.

PR #2295, opened 2020-07-26
2.x – Improve documentation for pagination

gchtr:
## Issue

When looking at a couple of issues here in Timber, I felt that there was often confusion about pagination, because there are actually 3 types of pagination (that I know of).

## Solution

I tried to describe the different sorts of pagination and provide examples for how to use them. You can have a look at the updated guide here: https://github.com/timber/timber/blob/2.x-docs-pagination/docs/v2/guides/pagination.md.

I also updated some of the DocBlocks for the related methods.

I’m doing the pull request against the `2.x-docs-api` branch (#2073), because this affects the pagination for posts, which should work with `Timber::get_posts()` in the future. In v1, it only worked with `Timber\PostQuery`.

## Impact

Hopefully, this will provide better guidance.

## Usage Changes

None.

## Considerations

I wonder whether we actually need `Timber\Post::paged_content()`, or whether this could be baked into `Timber\Post::content()` so developers wouldn’t have to use a different method to make this work. In earlier versions of WordPress, you had to know about `<!--nextpage-->` to use this feature, but with the Block Editor, the Page Break block is always available. So this is not a hidden feature anymore. It could be confusing for end users if it doesn’t do anything when `{{ post.content }}` is used.

## Testing

Tested the code snippets in combination with the starter theme.

### DIFF
--- a/docs/v2/guides/pagination.md
+++ b/docs/v2/guides/pagination.md
@@ -5,15 +5,15 @@ order: "1100"
 
 There are three types of pagination, which all work differently, but can easily be confused with each other, because they use similar terms:
 
-1. Pagination for archive templates.
-2. Adjacent post pagination for singular templates.
-3. Paged content for posts.
+1. Pagination for archive templates
+2. Adjacent post pagination for singular templates
+3. Paged content within a post
 
 Let’s look at them in more detail.
 
 ## Pagination for archive templates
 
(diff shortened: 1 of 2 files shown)


jarednova:
@gchtr this does a great job of clarifying these three very different functionalities all sometimes casually referred to as "pagination" (and a good reference for us when people file reports). I just made some minor tweaks to further clarify the details. Please merge if those look good to you!

PR #2296, opened 2020-07-27
Remove all conditionals from CI configuration

szepeviktor:
From one side it is nice, from another side it is ugly

Now what??

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,14 +27,19 @@ env:
 matrix:
     include:
     - php: 5.6
-      env: WP_VERSION=4.9.8 WP_MULTISITE=0 PREFER_LOWEST="--prefer-lowest --prefer-stable" COVERAGE=1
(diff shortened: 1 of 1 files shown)


jarednova:
Hmmmm ... now why on earth does this boost coverage by 1%?

My big question with this @szepeviktor is whether this is better? There's quite a number of repeated commands. But I can also see how it's nice to be able to read through the 5.6 vs. 7.4 scripts in one place. You're carrying the burden here, so I trust your judgement; but want to make sure we're considering the ideal outcome

szepeviktor:
> boost coverage

Coveralls must agreeing with us in CI configuration: this must be its _bonus_ :)

> the 5.6 vs. 7.4 scripts in one place

Yes, I'd also stress **in one place**.

And generally: ideally separate thing (imagick, coverage, plain unit tests, static analysis) are in separate CI config files.

PR #2300, opened 2020-07-31
2.x menu factories

acobster:
Addresses the rest of the points in #2089.

## Impact

Better, more flexible Users and Menus APIs!

## Usage Changes

Yes, as we've discussed at length in #2073. :)

## Considerations

I implemented Class Maps for `MenuItem` slightly different from the [docs](https://github.com/timber/timber/blob/2.x-docs-api/docs/v2/guides/class-maps.md#the-menuitem-class-map), so that the callback gets a second arg of the menu item (a `WP_Post` instance) itself.

There are also a couple merge conflicts which I wasn't sure how to resolve.

**Next up, The Big One: the new Posts API.** I'll circle back to `PagesMenu` after Posts, which are more important.

## Testing

Updated a lot of tests to stop directly instantiating `Menu` instances but tried to mostly leave tests alone. Introduced a few helper functions in `Timber_UnitTestCase` that are a little more flexible, which we may want to migrate the rest of our tests to at some point. Also introduced tests for the new Factories.


jarednova:
wowza! Lots to dive into here. One quick Q: did you mean to set `2.x-docs-api` as the base branch?

acobster:
@jarednova yeah - I thought that's the branch we were targeting with all the Factory-related stuff until all of it is ready to merge into `2.x`. But I may be remembering wrong, as it's been approximately 300 years in quarantine since we talked about that. @gchtr ?

gchtr:
@acobster I think we said we'll use ‘2.x-docs-api‘ only for documentation, and ‘2.x-factories‘ for code related to factories.

I can review next week 👍

acobster:
>  "2.x-factories" for code related to factories.

Oh, that's right. Thanks for updating.

lib/Menu.php
```diff
@@ -279,32 +250,33 @@ public function find_parent_item_in_menu( $menu_items, $parent_id ) {
...
 				unset($item->title);
 			}
+
+			// Check it we're working with a post
 			if ( isset($item->ID) ) {
-				if ( is_object($item) && get_class($item) == 'WP_Post' ) {
-					$wp_post_menu_item = $item;
-					$item = new $this->PostClass($item);
-				}
-				$menu_item = $this->create_menu_item($item);
```
gchtr, line 296: I guess the `create_menu_item()` method can be removed, then?

lib/Timber.php
```diff
@@ -401,6 +396,87 @@ public static function get_user_by( string $field, $value ) {
...
+			]);
+			if ($all_menus && is_array($all_menus) && isset($all_menus[0]->term_id)) {
+				$menu = $all_menus[0]->term_id;
+			}
+		}
+
+		return $factory->from($menu, $options);
+	}
+
+	public static function _get_menu_id_from_locations( $slug, $locations ) {
```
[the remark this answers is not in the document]
acobster: oh, I totally meant to push that down into the Factory layer, or else into `Menu` itself. I definitely don't think `Timber` is the right place for it.

lib/Menu.php
acobster: Removed!

lib/Factory/MenuFactory.php
```diff
@@ -0,0 +1,71 @@
...
+		$term = wp_get_nav_menu_object($id);
+
+		if (!$term) {
+			return false;
+		}
+
+		return $this->build($term, $options);
+	}
+
+	protected function from_term_obj(object $obj, array $options) : CoreInterface {
```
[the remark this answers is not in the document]
acobster: Good call, done.

lib/Factory/MenuItemFactory.php
```diff
@@ -0,0 +1,57 @@
+<?php
```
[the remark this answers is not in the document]
acobster: Fixed.

lib/Menu.php
```diff
@@ -140,6 +157,9 @@ protected function init( $menu_id ) {
 		}
 
 		if ( $menu ) {
+			// @todo do we really need to call this fn? It's marked as "private" in the WP docs.
+			// Commenting out this line only breaks a single test: TestTimberMenu::testMenuTwigWithClasses
+			// ...maybe that means there's a way to accomplish what we need without calling a "private" fn.
 			_wp_menu_item_classes_by_context($menu);
```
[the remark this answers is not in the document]
acobster: #2304

lib/Timber.php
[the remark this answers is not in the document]
acobster: Pushed down into `MenuFactory`.

acobster:
@jarednova @gchtr I think this should be good to go now. Not sure what that test error is but I don't think it's related to Menus or Factories.

lib/Timber.php
```diff
@@ -428,28 +428,6 @@ public static function get_menu( $menu = null, array $options = [] ) {
 		return $factory->from($menu, $options);
 	}
 
-	public static function _get_menu_id_from_locations( $slug, $locations ) {
-		if ( isset($locations[$slug]) ) {
-			$menu_id = $locations[$slug];
-			// TODO do this in filter/Integration
```
gchtr: I see that you removed the special WPML check and just want to confirm: Can we really just ditch this? It was added in #1437.

The tests that we have would suggest that this still works. But if we still need to check this thoroughly, we can maybe do it in a separate issue.

acobster: Ack, you're right, I missed closing that loop because all the tests still passed.

@jarednova what needs to happen here? I'm not familiar with WPML and I'm not sure why it wouldn't fail once we take out those lines.

jarednova: @acobster I was scratching my head over that this morning too. Unless it breaks something in your work (or requires additional refactor), I'd like to restore these lines here ...

```php
if ( function_exists('wpml_object_id_filter') ) {
				$menu_id = wpml_object_id_filter($locations[$slug], 'nav_menu');
			}
```

We can then take it up on a separate ticket or PR about whether those tests are still valid, or perhaps some other check is necessary. This was a fix contributed specifically by the WPML team, so we can also follow-up with them if something in their code might have changed to make this no longer necessary.

acobster: Restored as a proper integration. I had to change the mock `wpml_object_id_filter()` implementation in `test-timber-wpml.php` but I think that's only because `TestMenuFactory::testGetMenuFromLocation()` introduced some coverage that wasn't there before: getting a `Menu` by _location name_ when the hard-coded `extra-menu` location is not defined:

```php
function wpml_object_id_filter( $element_id, $element_type = 'post', $return_original_if_missing = false, $language_code = null ) {
	$locations = get_nav_menu_locations();
	if (isset($locations['extra-menu'])) {
		return $locations['extra-menu'];
	}
	return $element_id; // <--- had to add this
}
```

jarednova:
The remaining build error is related to the (now resolved) bug with `twig_array_filter` — thus, the branch build fails while the PR test passes in the above checks

Issue #2301, opened 2020-08-05
Twig 2.13.1 introduces a third parameter for twig_array_filter()


gchtr:
This issue was raised in https://wordpress.org/support/topic/args-passed-to-twig_array_filter-must-be-an-instance-of-twigenvironment/.

In [this commit](https://github.com/twigphp/Twig/commit/d8259519c057360d1ba0fcb4ae9c3808977915d6#diff-6a6b43ace196ec261ae9f552264f9443R1696), they changed the parameters from

```php
function twig_array_filter($array, $arrow) {}
```

to

```php
function twig_array_filter(Environment $env, $array, $arrow) {}
```

Yes, they prepended a parameter. This now conflicts with the patch we introduced in [Timber 1.14.0](https://github.com/timber/timber/releases/tag/1.14.0), in #2124 to be exact. The tricky thing would be to get the Twig environment in there. Or maybe we need to take a completely different approach.

@palmiak Any idea how we could solve this?

palmiak:
Prepending the parameter was a great idea...

TBH - the only thing that comes to my mind would be to check Twig version and based on it use 2 or 3 params. I think it sucks a bit, but can't think of anything better.

jarednova:
Ugggggggh. I think we might want to consider splitting the solution between 1.x and 2.x here. My proposal would suggest ...

### Version 1.x
1. Introduce a kludge like what @palmiak suggests. Ugly but workable. BUT what about the `Environment`? is that available to the static function `Helper::filter_array()`? I don't believe it is. Hopefully it still performs if sent a null value. 
2. If the above is not true (it fails when sent a `null` value for `$env`) — we'll need to issue some kind of error or warning when `Helper::filter_array()` is called with Twig >= 2.13.1

### Version 2.x
1. Assume/require that Twig 2.13.1+ as the minimum version and go to 3 args. BUT what about the `Environment`? is that available to the static function `Helper::filter_array()`? I don't believe it is. Hopefully it still performs if sent a null value. 
2. If the above is not true, I don't think it's worth doing heroics. Better instead to remove and require that users rely on either custom or native Twig filters for similar behaviors.

jarednova:
Update: just ran a test and sending `null` does NOT work. Without going through the history of that commit/issue, it's hard to understand why this breaking change suddenly landed in the code? 

However, the good news is that we can bypass the call to Twig in the first place by replacing the line to ...

```php
[...]

PR #2302, opened 2020-08-06
Better control over Pagination stops

IJMacD:
I had the case where I needed to hide the last pages of pagination (the very oldest posts).

I found the `end_step` (not documented anywhere) controlled how many pages are shown at either end of the pagination range.
Additionally I was not able to set this to zero to hide the pages completely.

My proposal is to allow zero pages to be shown at the ends of the list. Also to enable different values for the start and end. The start value defaults to the same as the end value for backwards compatibility.

 

**Ticket**: # 

## Issue

 I needed to hide the last pages of pagination (the very oldest posts).

## Solution

My solution is to add a new optional parameter to the paginate_links arg array allowing control over the start and end separately.
Also allow zero as a valid value for the ends of the list.

## Impact

The new parameter defaults to the value of the old one so maintains perfect backwards compatibility.


jarednova:
@IJMacD thanks for the PR! Would you be up for writing some tests in `TestTimberPagination` to cover the new parameters you added?

The good news is that it seems that nothing about the prior behaviors are broken. While Travis is failing right now, it's due to an unrelated issue about changes in Twig (#2301)

IJMacD:
I've added the tests and checked against the original code. I found that setting `mid_size` to 0 was previously permitted yet would cause a PHP Notice about an 'Undefined index: link' causing the test suite to fail so I fixed that too.

I was aware of #2301. If you need to see green ticks from the tests I can rebase if you'd like?

jarednova:
This is great, thanks @IJMacD ! Really appreciate all the permutations of tests/sizes so we can make sure things (keep) behaving into the future.

PR #2303, opened 2020-08-06
2.x Guides optimizations

gchtr:
This pull request fixes two things.

1. Adds the *Hosts & Servers* Guide that was added in #2250 to version 2 of the docs as well.
2. Optimizes the Gutenberg guide by including the official name "Block Editor" and adding a missing `order` parameter.

## Usage Changes

None.

## Considerations

We need to remember that if we change documentation in `master`, we need to port it over to `2.x` as well, because `2.x` is the current master for the [live documentation](https://timber.github.io/docs/).

### DIFF
--- a/docs/v1/guides/gutenberg.md
+++ b/docs/v1/guides/gutenberg.md
@@ -1,10 +1,10 @@
 ---
-title: "Gutenberg"
+title: "Block Editor (Gutenberg)"
 ---
 
-## Using Gutenberg with Timber
+## Using the Block Editor with Timber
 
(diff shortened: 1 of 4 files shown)


jarednova:
> We need to remember that if we change documentation in master, we need to port it over to 2.x as well, because 2.x is the current master for the live documentation.

Good reminder! I'll make sure to note next time this comes up. FYI, the Travis failures are based on #2301 — not related to this PR.

gchtr:
Thanks, Jared!

PR #2305, opened 2020-08-07
Update Helper.php / twig_array_filter

jarednova:
**Ticket**: #2301

## Issue
As pointed out by @gchtr:

This issue was raised in https://wordpress.org/support/topic/args-passed-to-twig_array_filter-must-be-an-instance-of-twigenvironment/.

In this commit, they changed the parameters from
```php
function twig_array_filter($array, $arrow) {}
```
to
```php
function twig_array_filter(Environment $env, $array, $arrow) {}
```
Yes, they prepended a parameter. This now conflicts with the patch we introduced in Timber 1.14.0, in #2124 to be exact. The tricky thing would be to get the Twig environment in there. Or maybe we need to take a completely different approach.

## Solution
Simply replace the call to the native PHP `array_filter` function. I copied the third parameter of the `\ARRAY_FILTER_USE_BOTH` constant as well to match what Twig is doing internally.

## Impact
Fixes a fatal error when that function is called right now. 

## Usage Changes
None.

## Considerations
This is short-circuiting certain Twig checks around sandboxing, etc. Was anyone using them? Out of tens of thousands of Timber sites, maybe _somebody_, but I doubt a mass.

## Testing
Existing tests that were failing now pass

### DIFF
--- a/lib/Helper.php
+++ b/lib/Helper.php
@@ -427,6 +427,8 @@ class Helper {
 	 * Uses native Twig Filter.
 	 *
 	 * @since 1.14.0
+	 * @deprecated since 1.17 (to be removed in 2.0). Use array_filter or Helper::wp_list_filter instead
+	 * @todo remove this in 2.x
 	 * @param array                 $list to filter.
 	 * @param callback|string|array $arrow function used for filtering,
 	 *                              string or array for backward compatibility.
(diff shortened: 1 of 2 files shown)


jarednova:
As @gchtr pointed out in #2301 — the "source" Twig filter in `lib/Twig.php` was deprecated anyway in 2.x. I carried this deprecation along to `Helper::filter_array` so we remember to address that function as well.

gchtr:
👍

PR #2307, opened 2020-08-12
2.x Iterate on calling actions from Twig

gchtr:
**Ticket**: #1773

## Issue

@pascalknecht pointed out to me that we could use `do_action_ref_array()` instead of `call_user_func_array()` when we want to call an action with an array of arguments. We already changed the arguments logic in #1773. This pull request iterates on that.

## Solution

Apart from switching to `do_action_ref_array()`, I also found that we can use [arguments unpacking](https://www.php.net/manual/en/functions.arguments.php#functions.variable-arg-list), which is available since PHP 5.6. This syntax is similar to what some of us already know from modern JavaScript.

Without, it would look like this:

```php
$twig->addFunction( new TwigFunction( 'action', function() {
    $args        = func_get_args();
    $action_name = $args[0];

    array_shift( $args );

    do_action_ref_array( $action_name, $args );
} ) );
```

Compared to this, it seems like the new way is a lot easier to read:

```php
$twig->addFunction( new TwigFunction( 'action', function( $action_name, ...$args ) {
    do_action_ref_array( $action_name, $args );
} ) );
```

## Impact

None.

## Usage Changes

None.

## Considerations

None.

## Testing

I improved the tests by removing the called hooks after they were tested to prevent side effects when running all tests.


jarednova (approved):
Nice! Much cleaner. To clean up the tests further I cloned `Timber_UnitTestCase::add_filter_temporarily` into `Timber_UnitTestCase::add_action_temporarily`. That means you don't need to locally store or remember to do `remove_action`s in tests. Just `$this->add_action_temporarily(....`

gchtr:
> To clean up the tests further I cloned `Timber_UnitTestCase::add_filter_temporarily` into `Timber_UnitTestCase::add_action_temporarily`. That means you don't need to locally store or remember to do `remove_action`s in tests. Just `$this->add_action_temporarily(....`

@jarednova  Great idea! But did you miss to push your commits? I don’t see any commits from you yet.

I’ll leave it to you to merge this in.

jarednova:
Oh wait, I needed to actually _push_ my code? 🤦

PR #2309, opened 2020-08-14
fix #2216: delete sideloaded image tmp files

### LINKED ISSUE #2216: Sideloaded image tmp files are not cleaned up
## Expected behavior

After having sideloaded an external image, the temporary file created by `download_url` should be deleted.

Timber downloads the image using [`download_url`](https://developer.wordpress.org/reference/functions/download_url/) here https://github.com/timber/timber/blob/1.15.1/lib/ImageHelper.php#L376. If you read the docs of the function, it says:

> Please note that the calling function must unlink() the file.

This is never done and should be added below this line https://github.com/timber/timber/blob/1.15.1/lib/ImageHelper.php#L388.

## Actual behavior

Temporary files are never deleted.

Sorry, i dont have a real example at the moment but I can come back to this issue and update this later.

PHP 7.2
Timber: master

Github

oxyc:
**Ticket**: #2216

## Issue

After having sideloaded an external image, the temporary file created by download_url is not deleted.

## Solution

This deletes the temporary file after it has been uploaded to wordpress.

## Impact

No impact, bug fix.

## Usage Changes

None

## Considerations

None

## Testing

No unit tests added.

### DIFF
--- a/lib/ImageHelper.php
+++ b/lib/ImageHelper.php
@@ -378,14 +378,15 @@ class ImageHelper {
 		$file_array = array();
 		$file_array['name'] = PathHelper::basename($matches[0]);
 		$file_array['tmp_name'] = $tmp;
-		// If error storing temporarily, unlink
+		// If error storing temporarily, do not use
 		if ( is_wp_error($tmp) ) {
-			@unlink($file_array['tmp_name']);
 			$file_array['tmp_name'] = '';
 		}
 		// do the validation and storage stuff
 		$locinfo = PathHelper::pathinfo($loc);
 		$file = wp_upload_bits($locinfo['basename'], null, file_get_contents($file_array['tmp_name']));
+		// delete tmp file
+		@unlink($file_array['tmp_name']);
 		return $file['url'];
 	}
 


jarednova:
Thanks @oxyc ! Before getting into tests and such, I noticed that you moved the `unlink` call to within a conditional for `is_wp_error($tmp)` (see: https://github.com/timber/timber/pull/2309/files)

Wouldn't this only unlink/remove the file only when there's an error? In theory, the later call after the conditional _should_ remove the file whether an error is triggered or not.

oxyc:
Actually I moved it out of the conditional so that it happens regardless if it's an error or not.

jarednova:
@oxyc I am such a dumb-dumb and read that backwards. Yes! This does exactly what it should then. I'd still like to write a test to verify though. Is that something you'd be up for taking a stab at?

oxyc:
If no one else gets to it before, I can take a look when I have time. Unfortunate I’m spending most of my free time on other projects at the moment.

jarednova:
I took care of a test in #2317 and merged. Thanks again @oxyc !

Issue #2308, opened 2020-08-13
Implement new Post Collection(s) API


acobster:
Per discussion here: https://github.com/timber/timber/pull/2073#issuecomment-538048569

I believe that the best approach here is to define a `PostCollectionInterface`, and return a different concrete instance depending on whether we've been passed an array or a `WP_Query`. Here's the API I propose, based on what's currently in `PostQuery`, `PostCollection`, and `QueryIterator`:

```php
interface PostCollectionInterface extends Traversable, Countable, ArrayAccess {
  public function pagination( array $options = [] );
}

// NOTE: was called PostQueryIterator in previous draft, but I think this is less confusing.
class PostQuery implements PostCollectionInterface {
  public function pagination( array $option = [] ) {
    return new Pagination($options, $this->_query);
  }
  /* all the same traversal/counting methods currently in QueryInterface, plus ArrayAccess methods... */
}

class PostArrayObject extends ArrayObject implements PostCollectionInterface {
  public function pagination( array $options = [] ) {
    return null;
  }
}
```

A couple things to note:

* I chose to extend those interfaces specifically because they are close to the lowest common denominator of `ArrayObject` (which `PostCollection` extends) and `Iterator`/`Countable` (which `QueryIterator` implements). Also because `ArrayAccess` is nice, and easy to implement.
* `PostArrayObject` already implements `Traversable` etc. by virtue of extending `ArrayObject`.
* A Factory can decide which class to instantiate based on whatever was passed to `::from()`.

We should document `PostCollectionInterface` and the classes that implement it. Maybe a future version of Timber can expose a filter for overriding the concrete instance of `PostCollectionInterface` returned from `Timber::get_posts()` (i.e. a Class Map filter), but we don't need that right now.

jarednova:
all still seems to hold! I'm eager to see if things work out in practice as clean as they look in theory above — 🤞

gchtr:
@acobster Just checking: with #2311 being merged, can this issue be closed?

> We should document `PostCollectionInterface` and the classes that implement it.

Is this also done, or does it need some work?

acobster:
@gchtr Yep! Already documented [here](https://github.com/timber/timber/blob/2.x-docs-api/docs/v2/guides/posts.md#types-of-post-collections).

Issue #2207, opened 2020-03-01
Create a multisite guide


gchtr:
**Ticket:** #1434 

There are some limitations when getting posts from other pages in a multisite network. Specifically, when you fetch posts from other blogs, you’ll have to prepare the data in advance.

The linked issue has a very good example for this, that I’ve run into myself, too. I think we should document that in a new Multisite Guide.

There’s also some functionality in the [`Timber\Site`](https://timber.github.io/docs/reference/timber-site/) class that we could document from a different perspective in the new guide.

Issue #1434, opened 2017-05-24
{{ post.link }} not working when retrieving posts from subsites in multisite


kamontander:
### Expected behavior
I am working on a multisite install and I'm trying to list most recent posts from subsites on the main site home page with links to posts on the respective subsites.

### Actual behavior
{{ post.link }} has no output.

### PHP
    $context = Timber::get_context();
    $context['posts'] = Timber::get_posts();
    $templates = array( 'index.twig' );

    if ( is_main_site() ) {
      $context['news'] = array();
      for ($id=2; $id <= get_site_count(); $id++) { 
        switch_to_blog( $id );
        $args = array(
          'orderby'        => 'post_date',
          'posts_per_page' => 1
        );
        $latest_post = Timber::get_posts($args);
        $context['news'][] = $latest_post[0];
        restore_current_blog();
      }
    }

    Timber::render( $templates, $context );

### Twig
    {% for news_item in news %}
      <h3>
        <a href="{{ news_item.link }}"> // this is blank
          {{ news_item.title }}
        </a>
      </h3>
      <p>{{ news_item.content }}</p>
    {% endfor %}

### What version of WordPress, PHP and Timber are you using?
WordPress 4.7.5, PHP 5.4.19, Timber 1.2.4

### How did you install Timber?
Upgraded to newest version via plugin updater in WordPress dashboard

hatto:
not my preferred solution, but it does work
<a href="{{ TimberPost(news_item.ID).link }}">

designrubenz:
I just encountered the same problem – not with a multisite but a regular wordpress install.
WordPress: 5.3.2
timber: 1.10.1

### page.php
```
$feed = get_posts(array(
    'posts_per_page'    => -1,
    'post_type'         => 'post',
    'category_name'     => $category_name
));
$context['feed'] = $feed;
```

### page.twig

```
{% for feed_item in feed %}  
        {{ feed_item | print_r }}
        <h4><a href='{{ TimberPost(feed_item.ID).link }}'>{{ feed_item.post_title }}</a></h4>
 {% endfor %}. 
```

### debugging output (`feed_item | print_r`)
`
[...]

PR #2310, opened 2020-08-15
2.x post collections

acobster:
@jarednova @gchtr I wanted to check in about this before I get too far along, because it involves a pretty deep design change that I want to validate. This is the first attempt at addressing #2308.

## Summary

Right now, the logic for building and running `WP_Query`s is spread around `PostQuery`, `PostGetter`, `QueryIterator`, and elsewhere. Part of the goal of Factories is to consolidate all of that logic in one place: `PostFactory`. To allow that, `PostQuery` needs to give up some power, because right now it queries for posts itself via `PostGetter::query_posts()`. But if the Factory already knows what it's dealing with, there's no need for this - the factory can simply instantiate a `WP_Query` directly and wrap it with a `PostQuery`. Then there's exactly one class responsible for querying for collections of posts.

## What this means

* As of this PR, `PostQuery` now extends `ArrayObject` directly and implements the `PostCollectionInterface`.
* This means we can also get rid of the `PostCollection` class completely - just need to make sure the `::maybe_set_preview()` logic makes it out of there.
* I _think_ we can get rid of `QueryIterator`. More on that below.
* No more direct instantiation of `Timber\PostQuery` except when passing a `WP_Query`.
* I _think_ we can do away with the array wrapping the `WP_Query` in an array (i.e. `[ 'query' => $wp_query ]`) when passing to the constructor. The `post_class` option is a precursor to the new Class Maps design and the `merge_default` seems like something that would better be implemented as an option to `Timber::get_posts()`.
* As a nice bonus, we can save on performance and insantiate Posts inside of collection lazily, meaning while we already have the `WP_Post` objects they wrap, the `Timber\Post` instances themselves are not actually created until they're needed. See `PostsIterator` for how this works.
* Finally, we can completely get rid of `PostGetter`. I want to be careful here as a lot of high-level functionality expected from the `Timber` class is tested at the level of `PostGetter` instead, so I just want to be extra mindful about keeping that test coverage without lifting redundant tests up to the `Timber` level.

## A note about `QueryIterator`

As far as I can tell, this class has four responsibilities:

1. Implementing logic around `WP_Query` that (mostly) lives in `PostFactory` now
2. Of the special query logic mentioned above, the stuff that doesn't live in `PostFactory` are various query filter callbacks, namely `fix_number_posts_wp_quirk`, `fix_cat_wp_quirk`, and `handle_maybe_custom_posts_page`. This stuff should probably be moved 
3. Implementing a way to iterate over and count `WP_Query` results, which we already have in `PostQuery`.
4. Creating `Pagination` objects, which again, we can do directly in `PostQuery` (which is added in this PR).

If I'm missing something here and we need this class after all, please let me know! But if we can move out the logic we want to keep I'm hoping we can just get rid of it.

## The Plan

This PR isn't quite ready to merge but it seemed like a good stopping point to have this wider design discussion, since the high-level pieces are there. But quite a few tests are broken or skipped. I imagine I'll keep skipping many of them until I'm able to switch the Posts API altogether to using Factories, but some are probably easy fixes.

I'm leaning towards getting this to a good transitional point with Collections and merging this in then, before fully switching to Factories. Just don't want the PR to get too huge. Thoughts?

### DIFF
--- a/docs/v2/guides/pagination.md
+++ b/docs/v2/guides/pagination.md
@@ -3,185 +3,79 @@ title: "Pagination"
 order: "1100"
 ---
 
-There are three types of pagination, which all work differently, but can easily be confused with each other, because they use similar terms:
+Do you like pagination? Stupid question, of course you do. Well, here's how you do it.
 
-1. Pagination for archive templates.
-2. Adjacent post pagination for singular templates.
-3. Paged content for posts.
+## Default pagination
 
-Let’s look at the two in more detail.
-
-## Pagination for archive templates
-
-The pagination for archive pages only works for template files with an active query of multiple posts, like **archive.php** or **home.php**. When you have a lot of posts, you wouldn’t list all of the posts on one page, because that would slow down your site too much. Instead, you would only show a defined number of results and use other pages that show the next set of results. Here’s an example:
-
-- `posts/` – Shows first 10 posts.
-- `posts/page/2` – Shows posts 11-20.
-
-### Default pagination
-
-**archive.php**
+This will only work in a php file with an active query (like `archive.php` or `home.php`):
 
 ```php
-<?php
-
-$context = Timber::context();
-
-Timber::render( 'archive.twig', $context );
+	<?php
+	$context = Timber::context();
+	$context['posts'] = new Timber\PostQuery();
+	Timber::render('archive.twig', $context);
 ```
 
-Because we’re on an archive page, Timber already prepared a `posts` variable in the context for us. You could then markup the output like so:
-
-**archive.twig**
+You can then markup the output like so  (of course, the exact markup is up to YOU):
 
 ```twig
 <div class="tool-pagination">
 	{% if posts.pagination.prev %}
-		<a
-            href="{{ posts.pagination.prev.link }}"
-            class="prev {{ posts.pagination.prev.link|length ? '' : 'invisible' }}"
-        >Previous</a>
+		<a href="{{posts.pagination.prev.link}}" class="prev {{posts.pagination.prev.link|length ? '' : 'invisible'}}">Prev</a>
 	{% endif %}
-
 	<ul class="pages">
 		{% for page in posts.pagination.pages %}
 			<li>
 				{% if page.link %}
-					<a
-                        href="{{ page.link }}"
-                        class="{{ page.class }}"
-                    >{{ page.title }}</a>
+					<a href="{{page.link}}" class="{{page.class}}">{{page.title}}</a>
 				{% else %}
-					<span class="{{ page.class }}">{{ page.title }}</span>
+					<span class="{{page.class}}">{{page.title}}</span>
 				{% endif %}
 			</li>
 		{% endfor %}
 	</ul>
-
 	{% if posts.pagination.next %}
-		<a
-            href="{{ posts.pagination.next.link }}"
-            class="next {{ posts.pagination.next.link|length ? '' : 'invisible' }}"
-        >Next</a>
+		<a href="{{posts.pagination.next.link}}" class="next {{posts.pagination.next.link|length ? '' : 'invisible'}}">Next</a>
 	{% endif %}
 </div>
 ```
 
-### What if I’m not using the default query?
-
-If you want to overwrite the default query, you can do that by overwriting `posts` in the context with a new query. Make sure to include the `paged` parameter in the query.
-
-**archive-event.php**
+## What if I'm not using the default query?
 
 ```php
-<?php
-
-global $paged;
-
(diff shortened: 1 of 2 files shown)


acobster:
Not sure why there are so many old commits in here since I based this branch off of `2.x-factories`, but that's a problem for future Coby to solve...

gchtr:
> Not sure why there are so many old commits in here since I based this branch off of `2.x-factories`, but that's a problem for future Coby to solve...

That’s probably because [you merged `2.x-docs-api` into `2.x-factories`](https://github.com/timber/timber/pull/2194/commits/5393189a55e3c110e2da42e85cf434892fff2b6e). However, the `2.x-docs-api` branch is pulled against the `2.x` branch in #2073. That’s why all the commits from that PR also show up here.

gchtr:
And I realized it’s hard to look at what you changed if all those commits from #2073 are in here as well. You probably have to revert that somehow.

Up until here, what you write makes sense to me. But I also have to say that the exact inner workings of all the classes that are in involved here were always a bit hard to keep track of for me. So if we can reduce the complexity here, that would be great 👍. 

>     * I _think_ we can do away with the array wrapping the `WP_Query` in an array (i.e. `[ 'query' => $wp_query ]`) when passing to the constructor. The `post_class` option is a precursor to the new Class Maps design and the `merge_default` seems like something that would better be implemented as an option to `Timber::get_posts()`.

Yes! I think I already partially worked on this in #2145. Maybe you could build on that.

>     * As a nice bonus, we can save on performance and insantiate Posts inside of collection lazily, meaning while we already have the `WP_Post` objects they wrap, the `Timber\Post` instances themselves are not actually created until they're needed. See `PostsIterator` for how this works.

Doesn’t this have big implications on caching if the actual work is only done after fetching cached posts in PHP templates during `Timber::render()` or earlier?

And then this is probably also related to handling posts in multisite environments (see #1434 and #2207).

And from what I see in issues and StackOverflow questions, a lot of developers are relying on dumping post objects to see what’s in there. This of course wouldn’t be reliable anymore. I’m not saying I wouldn’t add lazy-loading, but we need to check what it means for other parts of Timber and probably add documentation about this in various places.

A related example: we already fetch meta values lazily. A lot of developers were used to work with `{{ dump(post.custom) }}` to see what’s there. Now, this won’t work anymore. See https://timber.github.io/docs/v2/guides/custom-fields/#accessing-all-meta-values.

> If I'm missing something here and we need this class after all, please let me know! But if we can move out the logic we want to keep I'm hoping we can just get rid of it.

I’m all in favor of simplifying this. What I figured when doing more complex stuff with iterators is that as soon as we have factories ready, we can move some logic regarding globals into `Post::setup()` method.

One specific example is this class: https://github.com/mindkomm/timber-integration-woocommerce/blob/master/lib/ProductsIterator.php. We discussed this in https://github.com/mindkomm/timber-integration-woocommerce/issues/6. With factories, we can get rid of it completley and a WooCommerce product object might solve WooCommerce globals like this:

```php
class Product extends Timber\Post {
    public function setup( $loop_index = 0 ) {
        parent::setup( $loop_index );

        wc_setup_product_data( $this->ID );

        return $this;
    }
}		
```

acobster:
> That’s probably because you merged 2.x-docs-api into 2.x-factories...You probably have to revert that somehow.

Dang, OK, I'll figure out how to do that.

> I think I already partially worked on this in #2145. Maybe you could build on that.

Awesome! I will take a look there, thanks!

> Doesn’t this have big implications on caching if the actual work is only done after fetching cached posts in PHP templates during Timber::render() or earlier?

I don't think I understand what you mean. :) Which code are you referring to as "fetching cached posts," and how is that affected by this change?

I agree we'd want to document the effects of the change if it actually affects the behavior of the API. But as far as I can tell this will be completely transparent to the caller. That is, by the time you have a `post` var within a Twig template, you've already done all the work that's needed. And, to be clear, this only affects how `Post`s are instantiated in `PostQuery` for now; in every other context they're instantiated eagerly. So I'm not sure why `{{ dump(post.custom) }}` wouldn't work.

Again, maybe I'm missing something...wouldn't be the first time. :) I'm also cool with being cautious and reverting the lazy instantiation piece for now. If I'm right and it's truly transparent, that means we can roll it out after 2.0 lands.

> I’m all in favor of simplifying this. What I figured when doing more complex stuff with iterators is that as soon as we have factories ready, we can move some logic regarding globals into Post::setup() method.

Makes sense to me! Good idea.

acobster:
OK, `git push --force` auto-closed this PR because I reset it to be exactly the same as 2.x-factories. Gonna cherry-pick my commits back into this branch and reopen. Sorry for the confusion.

PR #2314, opened 2020-08-24
#2301 Keep Twig native `filter` parity

nlemoine:
#2301

## Issue

The recent changes made in #2301 triggered many errors in my templates (`array_filter` running on `PostColelction` object). I use `filter` a lot and I use it against `PostCollection` which isn't an array. 

## Solution

Implement the native `twig_array_filter` method: https://github.com/twigphp/Twig/blob/1fb577363d3e3b5ce1b737b3255f2491b5f80833/src/Extension/CoreExtension.php#L1588-L1593

## Impact

It shouldn't because that method was used before that change.

## Usage Changes

None.

### DIFF
--- a/lib/Helper.php
+++ b/lib/Helper.php
@@ -446,7 +446,7 @@ class Helper {
 		}
 
 		// the IteratorIterator wrapping is needed as some internal PHP classes are \Traversable but do not implement \Iterator
-		return new \CallbackFilterIterator( new \IteratorIterator($array), $arrow );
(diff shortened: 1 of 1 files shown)


jarednova:
This is great, thanks for the tweak @nlemoine ! I'll merge and tag as a new release later today so you can start using asap

nlemoine:
🙏 Thanks a lot for the quick merge @jarednova!

PR #2315, opened 2020-08-26
Update Timber::get_posts() deprecated instances

jarednova:
A (potential) adjustment to #2311. This updates `Timber.php` to point users towards `Timber::get_posts()` instead of `Timber::query_post()` or `new Timber\PostQuery()` calls. This also handles for what comes back from `Timber::context()`


jarednova:
All this (and much more) seems covered in #2316

PR #2319, opened 2020-08-28
Add webp to recognized image extensions

jarednova:
## Issue
`webp` isn't in the list of recognized image extensions. So `webp` images will fail `Image::is_image()` check

## Solution
Add it!

## Impact
Should resolve unseen issues of it upper-level functions failing.

## Usage Changes
None

## Considerations
We've been unable to test against Webp previously on Travis. This would be a good time to figure that out (I've been able to test locally successfully, it seems to be because of how the Travis PHP build is configured)

## Testing
We've been unable to test against Webp previously on Travis. This would be a good time to figure that out (I've been able to test locally successfully, it seems to be because of how the Travis PHP build is configured)

### DIFF
--- a/lib/Image.php
+++ b/lib/Image.php
@@ -497,7 +497,7 @@ class Image extends Post implements CoreInterface {
 	 */
 	protected function is_image() {
 		$src = wp_get_attachment_url($this->ID);
-		$image_exts = array( 'jpg', 'jpeg', 'jpe', 'gif', 'png' );
+		$image_exts = array( 'gif', 'jpg', 'jpeg', 'jpe', 'png', 'webp' );
 		$check = wp_check_filetype(PathHelper::basename($src), null);
 		return in_array($check['ext'], $image_exts);
 	}


jarednova:
@szepeviktor I've been on a year-long journey of banging my head against the wall to get webp going with Travis (needing the `imagewebp()` function). Is this something you have any experience with?

szepeviktor:
Let me check this.
https://github.com/szepeviktor/debian-server-tools/blob/master/webserver/Continuous-integration-Continuous-delivery.md
Although my head is empty, maybe I find something in my notes....

szepeviktor:
> Exception: Function imagewebp() does not exist

I see!!!

szepeviktor:
Who's that guy asking that smart question??
https://travis-ci.community/t/how-to-enable-webp-image-format-support-in-php-7/3209/2

szepeviktor:
Basically we need another build of the PHP engine. TravisCI does not include WebP functions.

jarednova:
@szepeviktor thanks for looking into it! At least we know that there's no "easy way". Not sure if it makes sense to switch the entire build over to that given that this is just a handful of tests (and I'm able to verify them all locally if/when webp-specific issues come up)

szepeviktor:
> this is just a handful of tests

Oh nooooooo!
Let's create a CI step that runs WebP test on a certain PHP version.
Could you make its PHP part possible?

jarednova:
> Let's create a CI step that runs WebP tests on a certain PHP version.
I'm good with that — let's just recognize it as its own thing so it doesn't slow down the bulk of tests

> Could you make its PHP part possible?
... didn't quite follow you here ...

szepeviktor:
> ... didn't quite follow you here ...

I'd add a new CI step supporting WebP testing but I do not know how to start those tests only.

szepeviktor:
By a new `phpunit-webp.xml` file??

jarednova:
@szepeviktor ahhh, yes! You can send that as a param to php unit via `--configuration`. This StackOverflow has a little bit more:

https://stackoverflow.com/questions/3671685/how-to-run-a-specific-phpunit-xml-testsuite

szepeviktor:
:) You are tricky!

We do not have that file :-D

I cannot write those tests.

PR #2321, opened 2020-08-31
Run WebP related tests also

szepeviktor:
From https://github.com/timber/timber/pull/2319#issuecomment-683980238

@jarednova Here we(bp) go!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,7 +83,7 @@ jobs:
         # Install PHP with WebP support
         - LC_ALL=C.UTF-8 sudo add-apt-repository -y ppa:ondrej/php
         - sudo apt-get update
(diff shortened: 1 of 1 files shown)


szepeviktor:
> OK (6 tests, 18 assertions)

🍏

szepeviktor:
Will adding `--stop-on-skipped` make skipped tests error out??

jarednova:
WOW, that was quick!

> Will adding --stop-on-skipped make skipped tests error out??

It stops the tests, however the result isn't an error ...
Screen Shot 2020-08-31 at 10 04 18 PM

szepeviktor:
Sad!

Please merge.

jarednova:
Excellent! I like how this keeps the webp tests separate (and a breezy 2:29 to run). Even if a small amount of time could be saved by merging it with one of the other 7.2 tests; I think it's easier to manage as its own element to enable/disable/update as we need

PR #2322, opened 2020-09-02
2316/deprecate term getter

jarednova:
Deprecate the `TermGetter` class

### DIFF
--- a/lib/TermGetter.php
+++ b/lib/TermGetter.php
@@ -3,7 +3,10 @@
 namespace Timber;
 
 /**
- * Class TermGetter
+ * Class TermGetter 
+ *
+ * @deprecated 2.0.0 use Timber::get_terms or Timber::get_term()
+ * @codeCoverageIgnore
  */
 class TermGetter {
 	/**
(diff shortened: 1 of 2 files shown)


jarednova:
I think this is ready for a quick look @acobster or @gchtr. There is one failing test related to `PostQuery` being used in `Timber::context();` ...

```
1) TestTimberContext::testPostsContextSimple
Unexpected deprecated notice for Timber\PostQuery::get_posts()
Failed asserting that an array is empty.

/srv/www/wordpress-trunk/public_html/tests/phpunit/includes/abstract-testcase.php:494
/srv/www/timber/tests/Timber_UnitTestCase.php:17
/srv/www/wordpress-trunk/public_html/tests/phpunit/includes/abstract-testcase.php:506
```

... but this isn't connected to the code here and I believe is addressed in #2316

acobster:
@jarednova looks good to me (and you're correct about that `PostQuery` usage), although I was kinda planning on deleting this class (among others) as part of THE PURGE. :hocho: It's not documented AFAIK and will break expectations set in the docs if it's used. If you see a good reason to keep it around as deprecated, I suppose I can stay my bloodlust.

jarednova:
@acobster cool! let's take up the question of deprecation vs. blood over on #2323

PR #2324, opened 2020-09-15
Pass more variables to the `timber/loader/loader` filter

neojp:
## Issue

In the process of trying to override the `\Twig\Loader\FilesystemLoader` passed over on the WordPress filter `timber/loader/loader` it is required to pass over `$paths`, and `$rootPath` as arguments to the Loader class as seeing here: 

https://github.com/timber/timber/blob/master/lib/Loader.php#L149

```php
$fs = new \Twig\Loader\FilesystemLoader($paths, $rootPath);
```

The problem lies in the fact that the provided filter doesn't pass over the needed variables to accomplish this:

https://github.com/timber/timber/blob/master/lib/Loader.php#L150

```php
$fs = apply_filters('timber/loader/loader', $fs);
```

## Solution

This can be resolved by simply passing over the required variables like in this PR:

```php
$fs = apply_filters('timber/loader/loader', $fs, $paths, $rootPath);
```

## Impact

There should not be any impact to current users.

```php
add_filter( 'timber/loader/loader', 'custom_loader_override', 10, 3 );
```

## Usage Changes

Current users using this filter would keep getting the `$fs` variable by default by simply using the `add_filter()` function:

```php
add_filter( 'timber/loader/loader', 'custom_loader_override' );
```

New users that require these additional variables could request it in the `add_filter()` function like this:

```php
add_filter( 'timber/loader/loader', 'custom_loader_override', 10, 3 );
```

## Considerations

This is indeed only useful for advanced users, but it opens up possibilities to extend for custom templates paths to be used such as the ones from [Fractal Twig Adapter](https://fractal.build/).

## Testing

There won't be need for additional tests.

### DIFF
--- a/lib/Loader.php
+++ b/lib/Loader.php
@@ -147,7 +147,7 @@ class Loader {
 			$rootPath = null;
 		}
 		$fs = new \Twig\Loader\FilesystemLoader($paths, $rootPath);
-		$fs = apply_filters('timber/loader/loader', $fs);
+		$fs = apply_filters('timber/loader/loader', $fs, $paths, $rootPath);
 		return $fs;
 	}
 


jarednova:
@gchtr wanted to run this by you. It doesn't seem like it would conflict w anything going on in 2.x:

https://github.com/timber/timber/blob/2.x/lib/Loader.php#L301

... but interested if you had an opinion in regards to how the arguments are handled: Is there something that can/should be done to match some of the other argument patterns we're using with factories? Or would you consider this to be the wrong application of that concept? (Can't say I've encountered that argument style elsewhere with WP filters/hooks)

gchtr:
@neojp Thanks for filling out the pull request so properly!

@jarednova To me, it seems very reasonable to pass these parameters to that filter, if they are useful in the filter.

The only thing we could consider would be to pass these two variables in an array:

```php
$fs = apply_filters('timber/loader/loader', $fs, [
    'paths'     => $paths,
    'root_path' => $rootPath, // Will be $root_path in 2.x
] );

add_filter( 'timber/loader/loader', 'custom_loader_override', 10, 2 );
```

With function parameters, it’s sometimes better to pass them in an associative array, especially if you have many parameters. There’s also the problem of argument order, because if you for example have to change the last parameter, but not all others, you need to repeat default values for the ones you don’t need. But with filters, you *get* all parameters, so that’s not a problem here.

The `$paths` and `$pathInfo` variables are practically the only variables that we set in that function and there’s no indication that there will be a lot more info to pass to that filter in the future. If we knew or had a gut feeling that there would be 10 more valuable variables in the future, using an array would probably make sense here. But that’s not a problem now, and we could still deprecate the filter, or introduce a new filter with different info.

So I’m good with this as it is now 👍.

neojp:
@gchtr I agree. I can change the code to use an associate array. Like you said, this change should make it future-proof in case there is a need to pass additional variables.

gchtr:
> @gchtr I agree. I can change the code to use an associative array. Like you said, this change should make it future-proof in case there is a need to pass additional variables.

@neojp I meant that **we don’t have to use an array**, it’s just fine by passing the two variables. I’m usually just thinking out lout to come to a conclusion.

jarednova (approved):
Approving based on the convo between @neojp and @gchtr. To note: even though we're not using an associative array for arguments, the arguments aren't arbitrary: following the order of `Twig\Loader\FilesystemLoader`

PR #2325, opened 2020-09-16
Use WP-CLI stubs package

szepeviktor:
Make our repo a bit smaller.


phpstan.neon
```diff
@@ -1,5 +1,4 @@
 includes:
-    - phar://phpstan.phar/conf/bleedingEdge.neon
```
szepeviktor, line 2: This was experiemntal thus making analysis unstable.

jarednova:
Awesome, thanks @szepeviktor — I'm all for slimmer! There's an unrelated failing test for `TestTimberPagination::testPaginationEndLimits` that I'll take in a follow-up ticket.

PR #2327, opened 2020-09-28
upgrade Loader to use renamed Twig\TwigFunction

acobster:
If `WP_DEBUG` is disabled, a Fatal Error will be thrown: "Class 'Timber\Twig_Function' not found". This just fixes the name so we're using the right class.

## Issue

I'm just starting to integrate our budding 2.x alpha on a real project, and this is the first thing I ran into. Looks like a code path that's probably not covered by the tests, but I don't have time to figure out how to test using different constants. But the fix is about as simple as it gets.

## Solution

`use` the correct `Twig\TwigFunction` class within Loader.

## Impact

The thing will work now! :smile: 

## Usage Changes

None

## Considerations

How can we test this reliably? Is it worth it?

## Testing
None for now.

### DIFF
--- a/lib/Loader.php
+++ b/lib/Loader.php
@@ -7,6 +7,7 @@ use Twig\CacheExtension;
 use Twig\Environment;
 use Twig\Extension\DebugExtension;
 use Twig\Loader\FilesystemLoader;
+use Twig\TwigFunction;
 
 class Loader {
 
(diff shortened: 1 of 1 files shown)


jarednova:
d'oh! Great find. In terms of testing: unfortunately, there's no easy way to enable/disable PHP constants once they've been set. I banged my head against a wall for awhile trying to get things going with [PECL / Runkit](https://www.php.net/manual/en/function.runkit-constant-redefine.php) at one point. But after the bleeding stopped, I was no closer. I'll let @gchtr take a formal look at this, but my vote is to fix the problem and make testing against `WP_DEBUG` (or other constant considerations) a future `@todo`

gchtr (approved):
Looks good! 👍

@jarednova I’ll leave it to you to create an issue for a future todo about testing with `WP_DEBUG`.

PR #2330, opened 2020-10-05
2.x Use stable release of twig/cache-extension

gchtr:
## Issue and Solution

Up until some time ago, there wasn’t any stable release for [twig/cache-extension](https://github.com/twigphp/twig-cache-extension). But luckily, now [there are](https://github.com/twigphp/twig-cache-extension/releases). Let’s use them.

## Impact

Better control and predictability for when a new version of `twig/cache-extension` comes out.

## Usage Changes

None.

## Considerations

None.

## Testing

Should be covered by existing tests.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
     "php": "^7.0",
     "twig/twig": "^2.10",
     "composer/installers": "~1.0",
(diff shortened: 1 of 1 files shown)


jarednova:
most definitley!

Issue #2229, opened 2020-04-11
code style fixing ?


flip111:
Would it be a good idea to apply https://github.com/FriendsOfPHP/PHP-CS-Fixer ??

Maybe to setup a custom profile in case it needs some different syntax in some cases.

gchtr:
Hey @flip111, thanks for the suggestion. We’re aware that we can improve our coding style. This is something that we keep track of in #2166. We didn’t decide yet which tools we’re going to use, but PHP-CS-Fixer is surely one of those we’re going to look at.

I’m going to close this issue, because it’s already on our todo list ;).

flip111:
Thank you @gchtr

PR #2166, opened 2020-01-11
2.x Check coding style

szepeviktor:
There should be at least 3 basic checks

- syntax
- **coding style**
- static analysis

PR #2332, opened 2020-10-06
2.x WIP Update coding standards

gchtr:
## Issue

We’ve talked about applying WordPress Coding Standards as part of the 2.x update.

## Solution

This is very much work in progress. Part of this is also still in exploration. Currently, I’m trying to figure out

- which issues can be fixed with `phpcbf`
- which of the remaining issues after applying `phpcbf` we’ll want to enforce
- which adaptions to the default WordPress Coding Standards we’ll want to make

**Related pull requests and issues:**

- Fix problems reported by PHPStan #2246
- 2.x Check coding style #2166
- code style fixing ? #2229

## Impact

Clean code.

## Usage Changes

None yet.

## Considerations

We’ll see.

## Testing

We’ll see.


szepeviktor:
Automattic itself has a ruleset for PHP 7
https://github.com/Automattic/phpcs-neutron-ruleset

nlemoine:
@jarednova @gchtr Can we **quickly** adopt one random/well established standard while thinking of the future perfect one? At least for consistent syntax because it's currently a mess and a bit frustrating 😄 

I'm personally using [Easy Coding Standard](https://github.com/symplify/easy-coding-standard). It has [base default rule sets](https://github.com/symplify/easy-coding-standard/tree/main/config/set) you can pick from, it's fast and it can fix most things automatically. But any (modern) standard will do. 

> Done is better than perfect

🙏

szepeviktor:
> the future perfect one

🎉 a useful frankenstein 🥳
https://github.com/szepeviktor/phpcs-psr-12-neutron-hybrid-ruleset

jarednova:
@nlemoine thank you for flagging this as an important. I didn't realize how this was proving a blocker (or slowing things down).

I'm not familiar with Easy Coding Standard, but your endorsement obviously goes a long way. You and @gchtr are the most active right now in pushes, so I will support whatever you two determine as the "rule" (whether it's a recognized one or a modified one). Very much agree that perfect can always be accomplished at a later point!

nlemoine:
> I didn't realize how this was proving a blocker

It's not that important. A space before variable never stopped code from running. But it's always nice to read consistent written code and I feel like the quest for the perfect standard may last a long time so I decided to rush it a bit 😉

ECS is just PHP_CodeSniffer and PHP-CS-Fixer running together. But we could chose one or the other. From my experience with fixers, PHP-CS-Fixer is a better at fixing code than `phpcbf` that comes with `phpcs`.

I would start with a base standard such as PSR-12 and [a few quite common rules](https://github.com/symplify/easy-coding-standard/blob/main/config/set/clean-code.php) (short array syntax, etc.). We will add more rules further based on issues/needs. 

What do you think @gchtr?

szepeviktor:
> We will add more rules further based on issues/needs.

If you prune formatting rules from WPCS all you are left is - almost - security related ones!

gchtr:
> And I would add "Some is better than none" 😉 🙏
> (I'm aware there's a phpcs.xml file at the root of the repo but it's not used/checked)

Totally agree. We introduced a `composer lint` command some time back that uses the phpcs.xml, but we didn’t manage to add to the CI process yet.

I don’t have a strong opinion on which standard to choose. I initially started introducing the WordPress coding standards because with Timber, we work in the WordPress world. And even though WordPress isn’t renowned for modern PHP code, there are some rules in there that are related to security (as Viktor mentioned) or about function usage. For example, it might tell you where you could use a more suited function.

All in all, I think a standard should be here to make our lives easier. I just would want a standard that we can control, so that we can selectively apply rules or modify them.

The only thing I have a stronger opinion about is method and property naming. I wouldn’t change to camel case in function names and property names, just because it’s in a PSR standard. I would prefer it if the public facing part of Timber would match the WordPress naming conventions. But since that would introduce a lot of breaking changes I think we wouldn’t do it anyways.

@nlemoine If you feel like Easy Coding Standard would work well, especially in combination with CI and GrumPHP, then I’m all for it. And I’m happy if you want to work on this, because I have so many other things to do 😅.

jarednova:
> And I’m happy if you want to work on this, because I have so many other things to do 😅.

@gchtr based on this, can you re-assign to @nlemoine ?

gchtr:
> All right Lukas! **Timber Coding Standard** is born 👶

@szepeviktor Yeah, probably not a good idea of mine now that you state it like that 🙈.

@nlemoine I definitely don’t want to hold you back if you want to introduce solid coding standards. I’m sure you have more experience with this and I’m really glad if you can work on this. I’m probably too biased towards the WordPress Coding Standards because I’m really used to those and they work well for me. I don’t want you to see me as the decision maker here, I’m also only a contributor.

@jarednova I think this pull request is not a good foundation to continue working on the coding standards. I suggest to open a new pull request. With this one, I quickly realized it’s quite some work to figure out how to integrate all the parts. It looks like the Easy Coding Standards suggested by @nlemoine is a better foundation.

I’ll leave this open in case we need to continue the discussion.

szepeviktor:
In my unpopular opinion with a tool name "Easy" you cannot dig deep enough. 🪚

nlemoine:
> with a tool "Easy" in its name you cannot dig deep enough. 🪚

Not sure the name is to blame here 😄

ECS is just a tool that takes rules/sniffs from PHPCS and PHPCSF and run them. It has some valuable benefits:
- it's blazing fast
- it has nice features such as checking code styling in Markdown files
- it allows selecting rules from both existing tools
- it provides a far better DX than other tools

The only downside, IMHO, is that you can't use predefined sets of those tools (either XML or PHP files) unless you build them by your own which is indeed a big bummer we cannot ignore. 

It would be nice to have [PHP Compatibility](https://github.com/PHPCompatibility/PHPCompatibility), WordPress Compatibility(https://github.com/PHPCompatibility/PHPCompatibilityWP) and some other WP rules/sniffs in our CS.

This is a first shot here, I'll try to find a better alternative soon. Note that some standards might too "modern" for Timber as it is. [Inpsyde PHP Coding Standards](https://github.com/inpsyde/php-coding-standards) is a good candidate but enforces types everywhere, etc. that we are still missing.

jarednova:
Thanks @gchtr for resurrecting the convo and @nlemoine for getting the ball rolling on #2592. I agree with @szepeviktor that this is the moment to make the right changes to 2.x that will bring us more in line with modern standards. Now, how to best do that? Always the big question! I'll follow up there with some of the specifics that @nlemoine and @szepeviktor shared ...

nlemoine:
One last thing, in the priority list, I think passing the PHPStan analysis comes before enhancing the current standard. It will actually ease the choice of a better/stricter coding standard since many of PHPStan rules also are checked by more demanding standards (such as typed arguments, return types, etc.).

jarednova:
I'll sign on to _e mieux est l'ennemi du bien_! 

I'm happy to review #2592 asap as to unblock the great work here. I also totally agree with this:

> One last thing, in the priority list, I think passing the PHPStan analysis comes before enhancing the current standard. It will actually ease the choice of a better/stricter coding standard since many of PHPStan rules also are checked by more demanding standards (such as typed arguments, return types, etc.).

... far better to start at a place where we can give things the ✅ and then move progressively upwards through more exacting and specific standards one step at a time.

gchtr:
Very nice! I like this approach a lot 👍. I’m going to close this now so that we can continue the discussion in https://github.com/timber/timber/pull/2592.

PR #2335, opened 2020-10-07
2.x Document post serialization

gchtr:
**Ticket**: #2311

## Issue

As one of the final documentation issues in #2073, we should document JSON serialization of posts that was added in #2311.

## Solution

Take a stab at adding an example with a book post.

## Impact

Even more insights about posts and advanced Timber coding.

## Usage Changes

None.

## Considerations

@acobster I hope I got everything right!

Also, in https://github.com/timber/timber/pull/2311#issuecomment-677820224 you had an idea to implement `jsonSerialize()` in `Timber\Post`.

> > Yeah, in addition to providing some sensible default behavior in `Timber\Post` (I just wanted to avoid touching Post too much in this PR, and just focus on collections). Child implementations can easily call `parent::jsonSerialize()` to get the data to use as a basis, so we should capture that in the docs.
> 
> Right, good idea! We should definitely also have an example that explains that removing data would make sense, too, because you don’t want to fill the HTML of a page with data that has to be downloaded but that you will never need.

I’m still not sure about that, but I think it might be hard to figure out which data is added by default, because this depends sooo much on the use case. Maybe not defining a default isn’t all that bad after all?

## Testing

Nope.

### DIFF
--- a/docs/v2/guides/posts.md
+++ b/docs/v2/guides/posts.md
@@ -360,6 +360,11 @@ console.log(post);
 }
 ```
 
+Now you might think: Why do I have to add all the data manually. Could we not just add all the data from all the methods of a post? Well, technically we could. But all that data would end up in your HTML output, which might not be a good idea:
+
+- There could be sensible data that you don’t want to have publicly available in your HTML.
+- All the data you add to the HTML will make your page size bigger. For performance reasons, it makes sense to only load the data you need.
+
 ## Performance
 
 ### Consider using the `pre_get_posts` action
--- a/docs/v2/guides/posts.md
+++ b/docs/v2/guides/posts.md
@@ -306,7 +306,7 @@ Because `link` is a method of your post object, you wouldn’t have access to it
 console.log(post.link); // undefined
 ```
 
-Luckily, support for serialization is baked into Timber queries when you implements PHP’s [JsonSerializable](https://www.php.net/manual/en/class.jsonserializable.php) interface.
+Luckily, support for serialization is baked into Timber queries when you implement PHP’s [JsonSerializable](https://www.php.net/manual/en/class.jsonserializable.php) interface.
 
 Say you create a `Book` class that [extends](/docs/v2/guides/extending-timber/) `Timber\Post`. You define a `jsonSerialize()` method for that class. This method returns an array with all the data you want to use in JavaScript.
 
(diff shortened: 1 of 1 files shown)


docs/v2/guides/posts.md
```diff
@@ -283,6 +283,88 @@ If you’re used to using `get_posts()` instead of `WP_Query`, you will have to
 
 Of course, the other main difference is that instead of returning plain `WP_Post` objects, `Timber::get_posts()` returns instances of `Timber\Post`.
 
+## Serialization
+
+When you want to work with post data in JavaScript, you will want to convert it to JSON first.
+
+Under normal circumstances, this wouldn't be a problem. However, Timber posts are instantiated lazily. This means that most methods will only calculate and return a value the first time you call them. Take the `Timber\Post::link()` method for example. You use it to get the permalink for a post.
```
jarednova, line 290: Before this said: `Under normal circumstances, this would be a problem. ` — but I _think_ you meant the opposite

docs/v2/guides/posts.md
```diff
@@ -283,6 +283,88 @@ If you’re used to using `get_posts()` instead of `WP_Query`, you will have to
...
+$json = wp_json_encode( $post );
+```
+
+Because `link` is a method of your post object, you wouldn’t have access to it in JavaScript, because when you JSON-encode a post, you will only get its properties.
+
+```js
+console.log(post.link); // undefined
+```
+
+Luckily, support for serialization is baked into Timber queries when you implements PHP’s [JsonSerializable](https://www.php.net/manual/en/class.jsonserializable.php) interface.
```
jarednova, line 309:
```suggestion
Luckily, support for serialization is baked into Timber queries when you implement PHP’s [JsonSerializable](https://www.php.net/manual/en/class.jsonserializable.php) interface.
```

jarednova (changes requested):
Made a few tweaks. Then I remembered that I should do it as a review instead of just going and changing things. So one more tweak I found ("implements" vs. "implement"). Otherwise this is a super clear and helpful explanation and illustration of the problem/solution.

docs/v2/guides/posts.md
gchtr: > but I _think_ you meant the opposite

Yep 😊👍

gchtr:
@jarednova Thanks for checking! I think it’s totally fine to fix typos and grammatical issues directly. But suggestions work great as well 👍. I’ll leave it to you to merge this in.

jarednova:
👍

PR #2336, opened 2020-10-11
Name CI jobs

szepeviktor:
Just naming them :)

Looks good!

kép


jarednova (approved):
Had no idea you could do this, love it!

Issue #2333, opened 2020-10-07
Implement top-level image methods


acobster:
## Is your feature request related to a problem? Please describe.

When dealing with attachments in the new Posts API , instead of calling `new Image()` or `new Attachment()` directly, one would just call `Timber::get_post()`. This is a potential point of confusion, as in some cases it may not be obvious from looking at user code that we're actually getting an attachment:

```twig
<img src="{{ get_post( post.meta('my_img') ) }}">
```

This is arguably more intuitive:

```twig
<img src="{{ get_image( post.meta('my_img') ) }}">
{# or #}
<img src="{{ get_attachment( post.meta('my_img') ) }}">
```

## Describe the solution you’d like

Two small additions to the API:

1. Add `Timber::get_image()` and `Timber::get_attachment()` as aliases to `Timber::get_post()`
2. Add corresponding `get_image()` and `get_attachment()` Twig functions

The drawback here is that aliases are also potentially confusing, since it's possible you won't get an Attachment instance back from `::get_attachment()`. But this is probably less likely than the inverse case, since you're unlikely to use `::get_attachment()` in a situation where you don't already have an attachment ID. To cover out bases, in the docs we should say something like:

> Alias for Timber::get_post(). Honors Class Maps and query params in exactly the same way. That means it's possible to get something other than an Attachment (such a plain Timber\Post object) back if you pass something other than an attachment ID. Intended to clarify your code when you already know you have an attachment ID.

## Describe alternatives you’ve considered

Just using `::get_post()` for everything. Potentially confusing.

## Additional context

Discussion starts here: https://github.com/timber/timber/pull/2316#pullrequestreview-502502850

PR #2341, opened 2020-10-15
Update upgrade guide to reflect new image API

acobster:
Actual implementation to be done in #2333

Just some updates about the image/posts API that surfaced in #2316.


acobster:
see also: #2342

gchtr (approved):
This looks good. I fixed a small issue, because there never was a `TimberImage()` method on `Timber\Post`.

And then I realized the introductory wording for the updated Twig methods was off, too.

With all the changes coming in from the factories branch, I think it’s best I recheck the whole Upgrade Guide again, because updates were scattered in pull request against `2.x` and against `2.x-docs-api`.

I’ll do that separately.

PR #2342, opened 2020-10-15
2.x images api

acobster:
**Ticket**: #2333

## Issue

Discussed in detail in the linked issue and in #2316 .

## Solution

Aliases and docs for `::get_image()` and `::get_attachment()` and corresponding Twig functions.

## Impact

A slightly bigger but less confusing API.

## Usage Changes

MOAR METHODS

## Considerations

We threw out the idea of adding `::get_images()` and `::get_attachments()` (plural) for consistency. I don't see as strong a reason to have these in there, and they're potentially even more confusing because of the potential for getting back a mixed bag of concrete classes.

## Testing

Yes, test all the things!

### DIFF
--- a/lib/Timber.php
+++ b/lib/Timber.php
@@ -262,6 +262,34 @@ class Timber {
 	}
 
 	/**
+	 * Alias of Timber::get_post(). As such, will honor Class Maps and will NOT
+	 * necessarily always return an Attachment object. Intended for cases where you
+	 * already know you have an attachment ID (or WP_Post of type `attachment`,
+	 * etc.), to clarify your own code.
+	 *
+	 * @api
+	 * @see Timber::get_post()
+	 * @see https://timber.github.io/docs/v2/guides/class-maps/
+	 */
+	public static function get_attachment( $query = false, $options = [] ) {
+		return static::get_post( $query, $options );
+	}
+
+	/**
+	 * Alias of Timber::get_post(). As such, will honor Class Maps and will NOT
+	 * necessarily always return an Image object. Intended for cases where you
+	 * already know you have an attachment ID (or WP_Post of type `attachment`,
+	 * etc.), to clarify your own code.
+	 *
+	 * @api
+	 * @see Timber::get_post()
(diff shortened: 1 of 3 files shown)


lib/Timber.php
```diff
@@ -261,6 +261,34 @@ public static function get_post( $query = false, $options = [] ) {
...
+	 * necessarily always return an Attachment object. Intended for cases where you
+	 * already know you have an attachment ID (or WP_Post of type `attachment`,
+	 * etc.), to clarify your own code.
+	 *
+	 * @api
+	 * @see Timber::get_post()
+	 * @see https://timber.github.io/docs/v2/guides/class-maps/
+	 */
+	public static function get_attachment( $query = false, $options = [] ) {
+		return static::get_post( $query, $options );
```
gchtr, line 275: What if we did something like this to handle these cases?

```suggestion
		$post = static::get_post( $query, $options );
		
		return $post && $post instanceof \Timber\Attachment ? $post : false;
```

Wouldn’t that be the *expected* result?

Or wouldn’t that work well either, because somebody could use a class that doesn’t extend `Timber\Attachment`? Or would that be too much on an edge case and we could know that people know that they’re doing?

gchtr (commented):
I see your concerns with having these aliases and had an idea that I wanted to run by you (see suggestion).

acobster:
@gchtr I like it - nice and simple, and less surprising. @jarednova what do you think?

lib/Timber.php
jarednova: I keep going back-and-forth on this in my head. Either way, we're into edge-case land. I think the advantage is this suggestion is that many of the image/attachment methods are unavailable if the object itself _isn't_ an image/attachment. Thus, probably better to get a `false` than an object that can't be used in the expected way.

jarednova (approved):
This looks good @acobster! This is approved once you're able to resolve @gchtr's suggestion. I left my comments specifically on that proposed change

acobster:
@gchtr please see updates :)

gchtr (approved):
Looks good to me! 👍

Issue #2337, opened 2020-10-11
Gutenberg "Read More" Block option to hide excerpt is not working


Keysaw:
"Read More" Block in Gutenberg has an option to hide the excerpt when displaying full content. Toggling that option doesn't affect `{{post.content}}` at all.

This issue was supposedly resolved by https://github.com/timber/timber/pull/2256, but still does not work for me. However, changing the line https://github.com/timber/timber/blob/0a98324a5574484b4b9ccab1fec5fba7abeb20cd/lib/Post.php#L1244 to `if ( strpos($content, '"noTeaser":true') !== false ) {` (notice the double quotes around "noTeaser") fixes it.

## Expected behavior
The HTML output of `{{post.content}}` should be stripped of any tag above the `<!--more-->` tag from Gutenberg if "noTeaser" option is true.

## Actual behavior
Toggling the option in Gutenberg does affect the output.

## Steps to reproduce behavior

- Add "Read More" block in Gutenberg to separate content.
- Check the option "Hide the excerpt on the full content page".
- Print the output using `{{post.content}}` in Twig file.

## What version of WordPress, PHP and Timber are you using?
WordPress 5.5.1, PHP 7.3.20, Timber 1.18.1.

## How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Installed via Composer.

jarednova:
Thanks @Keysaw! Sounds like it's possible that there are some instances of WP where single vs. double quotes are used. Probably easiest is just to watch for both single and double quotes. Are you interested in doing a PR for this?

Keysaw:
Thanks for the reply, that seems to be the way to fix it indeed. Since I'm fairly new to GitHub I didn't want to go beyond reporting the bug, but I think I could take this opportunity to learn how to make a PR!

I will look into it very soon!

jarednova:
Excellent, glad you're up for the challenge! There are lots of guides out there, but in this case it's a pretty easy fork/edit/submit workflow. If you can get the ball started, I'm happy to add a unit test to confirm

errand:
Hello, for some reason i don't see `noTeaser":true` in the content output, but there is `<!--noteaser-->`, so why not to use it for detection?

I changed fuction to this:
```
protected function content_handle_no_teaser_block( $content ) {
		if ( strpos($content, '<!--noteaser-->') !== false ) {
			$arr = explode('<!--noteaser-->', $content);
			print_r($arr);
			return $arr[2];
		}
[...]

PR #2343, opened 2020-10-16
Adding double quotes check when using Read More block from Gutenberg

Keysaw:
**Ticket**: https://github.com/timber/timber/issues/2337

## Issue
Using `{{post.content}}` in a Post that contains a "Read More" block from Gutenberg with the option to hide the excerpt enabled sometimes does NOT hide the excerpt on full content page.

## Solution
Added an additional check for `'"noTeaser":true'` tag to cover the possibility that it contains double quotes.

## Impact
This should be 100% backwards-compatible, as it only adds a `strpos(...)` check with a `||` operator. The change will only correct the bug for instances of WordPress that use double quotes to surround `noTeaser` tag part.

## Usage Changes
None.

## Considerations
While repeating chunks of code (in this case `strpos(...)`) is not ideal, I tend to think that this is less error-prone in this case since dealing with quotes in a variable content can often lead to unexpected behaviour.

We could, however, first remove possible double quotes from the needle (`'"noTeaser":true'`), then use `strpos(...)` on it. I just wanted to keep things simple.

### DIFF
--- a/lib/Post.php
+++ b/lib/Post.php
@@ -1233,15 +1233,15 @@ class Post extends Core implements CoreInterface {
 	}
 
 	/**
-	 * Handles for an circumstance with the Block editor where a "more" block has an option to 
-	 * "Hide the excerpt on the full content page" which hides everything prior to the inserted 
+	 * Handles for an circumstance with the Block editor where a "more" block has an option to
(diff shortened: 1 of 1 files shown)


jarednova (approved):

jarednova:
Tested and confirmed @Keysaw. Thanks for your first PR!

PR #2350, opened 2020-10-19
Changed usage 'get_the_date' and 'get_the_time' filters according WP …

shvlv:
…core: third post id param is added

## Issue
Wordpress provide third param to  'get_the_date' and 'get_the_time' filters 
https://github.com/WordPress/WordPress/blob/master/wp-includes/general-template.php#L2540
https://github.com/WordPress/WordPress/blob/master/wp-includes/general-template.php#L2667
Timber has only two. And this is caused error for developer relying on core implementation.

## Solution
Add third param (post id).

## Impact
I think this is minor changes.


shvlv:
@jarednova thanks, I try to change outdated master.

gchtr:
@shvlv Thanks for taking the time to figure out the problem and providing a pull request with a fix! 👍

Just a note: This change is already included in `2.x`. I added it in #2150. But it definitely makes sense to backport it to `1.x` as well.

The following filters also changed their signature, so it might be worth to change them in this pull request as well:

- `get_the_modified_date`
- `get_the_modified_time`

jarednova (approved):

jarednova:
Looks good! Thanks for the spot @gchtr that the `get_the_modified_*` permutations have a diff't signature that requires a `WP_Post`:

We accept post IDs or `WP_Post`:
https://developer.wordpress.org/reference/hooks/get_the_time/
https://developer.wordpress.org/reference/hooks/get_the_date/

We only accept `WP_Post`:
https://developer.wordpress.org/reference/hooks/get_the_modified_time/
https://developer.wordpress.org/reference/hooks/get_the_modified_date/

Because WordPress.

Issue #2348, opened 2020-10-19
The content splitted by Read more block isn't rendered correct


jhhazelaar:
After the content is split by the option 'Hide the excerpt on the full content page' the content isn't rendered correct, because there is still a close tag of the more block in the content. The content return starts with the `<!-- /wp:more -->` and the rest of content isn't processed by gutenberg.

## What version of WordPress, PHP and Timber are you using?
WP 5.4, PHP 7.4, Timber dev-master

## How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Composer

jhhazelaar:
Related ticker: #2218 and #2337

gchtr:
I guess this can be closed.

PR #2351, opened 2020-10-20
#2348 Split content on wp:more tag instead of noteaser

jhhazelaar:
If content is split on noteaser the wp:more closing tag is still in the content. After the split the content isn't rendered correct.

**Ticket**: #2348


jarednova:
Thanks @jhhazelaar! This along with #2343 fixes up some stuff in teases that should make that 💯

Issue #2356, opened 2020-10-27
file_system_to_url() returns wrong url in Bedrock


Levdbas:
file_system_to_url() returns wrong url on specific WP installations, for example a bedrock based WP install where site_url returns domain.com/wp/ and the content lives in domain.com/app/

## Expected behavior
file_system_to_url() will return proper home url instead of site url.

## Actual behavior
Returns site url. When changing site_url in this function to home_url the issue is resolved.

## Steps to reproduce behavior
This breaks for example a bedrock based installation.  When an image has to be sideloader bij ImageHelper the image it tries to download resolves to http://yoururl.domain/wp/app/uploads/.... instead of http://yoururl.domain/app/uploads/.

## What version of WordPress, PHP and Timber are you using?
Timber 1.18.2
WordPress 5.5.1

## How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
composer package

jarednova:
Thanks so much for the issue and identifying the problem point! We've tested against Bedrock setups but there are some corners (like this) that we might have missed. Since you found the issue, I want to give you first crack at the PR opportunity

PR #2357, opened 2020-10-29
fix two cases where site_url() was used instead of home_url()

Levdbas:
**Ticket**: #2356 

## Issue
site_url() returns a wrong url on bedrock based installations, since site_url() returns https://yourdomain.com/wp since the WP installation lives in this directory.

Relevant file structure of a Bedrock based project:

├── Other stuff, no relevant for this PR.
└── web                        → Web root (document root on your webserver)
    ├── app                    → **wp-content equivalent**
    │   ├── mu-plugins   → Must use plugins
    │   ├── plugins         → Plugins
    │   ├── themes         → Themes
    │   └── uploads         → Uploads
    ├── wp-config.php    → Required by WP (never edit)
    ├── index.php           → WordPress view bootstrapper
    └── wp                     → **WordPress core**

## Solution
When using home_url() this issue is resolved.

## Impact
In my initial issue I found out that changing `site_url()` to `home_url()` in  `Timber\URLHelper::file_system_to_url()` fixed a sideloading img issue. This change is inside PR. While on it I found a second instance of `site_url` in `Timber\ImageHelper::_get_file_url`. to get a relative URL we need to remove `home_url()` from the $url instead of `site_url()` since uploads in a Bedrock based installation live in `https://yourdomain.com/app/uploads/` and not in `https://yourdomain.com/wp/app/uploads/`

## Considerations
It might be a good idea to write an extra set of tests based off a Bedrock based installation because I see the combination of Bedrock + Timber a lot in my freelance work.

### DIFF
--- a/lib/ImageHelper.php
+++ b/lib/ImageHelper.php
@@ -502,7 +502,9 @@ class ImageHelper {
 		}
 		$url = untrailingslashit($url).'/'.$filename;
 		if ( !$absolute ) {
-			$url = str_replace(home_url(), '', $url);
+			$home = home_url();
+			$home = apply_filters('timber/ImageHelper/_get_file_url/home_url', $home);
+			$url = str_replace($home, '', $url);
 		}
 		return $url;
 	}
--- a/lib/ImageHelper.php
+++ b/lib/ImageHelper.php
@@ -502,7 +502,7 @@ class ImageHelper {
 		}
 		$url = untrailingslashit($url).'/'.$filename;
 		if ( !$absolute ) {
-			$url = str_replace(site_url(), '', $url);
+			$url = str_replace(home_url(), '', $url);
 		}
 		return $url;
 	}
--- a/lib/Integrations/WPML.php
+++ b/lib/Integrations/WPML.php
@@ -8,6 +8,7 @@ class WPML {
 		add_filter('timber/URLHelper/file_system_to_url', array($this, 'file_system_to_url'), 10, 1);
 		add_filter('timber/URLHelper/get_content_subdir/home_url', array($this, 'file_system_to_url'), 10, 1);
 		add_filter('timber/URLHelper/url_to_file_system/path', array($this, 'file_system_to_url'), 10, 1);
+		add_filter('timber/ImageHelper/_get_file_url/home_url', array($this, 'file_system_to_url'), 10, 1);
 
 	}
 
--- a/lib/URLHelper.php
+++ b/lib/URLHelper.php
@@ -168,7 +168,7 @@ class URLHelper {
 	 */
 	public static function file_system_to_url( $fs ) {
 		$relative_path = self::get_rel_path($fs);
-		$home = site_url('/'.$relative_path);
+		$home = home_url('/'.$relative_path);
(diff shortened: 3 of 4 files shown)


jarednova:
Thanks @Levdbas ! looks like there's just one failing test we need to track down ...

```php
1) TestTimberURLHelper::testFileSystemToURLWithWPML
Failed asserting that 'http://example.org/wp-content/uploads/2020/11/arch.jpg' ends with "://example2.org/wp-content/uploads/2020/11/arch.jpg".
/home/travis/build/timber/timber/tests/test-timber-url-helper.php:62
```

Levdbas:
Hi @jarednova , I am kinda at a stale here... Tried for many hours but can't find a proper solution. 

Problem is as follows:

home_url() works in default and bedrock setups **without** WPML. 
site_url() works in default WP **with** WPML but not in Bedrock, because of the issues i started this issue anyway.
home_url() **does not** work **with** WPML, regardless of the setup.

WPML seem to hijack the home_url() function and add the home url for the current language.

So using home_url() is not an option because of WPML, and site_url() is not an option because of Bedrock. I am looking for a way to just get the base home url without the language and without any subdir where WP is living. And I cant find one. :(

Any ideas how we can work around this issue?

jarednova:
Oy. Always something, huh @Levdbas?

My thought is two steps ...

1. Because using `home_url()` seems like the correct "base" behavior the Timber code should be updated to begin there.
2. Okay, so now what about WPML? Like you said, it's doing funny stuff so let's isolate that into its own integrations file (like we have for ACF and a few others) that can hook into `ImageHelper` and `URLHelper` where needed. No doubt it might involve some URL acrobatics, but at least it will only be in play for people who are using a combo of those two plug-ins in tandem (not nobody, but certainly a vast minority of the user base)
3. If we get stuck, we've received help in the past from the WPML team to figure out what special sauce is needed to play nice (also seems like [an issue others have encountered/resolved](https://github.com/roots/bedrock/issues/30))

Levdbas:
Hi @jarednova , it was the test after all 🥇 . Updated the test case in the first commit.
in f38aa2f I've also added the specific WPML filter so ImageHelper keeps on working with WPML. Could you take a look if i applied the filters corretly this way with the right naming scheme?

jarednova:
Holy moly! Great stuff @Levdbas. I'll take a closer look this week to make sure the filter application is correct and keeping with naming, etc.

jarednova (approved):
The only thing worth considering is whether the preceding underscore in the filter name is weird. It is. But it matches with the format we use elsewhere, and not worth creating a one-off "rule" that omits it here.

Maybe in the future we'll clean up the function names in `ImageHelper` and a few other locations to remove those underscored functions. But let's not distract us from making this fix today.

gchtr:
> Maybe in the future we'll clean up the function names in `ImageHelper` and a few other locations to remove those underscored functions. But let's not distract us from making this fix today.

@jarednova As soon as we tackle coding standards, the underscores prefixes will go away.

jarednova:
👍  sounds good! much better to take those things on as a whole vs. ad-hoc bits here-and-there

PR #2358, opened 2020-10-29
support passing extra args to Timber::context()

acobster:
Here's something I've had kicking around in my head for a while. :)

In [Conifer](coniferplug.in/), which we use on pretty much every WP site at SiteCrafting, we have the nice convenience of being able to pass extra data to `Conifer\Site::context()`, which delegates to `Timber::context()` and just adds this merging capability on top. It means that this code:

```php
$data = Timber::context();
$data['extra'] = 'stuff';
```

Becomes:

```php
$data = $site->context([ 'extra' => 'stuff' ]);
```

I usually err on the side of not expanding an API if you can help it, but this is extremely convenient and I've found it makes template code way more readable, especially when you have a lot of extra stuff you need to get into your view data from your template. It turns a bunch of separate, imperative operations and turns it into one data _composition_ operation, which I think is closer to how folks tend to think about view data. So I think the extra readability is worth a tiny incremental expansion to the API. :)

This process is immutable, i.e. it doesn't affect the underlying context data in any way. So subsequent calls to `::context()` without extra data will return whatever it did before.

The proposal here is to simply push this behavior down into Timber, so we can do:

```php
$data = Timber::context([ 'extra' => 'stuff' ]);
```

### DIFF
--- a/lib/Timber.php
+++ b/lib/Timber.php
@@ -901,10 +901,13 @@ class Timber {
 	 * @api
 	 * @since 2.0.0
 	 *
+	 * @param array $extra any extra data to merge in. Overrides whatever is
+	 * already there for this call only. In other words, the underlying context
+	 * data is immutable and unaffected by passing this param.
 	 * @return array An array of context variables that is used to pass into Twig templates through
 	 *               a render or compile function.
 	 */
-	public static function context() {
+	public static function context(array $extra = []) {
 		$context = self::context_global();
 
 		if ( is_singular() ) {
@@ -913,7 +916,7 @@ class Timber {
 			$context['posts'] = Timber::get_posts();
 		}
 
- 		return $context;
+		return array_merge( $context, $extra );
 	}
 
 	/**
--- a/tests/test-timber.php
+++ b/tests/test-timber.php
@@ -258,6 +258,30 @@ class TestTimberMainClass extends Timber_UnitTestCase {
 		$this->assertInstanceOf( Timber\PostQuery::class, $context['posts'] );
 	}
 
+	function testContextWithExtraArgs() {
+		$pids = $this->factory->post->create_many(20);
+		$this->go_to('/');
+
+		$context = Timber::context([
+			'extra'  => 'stuff',
+			'fancy'  => [
+				'this' => 'can',
+				'be'   => 'whatever',
+			],
+		]);
+
+		$this->assertEquals( 'stuff', $context['extra'] );
+		$this->assertEquals( [
+			'this' => 'can',
+			'be'   => 'whatever',
+		],	$context['fancy'] );
+		$this->assertInstanceOf( Timber\PostQuery::class, $context['posts'] );
+
+		// Underlying context is immutable and unaffected by extra data.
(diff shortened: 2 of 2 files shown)


jarednova:
I think this works. Glad that it's based on a real use case versus "it would be cool if ...". Yes, it expands the API but it does so in a way consistent with how we've approached arguments and options elsewhere. @gchtr, what's your take?

gchtr (approved):
I like it! 😎

It’s another step towards even shorter PHP templates.

The only thing I’m asking myself is where in the documentation we’ll want to use this for examples.

Should we treat this as advanced usage, or as the default usage?
Should `Timber::context()` be used at the start of the template, or towards the end? Is this question too subjective?

Consider this (from the [Context Guide](https://github.com/timber/timber/blob/2.x/docs/v2/guides/context.md#archive-templates)):

```php
$context = Timber::context();

$context['posts'] = new Timber::get_posts( array(
    'query' => array(
        'post_type'      => 'book',
        'posts_per_page' => -1,
        'post_status'    => 'publish',
    ),
) );

Timber::render( 'archive.twig', $context );
```

Should we change it to this?

```php
$context = [
	'posts' => new Timber::get_posts( array(
	    'query' => array(
	        'post_type'      => 'book',
	        'posts_per_page' => -1,
	        'post_status'    => 'publish',
	    ),
	],
) );

Timber::render( 'archive.twig', Timber::context( $context ) );
```

acobster:
I'd propose:

```php
$context = Timber::context( [
	'posts' => new Timber::get_posts( [
	    'query' => [
	        'post_type'      => 'book',
	        'posts_per_page' => -1,
	        'post_status'    => 'publish',
	    ],
	],
] );

Timber::render( 'archive.twig', $context );
```

This way, `$context` refers to the entire context that gets passed to the view, rather than some intermediate/extra state, consistent with the mental model.

acobster:
> Should we treat this as advanced usage, or as the default usage?

This is pretty much the default in our usage of Conifer and I think it works well. I'm sure there are examples all over the docs, though, and I don't think it'd be too bad to circle back to those later.

> Should Timber::context() be used at the start of the template, or towards the end? Is this question too subjective?

In my experience, there isn't much that typically needs to happen after you fetch the context, except in the most advanced use-cases. So at the end is usually cleaner because it fits the mental model of "okay, we need to do some business logic and **then** gather all the info and send it to the view."

BTW - note that there's no more `query` array in the new Posts API; `::get_posts()` takes the query args directly. I've updated my previous comment accordingly.

gchtr:
Sounds good! I think I’ll update the usage examples.

> BTW - note that there's no more `query` array in the new Posts API; `::get_posts()` takes the query args directly. I've updated my previous comment accordingly.

Yeah, but it’s already updated with #2073. I constantly have to think about what’s already in #2073, so I think I should:

1. Resolve my outstanding documentation issues in #2073.
2. Merge #2073 into `2.x`.
3. Make sure that this pull request has the proper documentation and then merge it in.

acobster:
@gchtr cool, sounds like a plan. Thanks!

In the meantime, I'll go ahead and merge this.

Issue #1345, opened 2017-03-07
get_preview outputs "Read more"-link even if there aren't any more text to read


stianmartinsen:
When I'm using Timber's `get_preview`-function the "Read more"-link is always outputted – even if the post's text isn't longer than the preview.

I expected that there would at least be an option to stop this, but I couldn't find one. Am I missing something? Because why would a person want to click a link just to read the same text as he's already read?

certainlyakey:
Yes, there's an option to stop this — set the third argument as `null`. Had the same problem but took a look at the source code.

stianmartinsen:
[The third argument](https://github.com/timber/timber/blob/master/lib/Post.php#L390) is the text for the read more link. I do want read more links if the text is longer than the preview, so I don't think setting this to null is a good solution..?

certainlyakey:
you're right. i had somewhat different problem, tried to remove it whatsoever.

Djules:
The method `get_preview` is now deprecated, and using `{{ post.preview }}` doesn't allow any arguments, so this issue still exists.

If the user manually set an excerpt, it should not add the "Read more" link, no matter what, as the original WordPress function does…

tameroski:
#1015 Should fix this but it doesn't seam to work ... Or i'm missing something.

tameroski:
`{{post.preview}}` doesn't allow arguments but has modifiers like `{{post.preview.length(20).force.read_more("Keep Reading")}}`. See #1015 for details. Doesn't fix the issue though but may be usefull.

acobster:
Tagged this with 2.0 in case it's still an issue that would require a breaking change to fix...

gchtr:
Okay, I looked at it more closely and we need to do some changes to support this. We already have a `$trimmed` variable in the code that we can make use of. I already have some code ready locally. But I have several questions we should discuss first.

### 1. Should it be a global option or another argument for `Post::excerpt()`?

For a global option, we could use a filter. But since we already have a couple of parameter, using another parameter wouldn’t be that bad. Here’s how it could look like. (Be aware that this is how it would work in Timber 2.0).

```php
$excerpt = $post->excerpt( [
	'maybe_skip_read_more' => true,
] );
```

or in Twig

```twig
{{ post.excerpt({ maybe_skip_read_more: true }) }}
```

### 2. Should the option be enabled by default?

When we would enable it by default, it would be a breaking change.

> Because why would a person want to click a link just to read the same text as he's already read?

[...]

PR #2365, opened 2020-11-05
2.x Clean up post preview

gchtr:
## Issue

Functions that were deprecated in 1.x can be removed in 2.x. `Post::get_preview()` is one of these candidates.

## Solution

- Delete the method and update tests.
- Add an `@api` tag to the `PostPreview` class so that it appears in the documentation, because currently, it’s missing.

## Impact

Breaking change.

## Usage Changes

`Post::preview()` or `new PostPreview()` should be used.

## Considerations

When using `$preview = new PostPreview()` in PHP, you’ll have to do `(string) $preview` for it to be converted to string (so that `__toString()` kicks in), because `PostPreview::run()` is protected. I don’t like this API very much. Are there better solutions?

- `$preview->to_string()`?
- `$preview->get()`?
- others?

## Testing

There’s a failing test because when we use an empty string as the value for the `readmore` option, a Readmore link will still be added. I’m not sure whether this behavior was introduced only lately, because the tests tell us that an empty string should probably be evaluated as `false`, resulting in no link being added. @jarednova I would be glad if you could have a look.

Oh, and maybe ignore whitespace when looking at the changes, because I fixed a wrong indentation in the tests, too.

### DIFF
--- a/docs/v2/upgrade-guides/2.0.md
+++ b/docs/v2/upgrade-guides/2.0.md
@@ -577,6 +577,18 @@ $genre->posts( [
 ] );
 ```
 
+## Post Excerpts
+
+### "Excerpt" instead of "Preview"
+
+We renamed the `Timber\PostPreview` class to `Timber\PostPreview` and also changed all instances of the term "preview" to "excerpt". The reason for this change is that the term "preview" in WordPress is mainly used for previewing a post before you save changes in the admin. For the short texts that are used in post teasers, the term "excerpt" is used. We wanted to follow the WordPress terminology here.
+
+For this, we…
+
+- renamed the `Timber\PostPreview` class to `Timber\PostExcerpt`.
+- added a `Timber\Post::excerpt()` function that you should use instead of `Timber\Post::preview()`.
+- Changed filter names.
+
 ### `post.excerpt` takes arguments in array/hash notation
 
 The `{{ post.excerpt }}` function was added as a replacement for `{{ post.preview }}`. When using `{{ post.preview }}`, you could pass in arguments by chaining them. It’s now possible to pass in arguments as an array in PHP or in hash style in Twig. This finally cleans up how the previews of posts are handled. Here’s an example:
@@ -633,6 +645,7 @@ You should update the following hooks, because they will be removed in a future 
 **Timber\PostPreview**
 
 - `timber/post/preview/read_more_class`, use `timber/post/excerpt/read_more_class` instead
+- `timber/post/get_preview/read_more_link`, use `timber/post/excerpt/read_more_link` instead
 
 **Timber\Term**
 
--- a/docs/v2/upgrade-guides/2.0.md
+++ b/docs/v2/upgrade-guides/2.0.md
@@ -630,6 +630,10 @@ You should update the following hooks, because they will be removed in a future 
 - `timber_post_get_meta_field`, use `timber/post/meta` instead
 - `timber_post_get_meta`, use `timber/post/meta` instead
 
+**Timber\PostPreview**
+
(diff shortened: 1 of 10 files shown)


jarednova:
@gchtr fixed the test! Just a problem with `readmore` vs. `read_more` as the name for that argument. Two follow along questions:

- Should we change `PostPreview::$readmore` to `PostPreview::$read_more` so that's universal?
- Should we make `readmore` an alias of `read_more`?

As for your question, when I use that value w/o the `(string)` it still works because of `__toString()`. Checkout e86e941 for how I modified the test to see this in action. Is that what you're referring to? or a different use-case?

gchtr:
@jarednova 

> Just a problem with `readmore` vs. `read_more` as the name for that argument

😅 Phuu, glad it’s just that.

>     * Should we change `PostPreview::$readmore` to `PostPreview::$read_more` so that's universal?

Yes, please!

>     * Should we make `readmore` an alias of `read_more`?

Not necessarily. I probably used `readmore` instead of `read_more` because the deprecated `Timber\Post::get_preview()` used `readmore` as a parameter name. I had a window open with that function as a reference for the parameter order so I could change the tests.

I think this shouldn’t be a problem when we only use `read_more` in the code, in the inline documentation as well as the documentation.

> As for your question, when I use that value w/o the `(string)` it still works because of `__toString()`. Checkout [16450b0](https://github.com/timber/timber/commit/16450b0b270fa9ebfa117f36edb1c346ce4bfca2) for how I modified the test to see this in action. Is that what you're referring to? or a different use-case?

Okay, in that case I misunderstood it.

Do you want to make the final over this with fixing `readmore` → `read_more`, or do you want me to do it?

jarednova:
> Do you want to make the final over this with fixing readmore → read_more, or do you want me to do it?

Go ahead and take it home — then LTGM!

gchtr:
Cool, I’ll see what I can do.

From #1817, there are two more task:

- Rename `PostPreview` to `PostExcerpt` (in https://github.com/timber/timber/issues/1817)
- Maybe add tests for #1345 and fix it (in https://github.com/timber/timber/issues/1817#issuecomment-470706883)

gchtr:
Okay, I made progress here.

- I changed all instances of `readmore` to `read_more`.
- I renamed all relevant instances of "Preview" to "Excerpt".
- I optimized the code a bit to make it more readable, like [here](https://github.com/timber/timber/pull/2365/commits/f5afa7432db9521badc9813a372b2f013d4e5427) and [here](https://github.com/timber/timber/pull/2365/commits/a0b69f121f89552587ff94b3c8953b3530130eb2).
- Deprecated the `timber/post/preview/read_more_class` filter and added `timber/post/excerpt/read_more_class` instead in [`4c664cc` (#2365)](https://github.com/timber/timber/pull/2365/commits/4c664cc26813752edfbcda6e827b0d872e012a40).
- I found that the `timber/post/get_preview/read_more_link` filter that was removed in https://github.com/timber/timber/commit/a01086582ad0315ca1c90ec02d08a506f4f95d0e, so I added it back, deprecated it and added `timber/post/excerpt/read_more_link` instead in [`d36989b` (#2365)](https://github.com/timber/timber/pull/2365/commits/d36989b12cfbfdef9615dde47aafa94196cc449e).

And then there’s an additional pull request in #2375.

@jarednova I’d be glad if you could have a look at this.

jarednova (approved):
I found just a few tiny tweaks in 2a638bb. Will merge once those tests clear!

jarednova:
@gchtr all green over here! I'll let you do the honors ...

gchtr:
🙌

jarednova:
Niiiiiiiiiiiiiiiiiice

Issue #2323, opened 2020-09-09
THE PURGE


acobster:
Once #2316 and #2178 are wrapped up, delete any old code and tests we don't need anymore.

- [x] #2316 
- [x] #2178 

Potential purge candidates:

- [x] `PostGetter` & tests
- [x] `TermGetter` & tests
- [x] Any old `$CommentClass` props and the like
- [x] `QueryIterator` & tests
- [x] `Post::determine_id()` and possibly other methods
- [ ] `Timber_WP_CLI_Command.php` (or move to its own repo)

sdunham:
![THE PURGE](https://media.giphy.com/media/7ZpJgAQAyCKVa/giphy.gif)

jarednova:
🔪

jarednova:
Closing! A follow-up @todo is possibly moving Timber_WP_CLI_Command into its own repo. I want to check to see if there's a current best practice for what those files/repos/packages look like. But also: is it ever used? @szepeviktor I see your a contributor for WP-CLI. Have you ever used the Timber command?

szepeviktor:
> moving Timber_WP_CLI_Command into its own repo

+1

The _Keyboard User_ is a rare bird 🐦

szepeviktor:
> Have you ever used the Timber command?

I'm not a developer! :) and I haven't been is a Timber-based project up to now.
I just like to contribute here ❤️ 

BTW Monorepos are getting popular nowadays. We could have two (or more) Composer packages in one repository.

acobster:
> I just like to contribute here :green_heart:

Spotted: _Contributorus alrtuis_, one of the rarest birds of all! :)

Personally I don't really see a reason to separate `Timber_WP_CLI_Command` out to its own repo, and I like it when plugins implement their own WP-CLI commands. Having it in a separate repo seems like a barrier to entry for not much gain (and IMO Timber's barrier to entry from vanilla WP is already relatively high). Coordinating separate repos is also a lot more maintenance work. You need a good reason.

> We could have two (or more) Composer packages in one repository.

There are some cases where this is advantageous but even this presents a higher maintenance cost and I just don't see a compelling reason to do this here.

All that said, I have literally never run a `wp timber` command (except just now, to kick the tires) and I highly doubt any of my co-workers have either. I know we don't have reliable usage numbers, but do we have a channel for broadcasting out a survey or something, to get some kind of picture of what our usage looks like, and what users want?

szepeviktor:
[...]

PR #2366, opened 2020-11-05
2.x Remove functionality deprecated in 1.x

gchtr:
## Issue

Functionality that was deprecated in 1.x can be removed in 2.x. And I see that not every was catched by THE PURGE in #2323.

## Solution

Search for `@todo` tags that tell us to remove 1.x functionality and actually remove it.

## Impact

Breaking Changes.

## Usage Changes

Were already covered in 1.x.

## Considerations

Any?

## Testing

Maybe we still need to update some tests. Let’s see.

### DIFF
--- a/lib/Helper.php
+++ b/lib/Helper.php
@@ -625,33 +625,6 @@ class Helper {
 
 	/**
 	 * Filters a list of objects, based on a set of key => value arguments.
-	 * Uses native Twig Filter.
-	 *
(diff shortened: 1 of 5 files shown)


jarednova (approved):
Looks good! I just removed those last tests which were covering the old `filter` behavior style

gchtr:
Well of course, the tests! 👍 Thanks for looking at this.

Issue #1908, opened 2019-01-15
Multisite Timber::get_sites() in admin with same locale


jhhazelaar:
We have a WP multisite installation with Multilingualpress plugin for multi language support. We now want to extend a custom plugin that gets all sites and add some ACF fields per site with the locale as part of the key.

### Expected behavior
`Timber::get_sites()` get two site. One with 'en-US' en second with 'nl-NL'. In the theme we get this result. 

### Actual behavior
In the admin we get two site with different ids with the same locale 'en-US'

### What version of WordPress, PHP and Timber are you using?
Wordpress: 5.0.3, PHP: 7.2.13, Timber: 1.8.4

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Composer/Packages

highbelt:
Experienced the same issue, as 

https://github.com/timber/timber/blob/53cfc653d21c5d9a4b819040e9493c6699449154/lib/Site.php#L178

seems to return the initial site language (ignores `switch_to_blog(...)`).

Maybe use `$this->language = get_locale();`instead, as it seems to return the right value.

WP: 5.5.3, Timber 1.18.2 & Multilingualpress 3.5.0

gchtr:
Using `get_locale()` sounds like a good approach to this! When using `get_bloginfo( 'language' )`, then `determine_locale()` is used, which considers the language set in the user profile. This could lead to all the sites returning the same language when a user overwrites it in their profile on all sites. Using `get_locale()` is way more predictable, because it always uses the language setting used for a site.

@highbelt Do you feel comfortable enough to create a pull request for this?

gchtr:
That was a quick! With #2369 already being merged, we can close this one. The solution will be included in the next release.

PR #2369, opened 2020-11-11
Multisite Timber::get_sites() returns same locale for all sites (#1908)

highbelt:
**Ticket**: #1908 

## Issue
In a multisite environment `Timber::get_sites()` returns the same locale for all sites in backend (and frontend also, if user is logged in). This also happens while using `new Timber\Site()`.
When using `get_bloginfo( 'language' )`, then `determine_locale()` is used, which considers the language set in the user profile. This could lead to all the sites returning the same language when a user overwrites it in their profile on all sites. 

## Solution
Using `get_locale()` is way more predictable, because it always uses the language setting used for a site.

## Impact
No problems expected.

## Usage Changes
Use `Timber::get_sites()` and new Timber\Site() as before.

### DIFF
--- a/lib/Site.php
+++ b/lib/Site.php
@@ -175,7 +175,7 @@ class Site extends Core implements CoreInterface {
 		$this->rss = get_bloginfo('rss_url');
 		$this->rss2 = get_bloginfo('rss2_url');
 		$this->atom = get_bloginfo('atom_url');
-		$this->language = get_bloginfo('language');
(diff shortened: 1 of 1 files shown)


jarednova (approved):

jarednova:
Very interesting, thanks @highbelt for the background info here and in #1908. This looks good to me; congrats on your first contribution!

PR #2370, opened 2020-11-12
Make number_format support WordPress locale

nlemoine:
Hello Timber team!

## Issue

You took care of `date` filter but `number_format` seemed forgotten.

## Solution

Grabs WordPress locale settings (`decimal_point`, `thousands_sep`) and set [Twig number format](https://twig.symfony.com/doc/2.x/filters/number_format.html).

## Impact

No locale was set before this. So Twig applied the default PHP locale. If someone set custom `number_format` on the filter level, it will still work as exptected.

## Usage Changes

Not that I'm aware of. This setting didn't exists before and is automatic now because it depends on `$wp_locale`. 

## Considerations

Use PHP Locale would be great.

## Testing

Yes, taken from Twig's tests and adapted.

### DIFF
--- a/lib/Twig.php
+++ b/lib/Twig.php
@@ -249,7 +249,7 @@ class Twig {
 	/**
 	 * Overwrite Twig defaults.
 	 *
-	 * Makes Twig compatible with how WordPress handles dates, timezones and perhaps other items in
+	 * Makes Twig compatible with how WordPress handles dates, timezones, numbers and perhaps other items in
 	 * the future
 	 *
 	 * @since 2.0.0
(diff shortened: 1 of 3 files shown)


jarednova:
Thanks so much @nlemoine for the PR (with tests!!! :tada:!!!). While this looks good to me, I'd love a second set of eyes from @gchtr or @palmiak who might have more number formatting/localization experience.

jarednova (approved):

palmiak:
@nlemoine - it looks great to me too. Great job. Also I learned that there are many interesting stuff in $wp_locale

palmiak (approved):

PR #2374, opened 2020-11-19
Bugfix/wp cli

acobster:
WP-CLI is currently broken because `Timber_WP_CLI_Command` was removed in 354f0a9c but `Timber\Integrations` is still trying to add it as a command. This simply removes the registration code too.


jarednova (approved):
Yep, good catch!!

PR #2375, opened 2020-11-19
2.x Add options to better control read_more and end behavior in post excerpts

gchtr:
**Ticket**: #2365, #1345

## Issue

When using Timber’s excerpt functionality, the read more link is always added – even if the post’s content isn’t longer than the excerpt.

## Solution

First, there are two new options for post excerpts that can both be passed to the `Post::excerpt()` function as well as when using `new PostExcerpt()`:

- `always_add_read_more` – Controls whether a read more link should be added even if the excerpt isn’t trimmed (when the excerpt isn’t shorter than the post’s content). The default is `true` in order to not break backwards compatibility.
- `always_add_end` – Whether the end string should be added even if the excerpt isn’t trimmed. The default is `false`, because to me, that was a bug before.

More thoughts about why I chose these defaults can be found in https://github.com/timber/timber/issues/1345#issuecomment-726316686. We can still discuss this further in here.

And then, there’s a new `timber/post/excerpt/defaults` filter that can be used to update default options, including `always_add_read_more` and `always_add_end`.

It can also be used to change the defaults for other things.

```php
add_filter( 'timber/post/excerpt/defaults', function( $defaults ) {
    // Only add a read more link if the post content isn’t longer than the excerpt.
    $defaults['always_add_read_more'] = false;

    // Set a default character limit.
    $defaults['words'] = 240;

    return $defaults;
} );
```

I chose `always_add_*` instead of `always_show_*` (discussion in https://github.com/timber/timber/issues/1345#issuecomment-726794645), because the "showing" is very close to echoing. But a post excerpt is always "returned". So I think "add" would be cleaner when the meaning should be that a string is "assembled and returned".

## Impact

More versatility for post excerpts through the two new options and the new `timber/post/excerpt/defaults` filter.

## Usage Changes

- When the excerpt isn’t trimmed, then the end string is not added by default.

## Considerations

- The new options cannot be updated through chaining. Is that bad?
- When the `end` string is added, a read more links is always added to, no matter that the `always_add_read_more` option is set to. Does that make sense?

## Testing

Yes, I added tests. But do they cover all cases we need to cover?

## Todo

- [x] Update Upgrade Guide.

### DIFF
--- a/docs/v2/upgrade-guides/2.0.md
+++ b/docs/v2/upgrade-guides/2.0.md
@@ -589,6 +589,31 @@ For this, we…
 - added a `Timber\Post::excerpt()` function that you should use instead of `Timber\Post::preview()`.
 - Changed filter names.
 
+### Read more links and end strings will not always be added
+
+We updated the logic for when read more links and end strings (`…` by default) will be added. For example, if the excerpt is generated from the post’s content, but the post’s content isn’t longer than the excerpt, then no read more link and no end string will be added.
+
+You can control this behavior with these two new parameters for excerpts:
+
+- `always_add_read_more` – Controls whether a read more link should be added even if the excerpt isn’t trimmed (when the excerpt isn’t shorter than the post’s content). The default is `false`.
+- `always_add_end` – Whether the end string should be added even if the excerpt isn’t trimmed. The default is `false`.
+
+### Filter the defaults
+
+There’s a new `timber/post/excerpt/defaults` filter that can be used to update default options for excerpts, including `always_add_read_more` and `always_add_end`.
+
+```php
+add_filter( 'timber/post/excerpt/defaults', function( $defaults ) {
+    // Only add a read more link if the post content isn’t longer than the excerpt.
+    $defaults['always_add_read_more'] = false;
+
+    // Set a default character limit.
+    $defaults['words'] = 240;
+
+    return $defaults;
+} );
+```
+
 ### `post.excerpt` takes arguments in array/hash notation
 
 The `{{ post.excerpt }}` function was added as a replacement for `{{ post.preview }}`. When using `{{ post.preview }}`, you could pass in arguments by chaining them. It’s now possible to pass in arguments as an array in PHP or in hash style in Twig. This finally cleans up how the previews of posts are handled. Here’s an example:
--- a/lib/PostExcerpt.php
+++ b/lib/PostExcerpt.php
@@ -16,6 +16,10 @@ namespace Timber;
  * - have an ellipsis (…) as the end of the text.
  * - have a "Read More" link appended.
  *
+ * You can change the defaults that are used for excerpts through the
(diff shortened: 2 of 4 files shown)


jarednova:
@gchtr the code all looks golden and I totally agree with `always_add` vs. `always_show` I think that nails it on clarity and exactness (and is one character shorter!) In terms of those Qs ...

**The new options cannot be updated through chaining. Is that bad?**
Nope. RIP chaining

**When the end string is added, a read more links is always added to, no matter that the always_add_read_more option is set to. Does that make sense**
I understand the aim, but it feels a bit to Clippy to me ("it looks like you're trying to truncate an excerpt"). If I set the "read more" to be `false`, I feel like that should be respected even if there is a Very Good Reason why it shouldn't be so. Maybe there's a circumstance I'm not thinking of though.

... which brings up one last point ...

I know you campaigned in #1345 to set `$always_add_read_more = true;`. But in looking at the default WordPress behavior for `the_except();` it actually doesn't include those on very short posts. Check this example from the `twentyninteen` theme:
Screen Shot 2020-11-19 at 8 47 05 PM

My feeling here is that we should stick with default WP behavior unless there's a really good reason to deviate. This would be the right time to "fix" a 1.x quirk and get it back into WP conformity (also, it should be noted that we're in edge case land and it's NBD)
5018904-clippy-black-tar-heroin-memes-png-image-transparent-png-free-clippy-transparent-820_502

gchtr:
> I understand the aim, but it feels a bit to Clippy to me ("it looks like you're trying to truncate an excerpt"). If I set the "read more" to be `false`, I feel like that should be respected even if there is a Very Good Reason why it shouldn't be so. Maybe there's a circumstance I'm not thinking of though.
> 
> ... which brings up one last point ...
> 
> I know you campaigned in #1345 to set `$always_add_read_more = true;`. But in looking at the default WordPress behavior for `the_except();` it actually doesn't include those on very short posts.

I see, I see. Your argumentation makes total sense. I changed the default. And I moved the logic around a bit. And I updated the Upgrade Guide and some documentation. (Hell, I found that you can use custom text inside `<!-- more -->` tags).

Let’s see how the tests run, then this is ready for another review, @jarednova.

gchtr:
@jarednova Oh, and I again had to add a couple of typecasting `(string)` for excerpt objects in the tests. The objects didn’t convert themselves to strings.

jarednova:
Who would have thought something as simple as post ~previews~, er, excepts could be so complicated! I think this looks golden. Sorry for my delay over the last few weeks!

jarednova (approved):

gchtr:
> Who would have thought something as simple as post ~previews~, er, excepts could be so complicated! I think this looks golden. Sorry for my delay over the last few weeks!

Yeah, right, I didn’t think so either 😆. No worries, it was a busy end of the year for myself as well.

Issue #1098, opened 2016-06-30
External images are downloaded again each month


haroldangenent:
### Expected behavior

External images are downloaded to the uploads directory. Expected behaviour is that they are downloaded once and then served from this location in the future.
### Actual behavior

If an image is downloaded on 30-06-2016, it will be downloaded again on 01-07-2016. When a lot of images are involved, this can really fill up a hard drive.
### Steps to reproduce behavior
1. Use resize on an external image
2. Resize it again when a month passes
### Possible solution

If all images would be downloaded in the same folder, this wouldn't happen. You could also easily remove all sideloaded images that way.

This is not really a bug, but it would be great if this behaviour could be altered by filtering `get_sideloaded_file_loc()` or something.
### What version of WordPress, PHP and Timber are you using?

WordPress 4.5.3, PHP 7, Timber 1.0.5.
### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)

Timber Library via WPackagist (but doesn't really matter).

hbrok:
My solution (for now) was adding an action to [`ImageHelper::get_sideloaded_file_loc()`](https://github.com/timber/timber/blob/cbb8dc14ce47f5d7f24504675ea7390a8adca65b/lib/ImageHelper.php#L276):

```
public static function get_sideloaded_file_loc( $file ) {
		do_action('timber/img/external_upload_subdir'); // I added this line.
		$upload = wp_upload_dir();
		$dir = $upload['path'];
		$filename = $file;
		$file = parse_url($file);
		$path_parts = pathinfo($file['path']);
		$basename = md5($filename);
		$ext = 'jpg';
		if ( isset($path_parts['extension']) ) {
			$ext = $path_parts['extension'];
		}
		return $dir.'/'.$basename.'.'.$ext;
	}
```

Then in the action, I use the `upload_dir` filter to change the subdir to `external`.
```
add_action( 'timber/img/external_upload_subdir', function () {
	add_filter( 'upload_dir', function ( $upload ) {
		$upload['subdir'] = '/external';
		$upload['path']   = $upload['basedir'] . $upload['subdir'];
		$upload['url']    = $upload['baseurl'] . $upload['subdir'];
		return $upload;
	} );
} );
```

I did this so I would be editing as little of the Timber plugin files as possible so upgrades won't be a pain. But I'm referencing hundreds of external images so I needed a quick fix!

haroldangenent:
[...]

PR #2376, opened 2020-11-20
2.x Use a subdirectory for sideloaded images

gchtr:
**Ticket**: #1098

## Issue

Images from external URLs are downloaded in to the uploads folder directly. If a year-month-based folder structure is used, it will cause external images to be downloaded again each month. See #1098.

## Solution

- Adds a temporary filter for `upload_dir` that sets the subfolder to **sideloaded** when `ImageHelper::sideload_image()` is used.
- Adds a `timber/sideload_image/subdir` filter to change the behavior.
- Adds documentation to the Image Guide.
- Adds a note in the Upgrade Guide.

## Impact

- Prevents filling up the webspace with unnecessary image data.
- Makes it possible to remove all external images.
- No more need for custom solutions like documented in #1098.

## Usage Changes

This changes the behavior for sideloaded images. I don’t know if this is a breaking change, but it will certainly have an impact for certain sites.

To control the behavior, there’s the `timber/sideload_image/subdir` filter.

## Considerations

Should we use "sideloaded" as the default folder name. Or would "external" be a better name?

## Testing

I added all relevant tests I could think of.

It’s currently not possible to add a test and change the `uploads_use_yearmonth_folders` option in a test, because `wp_upload_dir()` uses a cache and it’s first called in https://github.com/timber/timber/blob/master/tests/bootstrap.php#L36. We’d probably have to restructure that a bit.


jarednova:
Ugh, SUCH an obvious solution! will review over the weekend

lib/ImageHelper.php
```diff
@@ -407,9 +429,61 @@ public static function sideload_image( $file ) {
...
+	 *
+	 * Used by ImageHelper::sideload_image().
+	 *
+	 * @internal
+	 * @since 2.0.0
+	 * @see   \Timber\ImageHelper::sideload_image()
+	 *
+	 * @param array $upload Array of information about the upload directory.
+	 *
+	 * @return array 		Array of information about the upload directory, modified by this
```
jarednova, line 448: Originally, this was `mixed`. This was my most substantive change — want to make sure that's correct!

jarednova (approved):

jarednova:
@gchtr finally found the time for this (and a bunch of other Timber things). Looks good! Such an obvious solution to the problems in #1098 (which was "known issue" for a while before that). To your question of ...

> Should we use "sideloaded" as the default folder name. Or would "external" be a better name?

I think so! I avoided that term when coding this initially because images are processed/stored locally so it felt not-quite accurate. But I think the average user browsing through their file system is more likely to understand the nature of files inside an `external` directory.

lib/ImageHelper.php
gchtr: @jarednova I’m not sure I follow you here. Which parameter was `mixed`?

gchtr:
@jarednova Thanks for reviewing. I also think that `external` is indeed more clear than `sideloaded`. I add the required changes for that.

I think as soon as we resolve your comment about the `mixed` thing, I can merge this :).

lib/ImageHelper.php
jarednova: Sorry, should have been more clear. The `@return` param was originally `mixed`.

jarednova (approved):

lib/ImageHelper.php
gchtr: You mean for the `upload_dir` filter in WordPress, or in Timber? Because I didn’t actually change a parameter type, or did I?

jarednova: In Timber, the (new?) method for `ImageHelper:: set_sideload_image_upload_dir` had a return type of `mixed`. I guess conceivably it _could_ be mixed if there's an error or given a non-array value from a different function that's also hooked into `upload_dir`

gchtr: Hm... when did it have that? I guess I still don’t where exactly it had or could have had a `mixed` return type. It’s a function that is going be first introduced with this pull request, right? I’m sorry if I still don’t understand 😅.

When I look at the `upload_dir` filter, it specifically requests an `array` type that contains certain values and can’t be `mixed` either. If there would be an error, it should be inside the array with the `error` key.

```php
/**
 * Filters the uploads directory data.
 *
 * @since 2.0.0
 *
 * @param array $uploads {
 *     Array of information about the upload directory.
 *
 *     @type string       $path    Base directory and subdirectory or full path to upload directory.
 *     @type string       $url     Base URL and subdirectory or absolute URL to upload directory.
 *     @type string       $subdir  Subdirectory if uploads use year/month folders option is on.
 *     @type string       $basedir Path without subdir.
 *     @type string       $baseurl URL path without subdir.
 *     @type string|false $error   False or error message.
 * }
 */
$uploads = apply_filters( 'upload_dir', $cache[ $key ] );
```

jarednova: Yeah, I think we're good here. For posterity, you can see from this [prior commit](https://github.com/timber/timber/blob/302acbfdc7faef8998d92620a9f6ebb583943cad/lib/ImageHelper.php#L448) where it was `mixed` earlier in this PR. But regardless, it's correct as it stands right now so good to merge!

gchtr: Ooooh, now I see! Thanks for searching for that older commit. I think that `mixed` might have been dropped in through code completion in PhpStorm and I corrected it to `array` in a later commit.

PR #2377, opened 2020-11-20
Fix a very annoying test

jarednova:
## Issue
There's a routine test around date comparisons that fails ~2% of the time due to some fraction of a second elapsing between two calls to the unix epoch timestamp.

## Solution
Round things by 10 to reduce/eliminate this scenario.

## Impact
No more randomly failing tests (at least for this one, hopefully)

## Usage Changes
None

## Testing
Updated/renamed one test and the docblock info


gchtr (approved):
Brilliant! 👌🧙‍♂️

Issue #90, opened 2013-10-22
Errors after uploading timber theme to server


daslicht:
Hello,
I have created a basic Timber theme and would like to try it on my Linux server, 
but I have noticed that it is not working for some reason, no script / style is enqued nor do I get any error on the main page.

Any idea how to debug such errors ?

If I open a post however I get :  

```
Fatal error: Class 'TimberHelper' not found in /home/wensauer.info/WordPress/wp-content/themes/mwt/single.php on line 16
```

PR #2382, opened 2020-12-16
How to split into chunks flexible content fields

19h47:
Add an example on how to split into small chunks flexible content fields thanks to `acf_fc_layout` value, it could be a way to build a landing page for instance.

It may be related to this issue [#90](https://github.com/timber/starter-theme/issues/90)

What do you think about it?

### DIFF
--- a/docs/v2/integrations/advanced-custom-fields.md
+++ b/docs/v2/integrations/advanced-custom-fields.md
@@ -186,7 +186,7 @@ Assuming block Twig files are inside a `blocks` folder:
 {% endfor %}
 ```
 
-> The filter [sanitize](https://timber.github.io/docs/v2/guides/filters/#sanitize), will slugify the block name. Consider this example for a Flexible Content Field named `credit` containing a `text` field, inside the `blocks/photo.twig` file:
+The filter [sanitize](https://timber.github.io/docs/v2/guides/filters/#sanitize), will slugify the block name. Consider this example for a Flexible Content Field named `credit` containing a `text` field, inside the `blocks/photo.twig` file:
 
 ```twig
 <p>Photo by: {{ block.credit }}</p>   
--- a/docs/v2/integrations/advanced-custom-fields.md
+++ b/docs/v2/integrations/advanced-custom-fields.md
@@ -186,12 +186,10 @@ Assuming block Twig files are inside a `blocks` folder:
 {% endfor %}
 ```
 
-> The filter [sanitize](https://timber.github.io/docs/v2/guides/filters/#sanitize), will slugify the block name
-
-And for example for a Flexible Content Field named `text` containing a `text` field, inside a `blocks/text.twig` file:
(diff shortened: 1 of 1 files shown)


docs/v2/integrations/advanced-custom-fields.md
```diff
@@ -171,6 +171,29 @@ When you run `meta` on an outer ACF field, everything inside is ready to be trav
...
+{% endfor %}
+```
+
+> The filter [sanitize](https://timber.github.io/docs/v2/guides/filters/#sanitize), slugify the block name
+
+And for example for a flexible content named `text` containing a `text` field, inside a `blocks/text.twig` file:
+
+```twig
+<p>{{ block.text }}</p>   
+```
```
jarednova: @19h47 I'm finding this example a little hard to understand. I think it might clarify if we use different names? for example:

A Flexible Content Field named `credit` containing a `text` field, inside a `blocks/photo.twig` file:

```twig
<p>Photo by {{ block.credit }}</p>   
```

... of course, I want to make sure that's correct and what you're actually trying to demonstrate

jarednova (changes requested):
Thanks @19h47! I made a few capitalization/wording tweaks and then left one Q for you on your example. Very cool way of utilizing the layout field. Def. will use this pattern in the future

gchtr:
@jarednova Before you merge this in, I have some smaller tweaks I want to add. Will do over the holidays ;).

docs/v2/integrations/advanced-custom-fields.md
19h47: Hello @jarednova, you totally right, my example with only the word `text` is weird and very redundant. A more real world example with a block named `photo` with a `credit` field is way better. 

Thank you for pointing this!

19h47:
@gchtr Can't wait to see this. And happy holidays everyone!

jarednova (approved):

jarednova:
cool, back to you @gchtr !

gchtr (commented):
Nice! I added some changes in #2388, see [this commit](https://github.com/timber/timber/pull/2388/commits/69aed88a0ce76d384e10d5bb87e6a1fe106eacde).

I didn’t know you could use variables in Twig strings with `#{variable}`. I always used concatenation with `~` or the `|format` filter, which works like `sprintf()`. 👍

19h47:
Thank you! 🙏🏼

Yes, I like to see this interpolation like an ES6 template literals, but `sprintf` do the job as well

gchtr:
I feel like I should document all these small Twig things in the [Twig cookbook](https://timber.github.io/docs/guides/cookbook-twig/) eventually.

19h47:
> I feel like I should document all these small Twig things in the [Twig cookbook](https://timber.github.io/docs/guides/cookbook-twig/) eventually.

Totally agreed

About that, in the [Filters](https://timber.github.io/docs/guides/wp-integration/#filters) section, the Twig `filter` tag is deprecated for the [`apply`](https://twig.symfony.com/doc/3.x/tags/apply.html) tag. 

But I'm not sure `apply_filter` still work, I have to test it

PR #2383, opened 2020-12-21
📝 Add to menus.md: Code samples using > 1 menu

hdevilbiss:
**Add 2 code samples, which show how to use the global $context for more than one menu.**
1. Declare plural menus in `functions.php`.
2. Use one menu of several in Twig markup.

## Issue

The `docs/menus.md` did not have an example, in which more than one menu was declared in the global `$context`. As such, I created two menus, but with the latter overriding the former at `$context['menu']`. This led to the behavior that only one menu was available in the markup.

## Solution

Write a Markdown file with examples showing how to declare more than one menu. Use the example menu in Twig markup.

## Impact

This Pull Request does not affect the PHP codebase. It is an addition of 30 lines to `/docs/menus.md`.

## Usage Changes

I do not think there are usage changes. Please correct me if I am wrong.

## Considerations

The `menus.md` file will look more fleshed out as a result of this addition.

## Testing

The syntax highlighting mimics the same syntax highlighting as the other examples, using `php` and `twig`, respectively.

### DIFF
--- a/docs/guides/menus.md
+++ b/docs/guides/menus.md
@@ -88,6 +88,19 @@ $context = Timber::context();
 Timber::render( 'index.twig', $context );
 ```
 
+When adding multiple menus to the global `$context`, declare unique names for each menu in the `add_to_context` function.
+
+**functions.php**
+
+```php
+public function add_to_context( $context ) {
+    $context['primary_menu']  = new Timber\Menu('Primary Navigation');
+    $context['footer_menu'] = new Timber\Menu('Footer Navigation');
+    // ...
+    return $context;
+}
+```
+
 ## Displaying the menu items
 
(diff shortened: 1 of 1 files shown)


jarednova:
Thanks @hdevilbiss ! I really appreciate this clear example (and a tidy PR to match). Looks good to me, but @gchtr is the official code owner on the `/docs` so I want to make sure he has a chance to glance too.

jarednova (approved):

gchtr (approved):
Hey @hdevilbiss, thanks for your first contribution to Timber!

This is a very nice addition! When you’re used to Timber, it’s harder to spot the things that are so obvious now, but might not be for people starting out with Timber.

Everything looks good in terms of formatting for the documentation as well. I’ll see that I can add this to the live documentation today.

If you find more things that you think could be documented better, you’re very welcome to create more pull requests. 👍

Issue #2368, opened 2020-11-10
2.x Document changes for Timber::context()


gchtr:
In #2358, we added a new parameter for `Timber::context()` that still needs some documentation.

- [x] Make passing the data into `Timber::context()` the new default (see https://github.com/timber/timber/pull/2358#issuecomment-724278274) and update code examples.
- [x] Update Upgrade Guide.

gchtr:
Resolved in https://github.com/timber/timber/pull/2393.

PR #2388, opened 2020-12-30
#2382 Optimizations for "How to split into chunks flexible content fields"

gchtr:
**Ticket**: #2382

I added some optimizations for #2368 

- Moved new content from section about Reapeater Fields to Flexible Content Field section.
- Used `include()` function instead of `include` statement, as this is recommended by Twig, see https://twig.symfony.com/doc/3.x/tags/include.html.
- Added some syntax optimizations.

### DIFF
--- a/docs/v2/integrations/advanced-custom-fields.md
+++ b/docs/v2/integrations/advanced-custom-fields.md
@@ -226,7 +226,8 @@ Similar to nested repeaters, you should only call the `meta` method once when yo
 
 We can break Flexible Content Fields into small blocks with include files utilizing the `acf_fc_layout` value. This way you have more flexibility and reusability for included sections. For instance, this is a great way to build landing pages that re-use the same blocks in different configurations.
 
-Assuming your block Twig files are inside a **blocks** folder:
+You could use a **blocks** subdirectory where you put all your Twig template files for your blocks. For example:
+**wp-content/themes/example-theme/views/blocks**.
 
 ```twig
 {% for block in post.meta( 'blocks' ) %}
@@ -236,10 +237,22 @@ Assuming your block Twig files are inside a **blocks** folder:
 {% endfor %}
 ```
 
-The [sanitize](https://timber.github.io/docs/v2/guides/filters/#sanitize) filter will slugify the block name. Consider this example for a Flexible Content Field named `credit` containing a `text` field, inside the **blocks/photo.twig** file:
+The [sanitize](https://timber.github.io/docs/v2/guides/filters/#sanitize) filter will slugify the block name.
+
+Consider this example for a Flexible Content Field named `Photo` with a text field named `credit`. You would create a **blocks/photo.twig** file which is automatically included:
 
 ```twig
 <p>Photo by: {{ block.credit }}</p>
+```
+
+To prevent errors with include files that can’t be found, you can optionally use the `ignore_missing` parameter for `include()`:
+
+```twig
+{% for block in post.meta( 'blocks' ) %}
+    {{ include("blocks/#{ block.acf_fc_layout|sanitize ) }.twig", {
+        block: block
+    }, ignore_missing = true) }}
+{% endfor %}
 ```
 
 ## Options Page
--- a/docs/v2/integrations/advanced-custom-fields.md
+++ b/docs/v2/integrations/advanced-custom-fields.md
@@ -186,7 +186,7 @@ Assuming block Twig files are inside a `blocks` folder:
 {% endfor %}
 ```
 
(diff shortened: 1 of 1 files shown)


19h47:
Didn't know about the `include` function from Twig 3 either, thank you! 😮

gchtr:
> Didn't know about the `include` function from Twig 3 either, thank you! 😮

It’s also available in Twig 2: https://twig.symfony.com/doc/2.x/functions/include.html.

docs/v2/integrations/advanced-custom-fields.md
```diff
@@ -232,7 +222,25 @@ Similar to nested repeaters, you should only call the `meta` method once when yo
 {% endfor %}
 ```
 
-* * *
+### Split Flexible Content Fields into includes / chunks
+
+We can break Flexible Content Fields into small blocks with include files utilizing the `acf_fc_layout` value. This way you have more flexibility and reusability for included sections. For instance, this is a great way to build landing pages that re-use the same blocks in different configurations.
+
+Assuming your block Twig files are inside a **blocks** folder:
```
jarednova, line 229:
```suggestion
Assuming your block Twig files are inside a **my_blocks** folder within your other Twig file directory (ex: `my_theme/views/my_blocks`).
```

docs/v2/integrations/advanced-custom-fields.md
```diff
@@ -232,7 +222,25 @@ Similar to nested repeaters, you should only call the `meta` method once when yo
...
+
+```twig
+{% for block in post.meta( 'blocks' ) %}
+    {{ include("blocks/#{ block.acf_fc_layout|sanitize ) }.twig", {
```
jarednova, line 233:
```suggestion
    {{ include("my_blocks/#{ block.acf_fc_layout|sanitize ) }.twig", {
```

docs/v2/integrations/advanced-custom-fields.md
```diff
@@ -232,7 +222,25 @@ Similar to nested repeaters, you should only call the `meta` method once when yo
...
+        block: block
+    }) }}
+{% endfor %}
+```
+
+The [sanitize](https://timber.github.io/docs/v2/guides/filters/#sanitize) filter will slugify the block name. Consider this example for a Flexible Content Field named `credit` containing a `text` field, inside the **blocks/photo.twig** file:
```
jarednova, line 239:
```suggestion
The [sanitize](https://timber.github.io/docs/v2/guides/filters/#sanitize) filter will slugify the block name. Consider this example for a Flexible Content Field named `credit` containing a `text` field, inside the **my_blocks/photo.twig** file:
```

jarednova (changes requested):
Looks good! Just one suggestion in regards to use of names. If I were building a theme, I'd probably keep everything named the same ("blocks") — but I think we should make super-clear to readers that there's no "magic" in naming things the same and they can use whatever names they need/want.

19h47:
Totally agreed. No "where do you put this folder" question

gchtr:
@jarednova I agree, we can make this clearer, but I’m always a little hesitant about using "my" in examples. I wouldn’t use `my_blocks` as a name for anything that I code myself. I’m sure it adds to the notion of "you can actually choose a name for this", but I think it’s not good practice for actually naming things. I also see that lots of developers copy-paste code from the documentation as it is. So I’d rather use real world code examples than examples where you’d have to rename stuff before using it.

But, maybe I’m all wrong 😊. Maybe if I’ve yet missed that using "my" is some best practice that’s established in the developer community.

> but I think we should make super-clear to readers that there's no "magic" in naming things the same and they can use whatever names they need/want.

Wouldn’t it be clear enough if we mention that you can name the folder as you want? Maybe like this:

`Assume your block Twig files are inside a folder within your other Twig template directory. For example, if you used a folder named **blocks**, then that would be the **wp-content/themes/example-theme/views/blocks** folder.`

jarednova:
@gchtr I definitely take `my_` as a "you can actually choose a name for this, but I think it’s not good practice for actually naming things." Though, agreed on the copy-and-pasted tendencies out there. As long as the "magic" problem is solved for, your solution of `Assume your block Twig files are inside a folder within your other Twig template directory. For example, if you used a folder named **blocks**, then that would be the **wp-content/themes/example-theme/views/blocks** folder` is totally 👍  w me!

gchtr:
@jarednova Cool, so I guess we’re on the same page here. I’ve updated the documentation once again a little in [`24c790a` (#2388)](https://github.com/timber/timber/pull/2388/commits/24c790ac33038add222bd381dde71eaaebfd4fe9). I hope it’s clearer now, even without using "my".

jarednova (approved):
LGTM!

Issue #2362, opened 2020-11-03
Term returning false if we have equal slug in different taxonomies.


emanuellopes:
## Expected behavior
Term Class should return the term searched by slug.

## Actual behavior
If we have to terms from different taxonomy this will return always the first based on SQL Query.

**What I did to fix this issue on my side:**

```php
<?php

namespace App\Models\Term;

use Timber\Term as TermModel;

class Category extends TermModel {

	public function __construct( $slug = null ) {
		$term = get_term_by( 'name', $slug, 'category' );
		if ( ! isset( $term->term_id ) ) {
			$term = $slug;
		}

		parent::__construct( $term, 'category' );
	}
}

```
## Steps to reproduce the behavior

1. Create a tag with name security and slug security
2. Create a category with name security and slug security.

Do the following:
new Timber\Term('security', 'category'); this will return false because we have two terms with the same name.

The Sql Query on Term Class
```php
/**
	 * @internal
	 * @param int $tid
	 * @return int
	 */
	protected function get_tid( $tid ) {
		global $wpdb;
		if ( is_numeric($tid) ) { // if is an ID exit here;
			return $tid;
		}
		if ( gettype($tid) == 'object' ) {
			$tid = $tid->term_id;
		}
		if ( is_numeric($tid) ) { // never enter here
			$query = $wpdb->prepare("SELECT * FROM $wpdb->terms WHERE term_id = %d", $tid);
		} else {
			$query = $wpdb->prepare("SELECT * FROM $wpdb->terms WHERE slug = %s", $tid);
		}
		$result = $wpdb->get_row($query);
		if ( isset($result->term_id) ) {
			$result->ID = $result->term_id;
			$result->id = $result->term_id;
			return $result->ID;
		}
		return 0;
	}
```
![Screenshot_2020-11-03 16 02 58_KGdoWM](https://user-images.githubusercontent.com/6977270/98009986-149a2380-1dee-11eb-8665-74656f50606a.png)
 

## What version of WordPress, PHP and Timber are you using?
- WordPress 5.5.3
- Timber 1.18.2

## How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Installed using Composer/Packagist

jarednova:
Thank you @emanuellopes for the issue and clear test case. I was able to confirm the issue in 1.x. However, in 2.x the syntax for getting a specific term like the one you show is via the new Factories classes and the issue does not occur.

[...]

PR #2390, opened 2020-12-30
Handle for duplicate term names in difft taxes

jarednova:
**Ticket**: #2362 

## Issue
When there are two identically named terms in two different taxonomies (which WordPress allows), attempting to fetch the newest one from a specific taxonomy will return false.

## Solution
Modify the `Term::get_tid` method. When >1 term_id is discovered, test it against possible taxonomies in `Term::get_term`.

## Impact
This will apply only to 1.x as 2.x handles this situation much more gracefully via Factories. The `Term::get_tid` method is protected, so modifying its return type is unlikely to cause issues. Theoretically, there could be an issue only if a user has extended the `Term` class specifically and is directly interacting with the method.

## Usage Changes
None. Should now work as anticipated.

## Considerations
This is a patch. The true "fix" is how 2.x handles this. I'd spend more time optimizing if this were going to be an enduring part of the code base. This definitely qualifies as a "bug" which is why I wanted to resolve for 1.x

## Testing
New test written to uncover and solve the issue described in #2362

### DIFF
--- a/lib/Term.php
+++ b/lib/Term.php
@@ -164,6 +164,19 @@ class Term extends Core implements CoreInterface {
 		}
 		$tid = self::get_tid($tid);
 
+		if ( is_array($tid) ) {
+			//there's more than one matching $term_id, let's figure out which is correct
+			if ( isset($this->taxonomy) && strlen($this->taxonomy) ) {
+				foreach( $tid as $term_id ) {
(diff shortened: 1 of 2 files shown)


gchtr (approved):
Looks fine to me for a patch that fixes this bug in 1.x 👍

jarednova:
Cool! Just waiting to get Travis going again then I'll merge ...

Issue #2359, opened 2020-10-30
::get_term_by() API method


acobster:
## Is your feature request related to a problem? Please describe.

We already have `::get_post_by()`, `::get_user_by()`, etc., and WordPress has a `get_term_by()` core function. So we should probably have a `Timber::get_term_by()` method.

## Describe the solution you’d like

Can probably just delegate to `Timber::get_term()` and `get_term_by()`:

```php
public static function get_term_by( string $field, $ident, string $taxonomy = '' ) {
  $wp_term = get_term_by( $field, $ident, $taxonomy );
  return $wp_term ? self::get_term( $wp_term ) : false;
}
```

Might be edge cases to flesh out.

Note that the `$taxonomy` arg is optional if `$field` is `term_taxonomy_id`. But is `term_taxonomy_id` ever not the same thing as `id`/`ID`? I don't know! This might harken back to [this edge case](https://github.com/timber/timber/issues/2087#issuecomment-578530137) in the old WP API (nice summary [here](https://wordpress.stackexchange.com/questions/357211/what-is-the-use-case-for-sharing-a-term-between-multiple-taxonomies)) when terms were shared across taxonomies. So if `$taxonomy` is required here for the same reason, and `term_id` is always the same as `term_taxonomy_id`, then that means we only need to require it if `$field` is `slug`, making the API little nicer than WP core's.

## Describe alternatives you’ve considered

Resistance is futile. :robot: 

## Additional context

I recently had to [do this](https://github.com/sitecrafting/greg/blob/main/src/api.php#L222) in Greg, a calendar plugin I'm building on top of Timber. The solution there works but it'd be much nicer if this logic just lived in Timber.

PR #2391, opened 2020-12-30
Introduce Timber::get_term_by() method

jarednova:
**Ticket**: #2359 

## Issue
There's `Timber::get_post_by`, `Timber::get_user_by` ... — let's wrap it up with `get_term_by`

## Solution
Copy the pattern from other get_x_by and apply it!

## Impact
More poly-morphizing

## Usage Changes
- [x] Need to add docs example

## Considerations
@acobster suggested in the write-up that we can improve upon WP functionality by better processing the `taxonomy` value. I agree. Even when writing the tests I was thrown off by its requirement even when there's only one term with a certain name.
- [x] Handle for what happens when a taxonomy isn't supplied

## Testing
Wrote a new test `testTermBy`

### DIFF
--- a/docs/v2/guides/terms.md
+++ b/docs/v2/guides/terms.md
@@ -21,18 +21,30 @@ $term = Timber::get_term( get_queried_object_id() );
 
 What you get in return is a [`Timber\Term`](https://timber.github.io/docs/v2/reference/timber-term/) object, which is similar to `WP_Term`.
 
+## Get term by field
+
+If you don’t have a term ID, you can also get a term by other fields, like `slug` or `name` through `Timber::get_term_by()`.
+
+```php
+// Get a term by slug.
+$term = Timber::get_term_by( 'slug', 'news', 'category' );
+
+// Get a term by name.
+$term = Timber::get_term_by( 'name', 'News', 'category' );
+```
+
 ## Twig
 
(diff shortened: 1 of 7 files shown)


gchtr (approved):
This looks all good from my side, but I’d be glad if @acobster could also look over it.

jarednova:
@acobster just wanted to make sure you had a chance to take a gander at this on.

In regards to merges (on this and others). Travis support is totally useless. I think we should plan to merge based on the results of local tests. If they aren't able to get things back running by end of the week, I'm going to look into migrating to GitHub Actions.

acobster:
@jarednova noted. I usually run the full suite locally before merging or creating a PR anyway.

acobster:
also, this did fall off my radar so thanks for pinging me! Hope to get to it this week.

jarednova:
Finally got Travis tests going (OMFG) and needed to adjust a few tests to match w most up-to-date return types and WP stuff. @acobster remember me? Hi! 👋 . This one is pretty non-controversial so I want to give a little time in case you'd like to do a final look

acobster:
@jarednova hmmmm your name rings a bell from some time before the decade or so I've spent doomscrolling in my cave....Hi! Taking a look at this now.

acobster (approved):
Looks great!

jarednova:
Ahhh, we were so young!

PR #2392, opened 2020-12-30
2.x Fix various instances of direct object instantiation

gchtr:
## Issue

I found several instances where we forgot to use the new API.

## Solution

Switch to new API in documentation where needed.

## Impact

No confusion.

## Usage Changes

None.

## Considerations

None.

## Testing

None, documentation only.


jarednova (dismissed):
good catches 👍 !

PR #2393, opened 2020-12-30
#2368 – Update context format in documentation

gchtr:
**Ticket**: #2368

## Issue

Now that we can pass an array directly to `Timber::context()`, it’s probably worth updating the documentation to the new format in a couple of cases.

## Solution

Update documentation.

## Impact

I think it’s actually more readable in a couple of places.

## Usage Changes

None.

## Considerations

I doesn’t make sense to change the format everywhere, especially if there is advanced logic involved.

## Testing

None.

## Todo

- [x] Update Context Guide.
- [x] Update Upgrade Guide.

### DIFF
--- a/docs/v2/guides/context.md
+++ b/docs/v2/guides/context.md
@@ -27,7 +27,9 @@ Timber::render( 'single.twig', $data );
 <p>{{ message }}</p>
 ```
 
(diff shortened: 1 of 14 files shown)


jarednova (approved):
Made two small fixes/tweaks — looks good to me!

acobster (approved):
Yep, looks good!

PR #2394, opened 2020-12-30
2.x Documentation optimizations

gchtr:
## Issue

Documentation formatting.

## Solution

- Removes a couple of horizontal rules (`* * *`) we don’t need.
- Optimizes where we add `<?php` in front of PHP code examples. In general: when it’s a whole template file, there should be an opening tag, otherwise, the surrounding text should suggest what language it is, e.g. when it’s introduced with a filename (single.php, functions.php, etc).
- Uses more consistent file name formatting by using bold formatting instead of code formatting.
- Fixes a couple of newline inconsistencies.
- Fixes a couple of incorrect array declarations, when an array was declared by assigning a key.

## Impact

More consistency. A little less clutter.

## Usage Changes

None.

## Considerations

None.

## Testing

None, documentation only.

### DIFF
--- a/docs/v2/getting-started/a-post-archive.md
+++ b/docs/v2/getting-started/a-post-archive.md
@@ -73,7 +73,7 @@ There are two new things that you see here:
 
 ## Using custom queries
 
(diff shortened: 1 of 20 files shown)


jarednova (approved):
Agree with all the steps! Very helpful to also establish some of those conventions (ex: filename in bold vs. code-formatted) to guide other docs too

gchtr:
@jarednova Cool, thanks! I guess I always used these conventions, which are based on "patterns" I already found in your documentation ;).

Issue #2385, opened 2020-12-24
PHP 8 Support


romanmartushev:
## Is your feature request related to a problem? Please describe.
This is an example of an error I am getting : `Deprecated: Required parameter $data follows optional parameter $sidebar in vendor/timber/timber/lib/Timber.php on line 462`

## Describe the solution you’d like
PHP 8 has deprecated required parameters after optional parameters.
for example: `public static function get_sidebar_from_php( $sidebar = '', $data )` the `$data` param can be given a default value to fix the error I am getting. This is the function I am getting an error on.

## Describe alternatives you’ve considered
At least the minimum amount of work done so that PHP 8 is supported throughout the project.

jarednova:
Thanks @romanmartushev ! PHP 8 is definitely something we'll be getting ready for. It feels like there should be a pretty limited number of these parameter issues, so please feel free to create a PR!

PR #2395, opened 2021-01-02
#2385 - PHP 8 Add default value to required parameter

marciojc:
**Ticket**: #2385

## Issue
Missing default value

## Solution
Add a default value to the missing required value

### DIFF
--- a/lib/Timber.php
+++ b/lib/Timber.php
@@ -459,7 +459,7 @@ class Timber {
 	 * @param array   $data
 	 * @return string
 	 */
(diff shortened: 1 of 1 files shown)


marciojc:
@jarednova maybe it's better to remove $data. It's not being used on the function.

jarednova:
@marciojc awesome, thanks! Like you say, what on earth is that `$data` var doing there in the first place? It's a good Q and one we should look into on a fresh PR (if warranted). It looks like junk, but I have this thought that there's some weird `global` piece that could be connected with the output buffer stuff.

Issue #2404, opened 2021-01-13
What is the official way of initializing Timber?


acobster:
## Is your feature request related to a problem? Please describe.

I found that now that Timber is a composer package only, I need to call `new Timber\Timber()` explicitly from my theme for anything to work. Is that the official way to initialize Timber now?

Also, do we want to refactor that process at all? It's a little awkward that the constructor calls `self::init()` statically, for example.

gchtr:
We could probably improve that and not do anything in the constructor, because obviously that is a pattern that leads to other problems.

However, we can’t init Timber automatically, because of compatibility issues with environments like Bedrock. With Bedrock, Composer initializes earlier, when WordPress is not loaded yet, which would lead to fatal errors. More details in https://github.com/timber/timber/issues/979#issuecomment-220094476 and https://github.com/timber/timber/pull/1041.

How about `Timber::init()` as the official way to initialize Timber?

twentyfortysix:
..how is it with initialisation of Timber with Bedrock now? I went thru all tickets I found and not even the tickets mentioned earlier helped me.
I have the Bedrock with Timber installed form github as library and following in functions.php in theme and I still get fatal error 
```
require_once( $_SERVER['DOCUMENT_ROOT'].'/../vendor/autoload.php' );
use Timber\Timber;

```

`Fatal error: Uncaught Error: Class 'Timber\Timber' not found...`

jameelmoses:
> ..how is it with initialisation of Timber with Bedrock now? I went thru all tickets I found and not even the tickets mentioned earlier helped me.
> I have the Bedrock with Timber installed form github as library and following in functions.php in theme and I still get fatal error
> 
> ```
> require_once( $_SERVER['DOCUMENT_ROOT'].'/../vendor/autoload.php' );
> use Timber\Timber;
> ```
> 
> `Fatal error: Uncaught Error: Class 'Timber\Timber' not found...`

I'd need more info of your exact directory structure, but assuming you have Timber in your overall project's composer.json as opposed to a composer.json in the theme, this is what I have in my wp-config.php: `require_once dirname( __DIR__ ) . '/vendor/autoload.php';`

twentyfortysix:
Hi jameelmoses

here is my composer.json
https://gist.github.com/twentyfortysix/5b88880c4e8e65ac7d012ff45a2b8294
Yes I try to load Timber as Library not as a plugin or use the composer inside the template directory.

[...]

Issue #2402, opened 2021-01-13
2.0 Integrations API


acobster:
## Is your feature request related to a problem? Please describe.

**This proposal has been edited for clarity and consistency with other APIs.**

We've discussed pulling integrations (e.g. ACF) out into separate libraries. This seems like overkill to me, but I think a happy medium would be to have an official, public Integration API that we then "dogfood," implementing the current integrations in terms of that API.

This is, IMO, fairly low-effort with a big benefit for people hoping to integrate their plugins of choice.

### 1. An `IntegrationInterface`

```php
namespace Timber\Integration;

interface IntegrationInterface {
  public function should_init() : bool;
  public function init() : void;
}
```

### 2. Re-implement existing integrations in terms of this interface

Example ACF integration:

```php
use ACF;
use Timber\Integration\IntegrationInterface;

class AcfIntegration implements IntegrationInterface {
  public function should_init() : bool {
    return class_exists( ACF::class );
  }

  public function init() : void {
    add_filter('timber/post/pre_meta', [static::class, 'post_get_meta_field'], 10, 5);
    // ...all the stuff that Timber\Integrations\ACF::__construct() currently does...
  }
}
```

### 3. Register the integrations:

```php
add_filter('timber/integrations', function(array $integrations) {
  return array_merge($integrations, [
    AcfIntegration::class,
    CoAuthorsPlusIntegration::class,
    CarbonFieldsIntegration::class,
  ];
});
```

The above code would live in `Timber::init()`, where `new Integrations()` is currently called. Other theme-level code would follow the exact same pattern.

`Timber\Timber` itself then loops over the result of applying this filter, activating each one accordingly:

```php
foreach (apply_filters('timber/integrations', []) as $integration) {
  if ($integration->should_init()) {
    $integration->init();
  }
}
```

No more `Integrations` junk-drawer class; no more `maybe_init_integrations()` method. :smile: 

## Describe alternatives you’ve considered

Keeping things the same? Where's the fun in that?

## Additional context

See discussion in #2397

gchtr:
❤️

> We've discussed pulling integrations (e.g. ACF) out into separate libraries. This seems like overkill to me, but I think a happy medium would be to have an official, public Integration API that we then "dogfood," implementing the current integrations in terms of that API.

Exactly my thoughts as well.

Having a filter for all integrations and putting the check whether an integration should be active into a function is exactly what I thought of when I thought about how integrations could be controlled better.

I have only two things that come to my mind:

[...]

PR #2405, opened 2021-01-13
Integrations API

acobster:
**Ticket**: #2402

## Issue

New Integrations API!

## Solution

* For consistency with `Factory` and `Image`, renamed the namespace to `Timber\Integration`
* Implemented the API per proposal #2404
* Preferring Gravatars in CoAuthors Plus meant you had to set `CoAuthorsPlus::$prefer_gravatar = true` which is awkward and non-idiomatic. Updated this to use a plain old filter instead.

## Left to do

- [x] Document how to build a custom integration
- [x] Fix an issue with image URLs in WPML integration
- [x] Fix potential issues in `test-timber-meta-deprecated.php`
- [x] Not sure what `Timber\Integrations\Command` is for - I haven't dug into that code yet. But it's not one of the ones that was previously initialized in `Integrations::__construct()`...

## Impact

Cleaner internal code, easier external integrations.

## Usage Changes

This wasn't an external API until now, so usage changes will be covered by the introduction of docs for this new API.

## Considerations

Please look carefully at the WPML integration and how it's tested. I'm less confident in that than I am in the others, because WPML has a bigger scope.

## Testing

Reworked tests to use the new API. Tests were mostly unaffected.


acobster:
@gchtr the image resize test that's failing now (`vendor/bin/phpunit --group integrations`) is confusing to me. See my most recent commit. It seems to have to do with the "sideloaded" images that I believe you were working on recently - can you take a look?

gchtr:
> @gchtr the image resize test that's failing now (`vendor/bin/phpunit --group integrations`) is confusing to me. See my most recent commit. It seems to have to do with the "sideloaded" images that I believe you were working on recently - can you take a look?

That was [a missing `ICL_LANGUAGE_CODE` constant for the WPML test](https://github.com/timber/timber/pull/2405/commits/8e275e8caef8895cf0d5e234c26c39b9d6a581b1#diff-124fa5993acee7315571fbe0f1dd6a3748c913adbc823807454a7f68536e0696R45). I hope that works now.

I moved all WPML-related tests to a new tests file that matches other integration tests. But I guess some things could still be moved around. We could probably use a `setUp()` function for the `ICL_LANGUAGE_CODE` constant, because that is a constant that is always set in a WPML environment.

acobster:
> We could probably use a setUp() function for the ICL_LANGUAGE_CODE constant, because that is a constant that is always set in a WPML environment.

Couple quick thoughts -

1. bootstrap.php is probably the better place for constants
2. If the constant is always set in a WPML environment, then the [defined](https://github.com/timber/timber/pull/2405/files#diff-e7a8f8a99b028cf01dae8110b5af1749523fd8231f1936a8a0e6a972d523f166R20) check will never fail, and we don't have a way to test the other code path (i.e. when `ICL_LANGUAGE_CODE` is not defined) anyway. Should we just remove that check?

gchtr:
@acobster Yeah, probably **bootstrap.php** is a better place. Since we don’t actually install WPML when testing, setting that constant should work fine and then we wouldn’t need the `defined()` check.

acobster:
@gchtr thanks for taking a look and consolidating those tests! Still not sure why the meta tests are failing. Also, please take a look at the new comments in **test-timber-user.php** - the tests pass now but if I add those same filters in a `setUp` method they fail, which indicates some implicit dependency to me. Haven't figure it out yet. I'd prefer to do it in `setUp()` because A.) it's cleaner, and B.) it uses the default `Timber\User` class across the board. Currently it's a `CoAuthorsPlusUser` instance, since the integration is global.

gchtr:
🤔 It works for me if I use `setUp()`, including `parent::setUp()`:

```php
function setUp() {
    parent::setUp();

    $this->add_filter_temporarily('timber/user/classmap', function() {
        return User::class;
    });
}
```

acobster:
Oh, duh! I was forgetting to call `parent` 😛

gchtr:
🙈😅 Sometimes it’s that, sometimes it’s the order in which tests are called, sometimes it’s something else. No worries!

jarednova:
> Oh, duh! I was forgetting to call parent 😛

I just banged my head against a wall for 30 minutes on #2413 on this. @acobster is this something you see as a required 2.0 element? or something that belongs in the 2.1 camp?

acobster:
I think 2.0 because we will need to refactor some of the ACF/meta stuff on top of this, and that stuff has at least one breaking change.

acobster:
I will get around to fixing those User tests soon and then I think this will be solid enough to merge in. The meta tests that are failing will probably need to be revisited anyway as we change the API/integration with ACF etc.

acobster:
@gchtr work came knocking for me too, sorry for falling off the face of the Earth! This should be ready to go now, we'll just need to resolve the meta test failures on #2397 as planned.

lib/ImageHelper.php
```diff
@@ -533,6 +533,9 @@ public static function analyze_url( $url ) {
 		$parts = PathHelper::pathinfo($tmp);
 		$result['subdir'] = ($parts['dirname'] === '/') ? '' : $parts['dirname'];
 		$result['filename'] = $parts['filename'];
+		// @todo remove this debug dump
```
gchtr, line 536: Do we still have to fix something here?

lib/Timber.php
```diff
@@ -173,6 +184,21 @@ class_alias( 'Timber\Timber', 'Timber' );
...
+	 *
+	 * @internal
+	 */
+	protected static function init_integration(
+		Integration\IntegrationInterface $integration
+	) : void {
+		if ($integration->should_init()) {
+			$integration->init();
+		} else {
+			echo 'DID NOT INIT ' . get_class($integration);
```
gchtr, line 198: I guess we shouldn’t leave a stray echo here?

gchtr (commented):
> @gchtr work came knocking for me too, sorry for falling off the face of the Earth!

Same here :)

I quickly checked and found two things that look like stray debugging code (in separate comments).

And what’s the status of the following "Left to do" list?

> ## Left to do
> 
>     * Document how to build a custom integration
> 
>     * Fix an image with image URLs in WPML integration
> 
>     * Fix potential issues in `test-timber-meta-deprecated.php`
> 
>     * Not sure what `Timber\Integrations\Command` is for - I haven't dug into that code yet. But it's not one of the ones that was previously initialized in `Integrations::__construct()`...

I think an Integration guide would be great. I’ll add it to my list of things to improve, but can’t guarantee when I can work on it 😅.

lib/Timber.php
acobster: Removed!

lib/ImageHelper.php
acobster: I think that was only temporary while we were figuring some image stuff out. Removed and it's working fine.

acobster:
@gchtr thanks for looking! Those have been cleaned up.

> * Document how to build a custom integration

Guide added!

> * Fix an image with image URLs in WPML integration

I think this had to do with that leftover debug statement and is now fixed.

> * Fix potential issues in `test-timber-meta-deprecated.php`

Still running into issues where the tests expecting specific meta values are seeing `null` instead. Repro: `vendor/bin/phpunit tests/test-timber-meta-deprecated.php`

> * Not sure what `Timber\Integrations\Command` is for - I haven't dug into that code yet. But it's not one of the ones that was previously initialized in `Integrations::__construct()`...

I looked and that's just the WP-CLI integration. It's not really an "integration" in the same sense (WP-CLI is not a plugin), although we could treat it as such:

```php
class WpCliIntegration implements IntegrationInterface {
  public function should_init() {
    return defined( 'WP_CLI' ) && WP_CLI;
  }

  public function init() {
    WP_CLI::add_command( 'timber', TimberCommand::class );
  }
}
```

I feel like it's fine to consider this out of scope here though. `Command` isn't used anywhere currently.

gchtr:
> > * Fix potential issues in `test-timber-meta-deprecated.php`
> 
> Still running into issues where the tests expecting specific meta values are seeing `null` instead. Repro: `vendor/bin/phpunit tests/test-timber-meta-deprecated.php`

I found the issue and fixed it in [`5560141` (#2405)](https://github.com/timber/timber/pull/2405/commits/556014171f3cd4ae62c7a5c72a6064c78b4deb90). Classic "hook setup in a class" bug. If you want to make filters removable, you can’t use `$this` when adding the filter, because the filter will be added on the instance. If you want to remove the filter, you’ll have to get the exact instance.

That’s why the filters of the ACF integration were probably set up with static classes in the first place. I reverted it to what it was before, because that seemed to be the easiest change to fix this.

Another way to solve this would be to save the result of the initialized integrations in the `Timber\Timber` class and add a `get_integrations()` method that can be used to get the very instance we would have to use to remove the filters. But this is probably not needed for a couple of deprecated tests that will be removed with the next major version.

> > * Document how to build a custom integration
> 
> Guide added!

Cool! I looked over it and saw that you added it in the /docs/v1 folder. I moved it to the v2 folder :).

And I added a note in the Upgrade Guide.

jarednova:
Hmmmm, it seems there's a failing test we'll need to resolve regarding how post meta data is stored/set ...

```php
1) TestTimberMeta::testNonNullReturnInPreMetaFilterDisablesDatabaseFetch
Failed asserting that true matches expected false.

Error: /home/runner/work/timber/timber/tests/test-timber-meta.php:182
```
https://github.com/timber/timber/blob/feature/integrations-api/tests/test-timber-meta.php#L182

gchtr:
@jarednova I fixed the failing test in [`17f1761` (#2405)](https://github.com/timber/timber/pull/2405/commits/17f17617602488eef0ed6d5b848ef81fa433a688). Apart from the decreased coverage, I think this one is good to merge.

gchtr (approved):

jarednova (approved):

Issue #2192, opened 2020-02-07
Opting in/out of certain Twig functions & filters


joelambert:
## Is your feature request related to a problem? Please describe.
Much of the work we do requires our Twig templates to be not too closely tied to Timber. Specifically, we share the templates between a Design System and WordPress.

As such, we don't as a rule make use of Timber specific filters and functions inside Twig, instead favouring more Timber work in our Controllers. This presents:

1. a potential challenge with developers making use of these less portable filters
2. an opportunity to reduce how much needs to be loaded on each request

## Describe the solution you’d like
I realise this is possibly a niche requirement but is there any appetite for a more configurable way to load the filters and functions?

Rough idea would be to modify `Twig::add_timber_functions()` and `Twig::add_timber_filters()` function to be more extendable. If these functions called private functions that return and associative array of keys and callables this could then potentially be filtered before adding to Twig.

Pseudo code example:

```
public function add_timber_functions( $twig ) {
    $functions = $this->get_timber_functions();

    $functions = apply_filters('timber/twig/functions/add`, $functions);

    foreach ($functions as $key => $callable) {
        $twig->addFunction(new Twig_Function($key, $callable); 
    }
}

protected function get_timber_functions() {
    $functions = [];

    $functions['bloginfo'] = 'bloginfo';
    $functions['foo'] = function () {};

    return $functions;
}
``` 

Is there any appetite for this sort of a change? If so I can put together a PR for it.

jarednova:
@joelambert I'm fine with a PR for this philosophically. But before going too far down the rabbit hole, I'm wondering if `remove_action`:

https://developer.wordpress.org/reference/functions/remove_action/

... could be used to target these particular functions within `Timber\Twig` so they don't interfere with your app

nlemoine:
I also think that filtering Timber filters & functions would be interesting. The issue with `remove_action` is that it's an all or nothing solution: you can only remove **all** functions/filters. 

Furthermore, Timber [makes it hard to apply the `remove_action`](https://github.com/timber/timber/blob/master/lib/Twig.php#L20) because methods are not `static` and the `Timber\Twig` object isn't publicly accessible.

I would also be in favor of filter like the one proposed by @joelambert.

Let me know if that fits for you, I can work on a PR.

gchtr:
Resolved in #2408. Big thanks to @nlemoine for leading the way! 🙌

Issue #2284, opened 2020-07-01
Should we reduce the number of `timber/twig/…` filters?


gchtr:
After the issue in #2276, I was thinking about whether there’s a specific reason why we have so many different `timber/twig/…` filters that all do practically the same?

- `timber/twig/filters`
- `timber/twig/functions`
- `timber/twig/escapers`
- `timber/loader/twig`
- `timber/twig`

Wouldn’t it be enough if we only had `timber/twig`? The only reason I can think of now is to control the order in which functionality is added. When we as developers add our own functionality, we only use `timber/twig`, even if we add functions or filters.

Maybe only filters would be enough? Something like:

- `timber/twig/core` or `timber/twig/internal`
- `timber/twig`

Or we could also use the priority parameter in `add_filter()`:

```php
add_filter( 'timber/twig', function( $twig ) {
	// Internal stuff here.

    return $twig;
}, 9 );
```

@jarednova Do you have any insight on the history for this? For example, the `timber/twig/functions` filter was added in https://github.com/timber/timber/pull/1391/commits/21464d37dfc996338862461e7911d4841f171944, but I don’t know if it was just to continue the existing pattern or if there’s something I’m missing.

jarednova:
I can't claim any grand plan for the exact organization. My recollection is that this set:

* `timber/twig/filters`
* `timber/twig/functions`
* `timber/twig/escapers`

... was added when doing a big re-organization of the `Timber\Twig` class with the thought of "hey, whatabout all these filters?" Well, it definitely succeeded in creating filters!

I'd agree that we could probably lose or drop these in a subsequent version and stick with something like `timber/twig` as what we emphasize to use. Of course there'd be a fair amount of effort to untangle and properly deprecate

jarednova:
... and being very selective when it comes to adding any _new_ filters. There's an attractive elegance to sticking them in a bunch of places. But as #2276 highlights, those additions come with unforeseen risks and potential for mayhem.

gchtr:
> I'd agree that we could probably lose or drop these in a subsequent version and stick with something like `timber/twig` as what we emphasize to use. Of course there'd be a fair amount of effort to untangle and properly deprecate

I agree. I moved this to be resolved in a future version.

gchtr:
Resolved in #2419.

PR #2408, opened 2021-01-15
#2192 - Allow users to filter Timber default Twig functions

nlemoine:
Hello!

Sorry about the PR flood. I have some spare time so I can finally propose changes I've been thinking/waiting for a long time.

## Issue

See #2192

## Solution

Provide a filter to remove some functions/filters. (PR is still a draft, I'll implement the same logic for filters if accepted).

## Impact

None. Obviously, if you are removing some default functions/filters, you are fully aware of consequences.

## Usage Changes

None.

## Considerations

Unfortunately, Twig does not support functions/filters removal once added.

## Testing

Not yet, I will if/when accepted.

## Side note

This PR also makes use of [native Twig `deprecated` option](https://twig.symfony.com/doc/2.x/advanced.html#deprecated-filters).


lib/Twig.php
```diff
@@ -22,12 +22,173 @@ class Twig {
 	public static function init() {
 		$self = new self();
 
-        add_action( 'timber/twig/filters', array( $self, 'add_timber_filters' ) );
-		add_action( 'timber/twig/functions', array( $self, 'add_timber_functions' ) );
-		add_action( 'timber/twig/escapers', array( $self, 'add_timber_escapers' ) );
+        add_action( 'timber/twig/filters', [ $self, 'add_timber_filters' ] );
+		add_action( 'timber/twig/functions', [ $self, 'add_timber_functions' ] );
+		add_action( 'timber/twig/escapers', [ $self, 'add_timber_escapers' ] );
```
gchtr, line 27: If we touch this, maybe we can reduce this to only use `timber/twig`, see https://github.com/timber/timber/issues/2284?

```php
add_action( 'timber/twig', [ $self, 'add_timber_filters' ] );
add_action( 'timber/twig', [ $self, 'add_timber_functions' ] );
add_action( 'timber/twig', [ $self, 'add_timber_escapers' ] );
```

Instead, we could use the `timber/twig/default_functions` filters etc. to control what is added?

lib/Twig.php
```diff
@@ -22,12 +22,173 @@ class Twig {
...
-  }
+	}
+
+	public function get_timber_functions() {
+
+		$postFactory = new PostFactory();
+
+		return apply_filters( 'timber/twig/default_functions', [
+			[
+				'name' => 'action',
```
gchtr, line 38: Did you choose a non-associative array on purpose? What if we would choose the function name as the key? For example:

```php
[
    'action' => [
        'callable' => function( $action_name, ...$args ) {
		    do_action_ref_array( $action_name, $args );
		}
    ]
]
```

This would also make it easier to remove or replace a function using the `timber/twig/default_functions` filter:

```php
add_filter( 'timber/twig/default_functions', function( $functions ) {
    // Remove.
    unset( $functions['action'] );

    // Or replace.
    $functions['action'] = [
        'callable' => 'other',
    ];

    return $functions;
} );
```

gchtr (commented):
I like this a lot for a solution to the underlying problem.

> Sorry about the PR flood. I have some spare time so I can finally propose changes I've been thinking/waiting for a long time.

Keep them coming ❤️! Your suggestions are very valuable!

> This PR also makes use of [native Twig `deprecated` option](https://twig.symfony.com/doc/2.x/advanced.html#deprecated-filters).

Nice! Good to know this option exists.

nlemoine:
> Keep them coming ❤️! Your suggestions are very valuable!

Thanks a lot! 😊  I'm a daily Timber user and I thought some improvements I made over projects could benefit or improve the existing code base. 

> Nice! Good to know this option exists.

I'm not an expert at error handling but it may replace actual deprecations functions?

```php
Helper::deprecated('{{ TimberPost() }}', '{{ get_post() }} or {{ get_posts() }}', '2.0.0');
```

lib/Twig.php
nlemoine: You're absolutely right, both are related and this could be the opportunity to tackle #2284. 

I saw that issue and thought about it too. But I wanted to get your feedback before doing too many changes so I kept this PR in the strict scope of making functions/filters filterable.

lib/Twig.php
nlemoine: I did. I first wrote the function name as key and the callable as value. While checking Twig's docs, I noticed `\Twig\TwigFunction` could have a third `$options` parameter that could be used to deprecate functions and filters and would avoid repeating deprecation and thought that it would be more consistent to have an array that reflects `TwigFunction` arguments.

Remove or adding functions could be quite easy too:
```php
// Remove.
$functions = array_filter($functions, function($fn) {
    return !in_array('action', $fn['name']);
});
// Add
$functions[] = [
    'name' => 'wpautop',
    'callable' => 'wpautop',
];
```

But I didn't consider replacement which would require a bit more code 😅

Maybe you're right, function name as key would be more convenient.

gchtr: Yeah, it’s great that we can pass more parameters.

I just think that finding and filtering out entries that have their key in the array itself is not as straightforward as when we have keys. Yes, there are so many ways you can filter that array, and I think they are all a little too complicated:

- Using `array_filter()`, like you mentioned
- A foreach loop
- Using `array_search()`
- Using `wp_list_filter()`

And of course – I didn’t realize this either – replacing an item is much much harder. :)

gchtr:
> I'm not an expert at error handling but it may replace actual deprecations functions?
> 
> ```
> Helper::deprecated('{{ TimberPost() }}', '{{ get_post() }} or {{ get_posts() }}', '2.0.0');
> ```

How does the error message look like? Have you tried it out?

The `Helper::deprecated()` method plugs in with the `@expectedDeprecated` tags in test DocBlocks, for example here: https://github.com/timber/timber/blob/922bab4effd9ce2d82991171d359142096326f89/tests/test-timber-twig-objects.php#L15-L17. This makes it possible for us test whether functions also appear as deprecated but still work as expected.

I guess that is the only critical thing to consider here.

As I read in https://twig.symfony.com/doc/2.x/recipes.html#displaying-deprecation-notices, to make this work we would maybe have to [install another package](https://github.com/symfony/phpunit-bridge). But that would make it even more complex.

gchtr:
> Thanks a lot! 😊 I'm a daily Timber user and I thought some improvements I made over projects could benefit to other users and improve the existing code base.

I think that’s why we are all here, after all 😊.

nlemoine:
@gchtr PR updated! I've not included #2284, this can be done a bit later.

jarednova:
This is looking really good @nlemoine! (and keep the flood of PRs coming!) Given the amount here that's touching Twig (and Timber) internals, we'll want to make sure this runs through C-I tests before we merge (at the moment our Travis plan + integration are down). 

@gchtr I concur with your review notes. Please let me know when this has your ✅  and we can see where we are on the C-I front at that point.

jarednova (approved):
This gets my vote. I made just one change to cover for a missing `get_users`. Since this touches a lot of stuff that @acobster and @gchtr are both active on I'd like one of their reviews as well before we merge

lib/Twig.php
gchtr: > I saw that issue and thought about it too. But I wanted to get your feedback before doing too many changes so I kept this PR in the strict scope of making functions/filters filterable.

Let’s follow that approach in https://github.com/timber/timber/pull/2419 and leave this PR as it is 👍 .

gchtr (approved):
This looks good to me 👍.

I’d leave it a that state and merge it in (@jarednova, I’ll let you do the honors).

I’ll add further iterations in https://github.com/timber/timber/pull/2419.

acobster (approved):
This looks great! I didn't run them but the tests look good to me.

PR #2411, opened 2021-01-17
Add |grayscale filter for images

jameelmoses:
## Issue
I've often had the need to apply grayscale to images, and this filter allows it to be handled on the server so there are no concerns regarding browser support.

## Solution
Timber already uses several PHP functions for image manipulation in ImageHelper, such as `imagewebp`. This merely adds `imagefilter` into the mix with `IMG_FILTER_GRAYSCALE`. One could argue if you're inclined to accept this PR that we may want to add support for the other `imagefilter`s and their associated parameters, but converting an image to grayscale seemed a common enough use case to try it in isolation.

## Impact
No breaking changes. Simply adds an additional filter.

## Usage Changes

I added the following documentation:

### Converting an image to grayscale

You can use Timber to generate a grayscale image:

```twig
<img src="{{ post.thumbnail.src|grayscale }}" />
```

You can use this in conjunction with other filters.

```twig
<img src="{{ post.thumbnail.src|grayscale|resize(300, 300) }}" />
```

Filters are executed from left to right. You’ll probably want to convert to grayscale before running the resizing, etc.

Unfortunately, it’s not possible to use the `|grayscale` filter with transparent images.

## Considerations
Consider supporting other `imagefilter`s

## Testing
Although I've tested this extensively locally, I didn't have the time to create a test for this. I suppose I could quickly make sure the file is generated and exists, but checking if the image was actually converted to grayscale would require other considerations.


jarednova:
@jameelmoses thanks for the contrib! I've often had the same thought about doing a grayscale conversion. In terms of tests, check out this test example for letterbox:

https://github.com/timber/timber/blob/master/tests/test-timber-image-letterbox.php#L12-L22

... I wrote a helper called `TestTimberImage::checkPixel` to test a pixel in an image against a color. You could use this to see if something like a red pixel went gray. Let us know if you have some time to give it a shot!

lib/Image/Operation/Grayscale.php
```diff
@@ -0,0 +1,85 @@
...
+ * - color to fill transparent zones
+ */
+class Grayscale extends ImageOperation {
+
+	private $quality;
+
+	/**
+	 * @param string $quality  ranges from 0 (worst quality, smaller file) to 100 (best quality, biggest file)
+	 */
+	public function __construct( $quality ) {
```
jarednova, line 21: I think this is the only big Q I have. The arg style we're using elsewhere is based on naming things in an associative array / object, rather than relying on arg order. Thus:

`{{ img.src | grayscale({quality:80}) }}`

... I think there could be some good questions about whether the grayscale filter _should_ consider quality (or if this should be handled by `| tojpg`, etc.). But given that this is how letterbox, resize, etc. all work let's start here.

jameelmoses: Agreed here. Removed.

jameelmoses:
@jarednova Unit test has been added, although alas, I'm on Composer 2.x so I couldn't actually run the test myself 😅

jameelmoses:
One note, if we did want to add transparency support (png, webp, etc.) we would need to use Imagick vs. GD. I'm assuming that's fine for processing image formats that support transparency ala `run_animated_gif()`, but I don't have a particular case right now where I need to worry about transparency.

jameelmoses:
@jarednova any movement on this? i've since added tests.

Levdbas:
Hi everyone, I stumbled upon this PR. Since [97%](https://caniuse.com/?search=filter%3Agrayscale) of browsers worldwide support the grayscale filter. Is this still something we want to implement?

@jarednova

Levdbas:
Closing this pr as part of a bit of householding.

Issue #2386, opened 2020-12-28
Resolve existence of `Term::from()` method


jarednova:
As I was starting to work on #2360 I noticed this. It looks like there may be an older semi-implementation of a factories pattern partially in Timber to resolve or remove in the method `Term::from()`. Based on the other 2.x Factories this is located in the object's `Factory` class (e.g.: `CommentFactory::from`, `PostFactory::from`). Indeed, there's a `TermFactory::from` that seems fully-functional.

Right now, this seems to be non-functional in 2.x, but I want to confirm before I remove. It appears to have been added in #1969. I'll push a branch to demonstrate the applicable tests that should also be pulled

jarednova:
Please see this branch/commit: https://github.com/timber/timber/commit/2138a2b9f99d77a8a179291a2fe810a6ce9d0eb2

acobster:
Yeah I noticed that too on my last sweep but forgot to write it down - I don't see any reason to keep it.

jarednova:
Cool, what I thought! Just wanted to make sure there wasn't any secret grand plan not immediately apparent.

jarednova:
Closed by #2413

PR #2413, opened 2021-01-18
#2386 Remove dormant Term::from() methods

jarednova:
**Ticket**: #2386 

## Issue
There's a dormant `Term::from()` method hiding out. Sort of a primordial implementation of Factories. Of course, we now have real Factories so it can go.

## Solution
Since it was public and `@api` I've deprecated the function

## Impact
I'm guessing it's a narrow number of users who were using this. The deprecation notices/warnings should help them find their way

## Testing
Re-enabled the tests that were connected. Since it's deprecated, I'll make it test the new Factories equivalents

### DIFF
--- a/lib/Term.php
+++ b/lib/Term.php
@@ -89,6 +89,7 @@ class Term extends Core implements CoreInterface, MetaInterface {
 	}
 
 	/**
+	 *
 	 * @deprecated 2.0.0, use TermFactory::from instead.
 	 *
 	 * @param $tid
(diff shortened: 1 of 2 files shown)


acobster (approved):
Looks good to me. I didn't run the tests myself.

jarednova:
Thanks for taking a look @acobster ! I wanted to make sure you had 👀  on the code. I just re-verified locally that all the tests ✅

Issue #342, opened 2014-09-19
"cache" directory in .gitignore?


becw:
The `.gitignore` file has a `cache` line, which means that new files in the `functions/cache/` directory will be ignored by git. Is this intentional? It was added as part of the composer integration in d37c684570b3d1c641b5489dbbe45a099bfd87e2; perhaps it should be more specific?

I noticed this because I pulled this plugin code into the repository for a site I'm working on; because of the `.gitignore`, the `functions/cache/` directory was not checked into my project repository. I've addressed the issue in my project by commenting out the `.gitignore` line, since I'm not sure what specifically it is supposed to apply to.

AlchemyUnited:
seems that maybe the folder should have just been named something less generic (and often ignored). i hear ya. top-level blah blah blah :) but a name tweak would be clear and avoid any confusion. 

next time :)

toonvandeputte:
On case-sensitive file systems (like MacOS) this .gitignore still results in `/lib/Cache` being ignored, which shouldn't happen.
To fix this, the following change to .gitignore does the trick:
```
/cache
!/lib/Cache
```

PR #2416, opened 2021-01-22
Update .gitignore

toonvandeputte:
Do not ignore `/lib/Cache` on case-insensitive filesystems (MacOS)

**Ticket**: #342

## Issue
On case-insensitive file systems, the current `/cache` line in .gitignore, results in `/lib/Cache` being ignored, which shouldn't happen.

## Solution
Add an extra line to un-ignore `/lib/Cache`

## Impact
Some contributors might have modified files in `/lib/Cache` which will pop up as merge conflicts, changed files or otherwise create mild confusion.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ twig-cache/*
 vendor
 build
 /cache
+!/lib/Cache
 *.swp
 *~
 wp-content


jarednova:
Excellent, thanks for that catch @toonvandeputte (and your first contrib)!

PR #2415, opened 2021-01-21
Github Actions tests (#2414)

nlemoine:
I'm not sure to be able to implement back all features from actual CI (pcov, etc. is unknown to me) but this is a starter workflow that runs main tests again on Github Actions.

Issue #2414, opened 2021-01-21
Migrate CI off of Travis to GitHub Actions


jarednova:
## Is your feature request related to a problem? Please describe.
Our project builds have been halted by @travis-ci due to a bug with their system. Their support has been unhelpful at resolving. The tests themselves are fine, we just need a new platform.

## Describe the solution you’d like
At this point, we should migrate to GitHub Actions for further CI test running. While I'm open to other platforms, keeping everything within GitHub seems to be the simplest solution.

## Describe alternatives you’ve considered
I've filed multiple tickets with Travis, tweeted at them and tried contacting their general inquiry email. I've remained polite, but in recent days it's been hard to hide my frustration. Perhaps with the rise of GH actions their business has taken a hit? Who knows, but it's clear it's not a priority for them (indeed, my tickets have been marked "low priority" by their triage).

## Additional context
I haven't done much with GH Actions so I'm marking this as "help wanted" and I mean it! If anyone has been interested in contributing but didn't quite know where to start — start here!

Issue #2407, opened 2021-01-15
twigphp/twig-cache-extension is not maintained anymore


gchtr:
As @nlemoine [mentioned](https://github.com/timber/timber/pull/2349#issuecomment-760306256), the [twigphp/twig-cache-extension](https://github.com/twigphp/twig-cache-extension) is not maintained anymore. Instead, [twigphp/cache-extra](https://github.com/twigphp/cache-extra) should be used.

The only information I found about this in [a tweet](https://twitter.com/fabpot/status/1346406037042171911).

We should check out what this means in more detail, and maybe add the relevant changes for 1.x and 2.x. And possibly bump the supported Twig version for 2.x, because twigphp/cache-extra requires a [minimum Twig version of 2.4](https://github.com/twigphp/cache-extra/blob/3.x/composer.json#L20).

acobster:
@gchtr looks like the current minimum Twig is ``^1.41|^2.10`` on [`master`](https://github.com/timber/timber/blob/master/composer.json#L32), but on [`2.x`](https://github.com/timber/timber/blob/2.x/composer.json#L32) it's already just `^2.10`. I agree we should look into the new extension though. Nice catch!

gchtr:
Okay, because this is connected to #2417, I checked it out a little more.

If we use [twigphp/cache-extra](https://github.com/twigphp/cache-extra), then we’d need to change one or two things in Timber. I didn’t look into it more closely, and I can’t quite assess now what this needs in terms of breaking changes or backwards compatibility. But I think there could be quite some work involved.

**Timber\Loader::_get_cache_extension**

https://github.com/timber/timber/blob/95e6cebf31c3b238f568507bfed029e75b407158/lib/Loader.php#L599-L615

**Timber\Cache\Keygenerator**

https://github.com/timber/timber/blob/95e6cebf31c3b238f568507bfed029e75b407158/lib/Cache/KeyGenerator.php#L1-L7

**Timber\Cache\WPObjectCacheAdapter**

https://github.com/timber/timber/blob/95e6cebf31c3b238f568507bfed029e75b407158/lib/Cache/WPObjectCacheAdapter.php#L1-L6

These are all features that I don’t find in the [twigphp/cache-extra](https://github.com/twigphp/cache-extra), because that new package is about providing a `cache` tag in Twig that allows to cache template fragments.

[...]

PR #2417, opened 2021-01-22
Github Actions tests (2.x)

nlemoine:
Here's the 2.x version of Github Actions (basically the same as #2415). #2414

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,132 @@
+name: Timber Tests
+
+on:
+  push:
+    branches:
+      - '*'
+  pull_request:
+    types: [opened, synchronize, ready_for_review]
+
+jobs:
+  phpunit:
+    runs-on: ubuntu-latest
+
+    services:
+      mysql:
+        image: mariadb:latest
+        env:
+          MYSQL_ALLOW_EMPTY_PASSWORD: true
+          MYSQL_ROOT_PASSWORD:
+          MYSQL_DATABASE: wordpress_test
+        ports:
+          - 3306
+        options: --health-cmd="mysqladmin ping" --health-interval=10s --health-timeout=5s --health-retries=3
+
+    continue-on-error: ${{ matrix.experimental == true }}
+    strategy:
+
+      fail-fast: false
+      matrix:
+        php: ['7.4', '7.3', '7.2'] # , '7.1', '7.0'
+        wp: ['5.3.6', 'latest']
+        multisite: ['0']
+        dependency-version: [prefer-stable] # prefer-lowest
+        webp: [false]
+        coverage: [false]
+        include:
+          # Multisite
+          - php: '7.4'
+            wp: 'latest'
+            dependency-version: 'prefer-stable'
+            multisite: '1'
+            coverage: false
+          # WP Trunk
+          - php: '7.4'
+            wp: 'trunk'
+            dependency-version: 'prefer-stable'
+            multisite: '0'
+            coverage: false
+            experimental: true
+          # Coverage
+          - php: '7.4'
+            wp: 'latest'
+            dependency-version: 'prefer-stable'
+            multisite: '0'
+            coverage: true
+
+    name: PHP ${{ matrix.php }} ${{ matrix.coverage && ' (with coverage)' || '' }} ${{ matrix.extensions && ' (with ' || '' }} ${{ matrix.webp && ' (webp)' || '' }} | WP ${{ matrix.wp }}${{ matrix.multisite == 1 && ' (MS)' || '' }} | ${{ matrix.dependency-version }}
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php }}
+          coverage: ${{ matrix.coverage && 'pcov' || 'none' }}
+          tools: composer:v2
+          extensions: curl, date, dom, gd, iconv, json, libxml, mysql
+          ini-values: pcov.directory=lib
+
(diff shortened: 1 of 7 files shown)


nlemoine:
Hi @jarednova, 

GitHub Actions now fully covers tests made in Travis (https://github.com/nlemoine/timber/actions/runs/742234302). Although I can't really test code coverage (I don't have access to your coveralls account), it should work or at least be easy to fix. 

GA will be a safer choice for the future.

Most importantly, as explained in #2436, I think tests should really be cleaned up/improved quickly in order to safely move forward. Commented variants from the matrix (PHP 7.0/PHP 7.1 / prefer-lowest) have been removed on purpose because they don't pass.

Let me know if it fits your needs.

szepeviktor:
How have you calculated GHA jobs?

TravisCI run these

kép

Multisite: 0, 1, 0, 1

.github/workflows/test.yml
```diff
@@ -0,0 +1,132 @@
...
+            wp: 'latest'
+            dependency-version: 'prefer-stable'
+            multisite: '1'
+            coverage: false
+          # WP Trunk
+          - php: '7.4'
+            wp: 'trunk'
+            dependency-version: 'prefer-stable'
+            multisite: '0'
+            coverage: false
```
szepeviktor, line 48: Be bravely explicit and say

```yaml
coverage: 'pcov'

coverage: 'none'
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,132 @@
...
+            experimental: true
+          # Coverage
+          - php: '7.4'
+            wp: 'latest'
+            dependency-version: 'prefer-stable'
+            multisite: '0'
+            coverage: true
+
+    name: PHP ${{ matrix.php }} ${{ matrix.coverage && ' (with coverage)' || '' }} ${{ matrix.extensions && ' (with ' || '' }} ${{ matrix.webp && ' (webp)' || '' }} | WP ${{ matrix.wp }}${{ matrix.multisite == 1 && ' (MS)' || '' }} | ${{ matrix.dependency-version }}
```
szepeviktor, line 57: `matrix.extensions` is not defined.

.github/workflows/test.yml
```diff
@@ -0,0 +1,132 @@
...
+        dependency-version: [prefer-stable] # prefer-lowest
+        webp: [false]
+        coverage: [false]
+        include:
+          # Multisite
+          - php: '7.4'
+            wp: 'latest'
+            dependency-version: 'prefer-stable'
+            multisite: '1'
+            coverage: false
```
szepeviktor, line 42: `false` is the default, so you may delete this line.

.github/workflows/test.yml
```diff
@@ -0,0 +1,132 @@
...
+        ports:
+          - 3306
+        options: --health-cmd="mysqladmin ping" --health-interval=10s --health-timeout=5s --health-retries=3
+
+    continue-on-error: ${{ matrix.experimental == true }}
+    strategy:
+
+      fail-fast: false
+      matrix:
+        php: ['7.4', '7.3', '7.2'] # , '7.1', '7.0'
```
acobster, line 30: Per https://github.com/timber/timber/issues/1863#issuecomment-760493487, I actually don't think we need to worry about anything other than 7.4 right now. (We can support 8.0 in a future version.)

.github/workflows/test.yml
nlemoine: When used in other steps in the workflow, makes more sense to me to use a boolean here.

.github/workflows/test.yml
szepeviktor: We should run tests on the lowest supported PHP version.

nlemoine: Good news! Done.

.github/workflows/test.yml
nlemoine: Done

.github/workflows/test.yml
nlemoine: Old stuff forgotten. I removed it. But this still might be useful for the future. 

Some tests are properly set up (webp, gd/imagick)

.github/workflows/test.yml
szepeviktor:
```json
  "require": {
    "php": "^7.0",
```

Please at least add 1 test on low PHP version.

acobster: Support for <= 7.3 will be dropped per discussion linked above. The current change is correct.

szepeviktor: # 😲

nlemoine: @szepeviktor Tests actually don't pass on PHP 7.0, neither on 7.1. Maybe should I update `composer.json`?

nlemoine: There's also a serious issue about Twig minimum version that's not tested at all (and don't pass).

szepeviktor: No, it is enough to run on PHP 7.4

acobster: @nlemoine that Twig thing is only an issue on older PHP versions though, right? FWIW, we will require Twig >= 2.10. See: https://github.com/timber/timber/issues/2407

nlemoine: [Not at all](https://github.com/nlemoine/timber/runs/2317169569?check_suite_focus=true). Please consider requiring a more recent version of Twig (at least `^2.4 || ^3.0`). 

From #2436 : 

`twig/cache-extension` has a `twig:^2.4` constraint and it would be a shame to release the brand new Timber 2.0 with an extension that's already not maintained anymore.

acobster: @nlemoine I see, thanks. I agree, and I think fixing #2407 will address your concern. The plan is to replace that extension with twig/cache-extra.

PR #2418, opened 2021-01-22
Replace Gitter refs with GitHub Discssions

jarednova:
## Issue
Gitter is mostly tumbleweeds. Even if it weren't, we don't have the ability to provide the "live" support that Gitter's interface suggests.

## Solution
Remove the refs (and replace with GitHub Discussions an a few spots)

## Considerations
We should also probably post a message and/or turn off the Gitter instance.

## Testing
None needed

### DIFF
--- a/.github/SUPPORT.md
+++ b/.github/SUPPORT.md
@@ -4,6 +4,6 @@ Please post your question to Stack Overflow (http://stackoverflow.com/questions/
 
 ## Other places
 
-You can use [Gitter](https://gitter.im/timber/timber) for user-to-user support and help.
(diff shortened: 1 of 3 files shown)


gchtr (approved):
> We should also probably post a message and/or turn off the Gitter instance.

A final goodbye message sounds good. But I wouldn’t turn it off, because it might still be useful as a reference for some questions people know were answered in there.

While we’re at it, could we add a sticky post to the [Timber support forum on WordPress](https://wordpress.org/support/plugin/timber-library/) that directs people to GitHub discussions as well?

jarednova:
Good call on those! Didn't realize you could sticky a post over at wp.org. Just posted one there along with Gitter (though I couldn't find a sticky/highlight ability over there)

PR #2421, opened 2021-02-06
Get dimensions also for SVG images

vyskoczilova:
**Ticket**: # 

## Issue

Load image dimensions also for SVG files.

## Solution

Modify `Timber\Image\get_dimensions()` function and add new private function `svgs_get_dimensions()` to retrieve dimensions. from svg file.

## Impact

Shouldn't touch performance or backward compatibility, just a new feature.

## Usage Changes

No usage changes accept that getting image.width and image.height will work for SVGs without a need to write own wrapper for it (thanks to @palmiak for helping hand!).

## Testing

I'm not sure how to run your test suite :( but if you'll point me the right way, I could write one.

Otherwise:
* all tests related to getting png/jpg width & height should pass
* test the same with an SVG with height and width parameters only
* test the same with an SVG with view box parameter only

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ docs/output
 /Timber.php
 wordpress
 composer.lock
+.vscode
--- a/lib/Image.php
+++ b/lib/Image.php
@@ -129,12 +129,49 @@ class Image extends Post implements CoreInterface {
 			return $this->get_dimensions_loaded($dim);
 		}
 		if ( file_exists($this->file_loc) && filesize($this->file_loc) ) {
-			list($width, $height) = getimagesize($this->file_loc);
-			$this->_dimensions = array();
-			$this->_dimensions[0] = $width;
-			$this->_dimensions[1] = $height;
+			if ( ImageHelper::is_svg( $this->file_loc ) ) {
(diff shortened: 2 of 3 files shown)


jarednova:
Thanks @vyskoczilova you nailed this on the first try. The test suite is a bit of a beast to get running locally:

Here's the test I added which quickly confirmed your work:

```php
 	function testSVGDimensions() {
		$pid = $this->factory->post->create();
		$filename = self::copyTestImage( 'icon-twitter.svg' );
		$attachment = array( 'post_title' => 'Twitter Icon', 'post_content' => '' );
		$iid = wp_insert_attachment( $attachment, $filename, $pid );
		$image = new TimberImage( $iid );
		$this->assertEquals( 23, $image->width() );
		$this->assertEquals( 20, $image->height() );
	}
```

vyskoczilova:
@jarednova My pleasure :) It was thanks to @palmiak but glad that it helps. Yeah, that's true with tests

Regarding the test - yeah, it's still a bit complicated for me to make them up and running together with WP. Thanks for doing that. However, I would test with two images:

1. SVG with viewbox
2. SVG with width & height

:)

PR #2424, opened 2021-02-13
2.x Improve Date/Time Guide

gchtr:
## Issue

When working with dates that have no times, the time zone is still relevant. When a datetime object is created, `wp_timezone()` should still be used. This was not mentioned in the Date/Time Guide before.

## Solution

Improve section about working with dates that have no times and some other parts that are related to this.

## Impact

None.

## Usage Changes

None.

## Considerations

None.

## Testing

None.


jarednova (approved):
👍  and I learned stuff too!

PR #2430, opened 2021-04-09
Remove PHP 5.6 remnants from Travis config

szepeviktor:
**Ticket**: I have only a speeding ticket.


jarednova:
Agree! FWIW we were already only running on 7.x, but this had a few defensive bits connected to a few `$PREFER_LOWEST` dependencies

jarednova (approved):

PR #2438, opened 2021-04-13
Fix MenuItem creation

acobster:
**Ticket**: None

## Issue

This addresses a couple issues:

1. MenuItemFactory::build() is less specific than it could be, causing PHPStan to complain.
2. MenuItem constructor is still public, allowing for direct instantiation which would forgo Class Maps.

## Impact

None.

## Usage Changes

Consistent with the rest of the Factory usage.

## Considerations

There are other Factories whose return values from `::build()` could also be more specific. These are out of scope here.

## Testing

Existing tests cover this functionality and are currently passing for me.

### DIFF
--- a/lib/Factory/MenuItemFactory.php
+++ b/lib/Factory/MenuItemFactory.php
@@ -49,10 +49,9 @@ class MenuItemFactory {
 		return is_object($item) && $item instanceof WP_Post;
 	}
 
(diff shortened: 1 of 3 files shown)


jarednova (approved):
Polymorphism! More polymorphism!

acobster:
POLYMORPHIZE ALL THE THINGS!

PR #2439, opened 2021-04-13
URLHelper::get_params() return value

acobster:
PHPStan considers implicit null returns a bug:

```
 ------ ---------------------------------------------------------------------------------------------------                                                                                                        
  Line   URLHelper.php                                                                                                                                                                                             
 ------ ---------------------------------------------------------------------------------------------------                                                                                                        
  533    Method Timber\URLHelper::get_params() should return array|string but return statement is missing.                                                                                                         
 ------ --------------------------------------------------------------------------------------------------- 
```

The fix is to always explicitly return, as well as declare the right `@return` type. I did some cleanup for conciseness, too.

No usage changes or other considerations I'm aware of. Tests still pass.

### DIFF
--- a/docs/v2/upgrade-guides/2.0.md
+++ b/docs/v2/upgrade-guides/2.0.md
@@ -1136,6 +1136,10 @@ add_filter( 'timber/twig/environment/options', function( $options ) {
 
 Read more about this in the [Escaping Guide](@todo).
 
+## URLHelper
+
+The `URLHelper::get_params()` method now returns `false` if passed an index that does not exist, whereas before it returned `null` in that case. This is for better consistency with other core API methods and will only affect code that is doing `===` or equivalent checks on return values from this method.
+
 ## Documentation
 
 We added a couple of new guides that you may want to read through in addition to this Upgrade Guide:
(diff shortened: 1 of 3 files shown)


jarednova:
Much cleaner! Just changed `false` => `null` in the function description to match

jarednova (dismissed):

jarednova (approved):

acobster:
whoops! Yeah I had thought to make it `false` to make it a little more consistent with other core functions, but `null` is more appropriate if you're thinking of it as an array. Still on the fence about this though. Thoughts?

jarednova:
Oh roger, that I thought you had settled on `null` based on the code. I think this is one of those 51/49 situations, but `false` would seem to be the more expected value returned (you could imagine using `get_params()` in conditionals — and I'm never confident about how `null` behaves in those situations)

acobster:
OK, I'll change it to `false` then. Bringing it up because it's _technically_ a breaking change, although it'll only break code that's checking `null === URLHelper::get_params(...)` which seems pretty unlikely and edge-casey.

acobster:
OK @jarednova - done. If you could give another once-over that'd be great.

jarednova:
Once-overed! Looks good from here

PR #2440, opened 2021-04-15
Add patch for SVG dimensions

jarednova:
Fixes behavior and test failures with SVG dimension loading in 2.x.

After merging in #2421, I inadvertently created a behavior conflict w how factories detect whether an image is an image (and whether an SVG counts as one). This manually moves some code from 1.x and adds a small modification to `Factory/PostFactory` to resolve this change.

### Considerations
As @gchtr notes, there could be a performance opportunity with `wp_check_filetype` to cache the results so we don't have to hit that function again and again

### DIFF
--- a/lib/Factory/PostFactory.php
+++ b/lib/Factory/PostFactory.php
@@ -98,7 +98,9 @@ class PostFactory {
 
 	protected function is_image(WP_Post $post) {
 		$src   = wp_get_attachment_url( $post->ID );
-		$check = wp_check_filetype( PathHelper::basename( $src ) );
+		$mimes = get_allowed_mime_types();
+		$mimes['svg'] = 'image/svg+xml';
+		$check = wp_check_filetype( PathHelper::basename( $src ), $mimes );
 
 		$extensions = apply_filters( 'timber/post/image_extensions', [
 			'jpg',
@@ -106,6 +108,7 @@ class PostFactory {
 			'jpe',
 			'gif',
 			'png',
+			'svg',
 			'webp',
 		] );
 
--- a/lib/Factory/PostFactory.php
+++ b/lib/Factory/PostFactory.php
@@ -98,7 +98,7 @@ class PostFactory {
 
 	protected function is_image(WP_Post $post) {
 		$src   = wp_get_attachment_url( $post->ID );
-		$mimes = get_allowed_mime_types();
+		$mimes = wp_get_mime_types();
 		$mimes['svg'] = 'image/svg+xml';
 		$check = wp_check_filetype( PathHelper::basename( $src ), $mimes );
 
--- a/lib/Factory/PostFactory.php
+++ b/lib/Factory/PostFactory.php
@@ -99,7 +99,9 @@ class PostFactory {
 	protected function is_image(WP_Post $post) {
 		$src   = wp_get_attachment_url( $post->ID );
 		$mimes = wp_get_mime_types();
+		// Add mime types that Timber recongizes as images, regardless of config
 		$mimes['svg'] = 'image/svg+xml';
+		$mimes['webp'] = 'image/webp';
 		$check = wp_check_filetype( PathHelper::basename( $src ), $mimes );
 
 		$extensions = apply_filters( 'timber/post/image_extensions', [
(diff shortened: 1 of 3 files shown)


lib/Factory/PostFactory.php
```diff
@@ -98,14 +98,17 @@ protected function get_post_class(WP_Post $post) : string {
 
 	protected function is_image(WP_Post $post) {
 		$src   = wp_get_attachment_url( $post->ID );
-		$check = wp_check_filetype( PathHelper::basename( $src ) );
+		$mimes = get_allowed_mime_types();
```
gchtr, line 101: I wonder whether we should use `wp_get_mime_types()` instead of `get_allowed_mime_types()`. I recently found a performance bottleneck with this when a lot images were processed with Timmy (https://github.com/mindkomm/timmy/commit/9812b9860bdd6f1500ad86685fc02b6c2794fefb).

The `get_allowed_mime_types()` includes a check with `user_can()`. But is this necessary for checking the file extension? I think we don’t need to check whether the current user is allowed to upload SVG images, that should have happened long before `is_image()` is called.

We already add the SVG mime type separately on the next line, so it actually doesn’t matter whether SVGs are allowed or not for the check whether an image is an SVG.

Also, quite a few plugins that manage the user rights to allow uploading SVG images hook into the `mime_types` filter, which is used by `wp_check_filetype()`. So maybe it’s worth to cache the result `wp_check_filetype()` so that we can make sure there’s no performance impact.

We could do this for all calls to `wp_check_filetype()` or `wp_check_filetype_and_ext()` in Timber, which would be a good separate issue.

jarednova: Awesome! didn't know about `wp_get_mime_types()` — that's def. the needed function (and why I didn't want to filter the `get_allowed_mime_types` which would interfere w the default security setting).

Agree w making the `wp_check_filetype()` caching a separate issue. For now this will allow 2.x to catch up w `master` (and a fresh reminder on why feature freeze should encompass even "innocent" ones like this)

lib/Image.php
```diff
@@ -127,6 +127,36 @@ protected function get_dimensions_loaded( $dim ) {
...
+			} elseif ( $attributes->width && $attributes->height ) {
+				$width  = (string) $attributes->width;
+				$height = (string) $attributes->height;
+			}
+		}
+
+		return (object) array(
+			'width'  => $width,
+			'height' => $height,
+		);
```
acobster, line 157: I don't know how far down this rabbit hole we want to go, but parsing XML can be expensive and there's a further opportunity for more caching here in case someone calls `get_dimensions()` a bunch of times. The strategy is basically equivalent to how the transient API works:

```php
static $svg_dimensions;
$svg_dimensions = $svg_dimensions ?? [];

$hash = md5($svg);
if (isset($svg_dimensions[$hash])) {
  return $svg_dimensions[$hash];
}

$dimensions = /* ...normal computation... */;

$svg_dimensions[$hash] = $dimensions;

return $dimensions;
```

acobster (commented):
This looks good to me and it sounds like you guys have a pretty good idea of what needs to happen here. I just have one comment about performance wrt. parsing SVGs, take it or leave it. 😃

lib/Image.php
jarednova: This is def. the way to go for a grand plan (with the same strategy for caching regular image width/height which might be similarly expensive). I think we should proceed to catch up on functionality for now. But after, @vyskoczilova — would you be able to do an optimization PR using the guidance here from @acobster ?

jarednova: ... for the next PR we should focus solely on 2.x

vyskoczilova: @jarednova I would have a look at it (but in May, I'm totally overloaded). What's the easiest way to work with 2.0?

jarednova: no big diff there from 1.x! You can pull the branch here: https://github.com/timber/timber/tree/2.x and build a PR off of it. Tests and other processes behave the same. The biggest change is really the "regular user" install where the WP plugin is no longer available (requiring composer installation as a theme dependency). More info here:

https://timber.github.io/docs/v2/installation/

PR #2454, opened 2021-05-26
multisite is_main_network removed

vrushalrt:
**Ticket**: # 

## Issue

On WordPress whenever site on MultiSite network and try to achieve clearing cache programmatically it just clear cache only for the main site not on other sites, I configured WordPress Multisite for multilingual and regional.

And I developed a plugin for Timber-Twig Cache cleaning but whenever use this feature it only deletes cache for the main site in my case default site is EN and so the cache for FR, DE, ES remains undeleted.

## Solution

lib/cache/Cleaner.php 68
```
if ( is_multisite() && is_main_network() ) 
{
    $records .= self::delete_transients_multisite();
}
```
Change to this 
```
if ( is_multisite()  ) 
{
    $records .= self::delete_transients_multisite();
}
```

### DIFF
--- a/lib/Cache/Cleaner.php
+++ b/lib/Cache/Cleaner.php
@@ -65,7 +65,7 @@ class Cleaner {
 
 		// Delete transients from multisite, if configured as such
 
-		if ( is_multisite() && is_main_network() ) {
+		if ( is_multisite() ) {
 
 			$records .= self::delete_transients_multisite();
 		}


jarednova:
Thanks @vrushalrt! Sorry I'm (quite) a bit behind on PRs but catching up here.

I'm 99% sure I understand the issue/solve, but just want to make sure before merging. The original code is quite old  and appears to be copy-and-pasted from another source leaving me a little unsure about why that check is there in the first place (#1190).

It appears the way it worked before is that the transients were only being deleted from the `main_network` site within a multisite install (is that site_id = 1?) and skipping any others (which in your case is all the multilingual/regional variants).

This PR removes that extra check so that transients are cleared regardless of whether it is the main_network site or not. 

Please let me know if any of that is incorrect (or better stated)

Levdbas:
Hi @vrushalrt, is this still an issue you run into? Please let us know if this still needs attention.

Levdbas:
Closing this since this is quite an old PR and we are not going to actively maintain v1

PR #2456, opened 2021-05-28
Package twig/cache-extension is abandoned,Use twig/cache-extra instead ?

davask:
## Issue
When installing Timber I got 
`Package twig/cache-extension is abandoned, you should avoid using it. Use twig/cache-extra instead.`

## Solution
Use `twig/cache-extra` instead as suggested 

## Impact
Not sure

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
     "twig/twig": "^1.41|^2.10",
     "upstatement/routes": "0.8",
     "composer/installers": "~1.0",
-    "twig/cache-extension": "^1.5"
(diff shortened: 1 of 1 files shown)


gchtr:
Related: https://github.com/timber/timber/pull/2456.

gchtr:
I looked into this more closely in https://github.com/timber/timber/issues/2407#issuecomment-910606627 and unfortunately, it’s not a simple as switching to the package mentioned in the notice. That’s why I’m going to have to close this pull request. Let’s continue exploring a solution in #2407.

PR #2463, opened 2021-07-01
Fix the `thumbnail()` method return type

titouanmathis:
## Issue
The returned value of the `Post::thumbnail()` method is considered to be `\Timber\Timber\Image` as we already are in the `Timber` namespace and `Timber\Image` is missing a leading `\`. This is making tools such as PHPStan to fail.

## Solution
The solution is to use the same type as the `Post::get_thumbnail()` method, which is directly using the `Image` class from the current namespace.

## Impact
It will improve compatibility with static analysis tools such as PHPStan.

## Testing
Nothing to update (I think).

### DIFF
--- a/lib/Post.php
+++ b/lib/Post.php
@@ -1570,7 +1570,7 @@ class Post extends Core implements CoreInterface {
 	 * ```twig
 	 * <img src="{{ post.thumbnail.src }}" />
 	 * ```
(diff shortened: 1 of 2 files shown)


jarednova (approved):

jarednova:
Thanks @titouanmathis; sorry this has taken me a while — but this looks good and will merge in!

Issue #1312, opened 2017-01-30
Problem with Timber\Image on Multisite


nexxome:
### Expected behavior
Image Url: http://www.example.com/wp-content/uploads/sites/2/image.jpg

### Actual behavior
Image Url: http://www.example.com/wp-content/uploads/image.jpg

### Steps to reproduce behavior
```php
switch_to_blog(1);

$image1 = new \Timber\Image( $imageFromBlog1 );
echo $image1->path();

restore_current_blog();

$image2 = new \Timber\Image( $imageFromBlog2 );
echo $image2->path();
```

### What version of WordPress, PHP and Timber are you using?
WordPress 4.7.2, PHP 7, Timber 1.2

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Packagist

The Problem occurs because Timber\Image stores the wp_upload_dir() in a static variable, so it doesn't 
recognize when you change the current blog:
```php
public static function wp_upload_dir() {
	static $wp_upload_dir = false;

	if ( !$wp_upload_dir ) {
		$wp_upload_dir = wp_upload_dir();
	}

	return $wp_upload_dir;
}
```

gchtr:
I ran into the same issue and submitted a pull request to fix this in #2478.

gchtr:
I guess I can close this. The fix is in the [1.19.1 release](https://github.com/timber/timber/releases/tag/1.19.1).

PR #2478, opened 2021-09-02
Fix bug when using switch_to_blog() in combination with Timber images

gchtr:
**Ticket**: #1312 

## Issue

`Timber/Image::wp_upload_dir()` caches the result of `wp_upload_dir()`. This causes problems when `switch_to_blog()` is used.

## Solution

Add a check whether we are in a multisite and whether we are in a switched site context. In that case, never return a cached result.

## Impact

None.

## Usage Changes

None.

## Considerations

We could think about whether the result should be cached at all or if we always should use a plain `wp_upload_dir()` call, because of all the side effects that might be involved.

To miminize performance impacts, we could think about passing the second `$create_dir` parameter as `false`, which would skip the checks whether the correct folder already exists. Of course we would have to check whether the check is already made by other functions earlier.

To actually remove the `Timber/Image::wp_upload_dir()` function, we would have to deprecate it first, because it’s public.

And sorry, I missed all the whitespace changes, because I set the settings in [Fork](https://git-fork.com/) to ignore whitespace in commits 😐.

## Testing

Yes.

### DIFF
--- a/lib/Image.php
+++ b/lib/Image.php
@@ -240,10 +240,18 @@ class Image extends Post implements CoreInterface {
 		return $url;
 	}
 
+	/**
+	 * Gets cached version of wp_upload_dir().
+	 *
+	 * Because wp_upload_dir() returns a different result for each site in a multisite, we shouldn’t
+	 * return the cached version when we switched to a different site in a multisite environment.
+	 *
+	 * @todo Deprecate this function in the future and use wp_upload_dir() directly.
+	 */
 	public static function wp_upload_dir() {
 		static $wp_upload_dir = false;
 
-		if ( !$wp_upload_dir ) {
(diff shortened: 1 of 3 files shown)


nlemoine:
Just a tiny performance suggestion: `wp_get_upload_dir` does the same but [doesn't attempt to create directories](https://developer.wordpress.org/reference/functions/wp_get_upload_dir/).

gchtr:
Yes, `wp_get_upload_dir()` would be better, but I didn’t check what’s involved here and whether directories need to be created, e.g. when using the sideloading feature for Timber images, which loads image from URLs into WordPress.

jarednova:
`ms_is_switched`? WordPress + Multisite is full of surprises. Thanks for taking care of this one along with the accompanying test. Now why won't Travis run it?......

jarednova (approved):

PR #2479, opened 2021-09-07
2.x Add hint about checking for Timber’s version

gchtr:
## Issue

When installing and trying out Timber v2, you might have to use checks for a specific Timber version if you still want to support both versions.

## Solution

Add a hint in the Upgrade Guide.

## Impact

None.

## Usage Changes

None.

## Considerations

None.

## Testing

No. just docs.


jarednova (approved):
Seems obvious, but a great tip!

PR #2480, opened 2021-09-14
#2417 Add GitHub Actions for 2.x and update version requirements

gchtr:
**Tickets**: #2417, #2414

## Issue

Since Travis isn’t working working correctly anymore, we’d like to switch to Github Actions (#2417). Updating the dependencies and fixing a couple of tests resparked the discussions about minimum version requirements in Timber (#2436). So in order to get the automated tests working again, we should update the minimum requirements first.

## Solution

This pull request builds on the work by @nlemoine in #2417 (thanks a lot, Nico!), updates the requirements to a working version and fixes some issues.

Instead of adding `^2.4|^3.0` as Twig version constraints, I used `^2.12|^3.0`, because `^2.12` was the lowest version I could make everything work.

## Impact

Automated tests will work again.

There’s a couple of things to consider when upgrading Twig to a higher version. For example, the the `filter` tag won’t work anymore and we have to use the `apply` tag. This is already [reflected in the documentation](https://timber.github.io/docs/v2/guides/wp-integration/#layouts-with-shortcodes). I’ll have to check whether I want to add these changes to the Upgrade Guide.

## Usage Changes

None.

## Considerations

See previous discussions.

## Testing

Yes, now with GitHub Action Tests.

## Todo

- [x] ~Update Upgrade Guide with a short list of Twig deprecations, e.g. the `filter` tag.~ Moved to #2482.


jarednova:
Oh sweet heavens! Retiring Travis has been the little thing eating at me for a long time.

It would also seem that the GH Actions tests run significantly faster than Travis. I'm seeing build times above in the 2-4 min ranges vs. the 6-7 I was used to with Travis.

I think the only open todo is to remove the `travis.yml` file so that Travis stops even attempting (which is triggering the single failed check with this PR). @gchtr do you want to ceremoniously slay the dragon?

jarednova (changes requested):
Remove remaining Travis vestiges
- [ ] Replace equivalent `README.MD` badge
- [ ] Remove `travis.yml`

jarednova (approved):

jarednova:
Oh yeah, backporting should be a lot of fun here. Well for now goodbye and good riddance! (mostly)

PR #2481, opened 2021-09-14
Don't prefer lowest for WebP tests

jarednova:
Current tests against `master` fail due to a setup error during the WebP tests. A starting hypothesis is that the `PREFER_LOWEST` might be the issue

https://app.travis-ci.com/github/timber/timber/jobs/536572310

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@
 
 os: linux
 
-dist: bionic
+dist: xenial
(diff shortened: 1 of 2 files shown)


jarednova:
This resolves current failed builds on `master`. There's no actual "fix" here — just fixing of how Travis runs the tests

PR #2486, opened 2021-09-21
Make cache extension optional and let users choose their own implemen…

nlemoine:
## Issue

Whatever comes out of #2407, I think users should be able to choose their cache implementation.

## Solution

Add a filter to control Timber Cache implementation.

## Impact

None because it defaults to `true`.

## Usage Changes

Users can use that filter to manually register their own cache extension.

## Considerations

use `twig/cache-extra` 😉

### DIFF
--- a/lib/Loader.php
+++ b/lib/Loader.php
@@ -427,7 +427,18 @@ class Loader {
 				return null;
 			}));
 		}
-		$twig->addExtension($this->_get_cache_extension());
+		/**
+		 * Filters the cache extension activation
+		 *
+		 * Allows users to disable the cache extension and use their own
+		 *
+		 * @since 2.0.0
+		 * @param bool $enable_cache_extension
+		 */
+		$enable_cache_extension = apply_filters('timber/enable_cache_extension', true);
+        if ( $enable_cache_extension ) {
+            $twig->addExtension($this->_get_cache_extension());
+        }
 
 		/**
 		 * Filters …


lib/Loader.php
```diff
@@ -427,7 +427,18 @@ public function get_twig() {
...
-		$twig->addExtension($this->_get_cache_extension());
+		/**
+		 * Filters the cache extension activation
+		 *
+		 * Allows users to disable the cache extension and use their own
+		 *
+		 * @since 2.0.0
+		 * @param bool $enable_cache_extension
+		 */
+		$enable_cache_extension = apply_filters('timber/enable_cache_extension', true);
```
gchtr, line 438: Because other hooks that are related to the cache all use the `timber/cache/…` namespace, I propose to use it here as well.

```suggestion
		$enable_cache_extension = apply_filters('timber/cache/enable_extension', true);
```

gchtr (changes requested):
Well, I can’t say too much about everything that’s involved, because I didn’t work often enough with Timber’s caching mechanism. But I feel like this could hold us back even more, despite the very valid points that @nlemoine raises in https://github.com/timber/timber/discussions/2436#discussioncomment-1359271.

For this pull request, I just have a suggestion for the naming of the filter, otherwise this sounds like a good compromise for the moment.

nlemoine:
> I can’t say too much about everything that’s involved, because I didn’t work often enough with Timber’s caching mechanism

Neither do I, but from I've read in the [Twig cache docs](https://twig.symfony.com/doc/3.x/tags/cache.html) and the [Symfony cache docs](https://symfony.com/doc/current/components/cache.html), the cache strategy (e.g. `KeyGeneratorInterface` based strategy) seems kind of outdated now. This goes in the right direction to me, leaving developer the responsability (and freedom) to set their own auto expiring cache keys.

Timber is about bringing Twig (and many other good practices) in WordPress, if some users wants extra features that are outside the Twig base scope, Twig already has reliable/heavily tested extensions you can freely add to your Timber environment.

So why not completely remove this extension? After all, the cache extension is the only one added by default in Timber. There are also some [handy](https://github.com/twigphp/html-extra) [ones](https://github.com/twigphp/string-extra) I use in all my projects. However, I think they belong to the right place: in my personal project `composer.json`. Maybe the cache extension should be too. 

Here's how I see this breaking change:
- Fully remove the cache extension from Timber
- Add twig/cache-extra to the `suggest` field of `composer.json`
- Add a basic example showing how to enable the cache-extra extension (basically what's in [the end of that page](https://twig.symfony.com/doc/3.x/tags/cache.html))
- Release a transient adapter in 2.x future

Otherwise, once 2.0 is out we'll be stuck with the `twig/cache-extension` composer dependency/code and will make everything harder.

gchtr:
> Here's how I see this breaking change for 2.0:
> 
>     * Fully remove the cache extension from Timber
> 
>     * Add twig/cache-extra to the `suggest` field of `composer.json`
> 
>     * Add a basic example showing how to enable the cache-extra extension (basically what's in [the end of that page](https://twig.symfony.com/doc/3.x/tags/cache.html))
> 
>     * Release a transient adapter in 2.x future
> 
> 
> Otherwise, once 2.0 is out we'll be stuck with the `twig/cache-extension` composer dependency/code and will make everything harder.

I think this is a fine plan 👌. Allowing users to add their own extensions is also what we do with Routing. We could still add the relevant documentation for how to add the cache extension.

I just want to hear what @jarednova is thinking about this.

nlemoine:
Thanks @jarednova for coming back on this topic. I'm fine with breaking changes. However, can you still approve this PR? It doesn't introduce a BC and still gives the freedom to opt in for a different cache extension.

Issue #2470, opened 2021-07-19
using 2.0.0-alpha.2, what's the purpose of $context['post'] in context on home page?


usamamuneerchaudhary:
<img width="754" alt="Screenshot 2021-07-19 at 12 41 26 PM" src="https://user-images.githubusercontent.com/24513380/126154854-906730ff-c991-41c6-adad-6b1eb840f2dc.png">

nlemoine:
When you set a static page as home page, WordPress will have a loop filled with `posts` **AND** a page stored in `queried_object`. 

You can access `posts` with `{{ posts }}` (recent posts) and the page (the `queried_object`) with `{{post }}`

usamamuneerchaudhary:
It's actually giving me an error on `setup()` method on bool, I had to revert back to version 1.18. 
<img width="429" alt="Screenshot 2021-07-19 at 1 45 35 PM" src="https://user-images.githubusercontent.com/24513380/126161699-f851dbc4-5cf5-41e8-b6bf-e74e16712525.png">


> When you set a static page as home page, WordPress will have a loop filled with `posts` **AND** a page stored in `queried_object`.
> 
> You can access `posts` with `{{ posts }}` (recent posts) and the page (the `queried_object`) with `{{post }}`

nlemoine:
What's your "reading" settings?

usamamuneerchaudhary:
It's the fresh WP install, I've not changed anything yet.

<img width="673" alt="Screenshot 2021-07-19 at 1 52 34 PM" src="https://user-images.githubusercontent.com/24513380/126162503-9e138372-5179-442f-8d3a-0768fef0becb.png">

gchtr:
@usamamuneerchaudhary Thanks for trying out 2.0.0-alpha.2. I too think this is a bug.

We will probably have to check first what `get_option( 'show_on_front' )` returns.

nlemoine:
Yes, Yoast has a good class for handling current context: https://github.com/Yoast/wordpress-seo/blob/8d93a5ad2f16628bb240c00ecb4e37203e320d00/src/helpers/current-page-helper.php#L40-L62

gchtr:
Hey @usamamuneerchaudhary. I just merged in a fix for this in #2490. This will be in the next 2.x release.

PR #2490, opened 2021-09-23
2.x Fix bug when page_on_front is not set in some cases

gchtr:
**Ticket**: #2470

## Issue

When no `page_on_front` is set, Timber tries to run `setup()` on a post that doesn’t exist, which leads to errors.

## Solution

Check for a post before running setup.

## Impact

Fixes a bug.

## Usage Changes

None.

## Considerations

I thought about whether I should add different checks there, like using `get_option('page_on_front')` or using [@nlemoine’s suggestion](https://github.com/timber/timber/issues/2470#issuecomment-911970671). But then I thought: why think about all the various cases we could run into if Timber already provides an easy way to check whether there’s a post – by using `Timber::get_post()` itself.

## Testing

Yes.

### DIFF
--- a/lib/Timber.php
+++ b/lib/Timber.php
@@ -624,7 +624,7 @@ class Timber {
 	 * ```
 	 */
 	public static function get_term( $term = null ) {
-		
+
 		if (null === $term) {
 			// get the fallback term_id from the current query
 			global $wp_query;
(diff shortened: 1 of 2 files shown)


jarednova (approved):
Nasty little bugger! Good find, catch and test!

PR #2498, opened 2021-09-27
Fix transforms on file and image

nlemoine:
## Issue

When `transform_value` is true, the meta value is passed directly to the factory. If this value is a(n empty) string, it will trigger a notice of incorrect usage of `Timber::get_post` (which is used under the hood by `Timber::get_image`).

## Solution

Return before passing value to the factory when empty.

## Impact

None.

## Considerations

Force types? `Timber::get_post()` only accept `array`/`bool`/`int`.

## Testing

Yes


gchtr (approved):
Thanks for catching those! That looks good to me 👍.

> Force types? `Timber::get_post()` only accept `array`/`bool`/`int`.

What do you mean with that consideration?

PR #2500, opened 2021-09-27
Update compatibility checks

nlemoine:
- Remove the Twig check, 2.x is now a pure PHP package
- Update PHP check

### DIFF
--- a/lib/Timber.php
+++ b/lib/Timber.php
@@ -100,12 +100,6 @@ class Timber {
 		if ( is_admin() || $_SERVER['PHP_SELF'] == '/wp-login.php' ) {
 			return;
 		}
-		if ( version_compare(phpversion(), '5.3.0', '<') && !is_admin() ) {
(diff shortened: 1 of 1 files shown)


nlemoine:
I event think PHP check should be skipped. Since you're supposed to install with composer, it will warn you if the package isn't compatible with your PHP version. No composer package does that check.

gchtr:
> I event think PHP check should be removed. Since you're supposed to install Timber with composer, it will warn you if the package isn't compatible with your PHP version. No composer package does that check.

Yes, that’s what I thought, too, when I checked your changes in this pull request 👍. I think the check can go.

jarednova (dismissed):
Agreed! This is a remnant of the WP install method

jarednova:
@nlemoine this is ✅ ! With your new powers/permissions can you hit the green button?

nlemoine:
Sorry @jarednova, the PHP check removal was missing, can you approve this again? 😅

jarednova (approved):

PR #2503, opened 2021-09-28
Update integration tests:

nlemoine:
Timber now officially supports PHP 8.0/8.1 🎉 
(I mean if @jarednova is ok with this 😅)

PR summary:

- Update PHPUnit 🎉 thanks to https://make.wordpress.org/core/2021/09/27/changes-to-the-wordpress-core-php-test-suite/
- Update/fix tests
- Test Timber on PHP 8.0/8.1 🎉

Side notes:
- PHPUnit has been removed as a dependency to let `yoast/phpunit-polyfills` handle the PHPUnit version to use as advised in link above
- You can now use PHPUnit ^9.0 assertions
- If you need to run tests on an older WP version (< 5.9), use `composer req phpunit/phpunit:^7.5 -W --dev` but don't commit changes made to `composer.json`
- When installing test suite, always use a branch (`5.8`, not `latest`, not `5.7.1`): `bash bin/install-wp-tests.sh wordpress_test <db_user> <db_pass> <db_host> 5.8`
- Read the [WordPress Core blog post](https://make.wordpress.org/core/2021/09/27/changes-to-the-wordpress-core-php-test-suite/) above for further details

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,7 +28,7 @@ jobs:
       fail-fast: false
       matrix:
         php: ['7.4']
-        wp: ['5.3.6', 'latest']
+        wp: ['5.3', '5.8'] # Always use branches not tags (e.g. 5.3 not 5.3.6), new test suite is only available in branches
         multisite: ['0', '1']
         dependency-version: [prefer-stable] # prefer-lowest
         webp: [false]
@@ -41,15 +41,21 @@ jobs:
             dependency-version: 'prefer-stable'
             multisite: '0'
             experimental: true
+          # PHP 8.0
+          - php: '8.0'
+            wp: 'trunk'
+            dependency-version: 'prefer-stable'
+            multisite: '0'
+            experimental: true
           # PHP with Imagick
           - php: '7.4'
-            wp: 'latest'
+            wp: '5.8'
             dependency-version: 'prefer-stable'
             multisite: '0'
             extensions: 'imagick'
           # Coverage
           - php: '7.4'
-            wp: 'latest'
+            wp: '5.8'
             dependency-version: 'prefer-stable'
             multisite: '0'
             coverage: true
@@ -99,6 +105,10 @@ jobs:
       - name: Install composer dependencies
         run: composer update --${{ matrix.dependency-version }} --prefer-dist --no-progress --no-interaction
 
+      - name: Downgrade PHPUnit to ^7.5 when WP < 5.9
+        if: ${{ matrix.wp < '5.9' }}
+        run: composer req phpunit/phpunit:^7.5 -W --dev
+
       - name: Install tests
         run: bash bin/install-wp-tests.sh wordpress_test root '' 127.0.0.1:${{ job.services.mysql.ports['3306'] }} ${{ matrix.wp }} true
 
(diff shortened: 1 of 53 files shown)


gchtr (dismissed):
Wow, Nico, this is amazing! 🙌🚀 I read that [post](https://make.wordpress.org/core/2021/09/27/changes-to-the-wordpress-core-php-test-suite/) and thought that is going to be quite some work. And you did it!  Thank you so much for putting in the effort.

The only thing that came to my mind is that the `latest` keyword in `https://github.com/timber/timber/blob/2.x/bin/install-wp-tests.sh` might not work either. But it doesn’t look like they changed it in https://github.com/wp-cli/scaffold-command/blob/master/templates/install-wp-tests.sh either, yet.

And I wonder whether we need to update https://github.com/timber/timber/blob/2.x/docs/v2/guides/testing.md. Maybe @jarednova could have a look at that.

Both comments could be handled in separate pull requests and shouldn’t hold us back from merging this in so you can continue working, Nico :).

nlemoine:
> thought that is going to be quite some work

It was not that, the only thing that took me while to understand is that the polyfilled test suite is only included in branches, not tags...

> The only thing that came to my mind is that the latest keyword 

This is only temporary, the [time that scaffold tests](https://github.com/wp-cli/scaffold-command/issues/299) are adapted to the new test suite. We could also modify `install-wp-tests.sh` so `latest` matches the latest branch, not tag.

But I think this is fine for now, I'll update these tests when a new WP version comes out or when scaffolded are updates.

> I wonder whether we need to update

I'm pretty sure they should be :) I don't install VVV anymore and it works fine if you have everything setup.

And maybe in the future, switch to https://github.com/Yoast/wp-test-utils which eases tests a lot. And split real unit tests from integration tests to speed thing up. But one thing at a time :)

nlemoine:
Since PHP 8.1 is out soon... I added some fixes so everything passes tests.

Just a tiny question, I had trouble with image resizing when Imagick is available and crop is set to 'default'. What does 'default' means when cropping? I found nothing about it: https://github.com/timber/timber/blob/master/lib/Image/Operation/Resize.php#L119-L149

jarednova:
So this appears to be a leftover highly opinionated artifact of what I thought a good "default" crop should be circa 2015. The key lines are here ...

https://github.com/timber/timber/blob/master/lib/Image/Operation/Resize.php#L116-L117

... where it centers on the X axis and then 1/6 from the top on the Y axis. A perfectly sensible crop (this would be good for cropping to where faces generally are in photos) — but I can see why we'd want to make that controllable at the least. I'm going to say that for now, rearchitecting crops should be a 2.x future task.

I'll review this over the next couple days, formal PHP 8 support is great!!!

jarednova:
@nlemoine great work! as @gchtr said: `thought that is going to be quite some work`!!!! I'm sure it was.

Thanks for linking to the WP.org blog post, that answered all the big Qs I had in my review. 

As @gchtr notes, this is a good time to get to those `testing.md` docs, or at least not tie them so directly to VVV. I've now got an Apple M1 that I've been meaning to get things set up with. What's your basic test setup? Docker? Or something else?

jarednova (approved):

nlemoine:
@jarednova I'm still on a mid-2012 macbook pro (non retina version)... Because this is the last one where you can still easily replace/upgrade your hard drive (I replaced the optical drive with a 2To SSD drive), memory and battery. It's running pretty good but it's getting white hairs and I try to avoid VM when I can. So basically, my setup is only `brew install php && brew install mariadb`.

The ideal way to run tests is [act](https://github.com/nektos/act) but [it's not supporting services yet](https://github.com/nektos/act/issues/173). Once it does, running tests will be quite easy and more importantly, we won't have to handle local env and Github Actions env.

jarednova:
Oh wow, you're going old school! Thanks for the links to act — it's great to have a way to execute those actions locally. I can't tell you how many hours I've spent pushing Travis commits just to see how it would use the build script.

Issue #2496, opened 2021-09-27
Menu and MenuItem factory inconsistencies


nlemoine:
Hi, 

I'm starting to play around with Timber 2.x and noticed some inconsistencies on the menu API.

**Inconsistency in the filter regarding other same kinds of filter**

https://github.com/timber/timber/blob/996a3244100a77ff3349f74d8f7c736f2213de51/lib/Factory/MenuFactory.php#L108-L112

This filter is taken from `post` and `term` except that it does not work in a similar way. Instead of returning a map, it returns a string. Maybe the name is wrong but as a user, I would except that: 
- a filter containing the word `classmap` should return an array
- all `timber/{object}/classmap` should behave the same

Same applies for the `timber/menuitem/classmap` filter.

**Inconsistency in the filter regarding filter argument**

Those filters (`timber/menu/classmap` and `timber/menuitem/classmap`) also have more arguments but in the first case (menu), it receives a `WP_Term` object whereas the 'menuitem' filter gets a `Timber\Post` object. I think they should get either WP core objects (`WP_*`) or Timber objects but not both mixed up.

gchtr:
Hmmm, yeah, this is an inconsistency. Discussion around this started in https://github.com/timber/timber/issues/2089#issuecomment-538764945, if I can remember correctly.

> as a user, I would except that:
> 
>     * a filter containing the word `classmap` should return an array

I didn’t see that before, but I agree with you.

Because menus and menu items are not really tied to a core type in WordPress, we can’t really apply class maps here. The same goes for users by there, for which we have a `timber/user/classmap` filter.

I guess we could rename the filters:

- `timber/user/classmap` → `timber/user/class`
- `timber/menu/classmap` → `timber/menu/class`
- `timber/menuitem/classmap` → `timber/menuitem/class`

> I think they should get either WP core objects (`WP_*`) or Timber objects but not both mixed up.

We’re merely passing in here what we already know. If we can provide a Timber object, because an object relies on one, we pass a Timber object. Otherwise, we only have a WordPress core object. Yeah, so being consistent here would mean we would have to pass a WordPress object to the `timber/menuitem/classmap` filter.

What if we passed in what we have as a first argument, and the Timber object as a second one?

**Before**

```php
$class = apply_filters( 'timber/menuitem/classmap', MenuItem::class, $item, $menu);
```

**After**

```php
$class = apply_filters( 'timber/menuitem/class', MenuItem::class, $wp_object, $timber_object, $timber_menu );
```

[...]

PR #2513, opened 2021-10-20
Add a timber/{object}/class to all objects types

nlemoine:
Since the `timber/{object}/class` exists in `Menu`, `MenuItem` and `User` factories, there's no reason it doesn't exist for other entities (`Post`, `Term`, `Comment`). It would more freedom to select a class on other data than the chosen ones for classmaps (`post_type` for `Post`, `taxonomy` for `Term`, `menu location` for `Menu`, etc.). 

See examples in the filters docblock and [my comment here](https://github.com/timber/timber/issues/2496#issuecomment-929089709). 

## Testing

Yes

### DIFF
--- a/lib/Factory/CommentFactory.php
+++ b/lib/Factory/CommentFactory.php
@@ -78,10 +78,10 @@ class CommentFactory {
 		$class = $class ?? Comment::class;
 
 		/**
-		 * Filters the comment class based on your custom criterias.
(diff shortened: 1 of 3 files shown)


jarednova:
Looks good @nlemoine! I just tweaked some of the docblock wording to make sure things were as clear as possible. You want to mash that green button?

Issue #2292, opened 2020-07-24
Implement new PagesMenu API


acobster:
Per discussion on #2089, implement the following:

- [x] A `Timber\PagesMenu` class.
- [x] A `timber/pages_menu/classmap` filter.
- [x] A `Timber::get_pages_menu()` function.
- [x] A `Timber\PagesMenuFactory` class - used under the hood in `::get_pages_menu()`, calls the `timber/pages_menu/classmap` filter.
- [x] Make `Menu` and `MenuItem` constructors protected.
- [x] Draft the relevant documentation as well as the notes in the Upgrade Guide for this.

There is already a `test-timber-pages-menu.php` test file. We may want to flesh this out with a few more tests, or redesign some of the existing tests.

gchtr:
I think the documentation is already done for this with #2254.

gchtr:
Implemented in #2562.

Issue #1820, opened 2018-10-17
Consider improving compatibility with wp_nav_menu()


gchtr:
This ticket comes out of a discussion in #1814.

Most normal WordPress themes make use of the `wp_nav_menu()` function to render a navigation menu. Many plugins hook into this function to add menu items or change properties on existing items, like CSS classes. The `wp_nav_menu()` function uses 7 filters that Timber doesn’t implement. For better compatibility with plugins, we could consider adding support for some of these. For example, in #1529, the [WP Menu Cart](https://wordpress.org/plugins/wp-menu-cart/) plugin is mentioned, which hooks into the `wp_nav_menu_{$menu->slug}_items` filter.

1. [x] `wp_nav_menu_args` – Filters the arguments used to display a navigation menu.  
	↬ Will not be needed as we don’t have menu args in Timber.
2. [x] `pre_wp_nav_menu` – Filters whether to short-circuit the wp_nav_menu() output.  
	↬ Probably won’t be needed.
3. [x] `wp_nav_menu_container_allowedtags` – Filters the list of HTML tags that are valid for use as menu containers.  
	↬ Will not be needed.
4. [x] `wp_nav_menu_objects` – Filters the sorted list of menu item objects before generating the menu's HTML.  
	↬ Will be fixed in #1814.
5. [ ] `wp_nav_menu_items` – Filters the HTML list content for navigation menus.  
	↬ Support for this would be nice.
6. [ ] `wp_nav_menu_{$menu->slug}_items` – Filters the HTML list content for a specific navigation menu.   
	↬ Support for this would be nice.
7. [ ] `wp_nav_menu` – Filters the HTML content for navigation menus.  
	↬ Pretty much the same as `wp_nav_menu_items`

Because the last three filters filter HTML and in Timber, we handle HTML through Twig, it might be a little tricky.

## Approach 1

We could use [a proper Twig example](https://timber.github.io/docs/guides/menus/#displaying-the-menu-items). If I’m not totally mistaken, the following might work. We’d have to provide the `default_args` and the `slug` as properties of `Timber\Menu`.

```twig
{% for item in menu.items %}
    <li class="nav-main-item {{ item.classes|join(' ') }}">
        <a class="nav-main-link" href="{{ item.link }}">{{ item.title }}</a>
        {% if item.children %}
            <ul class="nav-drop">
                {% for child in item.children %}
                    <li class="nav-drop-item">
                        <a href="{{ child.link }}">{{ child.title }}</a>
                    </li>
[...]

Issue #1440, opened 2017-06-01
Timber\Menu() returns default menu when using non-existent menu slug as parameter


onnimonni:
### Expected behavior
If I try to create a menu item with non-existent menu slug like here:
```php
<?php
// Get non-existent menu item
// Real menu slug was 'mobile'
$context['mobilemenu'] = new Timber\Menu("mobilemenu");
```

I would want to see error or no menu or a notice in the error log.

Relevant lines are here:
https://github.com/timber/timber/blob/4d19fa9bba16412bb012280cac714bdfb6be550d/lib/Menu.php#L82-L100

### Actual behavior
Current implementation just silently returns the default menu which makes debugging menu errors harder.

### What version of WordPress, PHP and Timber are you using?
php7.1, WP 4.7.5

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Composer

chrisgherbert:
I think this is expected behavior (to match WordPress standard behavior), but I absolutely hate it.

shaneoliver:
Conditionally add to context when the menu exists
```php
if( has_nav_menu( 'mobilemenu' ) ) {
    $context['mobilemenu'] = new Timber\Menu("mobilemenu");
}
```
Check the menu exists on the context before rendering
```twig
{% if mobilemenu%} 
   {% include "path/to/menu.twig" with {'menu': mobilemenu.get_items} %}
{% endif %}
```

ozanyolyapar:
Anti-pattern like behavior. Documentation says the following:

> Or pass nothing. This is good if you have only one menu. In that case Timber will just grab what you got.

Well, this is not okay. I got 8 contextual menus on a complex site. There is possibilities like @shaneoliver pointed out, but it is neither documented if the slug doesn't exist, nor it behaves like expected from an API as a dev.

nakashu:
I am willing to make an PR, if it will be clarified if the core team wants to add this change.
Personally I think current behavior is incorrect and can make bug in production if a needed menu is deleted - suddenly showing first default one.

Possible solutions
- add another option to constuctor `new Menu('slug', ['strict' => true])` or `mustExist`
- change current behavior to return Menu without items / empty

jarednova:
@nakashu — thank you so much for running that by before diving into a PR (nothing pains me more than saying "no thank you" to an otherwise well-meaning PR). 

[...]

PR #2517, opened 2021-11-10
Menu API enhancements

nlemoine:


lib/Factory/MenuFactory.php
```diff
@@ -94,20 +148,18 @@ protected function from_string($ident, $args) {
...
 
 		if ($obj instanceof WP_Term) {
+			$args['menu'] = $obj;
 			return $this->build($obj, $args);
 		}
 
-		throw new \InvalidArgumentException(sprintf(
-			'Expected an instance of Timber\CoreInterface or WP_Term, got %s',
-			get_class($obj)
-		));
```
nlemoine, line 110: Don't throw exception for consistency with other "from" mtehods.

lib/Menu.php
```diff
@@ -95,10 +94,148 @@ class Menu extends CoreEntity {
...
+
+		$nav_menu->items = $sorted_menu_items;
+		unset( $sorted_menu_items );
+
+		/**
+		 * Since Timber doesn't use HTML, serialize the menu object to provide a cacheable string
+		 *
+		 * @see wp_nav_menu()
+		 */
+		$_nav_menu = apply_filters( 'wp_nav_menu', serialize($nav_menu), $args );
```
nlemoine, line 236: Caching menu plugins are [misusing this filter](https://github.com/inpsyde/menu-cache/blob/master/index.php#L43-L44) to cache menus and excepts a string here, provide a serialized menu object.

lib/Menu.php
```diff
@@ -114,97 +251,56 @@ public static function build(WP_Term $wp_term, array $args = []) : Menu {
 	 *                            supported which says how many levels of hierarchy should be
 	 *                            included in the menu. Default `0`, which is all levels.
 	 */
-	public function __construct( $slug = 0, $args = array() ) {
+	protected function __construct( ?WP_term $term, array $args) {
```
nlemoine, line 254: Make constructor protected to force factory usage.

lib/MenuItem.php
```diff
@@ -264,36 +290,27 @@ public function import_classes( $data ) {
...
-		 *                                  walker. Instead, you get the options that were used to
-		 *                                  create the `Timber\Menu` object.
-		 * @param int              $depth   Depth of menu item.
+		 * @see Walker_Nav_Menu
 		 */
 		$this->classes = apply_filters(
 			'nav_menu_css_class',
 			$this->classes,
-			$this,
+			$this->core_object,
```
nlemoine, line 308: Providing a Timber object in a core filter should not happen.

lib/MenuItem.php
```diff
@@ -510,7 +527,11 @@ public function path() {
 	 */
 	public function title() {
 		if ( isset($this->__title) ) {
-			return $this->__title;
+			/**
+			 * @see Walker_Nav_Menu::start_el()
+			 */
+			$title = apply_filters( 'nav_menu_item_title', $this->__title, $this->core_object, $this->menu->args ? $this->menu->args : new \stdClass, $this->level );
```
gchtr, line 533: Nice 😎

lib/Timber.php
```diff
@@ -870,27 +870,68 @@ public static function get_user_by( string $field, $value ) {
...
 
 	/**
 	 * @todo implement PagesMenuFactory
 	 */
-	public static function get_pages_menu( array $pages = [], array $options = [] ) {
-		$menu = new Menu( $pages, $options );
-		$menu->init_as_page_menu();
-		return $menu;
+	public static function get_pages_menu( array $args = [] ) {
+		return Menu::build( null, $args );
```
gchtr, line 934: I commented it before, but I would leave this as a separate function. See https://github.com/timber/timber/issues/2292 and https://github.com/timber/timber/issues/2089#issuecomment-628977116.

lib/Timber.php
```diff
@@ -870,27 +870,68 @@ public static function get_user_by( string $field, $value ) {
...
+	 * $menu = Timber::get_menu_by( 'slug', 'primary-menu' );
+	 * ```
+	 *
+	 * @param string     $field The name of the field to retrieve the menu with. One of: `id`,
+	 *                          `ID`, `term_id`, `slug`, `name` or `location`.
+	 * @param int|string $value The value to search for by `$field`.
+	 *
+	 * @return \Timber\Menu|null
+	 */
+	public static function get_menu_by(string $field, $value, array $args = []) : ?Menu {
```
gchtr, line 906: I love this! 💚

gchtr (changes requested):
I think you’re way deeper into the menu logic than I will ever be. Thanks a lot for all your work! There’s lots of optimizations in there. I also like the part about the caching compatibility with other plugins.

I saw a couple of things that I think we discussed before. I added comments right in the code.

lib/Factory/MenuFactory.php
```diff
@@ -4,88 +4,142 @@
...
  */
 class MenuFactory {
-	public function from($params, array $args = []) {
-		if ($params === 0) {
-			// We know no nav_menu term exists with ID 0,
-			// so just look at the existing terms in the database.
+
+	/**
+	 * Try to get a menu by all means available in an order that matches the
+	 * most common use cases. It will fallback on the first menu found if no parameters are provided
```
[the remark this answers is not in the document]
nlemoine: The idea behind this was to mimic the `wp_nav_menu` behaviour which tries to find a menu by all means available.

Honestly, I'll personnaly never use that method so I'm fine with either option, just thought it would bring more parity with `wp_nav_menu`.

lib/Factory/MenuItemFactory.php
```diff
@@ -21,7 +21,7 @@ public function from($item, Menu $menu) {
 			return $this->build($item, $menu);
 		}
 
-		return false;
```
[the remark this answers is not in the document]
nlemoine: In my daily code practice, I find returning nullable types very handy. It allows null coalescing operator usage.

Let me know what's your option on this.

lib/Menu.php
```diff
@@ -95,10 +94,148 @@ class Menu extends CoreEntity {
...
+			try {
+				$nav_menu = unserialize($nav_menu);
+				if( $nav_menu instanceof Menu ) {
+					return $nav_menu;
+				}
+			} catch(\Throwable $e) {
+			}
+		}
+
+		// No menu term provided, get the pages menu
```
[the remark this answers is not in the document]
nlemoine: It follows the same idea mentionned in https://github.com/timber/timber/pull/2517#discussion_r762427431. `Menu::build` is basically `wp_nav_menu` adapted to Timber.

I read those threads about the PagesMenu API. Looks fine but I stayed in the current implementation scope. Those menus already bugged me enough :) I'll pass on this one.

lib/MenuItem.php
nlemoine: I have mixed feelings about this and some parts of https://github.com/timber/timber/issues/1820#issue-370922635. 

Timber is actually trying to get rid of those filters everywhere to handle HTML. And menu are a noticeably awful example of this practice. I think we shouldn't put too much effort in being 100% compliant with all those filters (`wp_nav_menu_items`, `wp_nav_menu_{$menu->slug}_items`, etc.). It might not be possible or it might come at too much code costs. 

If some user wants full compatibility with all core menu filters, skipping Timber implemented menus and using `wp_nav_menu` is quite easy. I actually had an idea for this that I didn't commit in that PR that could be a fair compromise between all Timber menus VS `wp_nav_menu` only.

A `__toString` method on the `Menu` class that will dump the HTML menu using walker and all those core stuff:

```php
public function __toString() {
	static $menu_id_slugs = array();

	$args = $this->args;

	$items = '';
	$nav_menu = '';
	$show_container = false;

	if ( $args->container ) {
		/**
			* Filters the list of HTML tags that are valid for use as menu containers.
			*
			* @since 3.0.0
			*
			* @param string[] $tags The acceptable HTML tags for use as menu containers.
			*                       Default is array containing 'div' and 'nav'.
			*/
		$allowed_tags = apply_filters( 'wp_nav_menu_container_allowedtags', array( 'div', 'nav' ) );

		if ( is_string( $args->container ) && in_array( $args->container, $allowed_tags, true ) ) {
			$show_container = true;
			$class          = $args->container_class ? ' class="' . esc_attr( $args->container_class ) . '"' : ' class="menu-' . $this->slug . '-container"';
			$id             = $args->container_id ? ' id="' . esc_attr( $args->container_id ) . '"' : '';
			$aria_label     = ( 'nav' === $args->container && $args->container_aria_label ) ? ' aria-label="' . esc_attr( $args->container_aria_label ) . '"' : '';
			$nav_menu      .= '<' . $args->container . $id . $class . $aria_label . '>';
		}
	}

	$items .= walk_nav_menu_tree( $this->sorted_menu_items, $args->depth, $args );

	// Attributes.
	if ( ! empty( $args->menu_id ) ) {
		$wrap_id = $args->menu_id;
	} else {
		$wrap_id = 'menu-' . $this->slug;

		while ( in_array( $wrap_id, $menu_id_slugs, true ) ) {
			if ( preg_match( '#-(\d+)$#', $wrap_id, $matches ) ) {
				$wrap_id = preg_replace( '#-(\d+)$#', '-' . ++$matches[1], $wrap_id );
			} else {
				$wrap_id = $wrap_id . '-1';
			}
		}
	}
	$menu_id_slugs[] = $wrap_id;

	$wrap_class = $args->menu_class ? $args->menu_class : '';

	$nav_menu .= sprintf( $args->items_wrap, esc_attr( $wrap_id ), esc_attr( $wrap_class ), $items );
	if ( $show_container ) {
		$nav_menu .= '</' . $args->container . '>';
	}

	return $nav_menu;
}
```

So you could simply do `{{ header_menu }}` in your Twig template while still keeping the ability to use `Menu` objects/factories. 

Let me know if think this could be handy.

lib/Factory/MenuItemFactory.php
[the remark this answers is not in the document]
gchtr: Let’s use `null`. Being able to use the null coalescing operator is a big plus. In the meantime, I also saw it as a standard in a couple of other projects. WordPress’s `get_post()` function also returns `null` if no post can be found.

The menu functions are just not the only place we need to change this, though. I’ll create a separate issue for this.

lib/Factory/MenuFactory.php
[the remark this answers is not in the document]
gchtr: Yeah, that’s the way to usually do it. But because this has led to unexpected behaviors and problems, I’d like to follow what we discussed https://github.com/timber/timber/issues/2089#issuecomment-586739839 and implement the changes listed in #2292.

lib/MenuItem.php
gchtr: That sounds like a great idea! I just wouldn’t use it with the `__toString()` method, but maybe add a `wp_nav_menu()` method?

```twig
{{ header_menu.wp_nav_menu }}
```

gchtr (approved):
I don’t want to keep this back much longer. There are lots of good things in here. I take three todos out of this:

- I’d like to just leave in the pages menu logic and refactor it with a separate issue (#2292).
- Then there’s the idea about the feature parity with `wp_nav_menu()`, which we found is quite hard to accomplish. A separate method that builds a menu using the nav menu walker is quite a good idea for people who need to use it for compatibility reasons (see https://github.com/timber/timber/pull/2517#discussion_r762430001).
- And finally, to return `null` instead of `false` from the factory methods so that they work with the null coalescing operator.

Once all the tests go green we can merge this in. Thank you so much for your work @nlemoine. Tell me if you still find the time to get this pull request ready more merging or if you’d like to pass it on.

jarednova:
Just want to flag the two failing tests we have remaining ...

```
There were 2 failures:

1) TestMenuItemFactory::testMenuGarbageInGarbageOut
Failed asserting that false is null.

Error: /home/runner/work/timber/timber/tests/test-menu-item-factory.php:53
phpvfscomposer:///home/runner/work/timber/timber/vendor/phpunit/phpunit/phpunit:60

2) TestTimberTerm::testTermWithNativeMetaNotExisting
Failed asserting that 'qux' is null.

Error: /home/runner/work/timber/timber/tests/test-timber-term.php:4[49](https://github.com/timber/timber/runs/5343496381?check_suite_focus=true#step:15:49)
phpvfscomposer:///home/runner/work/timber/timber/vendor/phpunit/phpunit/phpunit:[60](https://github.com/timber/timber/runs/5343496381?check_suite_focus=true#step:15:60)```

File: lib/MenuItem.php (hunk 5 above)
Comment (nlemoine) [reply lib/MenuItem.php:533]: I stand by the `__toString` method which makes more sense for the following reasons:
- The HTML menu is literally the string representation of the menu object
- The name `wp_nav_menu` is misleading, users could think it could take the same arguments than the native `wp_nav_menu` WordPress method, which is not

File: lib/Timber.php (hunk 6 above)
Comment (nlemoine) [reply lib/Timber.php:937]: I saw that comment a couple of times :) I have no idea what that `PagesMenu` is about and [will leave that to someone else](https://github.com/timber/timber/pull/2517#discussion_r762428881) if you're ok with that.

**Comment by nlemoine** [comment]
@jarednova If fixed those, ACF finally rolled back the [`get_field` change](https://www.advancedcustomfields.com/changelog/).

@gchtr I added everything but the part concerning the PagesMenu API. Don't mind the commit messages, I'll squash everything with a cleaner one once validated :)

File: lib/MenuItem.php [hunk 11]
```diff
@@ -107,8 +104,16 @@ public static function build( $data, Menu $menu = null ) : self {
...
 	protected function __construct( $data, $menu = null ) {
+		$this->core_object = $data;
 		$this->menu = $menu;
+
 		$data       = (object) $data;
+
+		// Items from WordPress can come with a $title property which conflicts with methods
+		$data->__title = $data->title;
+		// @todo Use case for unsetting title?
+		// unset($data->title);
```
nlemoine, line 115: @gchtr @jarednova Is it really needed? This prevents nav walker to print menu titles.

lib/Factory/MenuItemFactory.php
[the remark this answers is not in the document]
gchtr: See #2561.

lib/Menu.php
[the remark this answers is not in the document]
gchtr: > I read those threads about the PagesMenu API. Looks fine but I stayed in the current implementation scope. Those menus already bugged me enough :) I'll pass on this one.

All good. We can handle this in another pull request. I’ll try to see what I can do.

lib/MenuItem.php
gchtr: @nlemoine I see. Yeah, that makes sense. Let’s go with `__toString()` then 🙂.

lib/MenuItem.php
gchtr: Found the issue where the `$__title` property was introduced: https://github.com/timber/timber/issues/334.

The problem was when an ACF field named `title` existed. These kind of conflicts lead to the introduction of using `meta()` as the official way to get meta data from an object instead of directly using the field’s name. So if you have an ACF field for a menu item called title and you want to access it, you should use `{{ item.meta('title') }}`  instead of `{{ item.title }}`, because Timber wouldn’t know what to use.

We can definitely get rid of the `$__title` property and use `$title` directly.

Thanks for flagging this, @nlemoine!

lib/Factory/MenuFactory.php
```diff
@@ -4,88 +4,147 @@
...
+				$menu = $this->from_slug($params, $args);
+			}
+			if(!$menu) {
+				$menu = $this->from_name($params, $args);
+			}
 		}
 
-		// Fall back on the first nav_menu term we find.
-		return $this->from_nav_menu_terms($args);
+		if(!$menu) {
```
gchtr, line 52: While working on the Pages Menu functionality (#2562) I stumbled over this again. And I propose to remove this if-statement.

If we leave this out and just return `null` when no menu couldn’t be determined up until here, then we wouldn’t run into unexpected behaviors where the wrong menu is returned (see #1440).

WordPress uses the following functionality in `wp_nav_menu()`:

```php
// Get the first menu that has items if we still can't find a menu.
if ( ! $menu && ! $args->theme_location ) {
    $menus = wp_get_nav_menus();
    foreach ( $menus as $menu_maybe ) {
        $menu_items = wp_get_nav_menu_items( $menu_maybe->term_id, array( 'update_post_term_cache' => false ) );
        if ( $menu_items ) {
            $menu = $menu_maybe;
            break;
        }
    }
}
```

We replicate this functionality in `MenuFactory::from_nav_menu_terms()`. This is fine if no parameter is passed to `Timber::get_menu()` by choice. However, if a parameter is passed, I think we should return `null` if no menu is found. This would be what I expect the function to do.

Since we have a lot of custom menu logic, I think we don’t have to follow what WordPress does here.

lib/Menu.php
```diff
@@ -95,10 +94,148 @@ class Menu extends CoreEntity {
...
+			 */
+			$args = apply_filters( 'wp_page_menu_args', $args );
+
+			$nav_menu = new static(null, $args);
+			$nav_menu->init_pages_menu();
+
+			/**
+			 * @see wp_page_menu()
+			 */
+			$_nav_menu = apply_filters( 'wp_page_menu', serialize($nav_menu), $args );
```
gchtr, line 183:
```suggestion
		/**
		 * Since Timber doesn’t use HTML, serialize the menu object to provide a cacheable string.
		 *
		 * Certain caching plugins will use this filter to cache a menu and return it early in the
		 * `pre_wp_nav_menu` filter.
		 *
		 * We can’t use the result of this filter, because it would return a string. That’s why we
		 * don’t assign the result of the filter to a variable.
		 *
		 * @see wp_nav_menu()
		 */
		apply_filters( 'wp_nav_menu', serialize( $nav_menu ), $args );
```

lib/Factory/MenuFactory.php
nlemoine: Yes, I fully agree! I will update this.

tests/test-menu-item-factory.php
```diff
@@ -120,8 +120,60 @@ public function testMenuItemClassmap() {
...
+		$factory = new MenuItemFactory();
+
+		$this->add_filter_temporarily('timber/menuitem/classmap', function() {
+			return [
+				'custom' => MyMenuItem::class,
+			];
+		});
+
+		// Don't use instanceOf, it will return true whether it's MyMenuItem or MenuItem class
+		// and thus does not properly checks the classmap
```
nlemoine, line 173: @gchtr This comment may aloso apply to other classmaps tests

lib/MenuItem.php
```diff
@@ -14,7 +14,7 @@
  */
 class MenuItem extends CoreEntity {
 
-	protected WP_Post $core_object;
+	protected WP_Post $wp_object;
```
nlemoine, line 17: Update with proposed name (@gchtr in #2560)

lib/Factory/MenuItemFactory.php
```diff
@@ -38,7 +38,43 @@ protected function build(WP_Post $item, Menu $menu) : CoreInterface {
...
+		 *
+		 *     return array_merge( $classmap, $custom_classmap );
+		 * } );
+		 * ```
+		 *
+		 * @param array $classmap The menu item class(es) to use. An associative array where the key is
+		 *                        the location and the value the name of the class to use for this
+		 *                        menu item or a callback that determines the class to use.
+		 */
+		$classmap = apply_filters( 'timber/menuitem/classmap', [] );
```
nlemoine, line 66: While working with 2.x on a project, I found that it could be quite useful to have nav location based classmaps for menu items.

lib/Factory/MenuFactory.php
gchtr: I think now the logic is still the same, just switched around.

I propose that we don’t call `$this->from_nav_menu_terms($args)` in `MenuFactory::from()` if no menu can be found. With the current logic, if no menu could be found with the provided args, then Timber will *still* try to return the first menu it can find.

So instead of …

```php
if($menu) {
    return $menu;
}

return $this->from_nav_menu_terms($args);
```

at the end of the method, I would just do …

```php
return $menu;
```

This would fix the bug in #1440 and would be more in line with what I would expect the function to do.

lib/MenuItem.php
gchtr: Thanks!

gchtr (changes requested):
Nice, thanks for the additional updates @nlemoine. I like the new `timber/menuitem/classmap` filter 💪. And good catch with the `assertInstanceOf()` checks. I added two more issues to finish work on these in #2571 and #2572.

For me, the only remaining discussion is in https://github.com/timber/timber/pull/2517#discussion_r848621890. Apart from that it would be nice if we could merge this in, soon.

If you want to add additional features, I think it would be nice to add it in a separate pull request. I think it’s better to have many smaller pull requests instead of having one big pull request that keeps growing and growing.

gchtr:
Oh, and https://github.com/timber/timber/pull/2517#discussion_r820128943 is also unfinished 🙂. Do we want to move that to a separate issue as well, or do you want to solve it here, @nlemoine?

lib/MenuItem.php
nlemoine: Removing it is not so easy, because we need Twig to call the `title()` method in order to apply the `nav_menu_item_title`. If `$item->title` becomes a public property, `{{ item.title }}` won't call `title()` and no filter will be applied.

I added a `$title` and marked it as protected to solve this.

lib/Factory/MenuFactory.php
nlemoine: Sorry, this was not unfinished, this is now fixed!

nlemoine:
@gchtr I think we're finally good on menus 😅 

Please check my comment here: https://github.com/timber/timber/pull/2517#discussion_r861541645 for the `title` explanation.

lib/MenuItem.php
gchtr: @nlemoine Thanks for the update! Would it make sense to completely get rid of the `$title` property? Could this be a use case for #2560?

If we use `{{ item.title }}`, then the `title()` method would be called, were we could use `$this->wp_object->title` instead of `$this->title`:

```php
$title = apply_filters( 'nav_menu_item_title', $this->wp_object->title, $this->wp_object, $this->menu->args ? $this->menu->args : new \stdClass, $this->level );
```

Or might I be missing something?

gchtr: Good point! I see you’ve thought this through. Both solutions are fine for me then, too.

gchtr (approved):

lib/MenuItem.php
nlemoine: So change is required I guess :) Let me know if you still see something, the PR is ready to merge from my side.

nlemoine:
Do I still need to wait for @jarednova approval?

gchtr:
@nlemoine No, I think Jared would have commented already earlier in the process if he saw something. You can merge it in 🎉. Thanks for your work on this!

PR #2528, opened 2022-01-24
Add webp support to letterbox filter

ThomasBerends:
**Ticket**: None

## Issue
The letterbox class does not support webp images.

## Solution
Add support for webp images

## Impact
The 'imagewebp' & 'imagecreatefromwebp' functions were added in PHP 5.4. The Timber Wordpress page reports a minimum PHP version of 5.6, however, the current composer.json shows 5.3. I'm not sure if the composer.json is out of date or not.

## Usage Changes
There should be none

## Considerations
See Impact

## Testing
There doesn't seem to be specific tests for the different extensions.


jarednova (approved):
Looks great! Just fixed up the indentation

PR #2530, opened 2022-01-24
wp_body_open support

Vladisls:
## Issue
Proposal to add `wp_body_open` support similar to `wp_head` and `wp_footer`.

## Solution
Add `wp_body_open` to get_context() return array, similar to `wp_head` and `wp_footer`.

## Testing
There is not much to test as it just adds `wp_body_open` to the context array.

### DIFF
--- a/lib/Timber.php
+++ b/lib/Timber.php
@@ -265,6 +265,7 @@ class Timber {
 			 */
 			self::$context_cache['wp_head'] = new FunctionWrapper( 'wp_head' );
 			self::$context_cache['wp_footer'] = new FunctionWrapper( 'wp_footer' );
+			self::$context_cache['wp_body_open'] = new FunctionWrapper( 'wp_body_open' );
 
 			self::$context_cache = apply_filters('timber_context', self::$context_cache);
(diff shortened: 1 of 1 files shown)


gchtr:
Hey @Vladisls 

Thanks for your first pull request to the project and thanks for filling out the pull request template!

Unfortunately, this is some functionality that will be removed in the future. The proposed way to call `wp_header` and `wp_footer` is to use `{{ function() }}` (or `{{ fn() }}`. Check out the following example that I copied from the [Functions Guide](https://timber.github.io/docs/guides/functions/):

```twig
<html>
	<head>
	<!-- Add whatever you need in the head, and then...-->
	{{ function('wp_head') }}
	</head>

	<!-- etc... -->

	{{ function('wp_footer') }}
	</body>
</html>
```

In a similar way, you can call `wp_body_open()`:

```twig
<body>
	{{ function('wp_body_open') }}
```

Because we can use `function()` for a lot of functionality, we don’t have to add all these functions to the Timber context. That’s why I have to close this pull request, unfortunately. Sorry!

Issue #2533, opened 2022-01-31
Update composer/installers to 2.0


Levdbas:
## Is your feature request related to a problem? Please describe.
As other libraries are updating composer/installers to 2.0 such as roots/bedrock it becomes more difficult to use Timber in combination with said libraries. Installing Timber on a fresh bedrock install breaks the version constraints.

## Describe the solution you’d like
I suggest bumping the version and create a point release for 1.x and applying the same increase in the 2.x branch. I am not sure if we need to update the version constraints like this `"composer/installers": "~1.0|~2.0",` or to `"composer/installers": "~2.0,`

adamtomat:
As @Levdbas mentioned, this is causing a lot of difficulties (my difficulties I mean it's broken) for us using Timber and Bedrock. I'm happy to try and get some PRs together for 1.x and 2.x branches to help get this resolved

adamtomat:
Thanks for merging this in 👍  

Are you also able to tag this too, please (e.g. `1.19.2`)?

jarednova:
@adamtomat (and future Googlers): please see ....

https://github.com/timber/timber/releases/tag/1.19.2

PR #2544, opened 2022-02-21
Update composer/installers from v1 to v2

adamtomat:
**Ticket**: #2533

## Issue
Bedrock (from Roots) recently upgraded to composer/installers v2 and it makes it very hard to use Timber with Bedrock without also upgrading Timber's dependency.

## Solution
Upgrade Timber's "composer/installers" dependency to match Bedrock's (https://github.com/roots/bedrock/commit/43e2a97792104d141661193e478dfcb656bc1274)

**The upstatement/routes package also needs updating before this issue can be resolved.**

## Impact
Shouldn't have much impact. Anyone who still has a dependency on v1 of composer/installers would be able to use a previously tagged version of Timber if needed.

## Considerations
The alternative would be to add support for v1 or v2 like so:
`"composer/installers": "~1.0|^2.0"`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
   "require": {
     "php": "^7.4|^8.0",
     "twig/twig": "^2.12|^3.0",
-    "composer/installers": "^1.9",
+    "composer/installers": "^2.0",
     "twig/cache-extension": "^1.5.0"
   },
   "require-dev": {


gchtr (approved):
Good to go from me. We can merge this in as soon as we fixed the issues with the automated tests.

jarednova:
Automated tests are now clear — thanks for clearing through that @gchtr.

And thanks as always @adamtomat for this and #2543 to keep us up to date with these Composer dependencies!

PR #2547, opened 2022-02-21
2.x Fix tests

gchtr:
## Issue

Tests are broken, because of an error with PHP 8.1 in **wp-includes/block-supports/layout.php**. Here’s the relevant trac ticket: https://core.trac.wordpress.org/ticket/55119. The recommendation is to not use PHP 8.1 yet:

> WordPress core is not fully compatible with PHP 8.1 yet, there are still some deprecation notices. If you need a workaround, I would suggest switching to PHP 8.0 for now.

## Solution

Disable testing against PHP 8.1 when [WordPress doesn’t officially support it yet](https://make.wordpress.org/core/handbook/references/php-compatibility-and-wordpress-versions/) and let’s see if something else is broken, too.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,11 +49,11 @@ jobs:
             multisite: '0'
             experimental: true
           # PHP 8.1
-          #- php: '8.1'
-          #  wp: 'trunk'
-          #  dependency-version: 'prefer-stable'
-          #  multisite: '0'
-          #  experimental: true
+          - php: '8.1'
+            wp: 'trunk'
+            dependency-version: 'prefer-stable'
+            multisite: '0'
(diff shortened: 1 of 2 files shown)


gchtr:
Seems to do the trick 🙂. I’ll merge this in as soon as I get a quick review or opinion on why we should already test against 8.1.

gchtr:
I enabled the testing against PHP 8.1 again and skipped the two failing tests for PHP 8.1 in [`2eef553` (#2547)](https://github.com/timber/timber/pull/2547/commits/2eef5537725f284d8c8d4d5142660ae028da9a43).

jarednova (approved):
Skipping these for now makes sense!

PR #2550, opened 2022-03-01
Improve composer version ranges

gchtr:
**Ticket**: #2543

## Issue

- We should probably allow both version for `composer/installers` as also suggested in https://github.com/timber/timber/pull/2543 and https://github.com/timber/timber/pull/2546#pullrequestreview-894393017.
- We’re using an old pipe format for Composer versions.

## Solution

- Allow both versions `^1.0` and `^2.0` for `composer/installers`.
- Use modern pipe format using ` || ` instead of `|`. See https://getcomposer.org/doc/articles/versions.md#version-range for more information.

## Impact

Better compatibility in various environments.

## Usage Changes

None.

## Considerations

No.

## Testing

No.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
     "php": ">=5.3.0|7.*",
     "twig/twig": "^1.41|^2.10",
     "upstatement/routes": "0.8.*",
-    "composer/installers": "^2.0",
+    "composer/installers": "^1.0 || ^2.0",
     "twig/cache-extension": "^1.5"
   },
   "require-dev": {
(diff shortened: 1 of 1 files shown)


gchtr:
@jarednova Can I get a quick review to merge this in? Thanks!

jarednova (approved):

jarednova:
Yup yup! all set for you to merge @gchtr

adamtomat:
Thanks for merging this in 👍  Are you able to tag `master` with this in too please, as we can't benefit from it until then. (e.g. 1.19.2)

jarednova:
@adamtomat gladly! check it out at: https://github.com/timber/timber/releases/tag/1.19.2

Issue #2537, opened 2022-02-09
TimberImage::aspect() generates Fatal Error if image doesn't exist


slimndap:
## Expected behavior
Maybe `false` or `-1`?

## Actual behavior

A PHP warning: 
```
Fatal error: Uncaught DivisionByZeroError: Division by zero in /path/to/website/wp-content/plugins/timber-library/lib/Image.php on line 410
```
## Steps to reproduce behavior

1. Create a post.
2. Add an featured image.
3. Save post.
3. Add this in Twig template of single post: `{{ post.thumbnail.aspect }}`.
5. View the post on the front end. You should see a number.
4. Remove the thumbnail from disk (not from  media library).
5. View the post on the front end.
6. The PHP warning is triggered.

## What version of WordPress, PHP and Timber are you using?
WordPress 5.9, PHP 8.0, Timber 1.19.1

## How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)

Using 'add new plugin' inside the WordPress admin.

slimndap:
Additional information: the problem also occurs in PHP 7.3.5, but only trigger a _PHP warning_:

```
Warning: Division by zero in /path/to/website/wp-content/plugins/timber-library/lib/Image.php on line 410
```

czernika:
It happens because of step 6 - while you physically removed image from disk its metadata and attachment post type object itself still presents in database. You may check it if you call `wp_get_attachment_metadata()` function (if you know image id) - it will return array of data even if image doesn't exists anymore. When you're deleting image from Media uploader - post object will be deleted

So what causes the issue? The thing is while image is not present, its metadata is still "alive" and will return some value. At some point, when you are calling `aspect` method, Timber will call `get_dimensions()` method which will set image data **ONLY** if image is present. And if image is not present... Well Timber doesn't handle this so the method will return `null`

```
// lib/Image.php

protected function get_dimensions( $dim ) {
    if ( isset($this->_dimensions) ) {
        return $this->get_dimensions_loaded($dim);
    }

    if ( file_exists($this->file_loc) && filesize($this->file_loc) ) {
        if ( ImageHelper::is_svg( $this->file_loc ) ) {
            $svg_size             = $this->svgs_get_dimensions( $this->file_loc );
            $this->_dimensions    = array();
            $this->_dimensions[0] = $svg_size->width;
            $this->_dimensions[1] = $svg_size->height;
        } else {
            list($width, $height) = getimagesize($this->file_loc);
[...]

PR #2552, opened 2022-03-06
2.x images

czernika:
**Ticket**: [#2537](https://github.com/timber/timber/issues/2537)

## Issue
`Timber\Image` class generates DivisionByZeroError error when calling `aspect` method of image was deleted manually from uploads directory. While being deleted attachment object itself and its metadata still present in database and will return something. When calling `image.aspect` it defines image dimensions and in case if it is not exists returns `0` for both height and width which causes zero divided by zero issue

## Solution
Check if image height has `0` value and if it is return 0 as a result. This qay we will prevent site crashing. On `init()` method we will check if file exists and if not log it with `Image path/to/image doesn't exists`

Also added `protected $thumbnail` property to `Post` object and define it on `thumbnail` method as instance of `Image` - currently with `thumbnail()` method it will initialize `Image` every time it called

```twig
{% if post.thumbnail.aspect > 1.5 %}
    <img src="{{ post.thumbnail.src }}" alt="{{ post.thumbnail.alt }}">
{% endif %}
```

In this example `Image` will be initialized 3 times

## Impact
Fixes fatal error if image was deleted from uploads directory but not from WordPress media uploader. Also prevents Image object from initializing multiple times

## Usage Changes
All lefts the same

## Considerations
Partially discussed in [#2537](https://github.com/timber/timber/issues/2537)

## Testing
Steps to reproduce issue: [#2537](https://github.com/timber/timber/issues/2537)

- `{{ post.thumbnail.aspect }}` will return `0` if image file does not exists (but its object exists in database as an attachment post type)
- {{ post.thumbnail }} instantiation will log error `Image path/to/image doesn't exists` (not causing site crash but gives developer a clue what is may be wrong)
- Post thumbnail will be initialized once

### DIFF
--- a/lib/Image.php
+++ b/lib/Image.php
@@ -405,8 +405,13 @@ class Image extends Post implements CoreInterface {
 	 * @return float
 	 */
 	public function aspect() {
-		$w = intval($this->width());
 		$h = intval($this->height());
+
+		if ( 0 === $h ) {
+			return 0;
+		}
+
+		$w = intval($this->width());
 		return $w / $h;
 	}
 
--- a/lib/Image.php
+++ b/lib/Image.php
@@ -320,6 +320,11 @@ class Image extends Post implements CoreInterface {
 			$this->file = reset($this->_wp_attached_file);
 			$this->file_loc = $basedir.DIRECTORY_SEPARATOR.$this->file;
 		}
+
+		if ( ! file_exists( $this->file_loc ) ) {
+			Helper::error_log("Image {$this->file_loc} doesn't exists");
+		}
+
 		if ( isset($image_info['id']) ) {
 			$this->ID = $image_info['id'];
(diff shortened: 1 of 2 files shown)


czernika:
@jarednovan sorry for misleading. PR was intended for `1.x` version as it is causing fatal error. It was my bad I called branch `2.x` - to be honest I don't remember why does it happened but I have few pet projects with `2.x` version and some others with `1.x`. I guess I just lost myself in naming versions... Sorry again

Changes in `Image.php` file are compatible with version `2.x` though. Crucial part is this [one](https://github.com/timber/timber/pull/2552/files#diff-900140b378ca9109730a66837719810d77d3129c2d5f68eb8356c80f9b94dcd2R415) - it is the only required part of PR while others are just "recommended"

PR #897, opened 2016-03-09
Fixing issue with tojpg not being deleted

connorjburton:
Fixes #894 

#### Issue
Some JPG files are not deleted, when they were generated through [`tojpg` Twig filter](https://github.com/jarednova/timber/wiki/Image-cookbook#converting-images) that was used on a PNG file.

#### Solution
Change generated `tojpg` files to have `tojpg` in the filename.
Added two more calls to `delete_generated_files` with specific regex for the `tojpg`

#### Impact
Negligibly slower

#### Usage
No change

#### Considerations
I tried combining the two calls to one, but couldn't get it to work with `glob` correctly.

Issue #2535, opened 2022-02-02
img_to_jpg function doesn't delete JPGs generated from PNGs when original PNG is deleted 


4foot30:
## Expected behaviour
JPGs created from PNGs with a direct call to `ImageHelper::img_to_jpg()` should be deleted from the uploads directory when the original source PNG is deleted from the WordPress Media library.

## Actual behaviour
The generated JPGs remain in the uploads directory. 

## Steps to reproduce behaviour
I am generating the JPGs via my PHP, not in my Twig templates, and returning the filenames of the newly created JPGs to the Timber context for later output inside a Twig template inside a `<picture>` tag:

```
        $resized_image = ImageHelper::resize($image_original_file, $image_width, $image_height, ['center', 'center']);
        if ($to_jpg !== false) {
            $resized_image = ImageHelper::img_to_jpg($resized_image);
        }
        return $resized_image;
```

Everything works perfectly including the resizing and format conversion, but the converted JPGs are not deleted when the original PNG is. If I try the same code using an original JPG, with no need to call `ImageHelper::img_to_jpg()` then the resized JPGs *are* correctly deleted when the original source JPG is deleted from the Media library. The problem seems to only occur when a file format conversion has happened. 

Basically I've got all the same symptoms as this issue: https://github.com/timber/timber/issues/894 - but I'm converting the image type using `ImageHelper::img_to_jpg()` in the PHP rather than the `|tojpg` filter in Twig. I can see that I'm using a new enough Timber version that it includes the fixes from https://github.com/timber/timber/issues/894 and https://github.com/timber/timber/pull/897 but I haven't had any luck debugging `TimberImageHelper::delete_generated_files`.

If it helps, I've used Timber and my resize/convert code in the past on other sites (I think WP 5.7 or so) and had confirmed that the converted JPGs were correctly deleted - but it's failing for me on the latest versions of Timber and WordPress on my local as well as on WPEngine hosting. 

Timber is superb btw, so thank you! :) 

## What version of WordPress, PHP and Timber are you using?
WordPress 5.9, PHP 7.4.1, Timber 1.19

## How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
From Composer, via `composer require wpackagist-plugin/timber-library`

jarednova:
@4foot30 thanks so much for the great bug report. Are you interested or able to keep going and see if you can debug and bring things to a PR?

gchtr:
[...]

PR #2556, opened 2022-03-14
imagehelper -  filename (towebp, tojpg)

EpicKau:
Fixes #897 

## Issue
files are not deleted, when they were generated through timber filter. (webp, png)

## Solution
Change generated filenames to have target fileformat in the filename.
Added more calls to delete_generated_files with specific regex for fileformat

I hope it helps.

### DIFF
--- a/lib/Image/Operation/ToJpg.php
+++ b/lib/Image/Operation/ToJpg.php
@@ -28,7 +28,8 @@ class ToJpg extends ImageOperation {
 	 * @return  string    the final filename to be used (ex: my-awesome-pic.jpg)
 	 */
 	public function filename( $src_filename, $src_extension = 'jpg' ) {
-		$new_name = $src_filename.'.jpg';
+		$newbase = $src_filename.'-tojpg';
+		$new_name = $newbase.'.jpg';
 		return $new_name;
 	}
 
--- a/lib/Image/Operation/ToWebp.php
+++ b/lib/Image/Operation/ToWebp.php
@@ -28,7 +28,8 @@ class ToWebp extends ImageOperation {
 	 * @return  string    the final filename to be used (ex: my-awesome-pic.webp)
 	 */
 	public function filename( $src_filename, $src_extension = 'webp' ) {
-		$new_name = $src_filename  . '.webp';
+		$newbase = $src_filename.'-towebp';
+		$new_name = $newbase  . '.webp';
 		return $new_name;
 	}
 
--- a/lib/ImageHelper.php
+++ b/lib/ImageHelper.php
@@ -274,7 +274,7 @@ class ImageHelper {
 
 
 	/**
-	 * Deletes the auto-generated files for resize and letterboxing created by Timber
+	 * Deletes the auto-generated files created by Timber (resize, letterboxing, jpg, webp, retina) 
 	 * @param string  $local_file   ex: /var/www/wp-content/uploads/2015/my-pic.jpg
 	 *	                            or: http://example.org/wp-content/uploads/2015/my-pic.jpg
 	 */
@@ -290,6 +290,8 @@ class ImageHelper {
 		self::process_delete_generated_files($filename, $ext, $dir, '-lbox-[0-9999999]*', '-lbox-[0-9]*x[0-9]*-[a-zA-Z0-9]*.');
 		self::process_delete_generated_files($filename, 'jpg', $dir, '-tojpg.*');
 		self::process_delete_generated_files($filename, 'jpg', $dir, '-tojpg-[0-9999999]*');
+		self::process_delete_generated_files($filename, 'webp', $dir, '-towebp.*');
+		self::process_delete_generated_files($filename, 'webp', $dir, '-towebp-[0-9999999]*');
 	}
 
 	/**
--- a/tests/test-timber-image-tojpg.php
+++ b/tests/test-timber-image-tojpg.php
@@ -23,7 +23,7 @@
 		function testPNGtoJPG() {
 			$filename = TestTimberImage::copyTestImage( 'flag.png' );
 			$str = Timber::compile_string('{{file|tojpg}}', array('file' => $filename));
-			$renamed = str_replace('.png', '.jpg', $filename);
+			$renamed = str_replace('.png', '-tojpg.jpg', $filename);
 			$this->assertFileExists($renamed);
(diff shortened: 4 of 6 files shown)


jarednova:
Thanks so much @EpicKau ! It looks like some tests need to be updated to match with this new naming convention. Could you give it a look?

EpicKau:
@jarednova So what I see is that most time the test failed because of "assertFileExists" check.  ("Failed asserting that file "XYZ" exists.")
That makes sense because the output file has another name than the input file. 
E.g. XYZ-tojpg.jpg instead of XYZ.jpg.

Adnanlagharidotcom:
Thanks a lot @EpicKau Here i Created a tool [About us page Generator](https://adnanlaghari.com/free-about-us-page-generator/)

jarednova (dismissed):

Levdbas:
So I ran into this issue #2748 as well and found this PR after some digging.  #2535 seems to be related as well.

I did some local tests and this pr fixes both problems for me. both the webp and and jpg generated images are removed.

Only files that are not accounted for are when running towebp and tojpg on crops that will result to filenames like these.
```
test-400x267-towebp.webp
test-400x267-tojpg.jpg
```

gchtr (commented):
With this change, websites that use the ToJpg and ToWebp operations will be left with quite a few stray webp and jpg images.

Do we have to consider this as a **breaking change**? Because I see the following problems here:

- It will trigger the images to be regenerated and for sites with a lot of images, right? This could create performance issues or even max execution time errors.
- It might have serious impact on the whole size of the website, which could lead to problems on websites that have a limited available space.

marcoluzi:
This is still an issue we run into every now and then. Is there an ETA for this fix?

enrikberisha:
@gchtr @pascalknecht Any chance we could get this merged? The installations are getting bigger and bigger. I could also bring some 🍻 to your office. 😊

svenvonarx:
@EpicKau I tested it with WebP images generated with `Timber\ImageHelper::img_to_webp` and it doesn't seem to add the "-towebp" suffix. So these generated versions wouldn't be deleted by this addition. Also I found another issue that seems to come up, if the pictures have the WordPress native suffix "-scaled". That would need to be accounted aswell, since these would not be deleted either.

```php
self::processDeleteGeneratedFiles($filename, 'webp', $dir, '-[0-9999999]*', '-[0-9]*x[0-9]*-c-[a-z]*.');
self::processDeleteGeneratedFiles($filename, 'webp', $dir, '-lbox-[0-9999999]*', '-lbox-[0-9]*x[0-9]*-[a-zA-Z0-9]*.');

// If scaled image is deleted, delete the original image webp as well
$filename_without_scaled = str_replace('-scaled', '', $filename);

self::processDeleteGeneratedFiles($filename_without_scaled, 'webp', $dir, '-[0-9999999]*', '-[0-9]*x[0-9]*-c-[a-z]*.');
self::processDeleteGeneratedFiles($filename_without_scaled, 'webp', $dir, '-lbox-[0-9999999]*', '-lbox-[0-9]*x[0-9]*-[a-zA-Z0-9]*.');
```

With these addition we seem to catch all files. Would be great to see this implemented by default, if you need any help on this, I am happy to help with the update of the pull request.

marcoluzi:
@svenvonarx 
> I tested it with WebP images generated with `Timber\ImageHelper::img_to_webp` and it doesn't seem to add the "-towebp" suffix.

The `-towebp` and `-tojpg` suffix have been added in this PR. See the `lib/Image/Operation/ToJpg.php` and` lib/Image/Operation/ToWebp.php` in the changed files.

Levdbas:
> @svenvonarx
> 
> > I tested it with WebP images generated with `Timber\ImageHelper::img_to_webp` and it doesn't seem to add the "-towebp" suffix.
> 
> The `-towebp` and `-tojpg` suffix have been added in this PR. See the `lib/Image/Operation/ToJpg.php` and` lib/Image/Operation/ToWebp.php` in the changed files.

Yes, source files that are changed to webp, but not crops that are changed to webp. Those still have to be addressed.

gchtr:
There’s no ETA on this fix yet. Because there’s more involved here than simply merging this PR. But it is definitely something we want to work on next.

As mentioned in https://github.com/timber/timber/pull/2556#pullrequestreview-1446346066, this "fix" would cause many websites to generate a lot of images, while existing images will not be deleted automatically. If we don’t classify this as a breaking change, many websites could run into problems.

I wrote an answer to a similar issue in https://github.com/timber/timber/pull/2876#issuecomment-1914234521, which outlines the problems we have to solve:

> As with all the PR involving file name changes for images, I wonder how we should handle this properly. Because this will **cause a lot of files to be regenerated**. It could cause lots of max execution errors and could quickly fill up space on some hosts and leave the uploads folder with lots of unused files.
> 
> Could we somehow add a migration path so that developers have a way to fix this on their sites?
> 
> * Provide a way to delete all deprecated (or even all) WebP files using the WP CLI.
> * But then again, not all hosts allow you to run WP CLI. So maybe we could add integrations for popular [image regeneration plugins](https://wordpress.org/plugins/search/regenerate+thumbnails/) that could handle the migration as well.
> * Maybe save generated WebP files to a different folder, as envisioned in [Big ticket about Image handling #2866](https://github.com/timber/timber/issues/2866)
> 
> I wonder whether we even should put this change behind a filter hook. The filter could be applied for all new Timber installations. For existing installations, we could maybe work with a flag that is saved in the `wp_options` database and is deleted once the change is through. Depending on the flag, we could show a notification in the admin dashboard.

So, for this issue to be fixed more quickly, I think the **most straight-forward way** to go would be to **add the fixes as an opt-in feature through a WordPress filter hook in a new PR against the 2.x branch.**

Something like:

```php
add_filter('timber/image/advanced_file_names', '__return_true');
```

gchtr:
@svenvonarx @marcoluzi @enrikberisha Does any one of you have the capacity to work on a basic PR for this? If we had the base code, this would surely speed up a fix for this. If you can’t include tests, that would be fine as well. We could work on those.

@enrikberisha Thanks for the offer for the 🍻😄. Let’s have on at [WordCamp Europe](https://europe.wordcamp.org/2025/) in Basel with the whole Timber team, if you can make it 😉.

marcoluzi:
@gchtr I can try to come up with something. 👍🏻

PR #2562, opened 2022-03-24
2.x Implement new PagesMenu API

gchtr:
**Ticket**: #2292

## Issue

See #2292.

## Solution

See #2292.

Please consider that this pull request is against the `2.x-menu-enhancements` branch, see #2517.

## Impact

See documentation updates in #2254.

## Usage Changes

See updates in #2254.

## Considerations

See review comments.

## Testing

Yes.


jarednova:
Amazing, thanks @gchtr ! I think @nlemoine will have better specifics than I. In reviewing the tests there is one failure ...

```php
There was 1 failure:

1) TestTimberContext::testPostsContextAuthor
Failed asserting that actual size 6 matches expected size 3.

Error: /home/runner/work/timber/timber/tests/test-timber-context.php:127
phpvfscomposer:///home/runner/work/timber/timber/vendor/phpunit/phpunit/phpunit:97
```

However, this error only occurs against trunk and I am not able to reproduce locally when running against 5.9.2. I am comfortable ignoring the failure. Related (though it would be a seperate PR) — should we continue to test against `trunk`? or `latest`?

jarednova (approved):

lib/PagesMenu.php
```diff
@@ -0,0 +1,60 @@
+<?php
```
nlemoine, line 1: None of the filters of [`wp_nav_menu`](https://developer.wordpress.org/reference/functions/wp_page_menu/) are used here. I think we could still provide some. Nor the ones from [`wp_list_pages`](https://developer.wordpress.org/reference/functions/wp_list_pages/) either.

We're skipping directly to `get_pages` and some features are lost along the way: `wp_nav_menu` -> `wp_list_pages` -> `get_pages`.

I'm not a big fan of how those features are implemented the WordPress way but since w're now supporting all `$args` and filters of `wp_nav_menu`, it would a shame / an inconsistency to not support them for the pages menu (`show_home`, etc.).

lib/Factory/PagesMenuFactory.php
```diff
@@ -0,0 +1,75 @@
...
+	 *
+	 * @internal
+	 *
+	 * @return string
+	 */
+	protected function get_menu_class( $args ) : string {
+		/**
+		 * Filters the class(es) used for different menus.
+		 *
+		 * Read more about this in the documentation for [Menu Class Maps](https://timber.github.io/docs/v2/guides/class-maps/#the-menu-class-map).
```
nlemoine, line 35: "Class filter" I guess

lib/Factory/PagesMenuFactory.php
```diff
@@ -0,0 +1,75 @@
...
+		 * @example
+		 * ```
+		 * add_filter( 'timber/pages_menu/class', function( $class ) {
+		 *     return ExtendedPagesMenu::class;
+		 * } );
+		 * ```
+		 *
+		 * @param array $class The pages menu class to use.
+		 */
+		$class = apply_filters( 'timber/pages_menu/class', PagesMenu::class );
```
nlemoine, line 49: Adding `$args` might be useful? See `timber/menu/class`

gchtr:
I added some more filters from `wp_page_menu()` and `wp_list_pages()` to make it more compatible with the WordPress filters. I think we can’t apply the filters from `wp_nav_menu()`, because the pages menu is really something different. I also didn’t add the `wp_get_nav_menu_items` filter, because that one requires a `$menu` object to be passed in, which we don’t have.

I think we can’t implement the `show_home` option, because when there’s a home page, it will already be included and if there is not, we would have to create a whole mock object for the home URL so that it will work as a `Timber\MenuItem`. And I think that would be overkill.

But I managed to make it work with the `depth` option and also added tests for that.

@nlemoine This is ready for another review. I’d be really glad if you could have another look at this.

lib/PagesMenu.php
gchtr: See https://github.com/timber/timber/pull/2562#issuecomment-1160760369.

lib/Factory/PagesMenuFactory.php
gchtr: Yep, that’s right. I updated it.

lib/Factory/PagesMenuFactory.php
gchtr: I added the `$args` parameter to the filter. Good call!

nlemoine:
> I think we can’t apply the filters from wp_nav_menu()

I'm fine with that 👍

nlemoine (approved):

Issue #2548, opened 2022-02-22
PHP8.1 | Deprecated: Return type compatibility


rubas:
**With PHP 8.1...**

Deprecated: Return type of Timber\QueryIterator::current() should either be compatible with Iterator::current(): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in 
`vendor/timber/timber/lib/QueryIterator.php on line 137`

Deprecated: Return type of Timber\QueryIterator::next() should either be compatible with Iterator::next(): void, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in 
`vendor/timber/timber/lib/QueryIterator.php on line 150`

Deprecated: Return type of Timber\QueryIterator::key() should either be compatible with Iterator::key(): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in 
`vendor/timber/timber/lib/QueryIterator.php on line 156`

Deprecated: Return type of Timber\QueryIterator::valid() should either be compatible with Iterator::valid(): bool, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /
`vendor/timber/timber/lib/QueryIterator.php on line 133`

Deprecated: Return type of Timber\QueryIterator::rewind() should either be compatible with Iterator::rewind(): void, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in 
`vendor/timber/timber/lib/QueryIterator.php on line 152`

Deprecated: Return type of Timber\QueryIterator::count() should either be compatible with Countable::count(): int, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in 
`vendor/timber/timber/lib/QueryIterator.php on line 202`

Deprecated: Return type of Timber\PostsIterator::current() should either be compatible with ArrayIterator::current(): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in 
`vendor/timber/timber/lib/PostsIterator.php on line 12`

## What version of WordPress, PHP and Timber are you using?
PHP 8.1, Timber 1.19.1

## How did you install Timber?
Composer

trsteel88:
composer.json php version also needs to be bumped to allow php8

trsteel88:
@jarednova is there any existing PR's to allow php8.1?

nlemoine:
[...]

PR #2563, opened 2022-03-25
Test against Latest WP Version instead of Trunk

jarednova:
**Ticket**: #2562 

## Issue
There appears to be an unrelated failure occurring due to `trunk` changes in WP on #2562. This begs the question — why test against trunk at all?

## Solution
Swap `trunk` for `latest`

## Impact
We'll be a bit behind should a breaking change come in WP's functions (whereas before we could have caught it before formal release). However, 

## Usage Changes
None

## Considerations
Should we maybe keep one trunk test in the matrix and allow failures?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -48,7 +48,7 @@ jobs:
             dependency-version: 'prefer-stable'
             multisite: '0'
             experimental: true
-          # PHP 8.1
+          # PHP 8.1 / expiremntal
           - php: '8.1'
             wp: 'trunk'
             dependency-version: 'prefer-stable'
@@ -56,6 +56,11 @@ jobs:
             experimental: true
           # PHP 8.0
           - php: '8.0'
+            wp: 'latest'
+            dependency-version: 'prefer-stable'
+            multisite: '0'
+          # PHP 8.1
+          - php: '8.1'
             wp: 'latest'
             dependency-version: 'prefer-stable'
             multisite: '0'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -44,19 +44,19 @@ jobs:
         include:
           # WP Trunk
           - php: '7.4'
-            wp: 'trunk'
+            wp: 'latest'
             dependency-version: 'prefer-stable'
             multisite: '0'
             experimental: true
           # PHP 8.1
           - php: '8.1'
-            wp: 'trunk'
+            wp: 'latest'
             dependency-version: 'prefer-stable'
             multisite: '0'
             experimental: true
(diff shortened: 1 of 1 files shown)


nlemoine:
Testing `trunk`, as well as latest PHP version is important to spot bugs before they are released (see https://github.com/timber/timber/issues/2548) which prevents from running into issues when it comes out.

I don't have stats on this but I feel that latest version adoption is quite fast nowadays.

jarednova:
I hear that @nlemoine — maybe I was being a bit too cavalier. I agree, we should preserve some amount of `trunk` tests. 

Perhaps the better "middle way" is now here with the latest commit: test against `latest` for most of the tests and keep one set of `trunk` tests. This way we can better identify what is a live issue in the code against WP vs. what could be a coming issue (or an accidental regression that the WP core team will likely resolve). What do you think?

nlemoine:
I would keep two `trunk`: 
- PHP 8.1/trunk: to ensure Timber is still working on edge configuration
- PHP 7.4/trunk: to ensure Timber is still working on edge WP

Both with allow failure on, so we can keep an eye on bugs to come but still let us merge things that are not ready yet to pass (or waiting that WP fixes some issues).

nlemoine:
I've updated the test configuration. 

I also think we should support PHP 8.1 as it's already out since a few months. It would be a shame to have a PHP 8.0.x only constraint. I already fixed the issues I've found for that version and It's only a matter of time before we have to support this.

gchtr:
So how do we manage the failed runs against WP trunk? Is there a state that they pass with the errors? Or do we manually check if it’s only those tests that fail and then overrule merging?



In the current test run, there’s also a PCOV error with [PHP 7.4 against WP latest](https://github.com/timber/timber/runs/5728064459?check_suite_focus=true).

nlemoine:
I don't think this related to GA trunk tests failing? It says "Review required"

gchtr (approved):

gchtr:
> I don't think this is related to GA trunk tests failing? It says "Review required"

😅  Now I feel stupid. Merging isn’t blocked now, but I think we still need to manually check whether any failing checks are against WP trunk before merging.

nlemoine:
> I think we still need to manually check whether any failing checks are against WP trunk before merging.

Not sure I understand your comment. If some failures happen on trunk, you will see them but they will not prevent the "Merge" button. A failure on non trunk test will disable the "Merge" button. 

Capture d’écran 2022-04-09 à 09 31 09

jarednova:
This works! So just to clarify/confirm, the _review_ is what creates the permission for the merge — not a setting in the config or matrix

Issue #2561, opened 2022-03-22
Use `null` instead of `false` as return type in factories


gchtr:
**Ticket:** https://github.com/timber/timber/pull/2517#pullrequestreview-803134223

We currently return `false` in the object factories when no object could be created. Two examples for posts:

- See [`Timber\Factory\PostFactory::from()`](https://github.com/timber/timber/blob/2.x/lib/Factory/PostFactory.php#L46)
- See [`Timber\Timber::get_post()`](https://github.com/timber/timber/blob/2.x/lib/Timber.php#L252)

We should change this to return `null`, because

- Being able to use the [null coalescing operator](https://stitcher.io/blog/shorthand-comparisons-in-php#null-coalescing-operator) on `Timber\Timber::get_post()` is a big plus.
- WordPress’s `get_post()` function also returns `null` if no post can be found.

This does not only apply to posts, but also other factories.

gchtr:
I tracked down where we discussed whether returning `null` or `false` in factories makes more sense.

The discussion started in #2160, where @jarednova first mentioned in https://github.com/timber/timber/pull/2160#discussion_r365606242 that he switched some instances from `false` to `null`. See commits [`aeba8af`](https://github.com/timber/timber/pull/2160/commits/aeba8afb862eaee884d6321739c65fe7b7c8e27b), [`d62ccde`](https://github.com/timber/timber/pull/2160/commits/d62ccde6a19e971f480c5f1539fa32551f51b6ac) and [`b16a764`](https://github.com/timber/timber/pull/2160/commits/b16a764b72ef5c7c7d5ee442f33120a16d3f4823).

> Take a look and let me know if you agree with my one substantive change (returning `false` instead of `null` when no user found via `Timber::get_user_by` and other similar methods). This is the pattern I saw in `CommentFactory` and what would make the most logical sense to me*. Merge if you agree! (or we can discuss what makes the most sense here and close that out)
> 
> `*` I realize there are a few other places in the 2.x codebase not related to factories where things are returning `null` instead of `false`. I'll take a separate PR/issue to investigate and resolve these
 
The discussion continued with two more comments:

- https://github.com/timber/timber/pull/2160#issuecomment-573452212.
- https://github.com/timber/timber/pull/2160#issuecomment-575781782

Then there were some more changes in #2073:

-  Commit [`e11566b` (#2073)](https://github.com/timber/timber/pull/2073/commits/e11566bb947570b7f1a8a6ce2ec110ca2139ad6f)
[...]

PR #2566, opened 2022-03-28
2.0 Use `null` instead of `false` as return type in factories

gchtr:
**Ticket**: #2561

### DIFF
--- a/docs/v2/upgrade-guides/2.0.md
+++ b/docs/v2/upgrade-guides/2.0.md
@@ -202,6 +202,10 @@ function get_post( mixed $query = false, array $options = [] );
 
 We deprecated the `$PostClass` parameter. If you want to control the class your post should be instantiated with, use a [Class Map filter](https://timber.github.io/docs/v2/guides/class-maps/#the-post-class-map). Instead, we now have an `$options` array.
 
+#### Changed return type `Timber::get_post()` on failure
+
+`Timber::get_post()` returned `false` if no post could be found in 1.x. Now, that function will return `null` if no post could be found.
+
 ### Timber::get_posts()
 
 #### Use `Timber::get_posts()` instead of `new Timber\PostQuery()`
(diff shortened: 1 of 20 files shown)


docs/v2/upgrade-guides/2.0.md
```diff
@@ -202,6 +202,10 @@ function get_post( mixed $query = false, array $options = [] );
 
 We deprecated the `$PostClass` parameter. If you want to control the class your post should be instantiated with, use a [Class Map filter](https://timber.github.io/docs/v2/guides/class-maps/#the-post-class-map). Instead, we now have an `$options` array.
 
+#### Changed return type `Timber::get_post()` on failure
+
+`Timber::get_post()` returned `false` if no post could be found in 1.x. Now, that function will return `null` if no post could be found.
```
jarednova, line 207:
```suggestion
`Timber::get_post()` returned `false` if no post could be found in 1.x. Now, that function will return `null` if no post could be found. We recommend using `null` instead of `false` for when no objects are found to match modern PHP practices.
```

jarednova (changes requested):
My only suggestion is to add a more universal note about `null` > `false` in the Upgrade Guide as we're also showing this as the prefered style like in `docs/v2/guides/extending-timber.md:287`

gchtr:
Thanks @jarednova. I committed your suggestion. With a final approval, this is ready to merge.

jarednova (approved):

jarednova:
Final approval granted, please merge away!

nlemoine (approved):

gchtr:
I’m waiting for the changes in #2517 to be merged before merging this PR in.

jarednova:
Thanks @gchtr. To confirm where we are on #2517, is the last pending edit the one described in this [comment/discussion](https://github.com/timber/timber/pull/2517#pullrequestreview-900994888), or are there any other items requiring resolution?

PR #2578, opened 2022-04-29
Allow the Timber\PostPreview::read_more to accept a boolean value

gerardo-rodriguez:
## Issue

The [Timber\PostPreview::read_more](https://timber.github.io/docs/reference/timber-postpreview/#read_more) docs say:
> Set this to `false` to not add a "Read More" link.

But `Timber\PostPreview::read_more` doesn't actually accept a boolean value, it only accepts a `string` value type. 😬 

Screen Shot 2022-04-29 at 11 37 53 AM

## Solution

Add `bool` to the `readmore` docblock.

## Impact

Minimal to none?

## Usage Changes

No usage changes, to my knowlege.

## Considerations

None come to mind.

## Testing

I did not add tests. Not sure how to do so.

### DIFF
--- a/lib/PostPreview.php
+++ b/lib/PostPreview.php
@@ -69,7 +69,7 @@ class PostPreview {
 	/**
 	 * Read more text.
 	 *
-	 * @var string
+	 * @var string|bool
 	 */
 	protected $readmore = 'Read More';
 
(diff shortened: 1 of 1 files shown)


nlemoine (approved):
Good catch @gerardo-rodriguez, thanks for fixing the docblock!

@timber/rangers Note for future releases, this should be revised with actual types.

PR #2579, opened 2022-05-03
2.x Clean up Timber\PostCollection

gchtr:
**Ticket**: #2311 

## Issue

As @acobster mentioned in #2311, the `Timber\PostCollection` class can be removed completely.

## Solution

This pull request:

- Removes the `Timber\PostCollection` class.
- Removes a last `Timber\QueryIterator` check that was always false.
- Updates the documentation and the Upgrade Guide.

## Impact

If someone directly used `Timber\PostCollection`, it will throw an error. This is probably clearer than a state where the class still exists, but doesn’t do anything.

## Usage Changes

Already handled in #2311.

## Considerations

The only thing that was still in the `Timber\PostCollection` and is now missing in Timber this is the `maybe_set_preview()` method.

The `maybe_set_preview()` method was introduced in https://github.com/timber/timber/commit/d137a698b00a0823c3b4a4379d59110a86ad0c77. But I didn’t find a reason why. And there aren’t any tests.

I asked myself what this method would do. Because get previews for a posts or posts in a collection of posts seems like an edge case to me. If I request a post preview, I usually do it on the post I’m editing. As soon as I click on a link when viewing the preview post, I’m not in preview mode anymore. So to me it would only make sense if the previewed post template also contained a list of posts where the previewed post itself appears as well. And in that list, the updated title, excerpt or content should show as well.

Or is there another use case?

One way to solve this would be to remove this functionality and see whether an issue pops up in the future. If it’s relevant, we could add it again.

## Testing

Tests already exist.


nlemoine (approved):
> One way to solve this would be to remove this functionality and see whether an issue pops up in the future. If it’s relevant, we could add it again.

I'm standing on this side. Further more, that's not something WP does natively. I don't think we should have such feature.

jarednova:
Thanks @gchtr and @nlemoine. In regards to the 🔥 hot topic 🔥 of post previews I'm simialrly at a loss. All the necessary preview loading of a post that one is looking at is already occuring in https://github.com/timber/timber/blob/2.x/lib/Post.php#L293-L346

I'm similarly unable to understand the use case based upon the comments left in https://github.com/timber/timber/commit/d137a698b00a0823c3b4a4379d59110a86ad0c77 

I agree that we should remove and restore if it has indeed broken some kind of essentiafunctionalityiy (at which point, also including the necessary supporting tests, etc.)

Issue #2572, opened 2022-04-12
2.x Ensure class map tests don’t use `assertInstanceOf()` to test the usage of a specific class 


gchtr:
- [ ] Check all class map tests to make sure that they don’t use `$this->assertInstanceOf()` but instead use `get_class()`. This way, we can ensure that a specific class is used. See https://github.com/timber/timber/pull/2517#discussion_r837078733.

PR #2591, opened 2022-06-20
2.x Use assertTrue() and get_class() instead of assertInstanceOf() in certain tests

gchtr:
**Ticket**: #2572

## Issue

#2572

## Solution

Update the tests where class maps are tested.

## Impact

Found an instance in the user class maps where a wrong class was returned because the CoAuthors Plus Integrations adds a filter that is also applied in the user tests:

https://github.com/timber/timber/blob/69aebbf42b9df6661129a6d92f6290691aa58529/lib/Integration/CoAuthorsPlusIntegration.php#L22-L24

## Usage Changes

None.

## Considerations

None.

## Testing

Yes.


jarednova:
good catch, ty!

Issue #2599, opened 2022-06-24
Add a .git-blame-ignore-revs file


gchtr:
**Ticket**: [#2592](https://github.com/timber/timber/pull/2592#pullrequestreview-1016411727)

> I recently stumbled over the possibility to ignore certain commits in blame views: [docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view). Back when we first talked about introducing coding standards, we feared that it would make it harder to find where code was changed. Would it make sense to add a `.git-blame-ignore-revs` file? We wouldn’t have to squash the commits then.

- [x] Add a `.git-blame-ignore-revs` file with code style commits from #2592.

PR #2606, opened 2022-07-13
Add a .git-blame-ignore-revs file

gchtr:
**Ticket**: #2599

## Issue

Coding style commits make it harder to use git blame.

## Solution

Ignore certain coding style commits with a .git-blame-ignore-revs file.

## Impact

Better navigation through code history.

## Usage Changes

None.

## Considerations

Can’t think of any.

## Testing

No.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,17 @@
+# Ignore commits from blame views.
+
+# Renamed lib/ to src/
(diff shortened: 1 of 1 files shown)


jarednova:
Oh nice! I've never seen that before. That will make some of those big moves less chaotic and disruptive

jarednova (approved):

Issue #2607, opened 2022-07-19
Caption on thumbnails is empty (2.x)


xavivars:
When trying to use a post's thumbnail caption is empty. The problem seems to be from Attachment class, that never sets a value for it.

## Expected behavior
Timber displays the caption

## Actual behavior
Caption is empty. It seems to never get defined in https://github.com/timber/timber/blob/2.x/src/Attachment.php

## Steps to reproduce behavior
On `single.twig` for a post with a featured image that has a caption, access ` {{ post.thumbnail.caption }}` and will be null.

## What version of WordPress, PHP and Timber are you using?
WordPress 6.1, PHP 7.4, Timber 2 alpha 6

## Workaround

`{{ post.thumbnail.post_excerpt }}` displays the right caption

gchtr:
Resolved by #2612. This will be in the next release.

PR #2610, opened 2022-07-19
Setup  on attachment

xavivars:
**Ticket**: 
Fixes: https://github.com/timber/timber/issues/2607

## Issue
Caption on images is empty, and not intitialized.

## Solution
Implements `Attachment::build` method that will be used in `PostFactory` so `caption` is defined.

## Impact
It's really a bugfix.

## Considerations
Not 100% sure this is the way that it's meant to be implemented. An alternative (that would break backwards compatibility) would be to just remove the field, and document it so `post_excerpt` is used instead.

## Testing
I didn't add any test.

### DIFF
--- a/src/Attachment.php
+++ b/src/Attachment.php
@@ -116,6 +116,22 @@ class Attachment extends Post
     public $caption;
 
     /**
+     * Create and initialize a new instance of the called Post class
+     * (i.e. Timber\Attachment or a subclass).
+     *
+     * @internal
+     * @return Timber\Attachment
+     */
+    public static function build(WP_Post $wp_post): self
+    {
+        $post = parent::build($wp_post);
+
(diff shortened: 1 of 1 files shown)


gchtr:
Hey @xavivars, thanks for reporting the issue and making this pull request. This is really something we missed when working on the factories as well as splitting `Timber\Image` into the two separate `Timber\Image` and `Timber\Attachment` classes. Nice catch!

> Not 100% sure this is the way that it's meant to be implemented. An alternative (that would break backwards compatibility) would be to just remove the field, and document it so `post_excerpt` is used instead.

I would keep `{{ image.caption }}` and not document that it’s saved in the post excerpt. I even think we should use a `caption()` method instead of the property and added a pull request that iterates on your pull request in #2612.

PR #2611, opened 2022-07-20
2.x Reintroduce Timber’s WP-CLI integration

### LINKED ISSUE #2364: ⁉️ COMMUNITY INPUT WANTED: Timber WP-CLI support ⁉️
We need input from the Timber community to answer a very simple question:

**Do you use the `wp timber` command?**

Please react with :+1: if you do, :-1: if you don't, and as always, feel free to leave a comment. We want to weight the usefulness of this WP-CLI tooling against its maintenance cost.

gchtr:
**Ticket:** #2364

## Issue

In the [2.0 purge](#2323) (also see #2353), the WP-CLI integration for Timber was removed. After asking the community in #2364, we decided we should bring it back, but with proper documentation.

## Solution

This pull request:

- Resolves #2364.
- Restores the changes made in #2353.
- Updates the CLI to match the new integration logic (https://github.com/timber/timber/commit/5aca2dec0d77f1c6d4c89b5ad47f5c42820c8f2d).
- Add Timber’s CLI as a default integration (https://github.com/timber/timber/commit/3981410edc5568b7f73f8f82b0862b6ccea7f15f).
- Removes the `Timber\Integration\Command` class and moves all its methods to `Timber\Cache\Cleaner` (https://github.com/timber/timber/commit/657e533196d342d0f316e733c047ac3b7b8eda06).
- Moves some repetitive object caching checks into a separate `Loader::is_object_cache()` method (https://github.com/timber/timber/commit/a693b58d779dfde87d54170144e4714458582a71).
- Improves return values for some caching functions (https://github.com/timber/timber/commit/81eca24a3a05ea9ea00bb82016d1cc3f49341c75). This makes the CLI return proper success and fail messages depending on whether the cache is actually enabled for either Timber or Twig.
- Adds tests (https://github.com/timber/timber/commit/5f633728418b7996a83437faf4230d65d86cd755).
- Adds a new WP-CLI Guide (https://github.com/timber/timber/commit/bb4d6973c9f08a58bd74ad8bd6528049ab328001).

## Impact and Usage Changes

Described in the updates to the Upgrade Guide (https://github.com/timber/timber/commit/1db521fb43c079b4fc2d23df069c711d910d29f8).

### Testing WP CLI commands

For testing the CLI command, I pretty much followed what Iain Poulson described in https://polevaultweb.com/2016/11/unit-testing-custom-wp-cli-commands/, with minor updates.

First, I added WP-CLI as a Composer dependency. But then I [found an issue](https://github.com/timber/timber/issues/687) that explains how having WP-CLI as a Composer dependency leads to problems in environments where WP-CLI is already loaded. So I switched from requiring WP-CLI as a Composer dependency to loading it in **bin/install-wp-tests.sh** and requiring the PHAR archive in **tests/bootstrap.php**. I’m not really sure whether this is a good solution, but it seems to work.

### More commands

There’s ideas for more commands in #2364, which could be added in the future.

## Testing

Yes. The tests don’t test actually running the commands from the command line. Jared already tried that in https://github.com/timber/timber/commit/c85b643d7c3e2ae98a68b188fbae316073014110. I didn’t try adding it. Do we need it?


src/Cache/Cleaner.php
```diff
@@ -2,8 +2,47 @@
 
 namespace Timber\Cache;
 
+use Timber\Loader;
+
 class Cleaner
 {
+    public static function clear_cache($mode = 'all')
+    {
+        if (is_array($mode)) {
```
nlemoine, line 11: Any reason `$mode` could be an array?

src/Integration/WpCliIntegration.php
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Class WpCliIntegration
+ *
+ * Adds a "timber" command to WP CLI.
+ */
+class WpCliIntegration implements IntegrationInterface
+{
+    public function should_init(): bool
+    {
+        return defined('WP_CLI') && WP_CLI;
```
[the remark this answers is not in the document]
gchtr: Good catch, thanks! 👍 Updated in [`8569f1c` (#2611)](https://github.com/timber/timber/pull/2611/commits/8569f1cc9e93ba66800554f1c0bd1e6e0d707c56).

src/Cache/Cleaner.php
gchtr: Probably not. I updated the function signature in 062540dd359e8598aeb3cf3fd8a54d38bf2e21dd and added a note to the Upgrade Guide.

nlemoine (commented):
The only comment I have left on this PR is related to the command names which may seem unfamiliar regarding the WP-CLI world. I'm not a huge fan of how these commands are structured (too many positional arguments, etc.) but consistency should always win over personal preferences IMHO.

I would rename them as:
- `wp timber clear-cache`
- `wp timber clear-cache twig`
- `wp timber clear-cache timber`

This would make WP-CLI users feel [more like at home](https://github.com/maheshwaghmare/wp-cli-commands-cheat-sheet).

I understand this implementation is not as easy as providing a class and letting method names be command names, please feel free to ignore my comment if you don't think it's relevant 😉

gchtr:
@nlemoine That’s a great input! I just renamed the command in 5ab06078e71f25fb98f35a4d52715a73bc976df6 and updated the documentation. I like this much better.

The `@alias clear_cache` in the DocBlock of the command will make the main command to clear all caches backwards-compatible with `wp timber clear_cache`.

nlemoine (commented):
This looks good to me, thanks @gchtr!

nlemoine (approved):

PR #2612, opened 2022-07-20
2.x Remove `Attachment::$caption` property in favor of `Attachment::caption()` method

### LINKED ISSUE #2607: Caption on thumbnails is empty (2.x)

gchtr:
**Ticket**: #2610

- Fixes #2607

## Issue

This pull request iterates on the work by @xavivars in #2610 (thanks, @xavivars). For 2.x, we missed that the caption of an attachment is set while initializing it. See this excerpt from 1.x:

https://github.com/timber/timber/blob/8d9bff23d2677652469badb75dcb607449a18204/lib/Image.php#L215-L217

## Solution

This pull request:

- Adds a test to confirm the bug in #2607.
- Removes the `Attachment::$caption` property and instead adds an `Attachment::caption()` method. I think this is the approach we choose more and more: Using methods instead of countless properties we have to initialize.
- Updates the Upgrade guide.
- Adds the `wp_get_attachment_caption` filter, which is used in the equivalent WordPress core function [`wp_get_attachment_caption()`](https://developer.wordpress.org/reference/functions/wp_get_attachment_caption/).

## Impact

This doesn’t break backwards compatibility, even if `Timber\Image::$caption` is used in PHP, because `Timber\Core::__get()` will automatically resolve to `Timber\Image::caption()`.

## Usage Changes

None.

## Considerations

Yes, `caption()` belongs in `Timber\Attachment` and not `Timber\Image`, because you can set captions in all kinds of attachments in WordPress, not just images.

## Testing

Yes.

### DIFF
--- a/docs/v2/upgrade-guides/2.0.md
+++ b/docs/v2/upgrade-guides/2.0.md
@@ -905,7 +905,7 @@ The following functions and properties were **removed from the codebase**, eithe
 - `get_src()` – use `{{ image.src }}` instead
 - `get_url()` – use `{{ image.src }}` instead
 - `url()` – use `{{ image.src }}` instead
-- `$caption` property – use `caption()` method instead. You can still use `{{ image.caption }}` in Twig.
+- `$caption` property – use `caption()` method instead. You can still use `{{ image.caption }}` in Twig. In PHP `$image->caption` also still works because magic properties will automatically resolve to `$image->caption()`.
 
 ### Timber\MenuItem
 
--- a/docs/v2/upgrade-guides/2.0.md
+++ b/docs/v2/upgrade-guides/2.0.md
@@ -848,7 +848,7 @@ The following functions are being deprecated and will be removed in a future ver
 
 ## Removed functions and properties
 
-The following functions were **removed from the codebase**, either because they were already deprecated or because they’re not used anymore.
+The following functions and properties were **removed from the codebase**, either because they were already deprecated or because they’re not used anymore.
 
 ### Timber\Timber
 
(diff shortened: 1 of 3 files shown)


xavivars:
Much better approach!

gchtr:
@nlemoine I’d be happy about a quick review here, too 😊, thanks!

nlemoine (approved):
I didn't know hummingbirds can’t walk! Thanks for that fun fact. Here's another one I just stumbled upon about `.ch` TLD: https://shkspr.mobi/blog/2022/07/dns-esoterica-why-you-cant-dig-switzerland/

Everything looks good :)

gchtr:
@nlemoine Yeah right? I found tests are a good place to hide fun facts, favorite movie quotes or other phrases that are stuck in your mind.

Whoa, and every day we learn something new. Didn’t know about the Chaosnet network protocol and how it’s connected to `.ch` a.k.a "Chaos" domains 😆.

Back on topic: Let’s merge this in, then! 🥂

PR #2613, opened 2022-07-20
2.x Add missing query() method to `Timber\PostQuery`

### LINKED ISSUE #2605: v2.0 A way to get the query object from Timber::get_posts()
Back in the v1 there was a way to get the currently running query object via `posts.get_query()` function. I've noticed that was removed from the new version without any notice in the upgrade guide. Is there any alternative for that in the v2.0?

I'd love to be able to grab the current query from the posts object with `posts.query` just like we can with `posts.pagination`.

I've tried grabbing the `$wp_query` via custom Twig function but it only works if you're using the default archive query (not a custom one).

## Additional context
I'd like to build a custom pagination utilising infinite loading and to do that I need to be able to pass the current `query_vars` as I'm reusing the same pagination component in multiple places so the scripts must know what the current query actually is.

gchtr:
**Ticket**:

- Fixes #2605.

## Issue

The [`get_query()` method](https://github.com/timber/timber/blob/8d9bff23d2677652469badb75dcb607449a18204/lib/PostQuery.php#L50-L56) didn’t make it over to 2.x.

## Solution

This pull request:

- Adds the missing method and renames from `get_query()` to `query()` to match the naming convention.
- Deprecates the `get_query()` method.
- Adds a test.
- Updates the Upgrade Guide.

## Impact and Usage Changes

Mentioned in the Upgrade Guide.

## Considerations

–

## Testing

Yes.

### DIFF
--- a/src/PostQuery.php
+++ b/src/PostQuery.php
@@ -133,7 +133,7 @@ class PostQuery extends ArrayObject implements PostCollectionInterface, JsonSeri
      * @since 2.0
      * @return WP_Query|null
      */
-    public function query()
+    public function query(): ?WP_Query
     {
         return $this->wp_query;
     }
@@ -144,7 +144,7 @@ class PostQuery extends ArrayObject implements PostCollectionInterface, JsonSeri
      * @deprecated 2.0.0, use PostQuery::query() instead.
      * @return WP_Query|null
      */
-    public function get_query()
+    public function get_query(): ?WP_Query
     {
         Helper::deprecated('Timber\PostQuery::get_query()', 'Timber\PostQuery::query()', '2.0.0');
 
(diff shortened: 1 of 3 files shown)


src/PostQuery.php
```diff
@@ -127,6 +127,30 @@ public function pagination($prefs = [])
...
+        return $this->wp_query;
+    }
+
+    /**
+     * Gets the original query used to get a collection of Timber posts.
+     *
+     * @deprecated 2.0.0, use PostQuery::query() instead.
+     * @return WP_Query|null
+     */
+    public function get_query()
```
nlemoine, line 147: Maybe start add return types for things we are pretty confident with? `: ?WP_Query`

gchtr: Yes 👍! I added it in ee3537f5d5c54763588f2364fac2ef54042606ef.

nlemoine (approved):

PR #2614, opened 2022-07-25
Use `yoast/wp-test-utils`

nlemoine:
- Simplifies `tests/bootstrap.php` and its maintenance
- Makes possible the addition of "pure"/faster unit tests (≠ integration tests) with BrainMonkey
- Update `Timber_UnitTestCase`

## Impact
None. Just getting some errors on WPML integration tests, not sure this is related.

## Usage Changes
None.


tests/test-timber-dates.php
```diff
@@ -286,7 +286,7 @@ public function testTimeAgoFutureTranslated()
             $this->change_locale('de_DE');
 
             $str = DateTimeHelper::time_ago('2016-12-01 20:00:00', '2016-11-30, 20:00:00');
-            $this->assertEquals('1 Tag ab jetzt', $str);
+            $this->assertEquals('1 day from now', $str);
```
gchtr, line 289: This test shouldn’t change, because we specifically want to test whether a German string is returned by `DateTimeHelper::time_ago()` here (see #2215).

If this is returning the English translation, it’s probably because the German translation wasn’t installed properly in the bootstrapping process.

The following line in **bootstrap.php** is only related to WPML in the sense that it’s also about translations. But it’s not directly related to WPML.

```php
// Make sure translations are installed.	
Timber_UnitTestCase::install_translation('de_DE');
```

nlemoine: I know this isn't related to WPML. I just tried to strip down the bootstrap file to the bare minimum and ignored that was used somewhere :)

However, I still see some major downsides to downloading whole language packs:
- Translations may change across WP versions and break tests (see how [WP includes lightweight translation files directly in tests folder](https://github.com/WordPress/wordpress-develop/tree/trunk/tests/phpunit/data/languages)). 
- Although it doesn't change much, we have to maintain the package installer function
- It slows down the tests

I removed the language pack installer function and added the latest

gchtr: If we can find a better solution that also make the tests faster, then I’m all for it.

> * Translations may change across WP versions and break tests (see how [WP includes lightweight translation files directly in tests folder](https://github.com/WordPress/wordpress-develop/tree/trunk/tests/phpunit/data/languages?rgh-link-date=2022-07-25T13%3A45%3A48Z)).

I added that functionality initially because we use the WordPress core translations. I didn’t want to add a `timber` text domain, because then we would have to manage a whole set of translations. And of course the core translations can change, which is exactly the reason I added the tests so we can detect these changes.

> I removed the language pack installer function and added the latest translation files in the tests folder.

I like that approach much better. But this way, we wouldn’t see when a translation changed, right? We would have to update the translations manually.

I see that the translation files are also downloaded into the temporary **wordpress-tests-lib/data/languages** folder. Maybe we could load or copy them from there?

nlemoine: > But this way, we wouldn’t see when a translation changed, right?

Do you mean the translation key (%s ago) or translation string? I don't think we should watch for translation change. Using core translation is usually discouraged. The test here is only asserting that the string is translated, hence having a "hardcoded" translation file.
Further more, translations are tied to a WP version. If some translation changed, we would need to do some version comparing to load the right translation. That's a path I'd rather not take.

> Maybe we could load or copy them from there?

Translations in the data/languages directory [are incomplete](https://github.com/WordPress/wordpress-develop/blob/trunk/tests/phpunit/data/languages/admin-es_ES.po). They just serve testing purpose, as explained above.

gchtr: > Do you mean the translation key (`%s ago`) or translation string (the actual translation)?

I would want to see when the translation key changes. I know core translations are discouraged. I tried to explore some options in https://github.com/timber/timber/issues/2214 and using core translations seemed like the best compromise that doesn’t involve developers to provide the translations.

> The test here is only asserting that the string is translated, hence having a "hardcoded" translation file.

Yes, because having untranslated strings was the initial issue in #2214.

> Further more, translations are tied to a WP version. If some translation changed, we would need to do some version comparing to load the right translation. That's a path I'd rather not take.

We talk about strings here that are very unlikely to change:

- `__('%s ago')`
- `__('%s from now')`
- `__('&laquo; Previous')`
- `__('Next &raquo;')`
- `__('&hellip;')`

If they should change, I’m sure we could find a better solution then.

> Translations in the data/languages directory [are incomplete](https://github.com/WordPress/wordpress-develop/blob/trunk/tests/phpunit/data/languages/admin-es_ES.po?rgh-link-date=2022-07-25T15%3A32%3A29Z). They just serve testing purpose, as explained above.

Oh, I see. Maybe we could just add a translation with the strings that Timber uses, then. I could prepare that, if you like. In the unlikely case the translation should change, we would see it in the tests.

What would your alternative solution look like? I would love to have a better approach that doesn’t put the burden on Timber developers 😊.

nlemoine: Thanks for your comments and especially the history details of this issue!

I think we're good right? If `%s ago` would change, then the string wouldn't be translated in german anymore and this would be caught by tests.

> I could prepare that, if you like

I've updated the translations and included a composer script command to update it. 

Note that a translation is missing from WordPress:
```
#: Archives.php:190
msgid "%1$s"
```

gchtr: > I think we're good right? If `%s ago` would change, then the string wouldn't be translated in german anymore and this would be caught by tests.

Exactly. Yes, I think with that, we’re good 😊. Thanks for your work on this!

> Note that a translation is missing from WordPress:
> 
> ```
> #: Archives.php:190
> msgid "%1$s"
> ```

Yeah, it’s actually weird to have a translation that is just a placeholder, right? Do we want to remove this one?

nlemoine: I think we do.

gchtr (approved):
Yet another step to a more robust testing setup. This looks great, thank you Nicolas! 🥳

PR #2615, opened 2022-07-26
2.x Add wp_object() method and $wp_object property for Core objects

### LINKED ISSUE #2560: Add $wp_object property for Timber core objects
**Ticket: [#2501](https://github.com/timber/timber/issues/2501#issuecomment-948792581)**

In various cases we need to get properties from WordPress core objects. For example, in a post, we need to use the `WP_Post` object to maybe overwrite the `$post` global for better compatibility within loops (#2545).

Or we have to access properties like `post_title` from the original post. Instead of saving all the values in separate properties in a Timber object, we could introduce a new `$wp_object` property for Timber core entities like `Timber\Post` or `Timber\MenuItem`.

In https://github.com/timber/timber/issues/2501#issuecomment-948792581, @nlemoine made a nice example for this:

```php
<?php

class Post {
    protected WP_post $wp_object = $post;

    public function core_object() : WP_Post {
        return $wp_object;
    }

    public function title() {
        return $wp_object->post_title;
    }

    public function date() {
        return new DateTimeImmutable( $wp_object->post_date );
    }

    // etc.
}
```

gchtr:
**Ticket**

- Resolves #2560

## Issue

We have no reference to the original WordPress core object for most of the Timber objects.

## Solution

This pull request:

- Adds a `wp_object()` method and a protected `$wp_object` property for all Timber objects that extend `Timber\CoreEntity`. I didn’t add the `wp_object()` method to `CoreEntity`, because then we couldn’t have the proper return types, which are different for each object.
- Adds a `CoreEntityInterface` with the `wp_object()` method. The interface doesn’t specify a return type, because the types are different for each object that implements the interface. I didn’t choose the `CoreInterface` to add the `wp_object()` method, because there are other classes that extend `Core` where it wouldn’t make sense to have a `wp_object()` method, like `Timber\Archives`.

## Impact

This shouldn’t add another breaking change, because the `CoreEntity` class is new in 2.0.

## Usage Changes

In cases where you need to use original WordPress object, we can now use `{{ object.wp_object.something }}`.

## Considerations

I just assumed that not setting the return type on an interface is fine. I don’t know whether this is bad practice or not.

We can also start using `$wp_object` for certain methods in Timber’s core, as discussed in #2560. However, in cases where a public property can be updated before `$wp_object` is used, this might be a breaking change.

For example: If we used `$wp_object` in the `Timber\Post::title()` method, then you wouldn’t be able to change the public `$post_title` property in `Timber\Post` before you use `Timber\Post::title()`.

```php
namespace Timber;

class Post
{
    public $post_title;

    public function title() {
        return apply_filters( 'the_title', $this->wp_object->post_title );
    }
}
```

```php
$post = Timber::get_post();

// "Something"
echo $post->post_title;

$post->post_title = 'Overwritten';

// "Something"
echo $post->title();
```

## Testing

Not yet. I first wanted to get some feedback on whether the logic is right.

## Todo

- [x] Add tests
- [x] Add documentation


gchtr:
@nlemoine I’d be glad if you could check out the changes here before I write the tests and the documentation.

nlemoine (dismissed):

nlemoine:
> However, in cases where a public property can be updated before $wp_object is used, this might be a breaking change.

It depends how you consider those properties. I personally think they should be `readonly` and you should not be able to change them. Since we're not using any ORM (`$orm->persist($post);`), I don't really see the point in changing model properties. Moreover, filters are here for this kind of purpose.

IMO, Timber is too permissive regarding properties right now (everything is `public`) and I'd like to move to stricter models. This would leverage easier memoization, etc.

But we'll be talking about this when we start writing 3.x branch ;)

gchtr:
> IMO, Timber is too permissive regarding properties right now (everything is `public`) and I'd like to move to stricter models. This would leverage easier memoization, etc.
> 
> But we'll be talking about this when we start writing 3.x branch ;)

Totally agree with you. And I also think that’s a problem to tackle in 3.x.

I added some tests and a note in the Upgrade Guide and I’m going to merge this pull request now.

Issue #2596, opened 2022-06-22
Implement "toavif" filter to convert images to .avif format


radex02:
## Is your feature request related to a problem? Please describe.
The ``avif`` image format has better compression than even the ``webp`` format and is now supported by [70% of browsers](https://caniuse.com/avif). More importantly, it can now be generated by GD starting from ``PHP 8.1`` using ``imageavif``. I have a lot of image heavy websites and this format could most likely reduce their bandwidth usage by ~25%.

## Describe the solution you’d like
I'd like a filter just like the ``tojpeg`` or ``towebp`` filter but for ``avif``.

jarednova:
Thanks @radex02 ! We're actually in the midst of talking about the best way to handle image transformation in the repos so we can be more flexible about what support is/isn't included — I know @nlemoine has some thoughts

Levdbas:
Closing this in favor of #2866

PR #2628, opened 2022-08-08
Added AVIF support

joebailey26:
**Ticket**: [#2596](https://github.com/timber/timber/issues/2596)

## Issue
Added a ```toavif``` filter for even smaller file compression

## Solution
Duplicated the ```towebp``` code but instead used PHP 8.1's native ```imageavif``` function.

## Impact
If ```imageavif``` isn't available, it should fail gracefully ensuring backwards compatibility. 

## Usage Changes
I've added in the docs for the ```toavif``` filter.

## Testing
I've duplicated the WEBP tests as they should be the same for AVIF.

### DIFF
--- a/docs/guides/cookbook-images.md
+++ b/docs/guides/cookbook-images.md
@@ -83,6 +83,19 @@ Similar to the `|tojpg` filter, there’s a `|towebp` filter.
 </picture>
 ```
 
+### Using AVIF images
+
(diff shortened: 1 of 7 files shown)


Levdbas:
Original PR was removed so closing this as well. Thanks for your effort @joebailey26

PR #2629, opened 2022-08-15
2.x Update .gitattributes

gchtr:
Some files are still downloaded when non-dev version of Timber is requested through Composer. This pull requests updates the **.gitattributes** file to not distribute these files to production releases.


gchtr:
Duplicate of #2650.

PR #2630, opened 2022-08-15
2.x Update PHPStan setup

gchtr:
## Issue

There are some PHP libraries we can use to provide function stubs for PHPStan instead of adding them ourselves.

## Solution

Use [php-stubs/acf-pro-stubs](https://github.com/php-stubs/acf-pro-stubs) and [php-stubs/wp-cli-stubs](https://github.com/php-stubs/wp-cli-stubs).

For now, this isn’t working because the function signature for `acf_get_field_type()` in [php-stubs/acf-pro-stubs](https://github.com/php-stubs/acf-pro-stubs).

```
  Line   Integration/AcfIntegration.php
  244    Result of function acf_get_field_type (void) is used.
  245    Result of function acf_get_field_type (void) is used.
  246    Result of function acf_get_field_type (void) is used.
  247    Result of function acf_get_field_type (void) is used.
  248    Result of function acf_get_field_type (void) is used.
  249    Result of function acf_get_field_type (void) is used.
  250    Result of function acf_get_field_type (void) is used.
  251    Result of function acf_get_field_type (void) is used.
  252    Result of function acf_get_field_type (void) is used.
```

I guess it’s because ACF uses a return type of `n/a` in the DocBlock for `acf_get_field_type()`, which is translated to `void`. But actually it should return `mixed`, when we look at the underlying `acf()->fields->get_field_type()` function.

For an easy fix, we could switch to using `acf()->fields->get_field_type()` instead of `acf_get_field_type()`, but maybe this is an issue that can be fixed in the [php-stubs/acf-pro-stubs](https://github.com/php-stubs/acf-pro-stubs) library.

I hope that @szepeviktor can help us out here 😊.

## Impact

Less code to manage.

## Usage Changes

None.


szepeviktor:
> Use [php-stubs/acf-pro-stubs](https://github.com/php-stubs/acf-pro-stubs) and [php-stubs/wp-cli-stubs](https://github.com/php-stubs/wp-cli-stubs).

I suspect these are fine packages :)

gchtr:
@szepeviktor Well, they apparently were made by someone who know his stuff 😉💪

szepeviktor:
> I guess it’s because ACF uses a return type of `n/a` in the DocBlock

I contacted Elliot and he was happy about docblock fixes but refused them.
Currently invalid docblocks are _dumbfixed_ by string replacement.
https://github.com/php-stubs/acf-pro-stubs/blob/7e44b653bd72e5eebf725f3af6b5afaae3039dcf/generate.sh#L30-L43

Fixing those could be also done with `sed`.

szepeviktor:
I would gladly apply a patch on top the released stubs if you provide me fixing functions in `includes/fields.php`.
That would go like this: `patch -p 1 < acf-pro-field-function-fixes.patch`

gchtr:
@szepeviktor Thanks for your insights!

> I contacted Elliot and he was happy about docblock fixes but refused them.

I see 🙃.

> Currently invalid docblocks are _dumbfixed_ by string replacement.
> [php-stubs/acf-pro-stubs@`7e44b65`/generate.sh#L30-L43](https://github.com/php-stubs/acf-pro-stubs/blob/7e44b653bd72e5eebf725f3af6b5afaae3039dcf/generate.sh#L30-L43)

Yes, I saw that :). I’m not familiar enough with `sed` though.

> I would gladly apply a patch on top the released stubs if you provide me one fixing functions in `includes/fields.php`.
> That would go like this: `patch -p 1 < acf-pro-field-function-fixes.patch`

So, you mean I could provide you with a `.patch` file that fixes that particular function signature and you could apply that when generating the stubs? I think I could prepare that.

szepeviktor:
> I think I could prepare that.

All right!
To be clear: please start with the stubs.

szepeviktor (dismissed):
Very nice work 💎

gchtr:
I decided to go the easy way and just use `acf()->fields->get_field_type()` instead of `acf_get_field_type()` to make it work with the [php-stubs/acf-pro-stubs](https://github.com/php-stubs/acf-pro-stubs) package.

PR #2631, opened 2022-08-15
2.x Make Timber\Request not implement Timber\CoreInterface

gchtr:
## Issue

While working on fixing some PHPStan issues, I found that the `Timber\Request` class extends `Timber\Core` and implements `Timber\CoreInterface`.

In the current state of Timber, I don’t see any reason why it should do so. Even when it was introduced in #853, it wasn’t clearly stated why it should extend `Timber\Core`.

The method required by `Timber\CoreInterface` are added as empty functions and they look like they were added only to serve the interface. But this request helper class doesn’t have anything to do with a Timber core object that usually matches core functionality in WordPress.

## Solution

Make `Timber\Request` not implement `Timber\CoreInterface`.

## Impact

This might break backwards compatibility with any class that extends `Timber\Request`. We will have to add a note to the Upgrade Guide.

## Usage Changes

None.

## Considerations

We could move this to a future release and not release this breaking change in 2.x. In that case, we would have to add a return to `__isset()` to fix the PHPStan issue.

I’d really like to get a second opinion on this one.

## Testing

No.

### DIFF
--- a/src/Request.php
+++ b/src/Request.php
@@ -5,9 +5,9 @@ namespace Timber;
 /**
  * Class Request
  *
- * Timber\Request exposes $_GET and $_POST to the context
+ * Timber\Request exposes $_GET and $_POST to the context.
  */
-class Request extends Core implements CoreInterface
+class Request
 {
     public $post = [];
 
@@ -15,7 +15,6 @@ class Request extends Core implements CoreInterface
 
     /**
      * Constructs a Timber\Request object
-     * @example
      */
     public function __construct()
     {
@@ -29,24 +28,5 @@ class Request extends Core implements CoreInterface
     {
         $this->post = $_POST;
         $this->get = $_GET;
-    }
-
-    public function __call($field, $args)
-    {
-    }
-
-    public function __get($field)
-    {
-    }
-
-    /**
-     * @return boolean|null
-     */
-    public function __isset($field)
-    {
-    }
-
-    public function meta($key)
-    {
     }
 }


nlemoine:
> I’d really like to get a second opinion on this one.

I would personally remove that class:
- It's not used internally in Timber
- It doesn't really provide any features
- It's (very) incomplete: it's missing cookies, files, headers, etc.
- It's not following any standard (PSR-7)
- There's plenty of maintained/well tested libraries that provides a Request object (https://github.com/laminas/laminas-diactoros, https://github.com/Nyholm/psr7, https://github.com/guzzle/psr7, etc.)

IMHO, 2.0 would be a good occasion to get rid of it :) Further more, as long as the class isn't used inside Timber, I don't see any valid argument to keep it, especially now we're composer only. If a user wants a request object, let's add a quick how to in the docs:

```php
// composer require nyholm/psr7-server

add_filter('timber/context', function($context) {
    $psr17Factory = new \Nyholm\Psr7\Factory\Psr17Factory();

    $creator = new \Nyholm\Psr7Server\ServerRequestCreator(
        $psr17Factory, // ServerRequestFactory
        $psr17Factory, // UriFactory
        $psr17Factory, // UploadedFileFactory
        $psr17Factory  // StreamFactory
    );

    $context['request'] = $creator->fromGlobals();

    return $contetx;
});

```

Now, if we have to keep it, I'll all for leaving `CoreInterface` out of it but as you said, it's so thin we would have to enhance it which would increase the code base to maintain.

nlemoine (approved):

gchtr:
@nlemoine I’m all for removing this 👍. Thank you for providing a different solution using one of the PSR-7 libraries. I added a PR in https://github.com/timber/timber/pull/2683.

Issue #2532, opened 2022-01-26
Timber\User `can` method does not match user_can function signature


kshaner:
The Timber\User::can method does not pass the arbitrary arguments to the WordPress core `user_can` function. This creates mismatched expectations for what the method can do.

## Expected behavior

I would expect that all Timber methods that are facades to WordPress core functions and use the same argument signatures.

## Actual behavior
Some capability checks always fail due to the lack of arguments. Certain capabilities that require arguments (edit_post, edit_user, etc) produces a PHP notice: 

```
Notice: Undefined offset: 0 in /var/www/wp-includes/capabilities.php on line 149
```

## Steps to reproduce behavior

This code does not work because `post.id` is not passed along to the `user_can` function.

```twig
{% if user.can( 'edit_post', post.id ) %}
    {# do something with privileges #}
{% endif %}
```

## Proposed change

Spread the arguments into the can method of the Timber\User class.

```php
	public function can( $capability, ...$args ) {
		return user_can( $this->ID, $capability, ...$args );
	}
```

## What version of WordPress, PHP and Timber are you using?
WordPress 5.8.3
PHP 7.4.27
Timber v2.14.7

## How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
composer require timber/timber:^2.0

gchtr:
Good catch! And thanks for taking the time set up such a clean issue. We should definitely pass those additional arguments down to `user_can()`.

kshaner:
@gchtr what is the official php version support for Timber at this point? I was going to make the PR but the composer.json still specifies 5.3+ and the spread operator is 7.4+. If it's older than 7.4 I'll go ahead and do a call_user_func implementation.

gchtr:
Hey @kshaner. It would be great if you could do a PR for this. Even if it’s without the tests. We can add those while reviewing the PR.

For Timber version 2, we will only support PHP 7.4+. For Timber version 1, we didn’t end support for version below 7.4 yet. We don’t want to invest too much time into 1.x, because we’re already thin on resources and want to make Timber v2 ready as soon as possible.

If you want to use the spread operator, I think we should add this to version 2 of Timber. You’ll need to work on the `2.x` branch instead of the `master` branch: see [https://github.com/timber/timber/blob/2.x/](https://github.com/timber/timber/blob/2.x/).

If you want to add it to Timber v1, then `call_user_func()` is the way to go.

gchtr:
Resolved in #2632.

[...]

PR #2632, opened 2022-08-21
2.x user can args

kshaner:
**Ticket**: #2532 

## Issue
fixes issue where `{{user.can('capability', arg)}}` doesn't work because arg is not passed to user_can

## Solution
Spreads arguments from `\Timber\User::can` to `user_can` function 

## Impact
No backward compatibility breaking changes (besides fixing the issue at hand)

## Usage Changes
The following now work:

```twig
{% if user.can('edit_post', post.id) %}
    {# do something with privileges #}
{% endif %}
```

```twig
{% if user.can('edit_term', term.id) %}
    {# do something with privileges #}
{% endif %}
```

```twig
{% if user.can('edit_user', user.id) %}
    {# do something with privileges #}
{% endif %}
```

```twig
{% if user.can('edit_comment', comment.id) %}
    {# do something with privileges #}
{% endif %}
```

## Considerations
The way timber 2.x tends to use this now is with `\Timber\Core::can_edit` https://github.com/timber/timber/blob/8d9bff23d2677652469badb75dcb607449a18204/lib/Core.php#L100-L108

However this method is in `\Timber\Core` and other classes such as `\Timber\User` and `\Timber\Term` also extend `\Timber\CoreEntity` (and thus `\Timber\Core`) and is not re-defined in sub-classes. This means that this code is checking against `edit_post` capability for the can_edit method of non Post classes. This code should should move to their specific sub-classes so there is no possibility of false negatives/false positives when id's match.

## Testing
Tests are written but un-tested. I've had some issues getting the tests to run correctly getting the error `Could not find /includes/functions.php`. Chasing down the error it has to do with the Yoast integration but I've hit a dead end in resolving that.

### DIFF
--- a/src/User.php
+++ b/src/User.php
@@ -374,6 +374,7 @@ class User extends CoreEntity
      * @since 1.8.5
      *
      * @param string $capability The capability to check.
+     * @param mixed ...$args Additional arguments to pass to the user_can function
      *
      * @example
      * Give moderation users another CSS class to style them differently.
@@ -384,11 +385,20 @@ class User extends CoreEntity
      * </span>
      * ```
      *
+     * @example
+     * Show edit link for posts that a user can edit
+     *
+     * ```twig
+     * {% if user.can('edit_post', post.id) %}
+     *     <a href="{{post.edit_link}}">Edit Post</a>
+     * {% endif %}
+     * ```
+     *
      * @return bool Whether the user has the capability.
      */
-    public function can($capability)
+    public function can($capability, ...$args)
     {
-        return user_can($this->ID, $capability);
+        return user_can($this->ID, $capability, ...$args);
     }
 
     /**
--- a/tests/test-timber-user.php
+++ b/tests/test-timber-user.php
@@ -73,9 +73,29 @@ class TestTimberUser extends Timber_UnitTestCase
             'user_login' => 'mbottitta',
             'role' => 'editor',
         ]);
+
+        $subscriber_uid = $this->factory->user->create([
+            'display_name' => 'Baberaham Lincoln',
+            'user_login' => 'blincoln',
+            'role' => 'subscriber',
+        ]);
+
+        $post_id = wp_insert_post(
+            [
+                'post_title' => 'Baseball',
+                'post_content' => 'is fine, I guess',
+                'post_status' => 'publish',
+            ]
+        );
+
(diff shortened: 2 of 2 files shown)


src/User.php
```diff
@@ -384,11 +385,20 @@ public function roles()
      * </span>
      * ```
      *
+     * @example
```
gchtr, line 388: Do you want to add your other examples that you added in the *Usage* section of your pull request here as well?

```twig
{% if user.can('edit_term', term.id) %}
    {# do something with privileges #}
{% endif %}

{% if user.can('edit_user', user.id) %}
    {# do something with privileges #}
{% endif %}

{% if user.can('edit_comment', comment.id) %}
    {# do something with privileges #}
{% endif %}
```

gchtr (approved):
Hey @kshaner 

Congratulations for your first pull request and thanks a lot for putting it together.

The tests seem to run okay. From my side, this is all good. I would just add all the examples you came up with in the DocBlock for `Timber\User::can()`.

And a very good catch with the `\Timber\Core::can_edit()` problem! I just added a new issue in #2645 for this. If you want, you could take this on, too.

src/User.php
```diff
@@ -384,11 +385,20 @@ public function roles()
...
+     * {% endif %}
+     * ```
+     *
      * @return bool Whether the user has the capability.
      */
-    public function can($capability)
+    public function can($capability, ...$args)
     {
-        return user_can($this->ID, $capability);
+        return user_can($this->ID, $capability, ...$args);
```
nlemoine, line 401: Since we're now storing core objects (WP_Post, WP_user, etc.) in our models, we should directly pass `$this->wp_object` in the `user_can` first argument. It would avoid WP getting the user (although it's already cached).

See https://github.com/WordPress/wordpress-develop/blob/2bb5679d666474d024352fa53f07344affef7e69/src/wp-includes/capabilities.php#L790-L802

nlemoine (changes requested):

src/User.php
gchtr: Resolved in #2647

src/User.php
gchtr: Thanks @nlemoine, good catch!

Resolved in #2647.

nlemoine (approved):

Issue #2580, opened 2022-05-25
PHP 8.0 Deprecation Warning


timnolte:
## Expected behavior
Classes updated in order to be compatible with future versions of PHP and not fill up logs with Deprecation warnings.

## Actual behavior
```
[25-May-2022 20:25:42 UTC] PHP Deprecated:  Required parameter $lineno follows optional parameter $only in /var/www/html/public/wp-content/plugins/timber-library/vendor/twig/twig/src/Node/EmbedNode.php on line 26
[25-May-2022 20:25:42 UTC] PHP Deprecated:  Required parameter $lineno follows optional parameter $only in /var/www/html/public/wp-content/plugins/timber-library/vendor/twig/twig/src/Node/IncludeNode.php on line 25
```

## Steps to reproduce behavior
Update a site to run under PHP 8.0.

## What version of WordPress, PHP and Timber are you using?
WordPress 5.9.3, PHP 8.0.16, Timber 1.19.1

## How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Composer via wpackagist.org

okadots:
I am also seeing this deprecation notice
`Deprecated: Method ReflectionParameter::isArray() is deprecated in /var/www/wp-content/plugins/timber-library/vendor/twig/twig/src/Node/Expression/CallExpression.php on line 257`

WP 5.9.3, php 8.0, Timber plugin 1.19.1

G-Rath:
It looks like these are fixed in the latest versions of Twig (across all majors), so this should be fixed when a new version is release (as I assume that gets packed with whatever the latest version of Twig is at the time) - hopefully that'll be soon 🤞

timnolte:
I was hopeful when I saw that 1.20.0 was out that these would be resolve but it doesn't look like it. :-/

JohnnyBravoTheBest:
I am getting lots of deprecated warnings : 

`Deprecated: Return type of Timber\QueryIterator::current() should either be compatible with Iterator::current(): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /www/kinsta/public/udemywtdwtat/wp-content/plugins/timber-library/lib/QueryIterator.php on line 137

[...]

PR #2638, opened 2022-09-02
Add PHP 8.0/8.1 to Timber 1.x matrix

nlemoine:
Related:
- https://github.com/timber/timber/issues/2580
- https://github.com/timber/timber/issues/2548


gchtr:
@nlemoine Thanks a lot for taking this on. Now, what do we need to move this out of draft status?

nlemoine:
Hey @gchtr 

I had some weird issues running tests (some "segmentation fault" with no valuable message). I solved them and finally almost fixed the tests. As you can see, some pass, some don't :
- PHP 7.4 | WP latest (MS) : some problems with the WP-CLI cache clearing
- PHP 7.4 | WP latest (with coverage) : I don't get why the clover file isn't readable, I didn't try much

Would you be interested to take this PR over?

gchtr:
@nlemoine I think this is ready now. I added some fixes for PHP 8.1 deprecation warnings, e.g. when cropping images without rounding the new width and height first.

And I found the fix for the missing clover.xml file in [`a7fdf59` (#2638)](https://github.com/timber/timber/pull/2638/commits/a7fdf59bf3e22dd548e0ead6936abcabdfec2d82).

There’s still a couple of warnings, but they are mostly related to WordPress itself, where we have no control.

Just need a final approval so we can merge this in.

nlemoine:
Good job @gchtr!

jarednova (approved):

jarednova:
@gchtr @nlemoine I confirmed that the tests all pass with the merge of #2640. I was able to do a brief review of the code and new testing elements and am so humbled by how much cleaner and more sophisticated this leaves things. Lots of green checks!!!!

PR #2639, opened 2022-09-13
External image (Fixes #2627)

### LINKED ISSUE #2627: Missing image constructor with file path in 2.x
## Expected behavior
Can create an image from a URL or local path

`$myImage = new Image( get_stylesheet_directory() . '/img/compass.png' );`

## Actual behavior
Constructor for images based on path no longer exists

PHP 7.4, WordPress 6.0, Timber 2.x beta 1

## More information
Images used to have an `init` method that was called from the image constructor.

https://github.com/timber/timber/blob/master/lib/Image.php#L265

All the underlying code is still there (now in `Attachment` class), but it's protected and can't be directly used: 

https://github.com/timber/timber/blob/2.x/src/Attachment.php#L129

xavivars:
**Ticket**: #2627 

## Issue
Adding a `Timber\ExternalImage` class, compatible with `Timber\Image` when using it in Twig, to get feature parity with Timber 1.x

## Solution
A new class, `Timber\ExternalImage`, that implements `Timber\ImageInterface`, compatible with `Timber\Image`

## Impact
There is some common logic that has been extracted from both `Timber\Image` and `Timber\Attachment` into their own classes, to avoid code duplication and being able to use it from `Timber\ExternalImage`.

## Testing
Not yet.

### DIFF
--- a/docs/v2/upgrade-guides/2.0.md
+++ b/docs/v2/upgrade-guides/2.0.md
@@ -831,7 +831,9 @@ The following functions are being deprecated and will be removed in a future ver
 
 - `get_pathinfo()` – use `pathinfo()` instead.
 - `get_dimensions()` – use `width()` or `height()` instead.
-- `get_dimensions_loaded()` – use `get_dimension_loaded()` instead.
+- `get_dimensions_loaded()` – use `width()` or `height()` instead.
+- `get_dimension()` – use `width()` or `height()` instead.
+- `get_dimension_loaded()` – use `width()` or `height()` instead.
 - `get_post_custom()` – use `meta()` instead.
 
 ### Timber\Term
--- a/src/ExternalImage.php
+++ b/src/ExternalImage.php
@@ -1,0 +1,484 @@
+<?php
+
+namespace Timber;
+
+/**
+ * Class ExternalImage
+ *
+ * The `Timber\ExternalImage` class represents an external image loaded via http.
+ *
+ * @api
+ * @example
+ * ```php
+ * $context = Timber::context();
+ *
+ * // Lets say you have an external image that you want to use in your theme
+ *
+ * $context['cover_image'] = Timber::get_external_image($url);
+ *
+ * Timber::render('single.twig', $context);
+ * ```
+ *
+ * ```twig
+ * <article>
+ *   <img src="{{cover_image.src}}" class="cover-image" />
+ *   <h1 class="headline">{{post.title}}</h1>
+ *   <div class="body">
+ *     {{post.content}}
+ *   </div>
+ *
+ *  <img
+ *    src="{{ get_image(post.custom_field_with_image_id).src }}"
+ *    alt="Another way to initialize images as Timber\Image objects, but within Twig" />
+ * </article>
+ * ```
+ *
+ * ```html
+ * <article>
+ *   <img src="http://example.org/wp-content/uploads/2015/06/nevermind.jpg" class="cover-image" />
+ *   <h1 class="headline">Now you've done it!</h1>
+ *   <div class="body">
+ *     Whatever whatever
+ *   </div>
+ *   <img
+ *     src="http://example.org/wp-content/uploads/2015/06/kurt.jpg"
+ *     alt="Another way to initialize images as Timber\Image objects, but within Twig" />
+ * </article>
+ * ```
+ */
+class ExternalImage implements ImageInterface
+{
+    /**
+     * File.
+     *
+     * @api
+     * @var string
+     */
+    private $alt_text;
+
+    protected function __construct($alt)
+    {
+        $this->alt_text = $alt;
+    }
+
+
+    /**
+     * File.
+     *
+     * @api
+     * @var mixed
+     */
+    public $file;
+
+    /**
+     * File location.
+     *
+     * @api
+     * @var string The absolute path to the attachmend file in the filesystem
+     *             (Example: `/var/www/htdocs/wp-content/uploads/2015/08/my-pic.jpg`)
+     */
+    public $file_loc;
+
+
+    /**
+     * Absolute URL.
+     *
+     * @var string The absolute URL to the attachment.
+     */
+    public $abs_url;
+
+    /**
+     * File extension.
+     *
+     * @api
+     * @since 2.0.0
+     * @var null|string A file extension.
+     */
+    public $file_extension = null;
+
+    /**
+     * Formatted file size.
+     *
+     * @api
+     * @since 2.0.0
+     * @var FileSize File size string.
+     */
+    private $file_size = null;
+
+    /**
+     * File types.
+     *
+     * @var array An array of supported relative file types.
+     */
+    private $image_file_types = [
+        'jpg',
+        'jpeg',
+        'png',
+        'svg',
+        'bmp',
+        'ico',
+        'gif',
+        'tiff',
+        'pdf',
+    ];
+
+    /**
+     * Image dimensions.
+     *
+     * @internal
+     * @var ImageDimensions stores Image Dimensions in a structured way.
+     */
+    protected ImageDimensions $imageDimensions;
+
+    /**
+     * Inits the ExternalImage object.
+     *
+     * @internal
+     *
+     * @param $url string URL to load the image from.
+     * @param $alt string ALT text for the image.
+     */
+    public static function build($url, $alt)
+    {
+        if (!is_numeric($url) && is_string($url)) {
+            $external_image = new ExternalImage($alt);
+            if (strstr($url, '://')) {
+                // Assume URL.
+                $external_image->init_with_url($url);
+
+                return $external_image;
+            } elseif (strstr($url, ABSPATH)) {
+                // Assume absolute path.
+                $external_image->init_with_file_path($url);
+
+                return $external_image;
+            } else {
+                // Check for image file types.
+                foreach ($external_image->image_file_types as $type) {
+                    // Assume a relative path.
+                    if (strstr(strtolower($url), $type)) {
+                        $external_image->init_with_relative_path($url);
+
+                        return $external_image;
+                    }
+                }
+            }
+        }
+
+        return null;
+    }
+
+
+    /**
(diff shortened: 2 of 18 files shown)


xavivars:
@gchtr @jarednova I'm trying to see why tests are failing, but I can't figure it out.

I can't manage to run tests locally (I tried lando and it doesn't work for me). Is there any easy way to run tests locally to make debugging easier?

gchtr:
I think it’s the following to errors:

```
There were 2 errors:

1) TestTimberImage::testImageSizes
Undefined property: Timber\ImageDimensions::$image

/home/runner/work/timber/timber/src/ImageDimensions.php:116
/home/runner/work/timber/timber/src/ImageDimensions.php:54
/home/runner/work/timber/timber/src/Image.php:238
/home/runner/work/timber/timber/tests/test-timber-image.php:197
phpvfscomposer:///home/runner/work/timber/timber/vendor/phpunit/phpunit/phpunit:97

2) TestTimberImage::testSVGDimensions
Undefined property: Timber\ImageDimensions::$image

/home/runner/work/timber/timber/src/ImageDimensions.php:116
/home/runner/work/timber/timber/src/ImageDimensions.php:[54](https://github.com/timber/timber/actions/runs/3048958254/jobs/4933686252#step:17:55)
/home/runner/work/timber/timber/src/Image.php:238
/home/runner/work/timber/timber/tests/test-timber-image.php:12[56](https://github.com/timber/timber/actions/runs/3048958254/jobs/4933686252#step:17:57)
phpvfscomposer:///home/runner/work/timber/timber/vendor/phpunit/phpunit/phpunit:97
```

You can run the tests locally without VVV or Lando.

First, you need to install the tests using **bin/install-wp-tests.sh**. Then, from the Timber root directory, run:

```bash
bash ./bin/install-wp-tests.sh timber-tests username password localhost trunk false
```

This will install a local database `timber-tests`. You need to change `username` and `password` to the corresponding credentials of your local database. And maybe you have a different host than `localhost`, too.

Then, you should be able to run the tests with

```bash
./vendor/bin/phpunit
```

xavivars:
@gchtr thanks a lot!

It seems I managed to fix tests for WP latest, but WP trunk are still failing. I'll try to investigate what's the problem. I was trying to avoid installing PHP & mysql on my machine (I use docker for everything, and didn't want to necessarily install them just to do timber development), but it's ok.

In any case, any thoughts on the approach taken? Is this solution something you'd be OK with?

xavivars:
I figured out what happened with trunk: WordPress 6.1 converts images to WebP by default when resizing them.

xavivars:
Actually.... they reverted 30 minutes ago :man_facepalming:

xavivars:
@gchtr I would appreciate a review!

src/ExternalImage.php
```diff
@@ -0,0 +1,479 @@
...
+{
+    /**
+     * File.
+     *
+     * @api
+     * @var string
+     */
+    private $alt_text;
+
+    protected function __construct($alt)
```
gchtr, line 59: I think I wouldn’t use alt in the constructor here, but introduce a `set_alt()` method so that we could also set an alt text from outside.

```php
$image = Timber::get_external_image($url);
$image->set_alt('Dogs are happy outside');
```

I would probably even not use the constructor at all. The pattern that we use for the factories for WordPress objects in Timber is that we use `new static()`, see https://github.com/timber/timber/blob/1b91b3c31589162b2bc1872d0ff946fa7b68665d/src/Post.php#L181.

src/ExternalImage.php
```diff
@@ -0,0 +1,479 @@
...
+
+    /**
+     * Inits the ExternalImage object.
+     *
+     * @param $url string URL to load the image from.
+     * @param $alt string ALT text for the image.
+     * @internal
+     *
+     */
+    public static function build($url, $alt)
```
gchtr, line 139: I would use an options array for the second parameter, that takes alt as array key. Who knows what other options we’ll need in the future?

```php
public static function build($url, array $args = [])
{
    $args = wp_parse_args( $args, [
        'alt' => '',
     ] );

    // …

    $image->set_alt( $args['alt'] );
}
```

src/ExternalImage.php
```diff
@@ -0,0 +1,479 @@
+<?php
+
+namespace Timber;
+
+/**
+ * Class ExternalImage
+ *
+ * The `Timber\ExternalImage` class represents an external image loaded via http.
```
gchtr, line 8: With this description and the name `ExternalImage`, I would assume that I could only use image URLs starting with `http(s)://` and that I could either use a path from my own site or load an image from a completely different domain.

The code in the build function says otherwise. Apparently I can load images:

- from URLs from the same website
- from absolute paths
- from relative paths

I think we should reflect this in the documentation.

I think this is the reason why I’m still not sure about the name `External`. External in this case would mean "outside of the WordPress attachment area, but still on the same site". But don’t we often use the term "external" to stuff from other websites?

To download external images, we already have a `ImageHelper::sideload_image()` that could be used in combination with `Timber::get_external_image()`. If we can also download external images, then I think I would be okay with the name `ExternalImage`.

src/ExternalImage.php
```diff
@@ -0,0 +1,479 @@
...
+            $this->file = URLHelper::remove_double_slashes(
+                ABSPATH . URLHelper::get_rel_url($url)
+            );
+            $this->file_loc = URLHelper::remove_double_slashes(
+                ABSPATH . URLHelper::get_rel_url($url)
+            );
+            $this->imageDimensions = new ImageDimensions($this->file_loc);
+            $this->file_size = new FileSize($this->file_loc);
+        } else {
+            $this->imageDimensions = new ImageDimensions();
```
gchtr, line 432: We‘re missing a required parameter `$file_loc` here.

src/ExternalImage.php
```diff
@@ -0,0 +1,479 @@
...
+                    if (strstr(strtolower($url), $type)) {
+                        $external_image->init_with_relative_path($url);
+
+                        return $external_image;
+                    }
+                }
+            }
+        }
+
+        return null;
```
gchtr, line 166: We should probably also return `null` if the local file doesn’t exist.

I guess we don’t have any check whether the file exists yet in the whole logic. Would it be good to have that?

docs/v2/upgrade-guides/2.0.md
```diff
@@ -831,7 +831,9 @@ The following functions are being deprecated and will be removed in a future ver
 
 - `get_pathinfo()` – use `pathinfo()` instead.
 - `get_dimensions()` – use `width()` or `height()` instead.
-- `get_dimensions_loaded()` – use `get_dimension_loaded()` instead.
```
gchtr, line 834: Nice simplification!

src/ExternalImage.php
```diff
@@ -0,0 +1,479 @@
...
+     * @api
+     */
+    public function alt()
+    {
+        return $this->alt_text;
+    }
+
+    public function caption()
+    {
+        return "";
```
gchtr, line 461: If we allow users to set an alt text, should we also allow them to set a caption?

gchtr (changes requested):
Thank you so much for taking this on, @xavivars.

In general, I’m very happy with the new setup and the decoupling of functionality into separate classes 💪. It’s also good to have an `ImageInterface`.

I have some thoughts about a couple of details that I added in separate comments.

And we still have to write the tests. You can tell us if you want to us to do that. I will also have a look at the documentation parts as soon as this is ready from your part.

And I would also like @nlemoine to have a look 😊. He’s usually better with all the high-level stuff.

src/ExternalImage.php
xavivars: Makes sense. Updated both.

However, I left `$alt` as an optional second parameter in `get_external_image`, as all images should really have an alt text (even if it's empty).

src/ExternalImage.php
xavivars: makes sense. Changed.

src/ExternalImage.php
xavivars: Yeah, external may not be the best name... And, actually, it doesn't really work 100% with external images.

I didn't know about `sideload`, but I'll integrate that, so external URLs also work

So if it works for:
- relative paths
- absolute paths
- arbitrary URLs (either same website, or different website)
would you feel more confortable with the `External` name?

src/ExternalImage.php
xavivars: Fixed. Made it optional

src/ExternalImage.php
xavivars: Both for URLs and local files?

src/ExternalImage.php
xavivars: Yes, we definitely should. Adding a setter.

xavivars:
@gchtr I'd love if you could take a look at the tests. I added a couple (just dealing with "ratio" for ImageDimensions), but it definitely needs more tests.

I mostly ported the existing Timber 1.x logic, adjusting it to 2.x, but it definitely lacks tests.

src/ExternalImage.php
gchtr: Yes, if we support all of those, then I’d be okay with the name `External` 👍

src/ExternalImage.php
gchtr: @xavivars I wouldn’t check external URLs, but since external URLs will be saved in the uploads folder and become local files when sideloaded, we could probably check those.

gchtr:
@xavivars I’ll have a look at the tests.

xavivars:
It would be great to get this merged...

gchtr:
I’m merging this so that I can add a pull request that adds the required tests and documentation.

src/ExternalImage.php
gchtr: Thinking further about this, I guess it’s not so bad if an image is displayed as a broken image on the website. That’s also what alt texts are good for. If we just return `null`, then it would be hard to track these images.

Let’s leave it like that and see if we can find a good solution in the future.

gchtr (approved):

Issue #2519, opened 2021-11-21
Post meta with value NULL is retrieved as Array when getting all meta via $post->meta()


gustavo-roganti:
I'm using Timber 2.0.0-alpha.4 

I have a post with some meta values. One of this meta values is set to NULL.

When I get all the meta at the same time via $post->meta() I expect this result:
`array (
  'foo' => 'bar',
  'baz' => NULL,
)`

Instead I am getting this
`array (
  'foo' => 'bar',
  'baz' => 
  array (
    0 => NULL,
  ),
)`

## Steps to reproduce behavior
Add some meta to a post, set one meta value as NULL (eg. using update_post_meta), then retrieve all meta at the same time via `$post->meta()`

## Analysis and potential fix 
(sorry I don't know how to create a PR and I don't know if my code is ok)

In the function `\Timber\CoreEntity::fetch_meta`  in the file `timber/lib/CoreEntity.php` at around line `180` a bit of code is commented with `// Mimick $single argument when fetching all meta values.`
The code tries to detect if a meta value is an array with a single value. To do this, it uses `isset()`. Isset fails when a variable _is_ set but is also null. 

In this case checking with array_key_exists could be better (see https://stackoverflow.com/a/68135825), so a potential fix could be to replace this line:
`if ( 1 === count( $meta ) && isset( $meta[0] ) ) {`
with this line:
`if ( 1 === count( $meta ) && array_key_exists(0, $meta)) {`

gchtr:
Resolved in #2643.

Thanks a lot for your investigation and the pull request, @gustavo-roganti. Very much appreciated! 🙌

PR #2643, opened 2022-09-19
#2519 Fix post get all meta when one value is null

gustavo-roganti:
**Ticket**: #2519

## Issue
As per the ticket, this solves the issue of getting all post meta via `$post->meta()` when one meta has value set to NULL
//Expected result:
array ( 'foo' => 'bar', 'baz' => NULL, )
//Actual result:
array ( 'foo' => 'bar', 'baz' => array ( 0 => NULL, ), )

## Solution
This implements the fix proposed in the issue, changing the check 
`isset( $meta[0] )` 
to
`array_key_exists(0, $meta)` 
I added a simple failing test in the test suite, then added the fix and saw the test passing.

## Impact
None, afaik. 

## Considerations
I think this is just a small bugfix, and shouldn't really impact much.

## Testing
A simple test was added, as per committed file.


gchtr (dismissed):
Hey @gustavo-roganti 

Congratulations on your first pull request! You mentioned in #2519 that you don’t know how to make one. But you figured it out, great! 🥳

And thank you so much for researching this issue and taking the time to put together a solution. You even added a correct test! 

The tests didn’t pass in your pull request because of some coding standard problems. But don’t worry, I fixed this in #2644.

To me, the solutions looks correct. I just want to get a quick second opinion from @nlemoine, before I merge in both pull requests.

nlemoine:
Looks ok to me, congrats @gustavo-rogantifor your first PR! 😄 

@gchtr I just added some tiny changes while looking at it:
- Let's use `get_metadata` instead of `call_user_func` to grab meta values since `get_post_meta`, `get_term_meta`, etc. are just shortcuts to `get_metadata`. It's easier to read and safer because it won't fail if `$object_type` is wrong (e.g. not one of `post`, `term`, `user`, `comment`) since `$object_type` is just a public property 🙈
- I added tome tests with `add_post_meta` because the same meta key can have multiple rows

nlemoine (approved):

gchtr (approved):

PR #2644, opened 2022-09-24
2.x Fixes for #2643

### LINKED ISSUE #2519: Post meta with value NULL is retrieved as Array when getting all meta via $post->meta()

gchtr:
- Resolves #2519
- Resolves #2643

Just some coding standard fixes for #2643 to make the tests pass.


nlemoine (approved):

gchtr:
There wasn’t a lot left in the end … thanks for the review @nlemoine!

PR #2647, opened 2022-10-09
2.x Updates to #2632

gchtr:
Updates #2632 with the following improvements:

- Updates DocBlock to include more examples.
- Passes the `WP_User` object to `user_can()` instead of the ID to skip fetching the user object again.


nlemoine (approved):
Thanks @gchtr! This looks good to me.

PR #2649, opened 2022-10-09
2.x Update .gitignore

gchtr:
- Removes some unused docs entries
- Adds .phpunit.result.cache
- Sorts lines


szepeviktor:
Hello @gchtr!

Please consider organizing by this hierarchy.
https://github.com/szepeviktor/debian-server-tools/blob/b2ae2c93c842e3b467d8c00a90a94707bba51eaf/.gitignore#L7-L12

gchtr:
@szepeviktor Thanks for the input, Viktor. I think that might help indeed. I added another commit and found some more entries we can get rid of:

- bin/deploy-to-wp-org.sh
- bin/timber.php
- /Timber.php

szepeviktor:
Brain update 🧠 for viktor.
Better place for "OS" and "IDE and local environment" is a file specified by git global configuration option `core.excludesFile`
Please see https://git-scm.com/docs/gitignore

Issue #2655, opened 2022-10-19
Bug with latest release


bwalker5:
We received the latest Timber update last night 1.21 and our web site wasn't available after that (error message on page). The issue was resolved by going back to 1.20. 

Here is part of the error that was in our WordPress logs - PHP Fatal error:  Uncaught Error: Class "Twig_Filter_Function" not found***

GaelenN:
Also having this problem with a client site.

mrfsrf:
you have to specify more information. Did you install Timber using composer and than switched to wp plugin version, or vise versa?
Have a look at the timber starter theme functions.php

gchtr:
We have a pretty good idea where this is issue might be coming from and are currently investigating.

bwalker5:
> We have a pretty good idea where this is issue might be coming from and are currently investigating.

do you need any info from me to help troubleshoot the issue?

jarednova:
@bwalker5 if you're interested in lending a hand, we just published to WP.org. Are you able to download and verify in one of your setups?

https://wordpress.org/plugins/timber-library/

gchtr:
It looks like this is the same issue as https://github.com/timber/timber/issues/2672. It will be fixed in https://github.com/timber/timber/pull/2679.

For a very quick fix you could use

```bash
composer require twig/twig ^2
```

gchtr:
We just [released a new 1.22.1 version](https://github.com/timber/timber/releases/tag/1.22.1). Could you check whether you still have your issue?

Luckyfella73:
@gchtr  I can confirm that my _Unknown "filter" tag_ issue is solved with TimberTwig 1.22.1 

Thanks for fixing the issue that fast! 🙏

jarednova:
Great to hear @Luckyfella73 ! thanks to @rmens for submitting the PR to help us here

Issue #2654, opened 2022-10-19
Should semantic version be followed in relation to breaking changes?


mmcev106:
The twig update in 1.21.0 introduced some breaking changes, which caused an old site of mine to break with errors like the following:

```
Uncaught TypeError: Argument 1 passed to Twig\Environment::addFilter() must be an instance of Twig\TwigFilter, string given
```

They were easy to solve, but they made me wonder if any breaking change should have necessitated a version 2.0.0 of this plugin per [semantic versioning](https://semver.org/).  I'm guessing it's just not worth your time to be too strict about these things, which is totally fine.  I was just curious, so I figured I'd ask.

gchtr:
Yes, of course, the idea is to follow semantic versioning. (That’s why we are still working on a version 2 that will [introduce a lot of breaking changes](https://timber.github.io/docs/v2/upgrade-guides/2.0/).)

We didn’t mean to introduce a breaking change and are currently investigating why this happened. We might have shipped Twig v2 instead of Twig v1 in the plugin by mistake. We hope that we can release a fix for this, soon. Sorry for that.

Would you mind giving us a code example that you had to fix? A short snippet of a code before and after? We could include it in our tests, then.

mmcev106:
I'm not an expert on the site in question yet (it was created by someone else years ago), but I believe this is a distilled example of part of the code that broke:

```php
add_filter('get_twig', function(){
	$twig->addFilter('myCustomFunction', new Twig_SimpleFilter('myCustomFunction', function($someArg){
		// Take some action
	}));
});
```

For all I know, that is incorrect usage of twig.  It did work with the version of twig included in 1.20.0 though.

mgruhn:
Just jumping into this ticket to note that, in addition to the filter issue identified above, 1.21 also broke custom loaders. I had custom loader set via the `timber/loader/loader` filter, an implementation of `\Twig\Loader\LoaderInterface`. But with 1.21, twig's `LoaderInterface` definition changed (renaming `getSource`), so the sites with this code all raised PHP errors. I think this supports the idea that 1.21 included an inadvertent twig version update.

gchtr:
@mgruhn The reason for this error is also that Twig version 2.0 was mistakenly included in the plugin version of Timber when we ran an automated script to build the plugin. This shouldn’t be a problem with the next release of Timber, where the most current version of Twig 1.x will be included. The interface didn’t change there.

Issue #2652, opened 2022-10-12
Uncaught Error: Class "Routes" not found


thangaraj-fuint:
## Expected behavior
HTTP 200 state of site.

## Actual behavior
Site down with fatal error 500

## Steps to reproduce behavior
Update the Timber Version from 1.20.0 to 1.21.0

## What version of WordPress, PHP and Timber are you using?
WP Version: 6.0.2
PHP Version: 8.0
Timber Version: 1.21.0

## How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Downloaded the timber from github: https://github.com/timber/timber/releases/tag/1.21.0

gchtr:
If you download the release zip from the release page, that won’t work, because then you’re missing all Composer dependencies.

If you’ve installed Timber with Composer, then you should update it through Composer by running

```bash
composer update timber/timber
```

If you installed Timber as a plugin, I’m afraid you’ll have to wait a little longer, because we didn’t add that release to the [WordPress plugin repository](https://wordpress.org/plugins/timber-library/) yet. But we will, soon.

For the different ways of installing Timber, please follow the official [Setup Guide](https://timber.github.io/docs/getting-started/setup/#installation).

thangaraj-fuint:
Thanks for your qucik response @gchtr

thangaraj-fuint:
Hi @gchtr 
I've tried your new wordpress release 1.21.0
Everything looks good expect one thing.
tiwg's "attribute" function not working.
When I'm reverted the wordpress release to 1.20.0 "attribute" function working properly.
Could you able to check & let me know if there is any syntax changes ?

gchtr:
> tiwg's "attribute" function not working.
> When I'm reverted the wordpress release to 1.20.0 "attribute" function working properly.

What do you mean, it’s not working properly? Is there an error message? Is there no output?
If it doesn’t work right, it doesn’t have to be about the `attribute()` function, but it could also be about how objects work, because you work with a `profile` object here.

What’s `profile`? How does that object look like? I would need some code to help you here.

thangaraj-fuint:
@gchtr There is no error message.
I'm calling profiles.twig file using attribute().
The problem is not calling profiles.twig
Because profile object returns NULL :(
I'm not sure how it returns as null, If I reverted the plugin to old version means my object having values

gchtr:
To add formatted code blocks, you can use triple backticks ```:

<pre><code>```twig
{# Your Twig code here. #}
[...]

PR #2658, opened 2022-10-24
Fix bugs with latest plugin release 1.21.0

gchtr:
Resolves:

- #2654
- #2655
- #2652
- #2661 

## Issue

When the plugin version 1.21.0 of Timber was released, we accidentally included Twig 2.15.3 instead of Twig 1.44.7. This caused a lot of bugs for sites that auto-updated the Timber plugin (not recommended). We deactivated the 1.21.0 as the default release to disable that version from being found by auto-updates.

## Solution

As a solution, this pull request makes sure that Twig version 1.44.7 is used when deploying the plugin version of Timber. When Timber is installed through Composer, the 2.x will still be installed.

We also added a new test configuration that tests PHP 8.0 and PHP 8.1 with version 1.44.7 of Twig by using `prefer-lowest` when installing Composer dependencies (thanks @nlemoine).

Unfortunately, I didn’t manage to get this fully working with PHP 8.1. A deprecation warning is somehow caught as content in the output in a shortcode test. I couldn’t find out how to "only" raise a PHP Deprecation warning. Does anybody have some ideas?

That’s why I’m hesitant about officially supporting PHP 8.1 in version Timber 1.x **in the plugin version.**

## Impact

This pull request should fix the errors that people are seeing because we will revert to Twig 1.44.7:

- `Uncaught TypeError: Argument 1 passed to Twig\Environment::addFilter() must be an instance of Twig\TwigFilter, string given` in #2654, because passing only names as the first argument in `addFilter()` was removed as of Twig 2.0.
- `Uncaught Error: Class "Twig_Filter_Function" not found` in #2655, because `Twig_Filter_Function` was removed as of Twig 2.0.
- Twig’s `attribute` function not working in #2652, because attributes can’t be used to call macros anymore as of Twig 2.0, as explained by in @mrfsrf https://github.com/timber/timber/issues/2652#issuecomment-1284581094.
- `Fatal error: Uncaught LogicException: Unable to register extension “Twig\Extension\DebugExtension” as it is already registered.` in #2661, because registering an extension twice will throw a `LogicException` as of Twig 2.0.

## Usage Changes

None.

## Considerations

I already noted in the readme.txt that the new version should be 1.22.0. Though I’m not sure whether this qualifies as a hotfix and should be tagged with 1.21.1 instead. If guess if we require WordPress 5.3 as a minimum WordPress version, that wouldn’t qualify as a hotfix anymore.

## Testing

Yes.

### DIFF
--- a/bin/deploy-to-wp-org.sh
+++ b/bin/deploy-to-wp-org.sh
@@ -10,7 +10,7 @@ function deploy () {
 	rm composer.lock
 
 	# Download dependencies for the maximum compatible PHP version.
-	composer config platform.php 8.0
+	composer config platform.php 7.2.5
 	composer install --no-dev --optimize-autoloader
 
 	# Install the lowest compatible version of Twig.
--- a/bin/deploy-to-wp-org.sh
+++ b/bin/deploy-to-wp-org.sh
@@ -8,8 +8,14 @@ function deploy () {
 	git clone git@github.com:Upstatement/timber-starter-theme.git
 	rm -rf ~/Sites/timber/timber-starter-theme/.git
 	rm composer.lock
-	composer config platform.php 5.6.20
+
+	# Download dependencies for the maximum compatible PHP version.
+	composer config platform.php 8.0
 	composer install --no-dev --optimize-autoloader
+
+	# Install the lowest compatible version of Twig.
+	composer update twig/twig:1.44.7 --no-dev
+
 	rm -rf ~/Sites/timber/vendor/upstatement/routes/.git
 	cd ~/Sites/timber-wp
 	mkdir tags/$1
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -139,8 +139,8 @@ jobs:
         if: ${{ matrix.wp < '5.9' }}
         run: composer req phpunit/phpunit:^7.5 yoast/wp-test-utils:^1.0 -W --dev
 
-      - name: Upgrade PHPUnit to ^9.0 when prefer lowest
-        if: ${{ matrix.dependency-version == 'prefer-lowest' }}
+      - name: Upgrade PHPUnit to ^9.0 when prefer lowest and PHP 8.0+
+        if: ${{ matrix.dependency-version == 'prefer-lowest' && matrix.php >= '8.0' }}
         run: composer req phpunit/phpunit:^9.0 yoast/wp-test-utils:^1.0 -W --dev
 
       - name: Install tests
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -128,7 +128,7 @@ jobs:
 
       - name: Get composer cache directory
         id: composer-cache
-        run: echo "::set-output name=dir::$(composer config cache-files-dir)"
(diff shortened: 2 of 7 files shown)


mrfsrf:
```"twig/twig": ">=1.44.7 || ^2.10",```
This installs Twig v.3.x.x. I believe logical or should be omitted.

**Comment by gchtr** [comment]
> `"twig/twig": ">=1.44.7 || ^2.10",`
> This installs Twig v.3.x.x. I believe logical or should be omitted.

@mrfsrf This installs Twig version 2, if you **install Timber through Composer** and the requirements are met. We’ve always allowed installing Twig version 2 this way.

For the **plugin version** of Timber, we explicitly request Twig 1.44.7, because we need to stay on version 1 for compatibility:
https://github.com/timber/timber/blob/3667a178feaf0b898a25c4d59e6cb387249c279c/bin/deploy-to-wp-org.sh#L16-L17

Timber v2 will allow you to install Twig 2 or 3.

File: bin/deploy-to-wp-org.sh [hunk 1]
```diff
@@ -8,8 +8,14 @@ function deploy () {
 	git clone git@github.com:Upstatement/timber-starter-theme.git
 	rm -rf ~/Sites/timber/timber-starter-theme/.git
 	rm composer.lock
-	composer config platform.php 5.6.20
+
+	# Download dependencies for the maximum compatible PHP version.
+	composer config platform.php 8.0
```
nlemoine, line 13: We should actually do the opposite. Install dependencies for the minimum compatible version.

`composer config platform.php 7.2.5`

Otherwise some dependency in the dependency tree might not be compatible with PHP 7.2.5. Although the opposite is also true for PHP 8.0 (some dependency might not be compatible with PHP 8), however, the consequences should not be as dramatic, e.g. a fatal error (VS a deprecation warning).

nlemoine (commented):
BTW, this example shows we should always run tests against the entire dependency range: meaning that `prefer-lowest` should be part of the matrix (for 2.x too).

bin/deploy-to-wp-org.sh
gchtr: Oh yeah, of course, that makes sense. I updated it in [`f3d92d6` (#2658)](https://github.com/timber/timber/pull/2658/commits/f3d92d6e72893c9fd8703efbef3bb49acc6d5183).

I think we better stay conservative here. People who really want to upgrade to the latest version of PHP can do so using Composer.

gchtr:
> BTW, this example shows we should always run tests against the entire dependency range: meaning that `prefer-lowest` should be part of the matrix (for 2.x too).

Yes, definitely!

jarednova:
Thanks so much @gchtr and @nlemoine. I was writing a long question about the 8.1 and the failing test, I see the notes in `readme.txt` so helpfully answer it already! While much ink could be spilled in regards to 1.21.1 vs. 1.22.0, I think it's good as-is. Will merge and then proceed with WP.org deploying

xavivars:
`"twig/twig": ">=1.44.7 || ^2.10",`
should be
`"twig/twig": "^1.44.7 || ^2.10",`

otherwise, twig 3 is installed (as it's > than 1.44.7)

PR #2659, opened 2022-10-25
2.x Fix PHPStan issues on level 0

gchtr:
## Issue

There are various issues raised by PHPStan on level 0.

## Solution

This pull request resolves some of them. Other issues will be resolved in separate pull requests.

- #2631
- #2660
- #2664

## Impact

More robust code.

## Usage Changes

Some return type changes.

## Testing

Should be covered through existing tests.

Except the new return types, which we should add tests for. Coverage shrinked because of this.

## Todo

- [x] Fix coverage

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -292,8 +292,6 @@ class Loader
          *
          * @link https://github.com/timber/timber/pull/1254
          * @since 1.1.11
-         *
-         * @param array $paths
          */
         $fs = apply_filters('timber/loader/loader', $fs);
 
@@ -333,7 +331,7 @@ class Loader
          *  *
          *  * @link https://twig.symfony.com/doc/2.x/api.html#environment-options
          *  *
-         *  * @param array $options An array of environment options.
+         *  * \@param array $options An array of environment options.
          *  *
          *  * @return array
          *  *\/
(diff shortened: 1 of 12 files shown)


gchtr:
I think we can ignore the decreased coverage. Some of it is caused by return types that should become stale again when #2669 is merged. I’m marking this ready for review.

src/Loader.php
```diff
@@ -333,7 +331,7 @@ public function get_twig()
          *  *
          *  * @link https://twig.symfony.com/doc/2.x/api.html#environment-options
          *  *
-         *  * @param array $options An array of environment options.
+         *  * \@param array $options An array of environment options.
```
nlemoine, line 334: Why the docblock backslash?

nlemoine (approved):

src/Loader.php
gchtr: @nlemoine I think I added this because this is a DocBlock inside an `@example` tag. If I remember correctly, this lead to problems when generating the documentation when I didn’t use the backslashes.

nlemoine: I was just curious about this :)

gchtr:
@nlemoine Thanks for the review! I’m merging this!

PR #2660, opened 2022-10-25
2.x Make object constructors final

gchtr:
Related:

- #2659

# Issue

PHPStan reports "Unsafe usage of new static()" for some objects on level 0. What this means and how to solve this can be read in https://phpstan.org/blog/solving-phpstan-error-unsafe-usage-of-new-static.

## Solution

Make the object constructors final. Not being able to overwrite a constructor seems like a good idea to me. As noted in the linked post:

> So the PHPStan rule protects you from accidentally breaking your code when extending the class and defining a different constructor.

By making the constructors final, I already ran into an issue in `PagesMenu`, where I used a custom constructor. Removing the constructor and just updating the values on the object (or through a custom method) works fine.

This change was already discussed in https://github.com/timber/timber/pull/2203#issuecomment-591834164. In that discussion, there’s an example that wouldn’t work anymore with the change.

> 2\. Making the constructor `final` precludes any user in the future doing anything involving a custom constructor. For example, I might want to do this:
> 
> ```
> class MyPost extends Post {
>   protected function __construct($stuff, $things) {
>     // do something special with $stuff and $things
>   }
> 
>   public static function build( WP_Post $wp_post ) {
>     [ $stuff, $things ] = SomeApi::special_sauce();
>     return new static($stuff, $things);
>   }  
> }
> ```

Am I wrong to assume that this could be rewritten with a custom method `do_something()` that doesn’t change the constructor?

```php
class MyPost extends Post {
  protected function do_something($stuff, $things) {
    // do something special with $stuff and $things
  }

  public static function build( WP_Post $wp_post ) {
    [ $stuff, $things ] = SomeApi::special_sauce();
    $object = new static($wp_post);

	$object->do_something($stuff, $things);
  }  
}
```

## Impact

With final constructors, it will be easier to update constructors in the future, because we will not break backwards compatibility automatically. Otherwise, we’d have to assume that any constructor could have been overwritten.

## Usage Changes

We won’t be able to overwrite the constructors in classes that extend the Timber core object. But it will still be possible to update the objects with other custom methods as shown above.

## Considerations

In case we need some other functionality to make it easier to solve a certain use case, I’m sure we can come up with a proper API for that that doesn’t involve custom constructors.

## Testing

Nothing in particular.


Levdbas (dismissed):
This seems like a logical thing to do. I do think we need to document this change in the upgrade guide as well for folks that extended these classes in the past and maybe have overwritten the constructor?

nlemoine (dismissed):
Looks good, thanks @gchtr !

gchtr:
@Levdbas I added the notes in the Upgrade Guide: [`73d96d8` (#2660)](https://github.com/timber/timber/pull/2660/commits/73d96d828826e50cc3058cc2337915ebe4ddf9cf). I’d be glad about a quick check by an additional pair of eyes 👀 .

After that, we can merge this.

Levdbas (approved):
Yes, this seems good. Maybe we can add a howto guide on what the recommended way is to extend a Timber object and how to call it from your code. But that is out of the scope of this PR.

gchtr:
@Levdbas Thanks!

I think we already covered how to extend Timber objects in https://timber.github.io/docs/v2/guides/extending-timber/, right?

mindfullsilence:
Please don't do this to the end users (me). I'm in the midst of using v2rc1 and came across this restriction - which seriously limits my ability to provide services to the class unless I instead write a bunch of setter methods or create an init function that I have to call every time I instantiate (both options aren't great). This removes any possibility of DI, autowiring, and I'm sure a lot of other things I can't think of right now. 

A few alternatives I'd like to propose:
1. Leave `final` in place, but provide an interface that we can use to write our own Post class. It doesn't appear there's a check in place right now under `PostFactory` to ensure the classmap only contains extentions of `Timber\Post`, but I imagine there's supposed to be (there was in v1). However, instead of checking for a concrete subclass of `Timber\Post`, checking instead for a `Timber\PostInterface` that doesn't define the constructor signature would give us a great deal more flexibility if you're truly sold on marking the `Timber\Post` constructor `final`. This would at least allow us to proxy the `Timber\Post` class methods via `__call` or some such while still having our own constructor. 
2. Remove `final` from the `__constructor`, and instead of restricting users from having their own constructor, restrict yourselves as authors from defining a constructor method (E.g. Laravel does this with quite a few of their classes to allow for users to define their own DI requirements - see models, controllers, jobs, queues, etc).
3. Remove `final` from the `__constructor`, and add migration documentation as needed when the parent constructor signature changes (probably the most common, easiest, and most flexible approach for all).

Pretty pretty please with sugar on top.

PR #2664, opened 2022-10-27
2.x Remove unused Term::get_term_from_query() function

gchtr:
Related:

- #2659

## Issue

I found a function that is not used anymore.

## Solution

Remove the function.

## Impact

Minimal. The function that gets the term from the main query is now covered through `Timber::get_term()` and the term factory.

https://github.com/timber/timber/blob/056c6a1c0c345f6c1f08857beed97730a3267187/src/Timber.php#L714-L718

Only the protected function `Term::get_term_from_query()` can’t be used anymore.

## Considerations

There’s an additional check for a tax query that was added in https://github.com/timber/timber/pull/521: 

https://github.com/timber/timber/blob/056c6a1c0c345f6c1f08857beed97730a3267187/src/Term.php#L142-L144

The logic there seems to be: When the global query includes a tax query, get the first term from that query. Is that a case we still want to support? Is this the intended behavior for `Timber::get_term()`? Or could it also be considered as unexcepted?

## Testing

Do we need more testing?

### DIFF
--- a/docs/v2/upgrade-guides/2.0.md
+++ b/docs/v2/upgrade-guides/2.0.md
@@ -411,6 +411,14 @@ function get_term($term = null)
 - We removed the `$taxonomy` parameter. Timber needs the ID of a term or a `WP_Term` object. It can figure out the taxonomy itself.
 - We removed the `$TermClass` parameter. If you want to control the class your term should be instantiated with, use a [Term Class Map filter](https://timber.github.io/docs/v2/guides/class-maps/#the-term-class-map).
 
+#### Removed edge case logic
+
+When you use `Timber::get_term()` without any parameter, Timber returns the queried term if the queried object is a term.
+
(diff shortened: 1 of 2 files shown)


nlemoine (dismissed):
> Is that a case we still want to support? Is this the intended behavior for Timber::get_term()?

I didn't event know that method existed 😅

This an absolute no to me, I don't see any valid use case where getting the term from the tax query should be returned with this kind of method. It should only return the current queried term.

gchtr:
@nlemoine Thanks for your feedback!

I added a note to the Upgrade Guide about the removed logic in [`ee42528` (#2664)](https://github.com/timber/timber/pull/2664/commits/ee4252853f76d209da890588db4c8a5688ceb842).

I’m sure in the past this logic made some developer’s life easier in very specific use cases. But these kind of gems are the ones that can lead to unexpected behavior quickly. So I’m happy to remove them.

PR #2665, opened 2022-10-27
Run tests with lowest dependencies

nlemoine:
- Add 'lowest' to the matrix to run tests with lowest dependencies.
- Replace DIY composer cache/install with `ramsey/composer-install@v2` which handles caching and installing dependencies.
- Bump ACF to `^5.0 || ^6.0`
- Bump Twig minimum version to `^2.15.3` as it improves PHP 8.1 compatibility and fixes some security issues (see screenshot above)
- Exclude WP 5.3 from PHP 8.0 and 8.1 tests as it's not compatible with those versions.

Capture d’écran 2022-10-26 à 22 51 30


tests/test-timber-integration-acf.php
```diff
@@ -106,6 +106,9 @@ public function testACFFieldObject()
 
     public function testACFFormatValue()
     {
+        // Avoid deprecation warning on ACF 5.6.x
+        remove_filter('acf_the_content', 'wp_make_content_images_responsive');
```
gchtr, line 110: Shouldn’t this filter removal be temporary to avoid side effects?

tests/test-timber-twig-date-function.php
```diff
@@ -1,141 +0,0 @@
-<?php
```
gchtr, line 1: I specifically added these tests in https://github.com/timber/timber/pull/2135 to make sure that the `|date` filter in Twig works

- in the way that Twig is expecting it to 
- with WordPress dates and times

There are some custom filters involved to make sure the correct WordPress time settings are used. It doesn’t test whether Twig’s date functions work, but **whether our custom filtering doesn’t break Twig’s basic date filter**. That’s why I would advocate to keep these.

composer.json
```diff
@@ -30,7 +30,7 @@
   "require": {
     "php": "^7.4 || ^8.0",
     "composer/installers": "^1.0 || ^2.0",
-    "twig/twig": "^2.12 || ^3.0"
```
gchtr, line 33: This needs an update to the Upgrade Guide.

tests/test-timber-integration-acf.php
nlemoine: Absolutely! good catch 😄

tests/test-timber-twig-date-function.php
nlemoine: I see, I restored the file.

gchtr (approved):
Thanks for the updates!

I think I found a bug with `remove_filter_temporarily()` that I fixed in [`c6ef9e4` (#2665)](https://github.com/timber/timber/pull/2665/commits/c6ef9e468b33ea792d19fa3f0fd243f21be2625b).

Other than that, this looks great to me 👌. I’ll leave you to merge this in.

PR #2668, opened 2022-11-04
2.x Fix PHPStan issues on level 2

gchtr:
## Issue

There are various issues raised by PHPStan on level 2.

## Solution

This pull request fixes some of them.

One special case was that for menu items, we set some properties dynamically. Menu items are `WP_Post` objects that are extended with the following properties: `object_id`, `menu_item_parent`,  `object` and `type`.

To allow these properties to be written dynamically, I added a [Class Reflection Extension](https://phpstan.org/developing-extensions/class-reflection-extensions) for PHPStan. You can see this in the new `tests/phpstan/WPNavMenuItemPost.php` class.

There are some issues that I couldn’t resolve yet.

### Type is not subtype of native type

I didn’t know how to solve this one. Any help is welcome.

```
------ -------------------------------------------------------------------------------------------
  Line   AccessesPostsLazily.php (in context of class Timber\PostArrayObject)
 ------ -------------------------------------------------------------------------------------------
  71     PHPDoc tag @return with type Timber\PostCollectionInterface is not subtype of native type
         Timber\PostArrayObject.
 ------ -------------------------------------------------------------------------------------------

 ------ -------------------------------------------------------------------------------------------------------------
  Line   AccessesPostsLazily.php (in context of class Timber\PostQuery)
 ------ -------------------------------------------------------------------------------------------------------------
  71     PHPDoc tag @return with type Timber\PostCollectionInterface is not subtype of native type Timber\PostQuery.
 ------ -------------------------------------------------------------------------------------------------------------
```

### Access to undefined property on WP_Post

```
 ------ -----------------------------------------------------------------------------------
  Line   Post.php
 ------ -----------------------------------------------------------------------------------
  505    Access to an undefined property WP_Post::$status.
  506    Access to an undefined property WP_Post::$id.
  507    Access to an undefined property WP_Post::$slug.
 ------ -----------------------------------------------------------------------------------
```

This is related to dynamically setting properties on a `WP_Post` object that will later be imported in the `Post::get_info()` function. It would be better to set these properties on the `Timber\Post` object after the other properties are imported. But there’s a `timber/post/import_data` filter, which might rely on these properties.

```php
$data = $post->get_info($wp_post);

$post->import(apply_filters('timber/post/import_data', $data));
```

I think there are a couple of solutions:

1. Ignore the errors. 
2. Don’t set the properties on `WP_Post` and update the filter.
3. Whitelist the properties through a [Class Reflection Extension](https://phpstan.org/developing-extensions/class-reflection-extensions).

## Impact

More robust code.

## Usage Changes

None.

## Considerations

None.

## Testing

Covered through existing tests.

## Todo

- [ ] Fix coverage


gchtr:
With the changes in #2674, the `Access to an undefined property` shouldn’t be a problem anymore.

src/MenuItem.php
```diff
@@ -136,7 +136,9 @@ protected function __construct(WP_Post $data, $menu = null)
         $this->wp_object = $data;
         $this->menu = $menu;
 
-        $data = (object) $data;
+        /**
```
Levdbas, line 139: Why is this doc added over here instead of in the top of the class defining the property?

src/Term.php
```diff
@@ -80,10 +80,13 @@ protected function __construct()
...
      * @internal
-     * @param \WP_Term the vanilla WP term object to build from
+     *
+     * @param \WP_Term $wp_term The vanilla WordPress term object to build from.
+     * @param array    $options Optional. Unused. An array of options.
+     *
      * @return \Timber\Term
      */
-    public static function build(WP_Term $wp_term, array $_options = []): self
+    public static function build(WP_Term $wp_term, array $options = []): self
```
Levdbas, line 89: Can you explain the $options parameter if unused?

Levdbas (commented):
I went through all the code except the tests. That is not really (yet) in my ballpark.

src/MenuItem.php
nlemoine: I think we should probably add types to declared properties, especially for the very predictable ones.

src/Term.php
nlemoine: I think it's related to `Menu` where it can take an additional `$args` argument. But I agree this feels weird.

nlemoine (commented):
Thanks for you work on this @gchtr. To pass PHPStan higher levels, we will probably have to be stricter regarding types, for method arguments as well as return types.

nlemoine (commented):
Thanks for you work on this @gchtr. To pass PHPStan higher levels, we will probably have to be stricter regarding types, for method arguments as well as return types.

src/MenuItem.php
gchtr: This is here because PHPStan couldn’t find a title property on the `WP_Post` object. The `title` property is set dynamically for menu items. I didn’t find an easier way to solve this :).

src/Term.php
gchtr: Hm... but `Menu` is a separate class that uses a `WP_Term` under the hood, but not actually `Timber\Term`.

I guess it comes from a state where `Timber\Menu` extended `Timber\Term`, see https://github.com/timber/timber/commit/21445a1f5b50606fb3f8789c981f698a0437b2b9.

I’d say we can remove this unused parameter in a separate pull request. I’ll create one.

Good catch, @Levdbas!

gchtr: See https://github.com/timber/timber/pull/2754

Levdbas (approved):

Levdbas (approved):

PR #2669, opened 2022-11-04
2.x Remove unneeded Timber\Image::is_image() method

gchtr:
Related:

- #2659

## Issue

While working on #2659, I found that we probably don’t need the `Timber\Image::is_image()` method anymore. We now have an `is_image()` function in the PostFactory:

https://github.com/timber/timber/blob/8545e2e807ec6de3c377dc5118de0147edd98895/src/Factory/PostFactory.php#L173

The logic is that if you use `Timber::get_post()`, you will only get an instance of `Timber\Image`, if it’s an image attachment and a `Timber\Attachment` instance if it’s a non-image attachment.

The  `Timber\Image::is_image()` method is used to check whether a certain function can be used, because it only makes sense for images. But now, if we’re inside `Timber\Image`, we can assume that we have an image.

## Solution

Remove the method and where it’s used.

## Impact

This is a breaking change, hence the note in the Upgrade Guide.

## Usage Changes

Instead of using `Timber\Image::is_image()`, we would have to use:

```php
if ( $post instanceof Timber\Image ) {

}
```

## Considerations

Using `instanceof` doesn’t work so well in Twig. Should there be an `is_image()` function in `Timber\Post`?

```php
public function is_image()
{
	return $this instanceof Timber\Image;
}
```

We could also just remove the `is_image()` checks in `Timber\Image` an keep the method.

Any opinions?

## Testing

Covered through existing tests.


nlemoine (dismissed):
This looks good to me!

Levdbas (approved):

gchtr:
Thanks for the feedback here, everyone!

PR #2670, opened 2022-11-04
2.x Use assertSame instead of assertEquals for certain values

gchtr:
Related: 

- #2659

## Issue

While working on #2659 I found that some tests use `assertEquals`. TIL this is only a `==` comparison. For a `===` comparison, we should use `assertSame`.

Using `assertEquals` could be a problem when we want to check for empty values or explicit return values like `null` instead of `false` or empty strings in methods:

```php
// Returns true
$this->assertEquals( '', null );
```

## Solution

Use `assertSame` for the following checks:

- `true`, `false` and `null`
- Empty string checks: `""`
- Numbers / Integers

## Impact

I found several type issues by changing to `assertSame()`. In some cases, I had to update the tests, in other cases I had to make typecast some return values to make sure they are what we expect.

For examples, the dimensions for SVG images will now be returned as integers instead of strings.

## Usage Changes

None.

## Testing

Yes.

### DIFF
--- a/src/Post.php
+++ b/src/Post.php
@@ -673,7 +673,7 @@ class Post extends CoreEntity implements DatedInterface, Setupable
      */
     public function comment_count()
     {
-        return get_comments_number($this->ID);
+        return (int) get_comments_number($this->ID);
     }
 
     /**
--- a/src/Image.php
+++ b/src/Image.php
@@ -128,13 +128,13 @@ class Image extends Attachment
      *
      * @internal
      * @param string $svg SVG Path
-     * @return array
+     * @return object
      */
     protected function get_dimensions_svg($svg)
     {
         $svg = simplexml_load_file($svg);
-        $width = '0';
-        $height = '0';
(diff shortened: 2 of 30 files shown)


src/Post.php
```diff
@@ -673,7 +673,7 @@ public function has_term($term_name_or_id, $taxonomy = 'all')
      */
     public function comment_count()
```
nlemoine, line 674: I would add return type as well.

nlemoine (changes requested):
Thank you @gchtr for this great work! I only have one tiny comment.

src/Post.php
gchtr: Resolved in [`97451ef` (#2670)](https://github.com/timber/timber/pull/2670/commits/97451efdb488ca97da192afc46c5006e7938bf3d).

gchtr:
Thanks for the feedback, @nlemoine! I’m merging this 😊.

Issue #2589, opened 2022-06-17
2.x Paged content within a post breaks core blocks that use wp-container (group, gallery)


drdogbot7:
## Expected behavior
Gutenberg blocks that make use of a "wp-container" block including core/group and core/gallery should work as expected even in paged content.

This refers to single posts that are split into multiple pages using the Page Break block, i.e. `<!--nextpage-->`.

## Actual behavior
These blocks display correctly on the first page only. After the first page break, they are broken.

Page 1 of post. Full-width group block inherits content width. Gallery displayed as 4 columns.
![image](https://user-images.githubusercontent.com/1050936/174395337-7630abb8-3818-484e-bb97-dc72b4199053.png)

Page 2 of post, using same content. Full-width group block content is full width. Gallery displayed as single column.
![image](https://user-images.githubusercontent.com/1050936/174395654-d0cf8ceb-f8f7-41e3-abd1-a60293aa9fa2.png)

Using the exact same content on page 1 and page 2 outputs different html.

Page 1 html:
![image](https://user-images.githubusercontent.com/1050936/174397046-b0763a51-9d53-4cf2-a732-5998e8dd9e70.png)

Page 2 html
![image](https://user-images.githubusercontent.com/1050936/174397186-ce53b22f-e0c1-48b1-85df-b30236a4693e.png)

I believe WP 5.9 introduced the `wp-container-{n}` class which is added to certain blocks, along with some inline styles.

## Steps to reproduce behavior
Follow examples here to set up paged content:
https://timber.github.io/docs/v2/guides/pagination/#paged-content-within-a-post

Screenshots show a custom theme I am developing, but I also tested this using the Timber Starter Theme (2.x), adding only the example code in order to enable paged content in post. This caused the same error.

I also tested paged content using the Twenty-TwentyTwo theme, Twenty-Nineteen, and Twenty-Seventeen.  These themes displayed the content without issue.

## What version of WordPress, PHP and Timber are you using?
WordPress 6.0, PHP 5.7, Timber 2.0.0-alpha-4

## How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
composer

gchtr:
Thanks for testing out 2.x and for providing such a detailed issue description!

We fixed this in https://github.com/timber/timber/pull/2673. It will be in the next 2.0 release.

PR #2673, opened 2022-11-09
2.x Fix bug with nextpage block

gchtr:
Resolves:

- #2589

## Issue

When the nextpage block was used, Timber doesn’t correctly split the content. The block on the second and subsequent pages are not properly converted, because there’s a stray `<!-- /wp:nextpage -->` block closer when we explode only on `<!--nextpage-->`.

A nextpage block looks like this:

```html
<!-- wp:nextpage -->
<!--nextpage-->
<!-- /wp:nextpage -->
```

## Solution

Take nextpage logic from WordPress to make it work with `<!--nextpage-->` only as well as nextpage blocks.

## Impact

Should be backwards compatible.

## Usage Changes

None.

## Considerations

WordPress ignores `<!--nextpage-->` if it’s at the beginning of a block. I added a test for this behavior as well.

## Testing

Yes.

### DIFF
--- a/src/Post.php
+++ b/src/Post.php
@@ -1170,22 +1170,47 @@ class Post extends CoreEntity implements DatedInterface, Setupable
         if ($len == -1 && $page == 0 && $this->___content) {
             return $this->___content;
         }
+
         $content = $this->post_content;
+
         if ($len > 0) {
             $content = wp_trim_words($content, $len);
         }
-        if ($page) {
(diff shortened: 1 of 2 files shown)


nlemoine (approved):
Thanks @gchtr! This is all good 👍

gchtr:
@nlemoine Nice, thank you!

Issue #2582, opened 2022-06-01
Fatal Error when previewing posts with images with PHP 8.0+


kshaner:
When there is an image inside a post, a preview triggers a fatal error and a entry in the error log:

```
PHP message: PHP Fatal error:  Uncaught Error: Attempt to assign property "post_content" on array in /var/www/wp-content/themes/theme-name/vendor/timber/timber/lib/Timber.php:168
```

In PHP < 8 this was considered a warning but is fatal in 8.0+. The get_info method of the `\Timber\Attachment` returns an array when all other classes return an object. https://github.com/timber/timber/blob/28f0d7bf9a6d49f9c0a840f2868e8882083b7d03/lib/Attachment.php#L186-L203

When previewing the post, the filter that modifies the object to their preview values is setting object properties directly here: https://github.com/timber/timber/blob/28f0d7bf9a6d49f9c0a840f2868e8882083b7d03/lib/Timber.php#L167-L171

As a workaround, I have added a new class that extends `\Timber\Attachment` and overloads the get_info method:

```php
class Attachment extends \Timber\Image {

	// this prevents a timber bug on php8 when accessing an object as an array and vice-versa
	protected function get_info( \WP_Post $wp_post ) {
		return (object) parent::get_info( $wp_post );
	}

}

```

I believe the proper solution here would be either cast the array to an object like my workaround does, or to check the whether the $data variable is an array or object when previewing.

## Expected behavior
A preview with all the correctly modified content

## Actual behavior
A fatal error is triggered.

## Steps to reproduce behavior
Create a post/page and upload an image to it. Then preview the page in a new window.

## What version of WordPress, PHP and Timber are you using?

* WordPress 6.0
* PHP 8.0.19
* Timber 2.0.0-alpha.4

## How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Installed with composer/packagist in the theme.

countfak:
Hi, I'm still having this issue. Here's my workaround:

```
add_filter(
  'timber/post/import_data',
  function ($data) {
    return (object) $data;
  },
  1
);
```

This has priority 1 because it has to run before the actual Timber filter, casting the $data variable to an object to prevent the error.

Not sure if I have to add additional checks to it, like make sure that I'm viewing a preview and not a regular post.

gchtr:
Thanks for reporting this issue! I added a pull request in #2674 to fix this. Your custom workaround will not be need anymore once this is released.

gchtr:
[...]

PR #2674, opened 2022-11-09
2.x Update Timber\Post::get_info() to work with a post data array only

gchtr:
Resolves:

- #2582 

## Issue

The different usages of `Timber\Post::get_info()` are not implemented consistently. Sometimes it’s implemented with an array as the parameter an other times with a `WP_Post` instance.

## Solution

This pull request changes the function signature of `Timber\Post::get_info()` to accept an array and return an array. The related `WP_Post` object is converted to an array before being passed to `Timber\Post::get_info()`. This way, we don’t set properties on a `WP_Post`, but update the array. This array will then be used to import properties to the `Timber\Post` instance.

I also moved a test related to previews to the corresponding test class.

This will also fix a related issue in #2668 where PHPStan warns about setting undefined properties on a post.

## Impact

If you use a custom `get_info()` method in a class that extends `Timber\Post`, you would have to update your code.  This is documented in the Upgrade Guide.

## Usage Changes

See Impact.

## Considerations

Maybe there’s a better solution than using arrays only. It will fix PHPStan issues, but it’s not really preventing us from setting undefined properties.

## Testing

Yes.


nlemoine (approved):
I agree with your considerations. But it works in the meantime 🙈

gchtr:
@nlemoine Thanks for the feedback 👍. I opened a new issue in #2704 where we can think about a better solution for this. I think we could do this more: Merge in pull requests if they work for the moment and create new issues where we can decide whether it needs to be fixed immediately or sometime in the future.

I will merge this in now, because the tests are fine (except for the tests with PHP 8.2).

gchtr:
@andrewwfesenko Can you create a new issue with a minimal setup so we can reproduce this? Because I can’t reproduce that bug (yet).

Issue #2646, opened 2022-10-05
Timber 2.x render caching is broken


silentdragonz:
## Expected behavior

Pages to be cached as their own transients when using the cache.

## Actual behavior

One transient is used to cache every page leading to the wrong page being returned from cache.

## Steps to reproduce behavior

Use a theme with no parent and the render cache.

## Fix
https://github.com/timber/timber/blob/1b91b3c31589162b2bc1872d0ff946fa7b68665d/src/Theme.php#L125 This line causes the context object to have recursion which means the json_encode for the cache key always returns false so every page has the same cache key. I saw this was added in #2264 to resolve other issues, but removing it fixes this issue.

gchtr:
Thanks for opening this issue 👍. I just added a test that confirms this bug. And I can also confirm that the recursion is what causes the problem. Now we only need to figure out how to solve this best.

This issue #2429 is a duplicate of this.

gchtr:
Fixed in #2675. This will be in the next 2.x release (hopefully soon).

Issue #2429, opened 2021-03-19
Timber 2.0 Twig Cache not working on render method


billybigpotatoes:
Using render method on single-post template - all posts are being cached - context appears to be ignored

## Expected behavior

I would expect to see different posts based on the post slug

## Actual behavior

All posts are showing the same content ignoring the slug / context. The item displayed seems to be the first item that was requested once the cache was enabled

## Steps to reproduce behavior

Following page controller:  single-post.php 

```
<?php
require_once( __DIR__ . '/vendor/autoload.php' );
$timber = new Timber\Timber();
$start = Timber\Helper::start_timer();

$context = Timber::context();
// $context['time'] = date();
// echo "<pre>";
// echo print_r($context, true);
// echo "</pre>";

// Render Page
get_header();
Timber::render( 'presenter.twig', $context, 60 );
get_footer();
echo Timber\Helper::stop_timer( $start);
```

Issue is resolved by setting cache from 60 to false - but at the expense of cache

## What version of WordPress, PHP and Timber are you using?

Not sure if this is related to the deprecated cache lib?

- WP 5.7. 
- PHP 7.4.1

- altorouter/altorouter        2.0.1           A lightning fast router for PHP
- composer/installers          v1.10.0         A multi-framework Composer library installer
- palmiak/timber-acf-wp-blocks 1.14            Create Gutenberg blocks from Twig templates and ACF fields.
- symfony/polyfill-ctype       v1.22.1         Symfony polyfill for ctype functions
- symfony/polyfill-mbstring    v1.22.1         Symfony polyfill for the Mbstring extension
- timber/timber                2.x-dev 2509b5c Plugin to write WordPress themes w Object-Oriented Code and the Twig Template Engine
- twig/cache-extension         v1.5.0          Cache fragments of templates directly within Twig.
- twig/twig                    v2.14.4         Twig, the flexible, fast, and secure template language for PHP
- upstatement/routes           0.8             Manage rewrites and routes in WordPress with this dead-simple plugin

## How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)

Composer

gchtr:
Hello @billybigpotatoes 

I’m trying to reproduce this issue and I think it would really help if you could also provide your Twig file. Are you looping over the posts in there?

In your example, I don’t see where your posts are coming from. In a singular template like **single-post.php**, there’s usually only one post to be displayed.

gchtr:
[...]

PR #2675, opened 2022-11-10
2.x Fix bug when Timber\Theme encoding breaks Timber caching

gchtr:
Resolves:

- #2429
- #2646

## Issue

In #2264, a change was added that adds a recursion for the `Timber\Theme` object. When serializing the object with `json_encode()`, this will cause the serialization to fail. Because of this, Timber can’t create a caching key and returns the same for different posts (see #2429 and #2646).

## Solution

This pull request:

- Implements the `JsonSerializable` interface for the `Timber\Theme` class so that it doesn’t break `json_encode()`.
- Adds a test for the serialization.
- Adds a check whether `json_encode()` was successfull and skips the cache if no cache key can be generated.
- Fixes testing some other theme functions (in https://github.com/timber/timber/commit/11f601e391a703edc4a4a844cb68c239db31e9b6, a change was added that caused a lot of theme tests to be skipped).

## Impact

None.

## Usage Changes

None.

## Considerations

I chose to go with `JsonSerializable` because this is a pattern that we already advise to use in cases like these: https://timber.github.io/docs/v2/guides/posts/#serialization.

This change may also be helpful for people who need theme data in a JSON-friendly version.

## Testing

Yes.


src/Theme.php
```diff
@@ -25,7 +25,7 @@
  * <script src="http://example.org/wp-content/themes/my-theme/static/js/all.js"></script>
  * ```
  */
-class Theme extends Core
+class Theme extends Core implements \JsonSerializable
```
nlemoine, line 28: `use JsonSerializable;`

Since we're making use of `use Something` elsewhere, let's be consistent and always choose imports.

nlemoine (changes requested):

src/Theme.php
gchtr: 👍 Added in [`ec534a1` (#2675)](https://github.com/timber/timber/pull/2675/commits/ec534a1cf262181d3baca22d6c47b393ebdb9452).

gchtr:
Tests only fail when running with coverage, but I couldn’t see why yet. I would merge this in and then check whether tests still fail on the 2.x branch.

gchtr:
I fixed the tests. Test only fail on PHP 8.2, which we ignore for now. I’m merging this.

Issue #2645, opened 2022-09-24
2.x `Timber\Core::can_edit` only handles post permissions


gchtr:
This issue was reported by @kshaner in #2632

> The way timber 2.x tends to use this now is with `\Timber\Core::can_edit`
> 
> https://github.com/timber/timber/blob/8d9bff23d2677652469badb75dcb607449a18204/lib/Core.php#L100-L108
> 
> However this method is in `\Timber\Core` and other classes such as `\Timber\User` and `\Timber\Term` also extend `\Timber\CoreEntity` (and thus `\Timber\Core`) and is not re-defined in sub-classes. This means that this code is checking against `edit_post` capability for the can_edit method of non Post classes. This code should should move to their specific sub-classes so there is no possibility of false negatives/false positives when id's match.

## Expected behavior

When using `{{ term.can_edit }}`, Timber should use the correct permission check for a term. The same goes for other core objects like comment and users.

## Actual behavior

Timber uses `current_user_can('edit_post', $this->ID)` under the hood. But it should use `current_user_can('edit_term, $this->ID)` instead.

## Possible solution

To solve this, we should move `can_edit()` to the specific Timber object classes.

gchtr:
Fixed in #2676.

PR #2676, opened 2022-11-11
2.x Fix can_edit() permission checks for Term, User, Comment and Menu classes

gchtr:
Resolves:

- #2645

## Issue

When using `{{ term.can_edit }}`, Timber should use the correct permission check for a term. The same goes for other core objects like comment and users.

## Solution

This pull request

- Moves `can_edit()` to the specific Timber object classes.
- Adds `can_edit()` to the `CoreEntity` interface.
- Adds new `edit_link()` functions to `Comment` and `User` class.

## Impact

Accurate capability checks.

## Usage Changes

None.

## Considerations

Checking the permissions on the `Menu` and `MenuItem` feels a bit weird first. They can’t be edited through the usual edit screens and need a check for the `edit_theme_options` capability. The question now is: Does it make sense to have the `can_edit()` check for the menu and menu items? Otherwise, we would have to remove the interface.

By adding `User::edit_link()`, it’s possible to get the profile URL for the current user, which is nice! I wonder whether we should add a `User::profile_link()` method as well. This would only return a link if the user object is the current user.

## Testing

Yes.

### DIFF
--- a/docs/v2/upgrade-guides/2.0.md
+++ b/docs/v2/upgrade-guides/2.0.md
@@ -1013,11 +1013,13 @@ The whole `Timber\Integration\Command` class was removed. Its methods were moved
 
 - `raw_meta()` – Gets a user meta value directly from the database.
 - `wp_object()` - Gets the underlying WordPress Core object.
+- `edit_link()` – Gets the edit link for a user if the current user has the correct rights.
 
 **Timber\Comment**
 
 - `raw_meta()` – Gets a comment meta value directly from the database.
 - `wp_object()` - Gets the underlying WordPress Core object.
+- `edit_link()` - Gets the edit link for a comment if the current user has the correct rights.
 
 **Timber\Menu**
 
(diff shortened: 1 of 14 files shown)


nlemoine (changes requested):
> Does it make sense to have the can_edit() check for the menu and menu items?

It doesn't from a usage POV but it does from technical. I would stand by the technical POV. `Menu` and `MenuItem` are special terms and posts and should implement the `CoreEntity` interface even if it's a bit weird in that one use case.

> I wonder whether we should add a User::profile_link() method as well

This makes sense!

Besides adding return types to the `edit_link` methods, this looks good to me!

gchtr:
> > I wonder whether we should add a User::profile_link() method as well
> 
> This makes sense!

Added as a future task in https://github.com/timber/timber/issues/2682.

> Besides adding return types to the `edit_link` methods, this looks good to me!

Yes, that makes total sense! Added in [`db206bb` (#2676)](https://github.com/timber/timber/pull/2676/commits/db206bb31cbaac024709f51f366d39381852a11c).

This is now ready for a final review.

nlemoine (approved):
All good to me! Thanks @gchtr

gchtr:
@nlemoine Thanks for the review! I’ll merge this in.

Issue #2672, opened 2022-11-08
1.22 updates Twig to 3.4.3


Levdbas:
## Expected behavior
Updating to 1.22 should install twig to 2.15.3

## Actual behavior
Updating to 1.22 updates twig to 3.4.3

This introduces breaking changes like a missing `filter` tag

First reported: https://github.com/timber/timber/pull/2658#issuecomment-1289531520

See https://getcomposer.org/doc/articles/versions.md#summary.

So this
```json
"twig/twig": ">=1.44.7 || ^2.10",
````

![afbeelding](https://user-images.githubusercontent.com/17764157/200786504-2273d56c-9af3-4643-9c62-2a172ac6c5f9.png)

Leads to the following packages: https://semver.madewithlove.com/?package=twig%2Ftwig&constraint=%3E%3D1.44.7+||+^2.10&stability=stable

## Steps to reproduce behavior

```bash
composer create-project roots/bedrock .
composer require timber/timber
```

## What version of WordPress, PHP and Timber are you using?

WordPress 6.1
PHP 8.0.22
Timber 1.22

## How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Composer

xavivars:
Same thing is happening. 

`"twig/twig": ">=1.44.7 || ^2.10",`
should be
`"twig/twig": "^1.44.7 || ^2.10",`

otherwise, twig 3 is installed (as it's > than 1.44.7)

gchtr:
Thanks for your insights and your help! And thanks for the link to the Semver checker, @Levdbas, that’s really helpful.

This will be fixed in #2679. I’ll see that we can publish a hotfix release for this as soon as possible.

jarednova:
Yep, in the midst of a deploy right now! Thanks to @Levdbas @xavivars and @gchtr for serving this all up on a silver platter for me

PR #2679, opened 2022-11-20
Ensure Twig 3.x is not installed

rmens:
**Ticket**: https://github.com/timber/timber/issues/2672

## Issue
There was an error with versioning in the composer.json file allowing Twig 3.4.3 to be installed. This introduces breaking changes like a missing `filter` tag.

## Solution
Make version selector more strict.

## Impact
This pull request fixes the errors that people are seeing in https://github.com/timber/timber/issues/2672 because we will revert to Twig 1.44.7 or 2.15.3.

## Usage Changes
None.

## Considerations
This is a small hotfix and might be considered version 1.22.1?

## Testing
Yes.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
   },
   "require": {
     "php": ">=7.2.5 || ^8.0",
-    "twig/twig": ">=1.44.7 || ^2.10",
+    "twig/twig": "^1.44.7 || ^2.10",
     "upstatement/routes": "^0.9",
     "composer/installers": "^1.0 || ^2.0",
     "twig/cache-extension": "^1.5"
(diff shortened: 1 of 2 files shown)


gchtr (approved):
@rmens Thanks for your pull request! 🙌

I’ll see that we can release this fix as version 1.22.1 as soon as possible.

gchtr:
Just a quick note that the failing test with PHP 8.1 outputs the following deprecation notice:

```
Deprecated: Return type of Twig\Markup::count() should either be compatible with Countable::count(): int, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /home/runner/work/timber/timber/vendor/twig/twig/src/Markup.php on line 38
```

This is not an error in itself, because PHP 8.1 outputs a lot of deprecation notices. This is expected.

PR #2684, opened 2022-11-24
2.x Fix hint about Co-Authors Plus in Upgrade Guide

gchtr:
We still support Co-Authors Plus, see https://github.com/timber/timber/blob/2.x/src/Integration/CoAuthorsPlusIntegration.php

I initially added the hint to the Upgrade Guide before we added back support.

Let’s remove this from the Upgrade Guide.


nlemoine (dismissed):

nlemoine:
Note that I'm about to add [Authorship](https://github.com/humanmade/authorship) support ;)

gchtr:
> Note that I'm about to add [Authorship](https://github.com/humanmade/authorship) support ;)

Didn’t know that, looks great!

PR #2690, opened 2022-12-13
Fix User entity

nlemoine:
## Issue

Setting public properties for user meta fields results in `{{ user.first_name }}` to return null. Because those properties are not `WP_User` properties and thus not imported.

Since those properties are declared, Twig will not retrieve them either: https://twig.symfony.com/doc/3.x/templates.html#variables

## Solution

Remove public properties matching user meta so `__call` in `Timber\Core` is actually called.

## Impact

None.

## Usage Changes

You can now do `{{ user.first_name }}`, `{{ user.last_name }}`, etc.

## Testing

Just changed the test with description to ensure this works fine.

### DIFF
--- a/docs/v2/guides/users.md
+++ b/docs/v2/guides/users.md
@@ -93,6 +93,44 @@ This allows you to check for a login state with an if statement.
 {% endif %}
 ```
 
+## User data
+
+If you need to access user data like the first and the last name, you can either load them from the properties set on user or with `{{ user.meta }}` for user values that are saved as meta values.
+
+Here’s a little cheat sheet:
+
+```twig
+{# Display name #}
+{{ user.name }}
+{{ user.display_name }}
+
+{# Nice name #}
+{{ user.slug }}
+{{ user.user_nicename }}
+
+{# Email #}
+{{ user.user_email }}
+
+{# Website #}
+{{ user.user_url }}
+
+{# Values loaded from user meta. #}
+{{ user.meta('first_name') }}
+{{ user.meta('last_name') }}
+{{ user.meta('description') }}
+
+{# Contact info #}
+{{ user.meta('facebook') }}
+{{ user.meta('instagram') }}
+{{ user.meta('linkedin') }}
+{{ user.meta('pinterest') }}
+{{ user.meta('soundcloud') }}
+{{ user.meta('tumblr') }}
+{{ user.meta('twitter') }}
+{{ user.meta('youtube') }}
+{{ user.meta('wikipedia') }}
+```
+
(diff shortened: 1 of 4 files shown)


gchtr (changes requested):
In the Upgrade Guide to 2.0, we advise to use `meta()` instead of directly accessing properties: https://timber.github.io/docs/v2/upgrade-guides/2.0/#meta.

To make it possible to use `{{ user.first_name }} or `{{ user.last_name }}`, I would add new methods on `Timber\User` instead. Here’s an example for the `first_name()` method.

```php
public function first_name(): string {
    return $this->meta( 'first_name' );
}
```

This way, we are explicit about where these values are coming from. And we will still have a reference for how to access these values in the documentation, as they will still be listed under https://timber.github.io/docs/v2/reference/timber-user/.

What do you think? If you want, I can add the methods.

nlemoine:
Hey @gchtr, 

I would personally leave it as it is. There are so many metas for users I'm afraid we won't be able to create methods for all of them, more precisely, we might not want to be exhaustive here. I think we can't stand half way here (some metas have direct shortcut methods and some don't), it would be more consistent if we adopt a stricter approach: no shortcut methods for metas, use `meta` instead.

Also note that `WP_User` is not behaving like other `WP_` objects (`WP_Post`, `WP_Term`, etc.). It's the only one that gets metas with magic `__get`.

gchtr:
> Hey @gchtr,
> 
> I would personally leave it as it is. There are so many metas for users I'm afraid we won't be able to create methods for all of them, more precisely, we might not want to be exhaustive here. I think we can't stand half way here (some metas have direct shortcut methods and some don't), it would be more consistent if we adopt a stricter approach: no shortcut methods for metas, use `meta` instead.

Huh, yes, you’re right. That would really be an exhaustive list. I added some documentation for how to access user values in [`4d2575f` (#2690)](https://github.com/timber/timber/pull/2690/commits/4d2575f84e287322b3ab3109e8c44a1908fb8435).

> Also note that `WP_User` is not behaving like other `WP_` objects (`WP_Post`, `WP_Term`, etc.). It's the only one that gets metas with magic `__get`.

The other objects also get meta values through the `__get` magic. This functionality is still here for backwards compatibility, but is discouraged. Or am I missing something?

gchtr (approved):
This looks good to me. I added the relevant changes in the Upgrade Guide in [`473839b` (#2690)](https://github.com/timber/timber/pull/2690/commits/473839bf1bef4815e55811c3016c2f1ed1874afc).

As soon as the checks go green, I’ll merge this.

PR #2695, opened 2023-01-07
2.x Add tests and documentation for external images

gchtr:
Continues work on:

- #2639

## Issue

Tests are still missing for external image functionality.

## Solution

This pull request:

- Adds tests for `Timber\ExternalImage` and make sure that `Timber\ImageDimensions` and `Timber\FileSize` also reach a Coverage of 100%.
- Fixes some bugs that showed up through these tests.
- Addes more parameter and return types where possible.
- Improves documentation.
- Updates the Upgrade Guide.

## Impact

None.

## Usage Changes

None.

## Considerations

I changed how `ExternalImage::srcset()` and `ExternalImage::img_sizes()` work. They now both return `null`. There would be too much work involved now to make these work.

Simply returning one image size in `ExternalImage::srcset()` doesn’t make sense either.

I would even propose to remove these to functions from `ExternalImage` als well as `ImageInterface`. We could add it in the future, when we try `ExternalImage` more compatible.

For now, we should get the feature parity with 1.x so that we can release 2.x.

## Testing

Yes.

@xavivars I would love it if you could also have a look at this, since you did most of the work on this feature. I would also like your opinion on the *Considerations* section above.

Again, thanks for your contribution 💪!


src/Attachment.php
```diff
@@ -203,10 +203,10 @@ public function path()
     public function src()
     {
         if (isset($this->abs_url)) {
-            return URLHelper::maybe_secure_url($this->abs_url);
+            return URLHelper::maybe_secure_url($this->abs_url) ?: null;
```
xavivars, line 206: Should we use coalesce operator instead? `??` instead of `?:`

xavivars:
Thanks a lot for putting this together!

I think your changes make sense, and I agree with your considerations too

src/Attachment.php
gchtr: @xavivars I intentionally used `?:`, because `$this->abs_url` or the result of `URLHelper::maybe_secure_url()` will probably never be `null`. But it could be empty, in which case I want to return `null` instead of an empty string.

xavivars: Makes sense

gchtr:
I fixed the tests in #2703. Now only the tests for PHP 8.2 fail, which we ignore for now. I’m merging this.

PR #2698, opened 2023-01-11
2.x Add `AllowDynamicProperties` attribute

mcaskill:
## Issue

Dynamic (non-explicitly declared) properties are deprecated as of PHP 8.2 and are expected to become a fatal error in PHP 9.0.

As you may have noticed in the CI logs, this generates a lot of noise both in error logs and test run logs.

## Solution

To reduce noise this pull request adds the `#[AllowDynamicProperties]` attribute to the [`Core`](src/Core.php) class.

## Considerations

Ideally, a magic `__set()` method should be added which will require adding a catch-all data property to store the non-declared fields.

```php
/**
 * Store dynamic meta fields.
 *
 * @var array<string, mixed>
 */
private $fields = [];

/**
 * Magic setter for dynamic meta fields, for convenience in Twig views.
 *
 * This method is not meant to be called directly.
 *
 * @example
 * ```php
 * $post = Timber\Post::get( get_the_ID() );
 *
 * $post->favorite_darkside_track = 'Any Colour You Like'
 * ```
 * ```twig
 * {# Since this property does not exist explicitly on the Post class,
 *    it will dynamically dispatch the magic __get() method with an argument
 *    of "favorite_darkside_track" #}
 * <span>Favorite <i>Dark Side of the Moon</i> Track: {{ post.favorite_darkside_track }}</span>
 * ```
 * @link https://secure.php.net/manual/en/language.oop5.overloading.php#object.set
 *
 * @param string $field The name of the property being mutated.
 * @param mixed $valuee The value to assign.
 *
 * @return void
 */
public function __set($field, $value)
{
    $this->fields[$field] = $value;
}
```

## Related

- [wordpress/wordpress#c033058](https://github.com/WordPress/WordPress/commit/c03305852e7e40e61cad5798eba9ebc3b961e27a)

### DIFF
--- a/src/Core.php
+++ b/src/Core.php
@@ -5,6 +5,7 @@ namespace Timber;
 /**
  * Class Core
  */
+#[\AllowDynamicProperties]
 abstract class Core
 {
     public $id;
--- a/src/PostType.php
+++ b/src/PostType.php
@@ -6,6 +6,7 @@ namespace Timber;
  * Wrapper for the post_type object provided by WordPress
  * @since 1.0.4
 */
+#[\AllowDynamicProperties]
 class PostType
 {
     /**
--- a/tests/test-timber-integration-acf.php
+++ b/tests/test-timber-integration-acf.php
@@ -48,7 +48,7 @@ class TestTimberIntegrationACF extends Timber_UnitTestCase
     public function testACFGetFieldTermCategory()
     {
         $tid = $this->factory->term->create();
-        update_field('color', 'blue', "category_${tid}");
+        update_field('color', 'blue', "category_{$tid}");
         $cat = Timber::get_term($tid);
         $this->assertEquals('blue', $cat->color);
         $str = '{{term.color}}';
--- a/tests/test-timber-site.php
+++ b/tests/test-timber-site.php
@@ -1,5 +1,6 @@
 <?php
 
+#[AllowDynamicProperties]
 class TestTimberSite extends Timber_UnitTestCase
 {
     public function testStandardThemeLocation()


nlemoine (changes requested):
H @mcaskill!

Thanks for this PR, I noticed those warnings myself and was about to take care of this.

There's still some dynamic properties issues on the co-authors integration, would mind handle them?

mcaskill:
Hi @nlemoine 

> There's still some dynamic properties issues on the co-authors integration, would mind handle them?

All other dynamic properties deprecations are related to third-party plugins which are out of Timber's control.

While looking for dynamic properties in the Co-Authors integration, I noticed that the [Upgrade to 2.0](docs/v2/upgrade-guides/2.0.md) mentions that [support for Co-Authors Plus was removed in January 2020](https://github.com/timber/timber/commit/ffd474d0e3485db4b488abf6deae6fa61239b64d#diff-9a0786fa7e0e9bfb55892c6028a2dacdcb9d9cbf65901415d36b5eda37eafb26R28-R32).

gchtr:
> While looking for dynamic properties in the Co-Authors integration, I noticed that the [Upgrade to 2.0](docs/v2/upgrade-guides/2.0.md) mentions that [support for Co-Authors Plus was removed in January 2020](https://github.com/timber/timber/commit/ffd474d0e3485db4b488abf6deae6fa61239b64d#diff-9a0786fa7e0e9bfb55892c6028a2dacdcb9d9cbf65901415d36b5eda37eafb26R28-R32).

Yep, support was initially removed. But there’s a pull request to fix that notice, because we still support CoAuthors: https://github.com/timber/timber/pull/2684.

gchtr (approved):
I think we’re good to go here. Thanks so much for leading the way, @mcaskill!

I don’t see any warnings in files that we can control ourselves, except for some float-string to int precision errors in ImageDimensions.php. But I already saw that errors like these will be captured by PHPStan in other pull requests.

nlemoine:
LGTM! thanks @mcaskill

nlemoine (approved):

drzraf:
Could you please backport to 1.x ?

danFWD:
> Could you please backport to 1.x ?

I second this. The way I’ve been adding properties to the $context for years now has now created hundreds of deprecated notices in my sites after upgrading to PHP 8.2. Perhaps I was doing it wrong all along (according to the PHP open-source committee tasked with making these changes), but that doesn’t change the fact that it would be almost impossible to de-tangle it from my sites if and when it becomes a fatal error. 

Appreciate your hard work on this plugin, it is literally indispensable to me.

Issue #2699, opened 2023-01-11
Method `Term::link()` gives `Term` instead of `WP_Term` to `get_term_link()`


mcaskill:
## Expected behavior

A hook on [`pre_term_link`](https://developer.wordpress.org/reference/hooks/pre_term_link/) or [`term_link`](https://developer.wordpress.org/reference/hooks/term_link/) expects to receive [`WP_Term`](https://developer.wordpress.org/reference/classes/wp_term/).

## Actual behavior

[`Term::link()`](https://github.com/timber/timber/blob/2.0.0-beta.1/src/Term.php#L325) passes itself instead of its related `WP_Term` object to [`get_term_link()`](https://developer.wordpress.org/reference/functions/get_term_link/).

```php
$link = get_term_link($this);
```

If a hook is type-hinted to `WP_Term`, PHP will throw a `TypeError`:

```
PHP Fatal error:  Uncaught TypeError: App\Models\Taxonomies\Category::filter_term_link(): Argument #2 ($term) must be of type WP_Term, Timber\Term given, called in web/wp/...
Stack trace:
#0 web/wp/wp-includes/class-wp-hook.php(308): App\Models\Taxonomies\Category->filter_term_link('https://examp...', Object(Timber\Term), 'category')
#1 web/wp/wp-includes/plugin.php(205): WP_Hook->apply_filters('https://examp...', Array)
#2 web/wp/wp-includes/taxonomy.php(4652): apply_filters('term_link', 'https://examp...', Object(Timber\Term), 'category')
#3 vendor/timber/timber/src/Term.php(325): get_term_link(Object(Timber\Term))
#4 vendor/twig/twig/src/Extension/CoreExtension.php(1635): Timber\Term->link()
```

## Steps to reproduce behavior

```php
add_filter( 'term_link', function ( string $link, WP_Term $term, string $taxonomy ) : string {
    // do something
    return $link;
}, 10, 3 );
```

## What version of WordPress, PHP and Timber are you using?

* PHP 5.6+
* WordPress 6.1+
* Timber 1+ or 2+

## Proposed solution

As far as I can see, there's three ways to resolve this:

1. Remove the type or use a less specific type:

    ```php
    add_filter( 'term_link', function ( string $link, object $term, string $taxonomy ) : string {
        // do something
        return $link;
    }, 10, 3 );
    ```

2. If in PHP 8+, use a union type:

    ```php
[...]

PR #2701, opened 2023-01-12
2.x Fix `get_term_link()` compatibility in `Timber\Term`

mcaskill:
**Ticket**: #2699

## Issue

A hook on [`pre_term_link`](https://developer.wordpress.org/reference/hooks/pre_term_link/) or [`term_link`](https://developer.wordpress.org/reference/hooks/term_link/) expects to receive [`WP_Term`](https://developer.wordpress.org/reference/classes/wp_term/) but receives [`Timber\Term`](src/Term.php).

## Solution

In [`Term::link()`](src/Term.php#L323-L326), pass `WP_Term` instead of `Timber\Term`.

## Impact

This is a breaking change.

Any implementation that hooks into `get_term_link()` expecting `Timber\Term` will be receive `WP_Term` (the type documented by WordPress).

### DIFF
--- a/src/Term.php
+++ b/src/Term.php
@@ -322,7 +322,7 @@ class Term extends CoreEntity
      */
     public function link()
     {
-        $link = get_term_link($this);
+        $link = get_term_link($this->wp_object);
 
         /**
          * Filters the link to the term archive page.


nlemoine (approved):
Thanks @mcaskill !

PR #2705, opened 2023-02-09
2.x Fix menu item compatibility with WPML

mcaskill:
## Issue

WPML uses a custom nav menu item class: `WPML_LS_Menu_Item`.

When this menu item is passed to `MenuItemFactory::from()`, the method ignores it and returns `null` which violates the return type of the array map closure in `Menu::convert_menu_items()` leading to a fatal error:

```
PHP Fatal error:  Uncaught TypeError: Timber\Menu::Timber\{closure}(): Return value must be of type Timber\MenuItem, null returned in vendor/timber/timber/src/Menu.php:289
Stack trace:
#0 [internal function]: Timber\Menu->Timber\{closure}(Object(WPML_LS_Menu_Item))
#1 vendor/timber/timber/src/Menu.php(288): array_map(Object(Closure), Array)
#2 vendor/timber/timber/src/Menu.php(213): Timber\Menu->convert_menu_items(Array)
#3 vendor/timber/timber/src/Factory/MenuFactory.php(276): Timber\Menu::build(Object(WP_Term), Object(stdClass))
#4 vendor/timber/timber/src/Factory/MenuFactory.php(101): Timber\Factory\MenuFactory->build(Object(WP_Term), Array)
#5 vendor/timber/timber/src/Factory/MenuFactory.php(47): Timber\Factory\MenuFactory->from_location('site-header...', Array)
#6 vendor/timber/timber/src/Timber.php(986): Timber\Factory\MenuFactory->from('site-header...', Array)
#7 web/app/themes/foobar/includes/Site.php(966): Timber\Timber::get_menu('site-header...')
#8 web/app/themes/foobar/includes/Site.php(296): App\Theme\Site->get_timber_menus()
#9 web/wp/wp-includes/class-wp-hook.php(308): App\Theme\Site->filter_timber_context(Array)
#10 web/wp/wp-includes/plugin.php(205): WP_Hook->apply_filters(Array, Array)
#11 vendor/timber/timber/src/Timber.php(1278): apply_filters('timber/context', Array)
#12 vendor/timber/timber/src/Timber.php(1169): Timber\Timber::context_global()
#13 vendor/locomotivemtl/wp-lib-theme/inc/Template/AbstractTemplate.php(40): Timber\Timber::context()
#14 web/app/themes/foobar/includes/Template/BaseTemplate.php(19): App\Theme\Template\AbstractTemplate->__construct()
#15 web/app/themes/foobar/theme/404.php(9): App\Theme\Template\BaseTemplate->__construct()
#16 web/wp/wp-includes/template-loader.php(106): include('...')
#17 web/wp/wp-blog-header.php(19): require_once('...')
#18 web/index.php(18): require('...')
#19 {main}
  thrown in vendor/timber/timber/src/Menu.php on line 289
```

## Solution

Added a filter `timber/menu/items` to `Menu::build()`, as [proposed](https://github.com/timber/timber/pull/2705#issuecomment-1439520248) by @gchtr, to allow one to filter all menu items. In turn, `WpmlIntegration` hooks into this filter to replace any occurrences of `WPML_LS_Menu_Item` with `WP_Post`.

<details>
<summary>Initial Solution</summary>

I propose adding a hook to `MenuItemFactory::from()` to allow integrations to convert any non-standard menu item into an acceptable menu item: `WP_Post`.

In `WpmlIntegration`, I've added a callback on this new hook to convert WPML's `WPML_LS_Menu_Item` into `WP_Post`.

The hook is named `timber/menuitem/pre_from` but the suffix, `pre_from`, is not my favourite name. I'm open to suggestions.

</details>

## Impact

Adding an extra hook could affect performance but I should be minor.

## Considerations

Alternatively, a filter to convert non-standard menu items could be applied to a late-callback on `wp_get_nav_menu_items` but that risks affecting any logic elsewhere that expects that custom menu item.

## Testing

A unit test for the filter has been added: `TestTimberMenu::testMenuItemsFilter()`.

### DIFF
--- a/src/Integration/WpmlIntegration.php
+++ b/src/Integration/WpmlIntegration.php
@@ -19,7 +19,6 @@ class WpmlIntegration implements IntegrationInterface
         add_filter('timber/url_helper/url_to_file_system/path', [$this, 'file_system_to_url'], 10, 1);
         add_filter('timber/menu/id_from_location', [$this, 'menu_object_id_filter'], 10, 1);
         add_filter('timber/menu/items', [$this, 'menu_items_filter'], 10, 1);
-        add_filter('timber/menuitem/pre_from', [$this, 'menu_item_pre_from_filter'], 10, 1);
         add_filter('timber/image_helper/_get_file_url/home_url', [$this, 'file_system_to_url'], 10, 1);
     }
 
--- a/src/Integration/WpmlIntegration.php
+++ b/src/Integration/WpmlIntegration.php
@@ -18,7 +18,7 @@ class WpmlIntegration implements IntegrationInterface
         add_filter('timber/url_helper/get_content_subdir/home_url', [$this, 'file_system_to_url'], 10, 1);
         add_filter('timber/url_helper/url_to_file_system/path', [$this, 'file_system_to_url'], 10, 1);
         add_filter('timber/menu/id_from_location', [$this, 'menu_object_id_filter'], 10, 1);
-        add_filter('timber/menu/items', [$this, 'menu_items_filter'], 10, 1);
+        add_filter('timber/menu/item_objects', [$this, 'menu_item_objects_filter'], 10, 1);
         add_filter('timber/image_helper/_get_file_url/home_url', [$this, 'file_system_to_url'], 10, 1);
     }
 
@@ -35,7 +35,7 @@ class WpmlIntegration implements IntegrationInterface
         return wpml_object_id_filter($id, 'nav_menu');
     }
 
-    public function menu_items_filter(array $items)
+    public function menu_item_objects_filter(array $items)
     {
         return array_map(
             fn ($item) => ($item instanceof WPML_LS_Menu_Item ? new WP_Post($item) : $item),
--- a/src/Integration/WpmlIntegration.php
+++ b/src/Integration/WpmlIntegration.php
@@ -2,6 +2,9 @@
 
 namespace Timber\Integration;
 
+use WP_Post;
+use WPML_LS_Menu_Item;
+
 class WpmlIntegration implements IntegrationInterface
 {
     public function should_init(): bool
@@ -15,6 +18,8 @@ class WpmlIntegration implements IntegrationInterface
         add_filter('timber/url_helper/get_content_subdir/home_url', [$this, 'file_system_to_url'], 10, 1);
         add_filter('timber/url_helper/url_to_file_system/path', [$this, 'file_system_to_url'], 10, 1);
         add_filter('timber/menu/id_from_location', [$this, 'menu_object_id_filter'], 10, 1);
+        add_filter('timber/menu/items', [$this, 'menu_items_filter'], 10, 1);
+        add_filter('timber/menuitem/pre_from', [$this, 'menu_item_pre_from_filter'], 10, 1);
         add_filter('timber/image_helper/_get_file_url/home_url', [$this, 'file_system_to_url'], 10, 1);
     }
 
@@ -29,5 +34,13 @@ class WpmlIntegration implements IntegrationInterface
     public function menu_object_id_filter($id)
     {
         return wpml_object_id_filter($id, 'nav_menu');
+    }
+
+    public function menu_items_filter(array $items)
+    {
+        return array_map(
+            fn ($item) => ($item instanceof WPML_LS_Menu_Item ? new WP_Post($item) : $item),
+            $items
+        );
     }
 }
--- a/src/Menu.php
+++ b/src/Menu.php
@@ -205,6 +205,28 @@ class Menu extends CoreEntity
          */
         $sorted_menu_items = apply_filters('wp_nav_menu_objects', $sorted_menu_items, $args);
 
+        /**
+         * Filters the sorted list of menu item objects before creating the Menu object.
+         *
+         * @since 2.0.0
+         * @example
+         * ```
+         * add_filter( 'timber/menu/items', function ( $items ) {
+         *     return array_map(function ($item) {
+         *         if ( is_object( $item ) && ! ( $item instanceof \WP_Post ) ) {
+         *             return new \WP_Post( get_object_vars( $item ) );
+         *         }
+         *
+         *         return $item;
+         *     }, $items);
+         * } );
+         * ```
+         *
+         * @param array<mixed> $item
+         * @param WP_Term $menu
+         */
+        $sorted_menu_items = apply_filters('timber/menu/items', $sorted_menu_items, $menu);
+
         // Create Menu object
         $nav_menu = new static($menu, (array) $args);
         $nav_menu->sorted_menu_items = $sorted_menu_items;
--- a/src/Menu.php
+++ b/src/Menu.php
@@ -211,7 +211,7 @@ class Menu extends CoreEntity
          * @since 2.0.0
          * @example
          * ```
-         * add_filter( 'timber/menu/items', function ( $items ) {
+         * add_filter( 'timber/menu/item_objects', function ( $items ) {
          *     return array_map(function ($item) {
          *         if ( is_object( $item ) && ! ( $item instanceof \WP_Post ) ) {
          *             return new \WP_Post( get_object_vars( $item ) );
@@ -225,7 +225,7 @@ class Menu extends CoreEntity
          * @param array<mixed> $item
          * @param WP_Term $menu
          */
-        $sorted_menu_items = apply_filters('timber/menu/items', $sorted_menu_items, $menu);
+        $sorted_menu_items = apply_filters('timber/menu/item_objects', $sorted_menu_items, $menu);
 
         // Create Menu object
         $nav_menu = new static($menu, (array) $args);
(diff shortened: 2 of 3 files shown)


gchtr:
Hey @mcaskill, that’s a good catch! Thanks for putting together this pull request!

I recently ran into a similar issue and I think a filter for menu items is a great solution.

@nlemoine and I recently discussed that factory methods don’t have strict parameter types. In the future, we might want to use a function signature that forces a type. We already started to do this in a lot of other places. So the `from()` method could then look like this:

```php
public function from(WP_Post $item, Menu $menu): ?MenuItem {}
```

(To handle post IDs, we would maybe need a from_id()` method.)

When the filter is inside the `from` method, it would not be possible to add strict types. So I’m thinking about a different place to do this. Maybe `Menu::convert_menu_items()` would be a better place for the filter?

Or even a filter in `Menu::build()`, before `$nav_menu->convert_menu_items()` is called? This way, we could make sure that as soon as the menu items are processed by Timber, they will always be `WP_Post` objects.

```php
$sorted_menu_items = apply_filters('timber/menu/items', $sorted_menu_items, $menu);

// Create Menu object
$nav_menu = new static($menu, (array) $args);
$nav_menu->sorted_menu_items = $sorted_menu_items;

// Convert items into MenuItem objects
$sorted_menu_items = $nav_menu->convert_menu_items($sorted_menu_items);
```

As you said, I wouldn’t hook into `wp_get_nav_menu_items` either.

I would also like to get @nlemoine’s opinion before proceeding with this.

nlemoine:
> Or even a filter in Menu::build(), before $nav_menu->convert_menu_items() is called?

Sounds good to me!

mcaskill:
So, should this filter be applied to each menu item or to the batch as a whole?


```php
$sorted_menu_items = array_map(function ($item) use ($args) {
    return apply_filters('timber/menu/pre_convert_menu_item', $item, $args);
}, $menu_items);

// Convert items into MenuItem objects
$sorted_menu_items = $nav_menu->convert_menu_items($sorted_menu_items);
```

```php
$sorted_menu_items = apply_filters('timber/menu/pre_convert_menu_items', $sorted_menu_items, $args);

// Convert items into MenuItem objects
$sorted_menu_items = $nav_menu->convert_menu_items($sorted_menu_items);
```

I actually stumbled upon [`PagesMenu::pre_setup_nav_menu_item()`](https://github.com/timber/timber/blob/9107027/src/PagesMenu.php#L197-L216) which could be added to `Menu` (or similarly named method) and have the filter called there.

nlemoine:
Feels better to apply the filter on the array of items, more similar to `wp_get_nav_menu_items`, what do you think @gchtr?

gchtr:
Intuitively, I would **loop over the whole array** and not over each item individually. This has the added benefit that you could filter items, too. But the intended usage would definitely be to convert the items.

> I actually stumbled upon [`PagesMenu::pre_setup_nav_menu_item()`](https://github.com/timber/timber/blob/9107027/src/PagesMenu.php?rgh-link-date=2023-03-01T16%3A06%3A49Z#L197-L216) which could be added to `Menu` (or similarly named method) and have the filter called there.

This method is used to convert items that are not native menu items yet. It works for posts. For terms, this it would have to be a little different for maximum compatibility. I wouldn’t use this method on all menu items. But it *could* be used for items added to a menu dynamically. 

We sometimes use a function that contains this logic in our projects if we want to add all child pages of a page as sub nav items. Or automatically adding terms a subitems to a nav menu item, if they contain posts.

A `Menu::pre_setup_nav_menu_item()` function could make sense if we filtered each item individually. But if we loop over the array, we prepare and convert the items before they are added to the `Menu` object. For working with the items in the `Menu` object, this could make things easier if can expect that they always have the same type.

Or did you have something different in mind?

mcaskill:
I've updated the pull requests initial description and force-pushed the new solution as [proposed](https://github.com/timber/timber/pull/2705#issuecomment-1439520248) by @gchtr which I had somehow missed/misread.

nlemoine (dismissed):
Thanks @mcaskill !

src/Integration/WpmlIntegration.php
```diff
@@ -15,6 +18,8 @@ public function init(): void
         add_filter('timber/url_helper/get_content_subdir/home_url', [$this, 'file_system_to_url'], 10, 1);
         add_filter('timber/url_helper/url_to_file_system/path', [$this, 'file_system_to_url'], 10, 1);
         add_filter('timber/menu/id_from_location', [$this, 'menu_object_id_filter'], 10, 1);
+        add_filter('timber/menu/items', [$this, 'menu_items_filter'], 10, 1);
+        add_filter('timber/menuitem/pre_from', [$this, 'menu_item_pre_from_filter'], 10, 1);
```
gchtr, line 22: This filter doesn’t exist anymore, right?

src/Integration/WpmlIntegration.php
```diff
@@ -15,6 +18,8 @@ public function init(): void
         add_filter('timber/url_helper/get_content_subdir/home_url', [$this, 'file_system_to_url'], 10, 1);
         add_filter('timber/url_helper/url_to_file_system/path', [$this, 'file_system_to_url'], 10, 1);
         add_filter('timber/menu/id_from_location', [$this, 'menu_object_id_filter'], 10, 1);
+        add_filter('timber/menu/items', [$this, 'menu_items_filter'], 10, 1);
```
gchtr, line 21:
```suggestion
        add_filter('timber/menu/item_objects', [$this, 'menu_item_objects_filter'], 10, 1);
```

src/Integration/WpmlIntegration.php
```diff
@@ -30,4 +35,12 @@ public function menu_object_id_filter($id)
     {
         return wpml_object_id_filter($id, 'nav_menu');
     }
+
+    public function menu_items_filter(array $items)
```
gchtr, line 39:
```suggestion
    public function menu_item_objects_filter(array $items)
```

src/Menu.php
```diff
@@ -205,6 +205,28 @@ public static function build(?WP_Term $menu, $args = []): ?self
...
+         *
+         *         return $item;
+         *     }, $items);
+         * } );
+         * ```
+         *
+         * @param array<mixed> $item
+         * @param WP_Term $menu
+         */
+        $sorted_menu_items = apply_filters('timber/menu/items', $sorted_menu_items, $menu);
```
gchtr, line 228:
```suggestion
        $sorted_menu_items = apply_filters('timber/menu/item_objects', $sorted_menu_items, $menu);
```

tests/test-timber-menu.php
```diff
@@ -474,6 +474,28 @@ public function testMenuOptionsInNavMenuCssClassFilter()
...
+        $menu_id = $term['term_id'];
+
+        $filter = function (array $items, WP_Term $menu) {
+            return array_map(function ($item) {
+                $item->classes[] = "test_{$item->ID}";
+                return $item;
+            }, $items);
+        };
+
+        $this->add_filter_temporarily('timber/menu/items', $filter, 10, 2);
```
gchtr, line 489:
```suggestion
        $this->add_filter_temporarily('timber/menu/item_objects', $filter, 10, 2);
```

src/Menu.php
```diff
@@ -205,6 +205,28 @@ public static function build(?WP_Term $menu, $args = []): ?self
          */
         $sorted_menu_items = apply_filters('wp_nav_menu_objects', $sorted_menu_items, $args);
 
+        /**
+         * Filters the sorted list of menu item objects before creating the Menu object.
+         *
+         * @since 2.0.0
+         * @example
+         * ```
+         * add_filter( 'timber/menu/items', function ( $items ) {
```
gchtr, line 214:
```suggestion
         * add_filter( 'timber/menu/item_objects', function ( $items ) {
```

gchtr (changes requested):
This looks good, thanks for the updates @mcaskill.

I was just thinking about the name again that I initially proposed, and I’m not sure `timber/menu/items` is the best name. When we talk about Timber menu items, I would think of items that are already `Timber\MenuItem` objects. But this filter is applied *before* they are converted to `Timber\MenuItem` objects.

If we ever wanted to introduce a filter for the items *after* converting them, we would probably want to name it `timber/menu/items`. So we couldn’t use it here.

That’s why I propose to change the filter name to `timber/menu/item_objects`. This is a general name for "various item objects that might have to be cleaned first".

What do you think? Better? Would that make sense? Or can you think of a better name?

The name `timber/menu/pre_convert_menu_items` would also work, but I think the `pre_` prefix is mainly used to short-circuit a filter in WordPress and return a result early, which isn’t the case here.

Sorry for the nitpicking 😊.

Side note: You don’t have to force push your changes whenever you make changes. It’s harder to read the progress on a pull request and I think we don’t mind have a couple more commits. If we wanted, we could squash the commits when merging. But then again, it’s also nice to have small bits and pieces in the future, when coming back and trying to find out why a change was made.

mcaskill:
Thanks for the suggestions.

I agree that `timber/menu/pre_convert_menu_items` gives the impression of short-circuiting given WordPress' usage of `pre_` in hook names.

The name `timber/menu/item_objects` seems like the best option so far.

gchtr (approved):
Nice, thanks for the updates @mcaskill! We can merge this in now.

Issue #2576, opened 2022-04-26
Menu : array_flip(): Can only flip string and integer


merijnponzo:
With Timber 2 , WPML and php8
I added multiple menu's but I got the message:
```
Warning: array_flip(): Can only flip string and integer values, entry skipped in //wordpress/websitename/vendor/timber/timber/lib/Factory/MenuFactory.php on line 186
```

My nav is outputting
```
array(4) {
  ["primary"]=>
  int(3)
  ["secondary"]=>
  NULL
  ["footer"]=>
  int(3)
  ["language"]=>
  NULL
}
```

So checking in MenuFactory.php on line 186 for NULL values would solve the problem
```
protected function get_menu_location(WP_Term $term) : ?string {
    // fix
    $safe_locations = get_nav_menu_locations();
    $safe_locations_ar = [];
    foreach($safe_locations as $key => $value){
        if($value){
            $safe_locations_ar[$key] = $value;
        }
    }
    $locations = array_flip($safe_locations_ar);
    return $locations[$term->term_id] ?? null;
}
```

nlemoine:
@merijnponzo This is fixed, It will be available as soon the PR mentioned above is merged (soon hopefully).

gchtr:
I’m closing this, because this was fixed in #2517. It will be available with then next 2.x release, though I can’t say when that will be exactly.

PR #2707, opened 2023-02-13
2.x Fix Menu theme locations array to enforce string or integers

mcaskill:
## Issue

With Timber v2 , WPML, and PHP 8, if any nav menu location happens to be `null`, this warning occurs:

```
Warning: array_flip(): Can only flip string and integer values, entry skipped in vendor/timber/timber/src/Menu.php on line 255
```

This is similar to issue #2576 where the warning was with `MenuFactory::get_menu_location()`.

## Solution

The solution is to apply the [same fix made to `MenuFactory`](https://github.com/timber/timber/commit/e4228795cf41a0e1d0c870fd70a9b969f2b44906); filter the nav menu locations array to enforce strings or integers.

## Testing

The `TestTimberMenu::testThemeLocationProperty()` method was updated to match `TestMenuFactory::testGetMenuLocation()`.

### DIFF
--- a/src/Menu.php
+++ b/src/Menu.php
@@ -252,7 +252,8 @@ class Menu extends CoreEntity
         }
 
         // Set theme location if available
-        $this->theme_location = array_flip(get_nav_menu_locations())[$term->term_id] ?? null;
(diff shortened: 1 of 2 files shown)


gchtr (approved):
Thanks for this fix, @mcaskill! 👍

PR #2708, opened 2023-02-19
Fix php warning

LoranRendel:
Fix php warning (Trying to access array offset on value of type int)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-  "name": "timber/timber",
+  "name": "loranrendel/timber",
   "type": "library",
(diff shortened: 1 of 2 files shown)


gchtr:
Hey @LoranRendel 

I’m sorry that I have to reject this, but we won’t change the package name :).

If you want to use Twig v3, you should use [Timber v2](https://timber.github.io/docs/v2).

If you wanted that PHP error to be fixed, we would need more information about the change. It would be best if you filled out the pull request template with all the information requested. What was the input that caused the error in the first place? How did that object property that caused the error look like?

You can tell us if you need help with that.

PR #2713, opened 2023-03-02
Make comment neutral.

Levdbas:
No need to use 'his' here. I'd like to think of code of something neutral.


gchtr (approved):
Yes, no-brainer! 👍

PR #2714, opened 2023-03-03
2.x clean init

nlemoine:
## Issue

- Register admin notice on `admin_init` action
- Remove `pre_get_posts` filters, I never paid attention to those filters but I don't think this should be Timber's responsability to provide some kind of query arguments normalization. 
- Filters integrations to ensure they implement `IntegrationInterface`
- Removes `test_compatibility()` which actually does nothing

## Impact

Moving from closures to static functions (not ideal but better) provides a better way to remove those filters if needed.

## Usage Changes

None.

### DIFF
--- a/docs/v2/guides/custom-fields.md
+++ b/docs/v2/guides/custom-fields.md
@@ -175,7 +175,7 @@ This example that uses a [WP_Query](http://codex.wordpress.org/Class_Reference/W
 
 ```php
 $args = [
-    'numberposts' => -1,
+    'posts_per_page' => -1,
     'post_type' => 'post',
     'meta_key' => 'color',
     'meta_value' => 'red',
--- a/docs/v2/guides/custom-integrations.md
+++ b/docs/v2/guides/custom-integrations.md
@@ -85,10 +85,10 @@ The hard part’s over. Now you need to tell Timber about your class:
 ```php
 use MyProject\MyAcfIntegration;
 
-add_filter('timber/integrations', function (array $classes): array {
-    $classes[] = MyAcfIntegration::class;
+add_filter('timber/integrations', function (array $integrations): array {
+    $integrations[] = new MyAcfIntegration();
 
-    return $classes;
+    return $integrations;
 });
 ```
 
--- a/src/Post.php
+++ b/src/Post.php
@@ -998,7 +998,7 @@ class Post extends CoreEntity implements DatedInterface, Setupable
         if (is_array($post_type)) {
             $post_type = implode('&post_type[]=', $post_type);
         }
-        $query = 'post_parent=' . $this->ID . '&post_type[]=' . $post_type . '&numberposts=-1&orderby=menu_order title&order=ASC&post_status[]=publish';
+        $query = 'post_parent=' . $this->ID . '&post_type[]=' . $post_type . '&posts_per_page=-1&orderby=menu_order title&order=ASC&post_status[]=publish';
         if ($this->post_status === 'publish') {
             $query .= '&post_status[]=inherit';
         }
--- a/src/Timber.php
+++ b/src/Timber.php
@@ -4,11 +4,12 @@ namespace Timber;
 
 use Timber\Factory\CommentFactory;
 use Timber\Factory\MenuFactory;
-
 use Timber\Factory\PagesMenuFactory;
 use Timber\Factory\PostFactory;
 use Timber\Factory\TermFactory;
 use Timber\Factory\UserFactory;
+use Timber\Integration\IntegrationInterface;
+
 use WP_Post;
 use WP_Query;
 
@@ -88,17 +89,6 @@ class Timber
     {
     }
 
-    /**
-     * Tests whether we can use Timber
-     * @codeCoverageIgnore
-     */
-    protected function test_compatibility()
-    {
-        if (is_admin() || $_SERVER['PHP_SELF'] == '/wp-login.php') {
-            return;
-        }
-    }
-
     protected function init_constants()
     {
         defined("TIMBER_LOC") or define("TIMBER_LOC", realpath(dirname(__DIR__)));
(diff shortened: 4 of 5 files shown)


nlemoine:
Hey @gchtr, 

Just went throw the `Timber::init` and made some changes. The most notable one targets integrations initializations. I went for object instead of class names which provides a better way to filter them out if not implementing `IntegrationInterface` (will avoid using Reflection in favor of `instanceof`).

Let me know if it's ok for you.

gchtr (changes requested):
Hey @nlemoine 

Thanks for these fixes. It looks like we forgot some stuff when we did the big cleanup. The changes for the integrations look good to me.

Some context for the `pre_get_posts` actions: We initially started discussing whether we want `Timber::get_posts()` to behave like `get_posts()` or like `WP_Query()` in https://github.com/timber/timber/pull/1989. That’s why we added a `timber/get_posts/mirror_wp_get_posts` filter to control that behavior in 1.x.

For 2.x, we chose to remove the fixes, see

- https://timber.github.io/docs/v2/guides/posts/#differences-from-wp-cores-get_posts
- https://timber.github.io/docs/v2/upgrade-guides/2.0/#removed-hooks
    > `timber/get_posts/mirror_wp_get_posts` – This filter was used so that `Timber::get_posts()` mimicks the behavior of WordPress’s `get_posts()` function. We removed it because `Timber::get_posts()` is the new official API. If you want the same behavior, there are [arguments that you can pass to this function](https://timber.github.io/docs/v2/guides/posts#differences-from-wp-cores-get_posts).

So, it’s good that we remove these `pre_get_posts` actions, too. However, I realized that we should take care of `numberposts`.  The `numberposts` argument is only used by `get_post()` and will be converted to `posts_per_page` there. We still use it in certain places in the documentation and many developers will have used it as well. To ease transition to 2.x, I want to add a removal notice for this in `Timber::get_posts()` in case someone uses `numberposts`.

I think it shouldn’t be a problem for `cat`, because that is also supported in `WP_Query`.

nlemoine:
>  However, I realized that we should take care of `numberposts`

When breaking change involves modifying "real" PHP code (filters, function signature, etc.), I totally understand deprecating it before removal. When the upgrade path is as easy and fast as a search/replace in your editor (as long as it's notified in the upgrade guide of course), I'm more on the pure removal side (I personally never used `numberposts` since 2009 😉)

Deprecations come with a small burden that means we have to write test for it, remember to remove it later, etc. Is it really worth here? 🤔

> The numberposts argument is only used by `get_post()`

You meant `get_posts` right? 😄

If you still want to deprecate it, I don't think this fix should reside in a `pre_get_posts` filter which applies on every queries. It should be handled just like it in [`get_posts`](https://github.com/WordPress/wordpress-develop/blob/5eefddf4afcd738abd52fe5b7ebbe163412532ad/src/wp-includes/post.php#L2397-L2399).

gchtr:
@nlemoine I added a new pull request in #2717 to clarify what I mean.

I wouldn’t deprecate the functionality in the sense that it’s still supported, but remove the functionality with a doing-it-wrong notice so that developers are automatically notified.

> Deprecations come with a small burden that means we have to write test for it, remember to remove it later, etc. Is it really worth here? 🤔

We have a lots of notices for things we can catch easily. With a big Upgrade Guide like for 2.0, it’s easy to miss a small thing like `numberposts`. And a notice added now might prevent us from having to read, triage and answer issues about it after the release.

> > The numberposts argument is only used by `get_post()`
> 
> You meant `get_posts` right? 😄

Of course 😅.

nlemoine:
> I added a new pull request in https://github.com/timber/timber/pull/2717 to clarify what I mean.

Yes, that's also what I had in mind. But I still disagree on this and strongly think we shouldn't bother taking care of this. This is also an arbitrage between time we can spend on OSS projects and "transition sweetness" we can provide in that time.

>  it’s easy to miss a small thing like numberposts

This one is super extra sweet 😄

> so that developers are automatically notified
> And a notice added now might prevent us from having to read, triage and answer issues about it after the release.

Catching those deprecations implies that you have setup your dev environment the right way, activated all kind of errors on dev mode, activated your debug log file, etc. It's WordPress, a world of very wide development skills and i bet most of them don't read their log file (me included), if it happens to exists  😉

Anyway, since you took care of this, let's merge this 🙂

gchtr:
> Yes, that's also what I had in mind. But I still disagree and strongly think we shouldn't bother taking care of this. This is also an arbitrage between time we can spend on OSS projects and "transition sweetness" we can provide in that time.

@nlemoine I totally get your point. This is probably still my habit from the early days of 2.x to try to make it easy to upgrade to 2.x. But I also see that stuff like this takes a lot of time, which we already don’t have a lot of. It’s good that you remind me of this.

> Catching those deprecations implies that you have setup your dev environment the right way, activated all kind of errors on dev mode, activated your debug log file, etc. It's WordPress, a world of very wide development skills and i bet most of them don't read their log file (me included), if it happens to exists 😉

Well, you would only need `WP_DEBUG` set to `true` for it to appear at the top of your website in a local development environment by default, if don’t set any other debugging constants. I can’t assess how many developers use it, but sure, you would need to have the right environment for this.

gchtr (approved):

nlemoine:
> This is probably still my habit from the early days of 2.x to try to make it easy to upgrade to 2.x

No problem. 

One last argument I have in mind but didn't explicitly stated. 

There's also an important change in Timber 2.x that has consequences on how we should handle such things. From my point of view, since it will be composer only and no more a "plug-and-play" plugin, it also means that we can expect a bit more from our users. Hence, I assume we don't necessarily have to "hold their hand" (sorry, it's a french expression 😉) in the upgrade process.

gchtr:
> It also means that we can expect a bit more from our users (installing through composer imply that you're already pretty confortable with PHP). Hence, IMHO, I assume we don't necessarily have to "hold their hand" (sorry, it's a french expression 😉) in the upgrade process.

Oh yes, that’s a good argument and we definitely can expect more from users. For future releases, we can surely skip deprecations and directly remove features.

"Holding someone’s hand", we have the same expression in German 😉.

PR #2715, opened 2023-03-06
Fix iterable case

nlemoine:
`Iterable` with capital `i` might let you think that it's a native class whereas it's a pseudo type

## Issue
Not really an issue but I think iterable return types might lead to confusion. Proof is one them was backslashed in a PHPDoc block.

## Solution
Lower case all of them. 

## Impact
None.

## Usage Changes
None.

## Considerations

Using iterables on every `get_{objects}` (terms, users, comments) is bit annoying. While it's legitimate on `posts` because we return true collection objects, all the other ones just return simple arrays.

But since the return type is `iterable`:
- you're discouraged to use `array_*` functions (outside of Twig) which are often very useful (map, filter, etc.), although it will work in the current implementation for users, terms and comments
- when using `array_*` functions, my IDE (and PHPStan) complaints about it ("Expected type 'array'. Found 'iterable'.") and they're right to
- you can't use [`iterator_to_array`](https://github.com/bpolaszek/php-iterable-functions/blob/7136af22d8f7165f173cbd04cfa224443ffacdb0/README.md?plain=1#L25-L27) if already an array

In the end, all these "iterables" are [converted to arrays in Twig](https://github.com/twigphp/Twig/blob/78c18ad58e48d3fff2f7812bb60b6176ee52689a/src/Extension/CoreExtension.php#L723-L727). 

So the underlying question is: do we really need `iterable`? I guess these types have been introduced to be future proof in case we might implement collections for WP objects. But will we for such object object types (terms/users/comments)?

### DIFF
--- a/src/Factory/CommentFactory.php
+++ b/src/Factory/CommentFactory.php
@@ -64,7 +64,7 @@ class CommentFactory
         ));
     }
 
-    protected function from_wp_comment_query(WP_Comment_Query $query): Iterable
+    protected function from_wp_comment_query(WP_Comment_Query $query): iterable
     {
         return array_map([$this, 'build'], $query->get_comments());
     }
--- a/src/Factory/PostFactory.php
+++ b/src/Factory/PostFactory.php
@@ -75,7 +75,7 @@ class PostFactory
         ));
     }
 
-    protected function from_wp_query(WP_Query $query): Iterable
+    protected function from_wp_query(WP_Query $query): iterable
     {
         return new PostQuery($query);
     }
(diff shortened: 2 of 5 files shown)


gchtr (approved):
> So the underlying question is: do we really need `iterable`? I guess these types have been introduced to be future proof in case we might implement collections for WP objects. But will we for such object object types (terms/users/comments)?

I don’t think we need that for terms, users and comments in the foreseeable future. And if we did, we would find a proper solution. So I’d say, we don’t really need it.

PR #2716, opened 2023-03-07
#2458 Add tests and fix merge conflicts

gchtr:
Related:

- #2458

This pull request fixes the merge conflicts in #2458 and adds the relevant tests to check whether the changes work correctly.


src/ThemeHelper.php
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * For example, consider a theme with the following tree structure (2.x default child theme structure):
+ * - my-theme-name/
+ * -> images/
+ * --> foo.jpg
+ * -> theme/
+ * --> functions.php
+ * --> style.css
+ *
+ * Theme template in style.css is my-theme-name/theme
```
nlemoine, line 21: Sorry, I don't really see why style.css would reside in a subdirectory? https://developer.wordpress.org/themes/basics/main-stylesheet-style-css/#location

Am I missing something?

gchtr: This is related to https://github.com/timber/starter-theme/issues/102 and the changes introduced in https://github.com/timber/starter-theme/pull/103. Apparently it’s possible to add theme files into a subdirectory of the theme.

I myself don’t use the starter theme, so I don’t have a strong opinion on this one.

nlemoine: I have very mixed feelings about his. Not sure this kind of structure:
- should be handled in core
- further more, should be proposed as default in the starter theme

PR #2719, opened 2023-03-08
Remove string helpers

nlemoine:
This is handled by PHP 8 and can easily be polyfilled. 

Timber is full of helpers that already exists either in WordPress or PHP (and can easily be polyfilled).


gchtr (approved):
Great! This all looks good to me. I only added a note in the Upgrade guide in [`0d65559` (#2719)](https://github.com/timber/timber/pull/2719/commits/0d65559102269b608762e99b9ff131cfdb50a3f7).

PR #2720, opened 2023-03-08
Improve scheme detection and take advantage of WP native functions

nlemoine:
Using `is_ssl`:
- covers more use cases than self made functions
- is already there


gchtr (approved):
Makes sense.

PR #2725, opened 2023-03-21
2.x Add ImageHelper filters to customize paths

mcaskill:
Related:

- #2458
- #2453

## Issue

As reported in aforementioned issue and pull request, the `ImageHelper::theme_url_to_dir()` method is unable to resolve a URL/path in a non-standard theme directory structure.

This is the case if your "stylesheet" directory is located in a subdirectory of your theme `my-theme/theme` as was initially proposed in Timber's v2 starter-theme and employed by various theme boilerplates.

## Solution

The solution, as [proposed](https://github.com/timber/timber/pull/2458#issuecomment-1471396358) by @gchtr, is to add filters in `ImageHelper` to allow one to customize the logic of converting a URL to a path.

> For example, we could add a `timber/theme_url_to_dir` filter that allows you to filter the result of `Timber\ImageHelper::theme_url_to_dir()`.

I've added four filters:

- `ImageHelper::analyze_url()`
	- `timber/image_helper/pre_analyze_url` to short-circuit the method's default URL analysis logic.
	- `timber/image_helper/analyze_url` to mutate the default analyzed URL components.
- `ImageHelper::theme_url_to_dir()`
	- `timber/image_helper/pre_theme_url_to_dir` to short-circuit the method's default path resolution logic.
	- `timber/image_helper/theme_url_to_dir` to mutate the default resolved path.

I've added filters to `analyze_url()` since it's the one that calls `theme_url_to_dir()` and also contains opinionated logic for analyzing the URL.

Maybe the short-circuit filters could still call their non-pre filters, this would involve wrapping the default logic in a "did pre-filter return null".

Here's how its being used in a client project:

```php
public function register_filters() : void {
	add_filter( 'timber/image_helper/pre_theme_url_to_dir', [ $this, 'filter_timber_image_helper_pre_theme_url_to_dir' ], 10, 2 );
}

public function filter_timber_image_helper_pre_theme_url_to_dir( ?string $path, string $src ): ?string {
	if ( is_null( $path ) && str_starts_with( $src, $this->theme_uri ) ) {
		$path = str_replace( $this->theme_uri, '', strtok( $src, '?' ) );
		$path = $this->theme_path . '/' . ltrim( $path, '/' );
	}

	return $path;
}
```

## Impact

Adding an extra hooks could affect performance but it should be minor.

## Usage Changes

The current logic of those methods does not change, filters are added to allow one to handle them early or mutate the default resulting values.

## Considerations

See discussions from #2458

## Testing

- Modified `install-wp-tests.sh` to ensure `TMPDIR` is canonical.
- Fixed tests of `TestTimberImageHelperInternals::testAnalyzeURLTheme()`.
- Added tests for filters in `ImageHelper::analyze_url()`.
- Added tests for filters in `ImageHelper::theme_url_to_dir()`.


gchtr:
Adding filters is definitely the way to go. Because this isn’t a breaking change and we want to focus on getting 2.0 ready, I switched this to `2.x Future`.

Moustachos:
@gchtr any update on this one now that 2.x is stable? :)

Thanks!

Levdbas:
Hi @Moustachos , we have this in mind for a first minor release for Timber 2. Release date is TBD.

Moustachos:
Thanks @Levdbas 👍

src/ImageHelper.php
```diff
@@ -516,26 +516,63 @@ public static function set_sideload_image_upload_dir(array $upload)
...
+         * @param array<string, mixed>|null $info The URL components array to short-circuit with. Default null.
+         * @param string                    $url  The URL pointing to an image.
+         */
+        $result = \apply_filters('timber/image_helper/pre_analyze_url', null, $url);
+        if (null === $result) {
+            $result = self::_analyze_url($url);
+        }
+
+        /**
+         * Filters the array of anlayzed URL components.
```
gchtr:
```suggestion
         * Filters the array of analyzed URL components.
```

src/ImageHelper.php
```diff
@@ -516,26 +516,63 @@ public static function set_sideload_image_upload_dir(array $upload)
...
+        return \apply_filters('timber/image_helper/analyze_url', $result, $url);
+    }
+
+    /**
+     * Returns information about a URL.
+     *
+     * @param  string $url A URL (absolute or relative) pointing to an image.
+     * @return array<string, mixed> An array (see keys in code below).
+     */
+    private static function _analyze_url(string $url): array
```
gchtr: According to [PSR-12](https://www.php-fig.org/psr/psr-12/), underscores should not be used for method names.

> Method names MUST NOT be prefixed with a single underscore to indicate protected or private visibility. That is, an underscore prefix explicitly has no meaning.

Yes, we still have some of those in the codebase, but for new private functions, we usually don’t introduce underscore prefixes.

This means that we should find a new name, which is the hard part here. Maybe we can use one of the following?

- `get_url_components()`
- `run_analyze_url()`

src/ImageHelper.php
```diff
@@ -576,16 +614,52 @@ public static function analyze_url($url)
...
+        return \apply_filters('timber/image_helper/theme_url_to_dir', $path, $src);
+    }
+
+    /**
+     * Converts a URL located in a theme directory into the raw file path.
+     *
+     * @param string  $src A URL (http://example.org/wp-content/themes/twentysixteen/images/home.jpg).
+     * @return string Full path to the file in question.
+     */
+    private static function _theme_url_to_dir(string $src): string
```
gchtr: Same as for `_analyze_url()`, I wouldn’t use an underscore prefix here.

```suggestion
    private static function get_dir_from_theme_url(string $src): string
```

gchtr (changes requested):
Thanks for taking this on, @mcaskill. And sorry for the longer wait.

I really appreciate that you took the time to carefully craft the DocBlocks for the filters with proper descriptions 🙌.

I just have on issue with the underscore prefixes in the private function names. Otherwise, this looks good to me and I think we can merge this for the next minor release.

src/ImageHelper.php
mcaskill: Renamed `_analyze_url` to `get_url_components`.

src/ImageHelper.php
mcaskill: Renamed `_theme_url_to_dir` to `get_dir_from_theme_url`.

gchtr (approved):

gchtr:
This looks good from my side 👍.

@nlemoine Can you have a quick look and merge this, if it looks alright to you, too?

nlemoine (approved):
Still feels a bit odd to handle such things but doesn't hurt 👍

Issue #2444, opened 2021-04-22
is_image() function not working with URL params appended to an image src


thetoine:
## Expected behavior

Image class method `is_image()` should return `true` when is called for supported file extensions. 

https://github.com/timber/timber/blob/384f130c7b5dc34363b8bea60eff8b50503b6f9f/lib/Image.php#L535

## Actual behavior

When the attachement SRC includes URLs parameters, `is_image()` method returns `false`.

Example :

https://cdn-name.imgix.net/2021/04/photo-1434871619871-1f315a50efba-scaled.jpeg?auto=format&fit=scale&h=1361&ixlib=php-3.3.0&w=2048&wpsize=2048x2048&s=fd1fd431490df94d4d338e1b1729f3fa

This is happening because WP internal method `wp_check_filetype()` does not support URL params. A 6 years old ticket is still open about this specific issue :

https://core.trac.wordpress.org/ticket/30377

## Possible fix and patch :

Strip URL params in `PathHelper` methods, check out my commit here :

https://github.com/thetoine/timber/commit/97390e202853ed783c0cc9508cc7eb4899ee83e9

## What version of WordPress, PHP and Timber are you using?

WP 5.7.1
Timber 1.18.2

## How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)

WP.org

nlemoine:
Salut @thetoine, 

This issue has been fixed in https://github.com/timber/timber/pull/2730. Unfortunately, we're probably not going to backport this change to Timber 1.x. 

Timber 2.x is almost out and we're putting all our efforts to make it happen ASAP.

thetoine:
Bonjour @nlemoine, thanks for the update!

PR #2730, opened 2023-04-25
Fix `PostFactory::is_image` incorrectly using `wp_check_filetype`

stayallive:
Related:

- #2444

## Issue

Images with query params are not correctly identified as images.

## Solution

The current way to detect an image is to use `wp_attachment_is_image` and use that with `wp_check_filetype` to detect if the file has an expected extension.

However, `wp_check_filetype` is not meant to be used with an URL but with a path instead. So this check fails currently if the URL has query params for example.

The fix is to use `get_attached_file` to get the actual file path instead of the URL matching what `wp_attachment_is_image` is doing.

## Impact

I don't think there are any performance or BC changes since this is a bug fix IMHO.

## Usage Changes

No.

## Considerations

Ideally `wp_attachment_is_image` would been used, but seeing there are custom filters in place this is (no longer) a viable solution so mimicking what `wp_attachment_is_image` is doing seems like the best choice at the moment.

## Testing

I have to see if I can figure it out, this code path seems untested at the moment and setting up a local dev environment requires a little more time then I can invest right now.

### DIFF
--- a/src/Factory/PostFactory.php
+++ b/src/Factory/PostFactory.php
@@ -172,7 +172,7 @@ class PostFactory
 
     protected function is_image(WP_Post $post)
     {
-        $src = wp_get_attachment_url($post->ID);
+        $src = get_attached_file($post->ID);
         $mimes = wp_get_mime_types();
         // Add mime types that Timber recongizes as images, regardless of config
         $mimes['svg'] = 'image/svg+xml';


nlemoine (approved):
Thanks for your PR @stayallive! You're absolutely right.

PR #2733, opened 2023-05-03
2.x Fix menu location compatibility with WPML

mcaskill:
## Issue

On projects using WPML and that customize the `Menu` class via the `timber/menu/classmap` filter, the custom classes won't be used because WPML only translates [`get_nav_menu_locations()`](https://developer.wordpress.org/reference/functions/get_nav_menu_locations/) ([`theme_mod_nav_menu_locations`](https://developer.wordpress.org/reference/hooks/theme_mod_name/)) in specific use cases such as during the filtering of `wp_nav_menu_args` and when we are in the Admin.

The (translated) menu's term ID won't match the (default language) menu ID assigned to the same location.

This affects `MenuFactory` and `Menu`.

## Solution

Since the issue affects multiple classes, and [based on suggestions](https://github.com/timber/timber/pull/2733#issuecomment-1552023083):

- Added a callback on the `theme_mod_nav_menu_locations` hook in `WpmlIntegration` to translate the menus mapped to locations.
- Added methods `get_menu_location()` and `get_menu_locations()` to `Timber` for convenient usage by `MenuFactory` and `Menu`.

## Impact

Adding an extra hook could affect performance but I should be minor.

Translating the menu IDs of menu locations (`theme_mod_nav_menu_locations`) could have a more noticeable impact on performance but should also be minor.

## Usage Changes

Externally, developers can take advantage of the added hook to fix WPML's edge case.

Timber and developers can rely on the new methods to easily retrieve filtered nav menu locations.

## Testing

Unit tests have not added for `Timber`, nor has the unit tests for `MenuFactory`, `Menu`, and `WpmlIntegration` been updated.

With the custom `wpml_object_id_filter()` function declared in the tests bootstrap, adding/updating tests will require more time than I have at the moment.

This pull request's functionality has been tested on a client project that has many menu locations and customizes the menus extensively.


nlemoine:
Hello @mcaskill, 

I have mixed feelings about adding another helper class. I'm actually trying to [get rid of them](https://github.com/timber/timber/pull/2719).

I don't use WPML, isn't there a way to use WordPress core filters to handle this? `get_theme_locations` is just a shortcut to `get_theme_mod` which [contains a filter](https://github.com/WordPress/wordpress-develop/blob/f7ca55eb038767c0be79a35565b3170a999bd0d5/src/wp-includes/theme.php#L1039) that could do the exact same thing.

Could you give it a try?

mcaskill:
Yeah, I was reluctant to adding another helper class but couldn't think of a quicker solution at the time.

I'll try to refactor this in the next month or so. There should indeed be a way to hook into the `theme_mod` filter from `WpmlIntegration` to avoid a helper class.

gchtr:
It would be great if we could find an easier way to fix this.

Moreover, I think that a fix for this doesn’t have to make it into the first 2.0 release. It looks like we can still solve this with a minor release. To me, it looks like if we find the right filter, this can be fixed without having to introduce a breaking change to Timber. That’s why I’m changing the label to `2.x-Future`.

mcaskill:
I've rebased the latest changes to the main branch and resolved issues.

* Add missing stub of `WPML_LS_Menu_Item` for PHPStan
* Fix mock `wpml_object_id_filter()` to avoid recursion with `theme_mod_nav_menu_locations` for WPML integration.

Most of the failed checks from GitHub Actions appear to be related to existing issues?

Levdbas:
Hi @mcaskill , tests are running again! It seems there are only a few failing tests that needs a fix. Mostly about the moved `get_menu_location` method, are you able to fix those?

mcaskill:
Hi @Levdbas, I've rebased and fixed the tests. Thanks.

src/Timber.php
```diff
@@ -1056,6 +1056,39 @@ public static function get_pages_menu(array $args = [])
         return $menu;
     }
 
+    /**
+     * Get the navigation menu location assigned to the given menu.
+     *
+     * @param  WP_Term|int $term The menu to find; either a WP_Term object or a Term ID.
+     * @return string|null
```
Levdbas, line 1066: We need to mark these new methods as @api functions to end up in the docs! Also probably a good idea to add the `@since 2.3.0` tag on both methods.

Levdbas (dismissed):
LGTM! Great work!

src/Timber.php
mcaskill: Good point! Done.

Levdbas (approved):

PR #2736, opened 2023-05-12
Add workflow to check docs links

nlemoine:
Related:

- https://github.com/timber/timber/pull/2732

## Issue

It may be hard to keep all links in docs up to date. 

## Solution

As suggested in PR above, setup a workflow to spot them.

It will check every week docs links and create an issue if bad links are found.


gchtr (approved):
Oh wow, we can automate this? Yes please!

PR #2746, opened 2023-05-21
Add @api tag to Helper

Levdbas:
- https://github.com/timber/timber/issues/2745

## Issue
Helper.php class is not in the current V2 docs. This seems to be caused by the missing @api tag. This also causes one of the missing links in #2745 

## Solution
Add the @api tag and deploy the new docs.

## Impact
More docs!


nlemoine (approved):
Nice! :)

gchtr (approved):
Good catch, thanks for this one!

PR #2749, opened 2023-05-26
Fix custom field test

nlemoine:
## Issue

While testing a new way to run tests, I stumbled upon this.

ACF functions shouldn't be used outside ACF integration tests.

## Solution

Use `update_post_meta`.

## Impact

None

### DIFF
--- a/tests/test-timber-meta.php
+++ b/tests/test-timber-meta.php
@@ -1039,8 +1039,8 @@ class TestTimberMeta extends Timber_UnitTestCase
             'post_content' => 'Cool content bro!',
             'post_date' => '2020-02-07 08:03:00',
         ]);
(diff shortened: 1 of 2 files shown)


nlemoine:
The new way of running tests (using SQLite) I'm trying also brought this issue regarding truncating tables to my attention.

Levdbas (changes requested):
See docs of update_post_meta()
`
update_post_meta( int $post_id, string $meta_key, mixed $meta_value, mixed $prev_value = '' ): int|bool`

First comes the id, then the key, then the value.

nlemoine:
Whoops, looks like I've been too fast on this 😅 Weird the tests pass, still.

It's fixed!

Levdbas (approved):

PR #2753, opened 2023-05-26
Refactor file models

nlemoine:
## Issue
First intended to only remove the `Filesize` class and refactor the `size()`/`size_raw()` methods for many reasons:
- `Attachment` is a model, it's not in its scope to return/format a raw file size into a human readable one. You might want to format a size in other places, this is filters role.
- The size method included an HTML breaking space (`&nbsp;`), which is fine on the Twig side but what if I need the text only version of it? Although Timber is mostly about Twig, models can also be used PHP side and in various contexts (JSON, etc.). The `Attachment` model should only by responsible for returning a raw file size you can manage yourself afterwards.
- The size was grabbed directly from filesystem, which is too bad because [WP 6.0 introduced a `filesize` key](https://make.wordpress.org/core/2022/05/02/media-storing-file-size-as-part-of-metadata/) which is set and stored on attachment first upload. The `Attachment::size()` method now try first to get that value before getting to filesystem which will lead to potentially better performances.

While working on the above issue, I noticed that `Attachment` was setting every possible properties when building the object (`get_info()`), getting values from database that might not be needed already and maybe won't at all.

I decided to refactor this class to only grab metadata values when they are called which might lead into more efficient code.

## Solution

- Remove Filesize class
- Refactor ExternalImage, Attachment, Image to add types and getting metadata in a more lazy way
- Add `format_bytes` filter to format raw file size into a human readable file size
- Fix/add tests
- Fix city-museum.jpg by removing exif data which was causing issues in tests

## Impact

- Image/Attachment/ExternalImage properties are now protected, return types have changed
- FileSize class has been removed (but it is a BC since it was introduced in 2.0?)

## Usage Changes

If you want to format a file size,  you will now have to use `{{ image.size|format_bytes(2) }}` to display a human readable file size.

## Considerations

Since metadata also contains image dimensions, ImageDimensions should also be refactored to grab those values from stored metadata rather than reading them from the filesystem.

## Testing

Yes!


nlemoine:
Note that I held myself back from removing more stuff 🙂

Things to consider:
- Remove `path()` methods in all models. Same reason as n°1 mentioned above: I don't think models need this method. If you want a relative path for something (`src`, `link`, etc.): use the `relative` filter or use a WP filter/plugin. Plus I don't think we need [a helper to get the relative](https://github.com/timber/timber/blob/99f7a1a10268a3c73e0570aff92faedd4f52f3f3/src/Twig.php#L348) URL, there's [already one in WP](https://developer.wordpress.org/reference/functions/wp_make_link_relative/). 
- Do you know what's the `abs_url` for? https://github.com/timber/timber/blob/99f7a1a10268a3c73e0570aff92faedd4f52f3f3/src/Attachment.php#L65 This feels redundant with `$image->src()`
- I think we should remove the [`maybe_secure_url`](https://github.com/timber/timber/blob/99f7a1a10268a3c73e0570aff92faedd4f52f3f3/src/Attachment.php#L187) helper. If the URL is `http`, then it's a somewhere else problem that will be hidden (and not fixed and brought to the developer attention) by this fix that's not Timber in Timber's scope IMHO.
- Same goes for [`remove_double_slashes`](https://github.com/timber/timber/blob/99f7a1a10268a3c73e0570aff92faedd4f52f3f3/src/URLHelper.php#L310): we shouldn't aim at fixing every issues whose causes are elsewhere in the stack.

I could further but I'd like your views on this first :)

src/Attachment.php
```diff
@@ -296,15 +279,12 @@ public function size_raw()
...
-
-            if (!empty($file_info['ext'])) {
-                $this->file_extension = strtoupper($file_info['ext']);
-            }
+        if (isset($this->file_extension)) {
+            return $this->file_extension;
         }
-
-        return $this->file_extension;
+        // TODO: why uppercase? should we really use wp_check_filetype()?
```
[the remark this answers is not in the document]
nlemoine: 💯  for basename, they shouldn't be any filtering in here since the object is built andalready passed quite some checks to get there.

gchtr:
I’ll look at this more closely soon, but I want to say that I totally agree with the changes. There are some details that we need to consider and I think I have some insight into some of your raised issues. This is definitely the direction we should take here. And we still can, because the `Attachment` class and the filesize features were all introduced in 2.x.

nlemoine:
> And we still can, because the Attachment class and the filesize features were all introduced in 2.x.

That's good news! 😁

src/Attachment.php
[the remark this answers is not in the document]
nlemoine: Went for `pathinfo` 😅

src/Twig.php
```diff
@@ -371,6 +371,11 @@ public function get_timber_filters()
                 },
             ],
 
+            // Numbers filters
+            'format_bytes' => [
```
gchtr, line 375: Can we use `size_format` instead of `format_bytes`? I feel like we don’t have to invent a new function name here.

src/ExternalImage.php
```diff
@@ -50,57 +50,62 @@ class ExternalImage implements ImageInterface
...
-    public $file;
+    protected string $file;
 
     /**
      * File location.
      *
      * @api
      * @var string The absolute path to the attachmend file in the filesystem
-     *             (Example: `/var/www/htdocs/wp-content/themes/my-theme/images/my-pic.jpg`)
+     *             (Example: `/var/www/htdocs/wp-content/uploads/2015/08/my-pic.jpg`)
```
gchtr, line 83: I would keep the `/themes/my-theme/images/` path, because that’s a typical location where an ExternalImage might live.

src/ExternalImage.php
```diff
@@ -221,6 +230,33 @@ public function path(): string
         return URLHelper::get_rel_path($this->file);
     }
 
+    /**
+     * Gets the relative path to the uploads folder of an attachment.
```
gchtr, line 234: If it’s not an attachment and live in a theme folder, will this still be relative path to the uploads folder?

src/ExternalImage.php
```diff
@@ -221,6 +230,33 @@ public function path(): string
...
+     *
+     * @return string
+     */
+    public function file(): string
+    {
+        return $this->file;
+    }
+
+    /**
+     * Gets the absolute path to an attachment.
```
gchtr, line 246:
```suggestion
     * Gets the absolute path to the image.
```

src/Image.php
```diff
@@ -64,7 +64,7 @@ class Image extends Attachment implements ImageInterface
      * @api
      * @var array An array of available sizes for the image.
      */
-    public $sizes = [];
+    protected array $sizes;
```
gchtr, line 67: This visibility change will require a note in the Upgrade Guide.

src/Attachment.php
```diff
@@ -86,39 +76,9 @@ class Attachment extends Post
...
-            $data['file_loc'] = $basedir . DIRECTORY_SEPARATOR . $data['file'];
-        } elseif (isset($data['_wp_attached_file'])) {
-            $data['file'] = $data['_wp_attached_file'];
-            $data['file_loc'] = $basedir . DIRECTORY_SEPARATOR . $data['file'];
-        }
-
-        $data['file_size'] = new FileSize($data['file_loc']);
-
-        return $data;
+        return (string) $this->src();
```
gchtr, line 81: Isn’t this typecast redundant?

nlemoine:
I removed the `$file` property which was useless and redundant with `$file_loc`, they both stored the absolute path.

There are still things I'm not fully ok with:
- image dimensions are still read from file, which could be avoided by getting image metadata for registered sizes and improve performance. But would require quite some refactoring (ImageDimensions doesn't takes some `$metadata` to look for dimensions)
- I noticed that `ImageHelper::sideload_image()` is very optimistic 😅 There are some places in that method that can go wrong and are not handled:
    -  the extension parsing is [not covering all possible cases](https://github.com/timber/timber/blob/2.x/src/ImageHelper.php#L442C13-L442C13)
    - the [`download_url` result ](https://github.com/timber/timber/blob/2.x/src/ImageHelper.php#L441C17-L441C29)should be handled as soon as it is is returned
    - there is no handling of the [possible `error` index](https://github.com/timber/timber/blob/2.x/src/ImageHelper.php#L452) of `wp_upload_bits()`

This is something we can probably fix later.

gchtr (approved):

gchtr:
Very nice, thanks a lot @nlemoine.

I added your notes as a separate issue in #2785. It would be great if we could resolve these issues as well.

Issue #2626, opened 2022-08-05
Source code contains to be resolved issues "todos"


szepeviktor:
Issue `grep -rFni todo src/`

## Expected behavior

No output.

## Actual behavior

41 todos

## What version of Timber are you using?

2.x

gchtr:
To resolve this, I would propose to go through the todos and:

- Create GitHub issues with the `2.0` label for issues we need to resolve before the 2.0 release. For issues that are easy to resolve right away, create a pull request.
- Create GitHub issues with the `2.x Future` label for the todos we can in a future release.
- Leave the todos that are about inline documentation as they are.

If someone wants to help with this, assign yourself to this issue or let us know before you start.

gchtr:
Resolved in #2757.

PR #2756, opened 2023-05-30
2.x Fix a bug when `the_post` hook runs twice on each post in a loop

gchtr:
Related:

- #2626

## Issue

While working through the todos, I found a skipped test. Upon running it, I realized that when we loop over a collection of posts, the `the_post` hook is actually called twice for each post.

This is because `PostsIterator::next()` calls `PostsIterator::current()`, which runs `Post::setup()` every time.

https://github.com/timber/timber/blob/f248f3c3e3e37a620b9af91ad0cd7602655aaa0c/src/PostsIterator.php#L60

https://github.com/timber/timber/blob/f248f3c3e3e37a620b9af91ad0cd7602655aaa0c/src/PostsIterator.php#L21-L43

https://github.com/timber/timber/blob/f248f3c3e3e37a620b9af91ad0cd7602655aaa0c/src/Post.php#L314

## Solution

I added a `$last_post` property to `PostsIterator`. It’s used to save the last post that was set up. Using `PostsIterator::current()` would lead to the same result, but because it always calls `Post::setup()`, it will call hooks that we don’t need to call.

## Impact

Less compatibility issues. More performance.

## Usage Changes

None.

## Considerations

I hope I didn’t miss a functionality that’s already baked into `ArrayIterator` that we could use for this.

## Testing

Yes.

### DIFF
--- a/src/PostsIterator.php
+++ b/src/PostsIterator.php
@@ -10,6 +10,12 @@ use Timber\Factory\PostFactory;
 class PostsIterator extends \ArrayIterator
 {
     /**
+     * @var \Timber\Post The last post that was returned by the iterator. Used
+     *                   to skip the logic in `current()`.
+     */
+    protected Post $last_post;
+
+    /**
      * Prepares the state before working on a post.
      *
      * Calls the `setup()` function of the current post to setup post data. Before starting the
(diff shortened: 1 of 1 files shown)


Levdbas (approved):
Nice find @gchtr !

nlemoine (approved):
Nice!

Issue #1689, opened 2018-03-09
Feature Request: UserQuery class


acobster:
I would love to be able to do something like:

```php
$users = new Timber\UserQuery([
  'role' => 'my-custom-role'
]);
```

...and iterate over a collection of `Timber\User` objects. I believe the closest I can get to this currently is by executing a raw `WP_User_Query` and mapping each object to a `Timber\User`, but it would be really nice to be able to do this directly.

I'd love to get started on a PR for this, as this would be valuable for a number of projects I'm working on. But I could see this being a non-trivial undertaking so wanted to get your thoughts first.

Thanks!

jarednova:
@acobster I'd love to do that as well! @gchtr is working on a similar `TermQuery` as we speak. `UserQuery` would be the piece that completes the puzzle.

In terms of guidelines, we want to follow [`WP_User_Query`](https://codex.wordpress.org/Class_Reference/WP_User_Query) as much as possible so we (or users) don't have to re-invent the wheel. Is this something you can start chipping away on?

acobster:
Absolutely! Totally agree about `WP_User_Query`; the API should be as transparent as possible.

gchtr:
I’ve looked into creating a new `TermQuery`. This actually goes much deeper than I initially thought and as @acobster has mentioned, is probably a «non-trivial undertaking». I think what we’d really want is to have parity not only for [`WP_Term_Query`](https://make.wordpress.org/core/2016/07/06/wp_term_query-in-wordpress-4-6/) and `WP_User_Query`, but also for `WP_Comment_Query`.

Currently, a PostQuery is built up from different objects, that I try to describe here:

- **PostQuery** – the public class to use that handles all the other classes behind the scenes.
- **PostGetter** – handles how posts are fetched from the database, depending on the object that is passed to `PostQuery`. It `PostQuery` is constructed with an array of posts, it will return a `PostCollection`. Otherwise it will return a `QueryIterator` object.
- **PostCollection** – a wrapper for an array posts that will automatically convert all posts to `Timber\Posts` (or custom objects, when requested).
- **QueryIterator** – handles queries defined as strings or arrays and will also return a `PostCollection` as soon as posts are requested.
- **PostsIterator** – handles what happens on each iteration when looping through a `PostCollection`. Basically, this sets up the `$post` global to make Timber compatible with (most) plugins.
- **Pagination** – handles pagination of posts for a `QueryIterator`.

[...]

PR #2757, opened 2023-05-30
2.x Cleanup todo in code

gchtr:
Related:

- #2626

## Issue

Source code contains a lot of todos.

## Solution

I worked through the todos in code as well as some skipped tests. I left the todos that are only about documentation.

I will add separate comments with the changes.

## Impact

Make sure we didn’t forget anything important for 2.x

## Usage Changes

None.

## Considerations

None.

## Testing

Yes.


src/Comment.php
```diff
@@ -564,7 +564,6 @@ protected function avatar_host($email_hash)
 
     /**
      * @internal
-     * @todo  what if it's relative?
```
gchtr, line 567: If this is an issue in the future, it will pop up as an issue.

src/Factory/UserFactory.php
```diff
@@ -86,7 +86,6 @@ protected function from_user_object($obj): CoreInterface
         ));
     }
 
-    // @todo return a UserCollection instance?
```
gchtr, line 89: Not at the moment. We discussed somewhere that user collections or collections for other types of content are not needed at the moment (https://github.com/timber/timber/issues/1689). We can add it if we really need it.

src/Loader.php
```diff
@@ -587,7 +587,6 @@ public function clear_cache_twig()
         $twig = $this->get_twig();
 
         // Get the configured cache location.
-        // @todo What if this is a custom caching implementation?
```
gchtr, line 590: This will pop up as an issue in case it becomes a problem.

src/Loader.php
```diff
@@ -596,7 +595,6 @@ public function clear_cache_twig()
         }
 
         if (is_string($cache_location) && is_dir($cache_location)) {
-            // @todo What if not all files could be deleted?
```
gchtr, line 599: Might be improved when we introduce better and streamlined error handling, see https://github.com/timber/timber/issues/2210.

src/Post.php
```diff
@@ -47,7 +47,6 @@
  *     </div>
  * </article>
  * ```
- * @todo implement JsonSerializable?
```
gchtr, line 50: Serialization can be added if needed. It’s documented in https://timber.github.io/docs/v2/guides/posts/#serialization.

src/PostsIterator.php
```diff
@@ -35,7 +35,6 @@ public function current()
          */
         $wp_post = parent::current();
         // Lazily instantiate a Timber\Post instance exactly once.
-        // @todo maybe improve performance by caching the instantiated post.
```
gchtr, line 38: Performance and caching is an area where we could improve in a lot of places. In case we want to improve post performance, we’ll find the right place to do it.

src/Timber.php
```diff
@@ -318,7 +313,6 @@ public static function get_attachment($query = false, $options = [])
     {
         $post = static::get_post($query, $options);
 
-        // @todo make this determination at the Factory level.
```
gchtr, line 321: We might improve this in the future. For the moment, it seems to work.

src/Timber.php
```diff
@@ -457,9 +450,6 @@ public static function get_posts($query = false, $options = [])
             );
         }
 
-        /**
-         * @todo Are there any more default options to support?
```
gchtr, line 461: We’ll figure them out if needed.

tests/test-timber-deprecated.php
gchtr, line 1: This file didn’t really do anything.

tests/test-timber-post.php
```diff
@@ -178,25 +178,20 @@ public function testNextCategory()
 
     public function testNextCustomTax()
     {
-        $v = get_bloginfo('version');
-        if (version_compare($v, '3.8', '<')) {
```
gchtr, line 182: We don’t need to support WordPress < 3.8.

tests/test-timber-static.php
```diff
@@ -68,20 +68,22 @@ public function testFrontPageAsPage()
...
             'post_type' => 'page',
         ]);
-        update_option('page_for_posts', $page_id);
-        $this->go_to(home_url('/?p=' . $page_id));
-        $children = $this->factory->post->create_many(10, [
+        $posts = $this->factory->post->create_many(10, [
             'post_title' => 'Timmy',
         ]);
+
+        update_option('show_on_front', 'page');
```
gchtr, line 80: This line was missing to get the test working.

tests/test-timber.php
```diff
@@ -578,22 +539,6 @@ public function testPostsPerPageBig()
         $this->assertCount(15, $posts);
     }
 
-    /**
-     * @expectedDeprecated Timber\Timber::query_post()
-     */
-    public function testQueryPost()
```
gchtr, line 584: I guess this somehow tests whether `Timber::query_post()` would change the post global so that `get_the_ID()` returns the queried post. This logic is not supported anymore.

tests/test-timber.php
```diff
@@ -772,16 +717,6 @@ public function testGettingEmptyArray()
         $this->assertEquals([], $collection->to_array());
     }
 
-    public function testFromFalse()
-    {
-        // We don't actually test this directly in TestTimberPostGetter::testGettingWithFalse();
-        // that test directly instantiates a collection.
-        $this->markTestSkipped('@todo what should this be?');
```
gchtr, line 779: With the newest version, when using `false` with `Timber::get_posts()`, the default query will be used. This test would test a deprecated functionality then. That’s why I removed it.

tests/test-timber.php
```diff
@@ -901,48 +833,6 @@ public function testThePostHook()
         $this->assertSame(3, $the_post_count);
     }
 
-    public function testGetAttachment()
-    {
-        $this->markTestSkipped('@todo seems like a lot of what gets tested here is core WP file mgmt. Is that what we want?');
```
gchtr, line 906: > @todo seems like a lot of what gets tested here is core WP file mgmt. Is that what we want?

I agree. Doesn’t seem to test anything worth testing that isn’t covered in other tests.

szepeviktor:
I feel a fresh smell.
Thank you!!

Issue #2737, opened 2023-05-16
Timezone Errors with post dates


asecondwill:
### Expected Behavior

I'm expecting the timezone setting in wp-admin to be used by the timber filters to output the  time_ago

### Actual behavior

My site is set to Australia/Adilade
 {{ post.date('j F Y H:i') }}      shows the published date in the correct timezone.  

Time Ago
`{{ post.post_date|time_ago }}`
Time ago says "In about 7 hours" at 1230 UTC when I'm testing, again the publish date is 1930 today (16th May) - Adilaide time, which was 3 hours ago.  

UPDATE:  if i try 
` {{ post.date|time_ago }}    `

I get 14hours ago.  which is still wrong, it's 2230 in Adilade so 3 hours ago. 

### Steps to reproduce behavior

Time ago
`{{ post.post_date|time_ago }}`

with composer twig: 
 `       "timber/timber": "2.0.0.beta-1",`

### Notes

I got the same behaviour in Timber 1, so upgraded thinking this had been fixed in 2 having read through the tickets here.

### What version of Timber are you using?

2.0.0.beta-1

### What version of WordPress are you using?

Wordpress 6.2

### What version of PHP are you using?

8.0

### How did you install Timber?

Installed or updated Timber through Composer

asecondwill:
Updated as I figured out the published date is

` {{ post.date('j F Y H:i') }}     `

not 
` {{ post.post_date|date('j F Y')}}`

gchtr:
Most of the date functionality was fixed in 2.x, but it could very well be that I missed the `|time_ago()` filter. We will first have to add a test to confirm this is still an issue.

Levdbas:
Did some local tests @gchtr and I recreated the same two outcomes that @asecondwill describes be it with different dates and time zones.

{{ post.post_date|time_ago }} does not seem to take the timezone into account because it relies on `time()` instead of using `current_time()` [see WP docs](https://developer.wordpress.org/reference/functions/current_time/)

{{ post.date }} as stated in the [Timber docs](https://timber.github.io/docs/reference/timber-post/#date) only returns the publication date without the time. So when using |time_ago on that it will always count use that day at midnight. That in combination with our first issue gives leads to inexplicable behavior. Using `{{post.date('j F Y h:i')|time_ago}}`, so explicitly force hours and minutes in, fixes this issue as well. So when you want to use |time_ago, it's best to use post_date.

Now I understand why i struggled so much with dates in the past as well.

[...]

PR #2758, opened 2023-05-31
2.x Fix a bug when `|time_ago` didn’t consider timezones correctly

gchtr:
Related:

- #2737

## Issue

The `|time_ago` filter doesn’t work correctly when timezone is set to Australia.

## Solution

- Don’t use `strtotime()`, but create DateTime objects with their corresponding timezone.
- Update some documentation entries for correct usage with `{{ post.date }}`. It’s important to use a time string including the time.

## Impact

The `|time_ago` filter should work more reliably.

## Usage Changes

None.

## Considerations

None.

## Testing

Yes.

### DIFF
--- a/docs/v2/guides/date-time.md
+++ b/docs/v2/guides/date-time.md
@@ -262,7 +262,7 @@ DateTimeHelper::time_ago($post->date());
 **Twig**
 
 ```twig
-{{ post.date|time_ago }}
+{{ post.date('U')|time_ago }}
 ```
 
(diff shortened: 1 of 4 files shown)


Levdbas (approved):
Nice work! I've updated the v2 doc as well

gchtr:
> Nice work! I've updated the v2 doc as well

Thanks for that, @Levdbas! Well spotted.

PR #2760, opened 2023-05-31
2.x Fix skipped tests

gchtr:
Related:

- #2626

### DIFF
--- a/tests/test-timber-function-wrapper.php
+++ b/tests/test-timber-function-wrapper.php
@@ -44,9 +44,8 @@ class TestTimberFunctionWrapper extends Timber_UnitTestCase
 
     public function testWPHead()
     {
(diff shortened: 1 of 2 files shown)


gchtr:
I first wondered why the tests suddenly fail. But then I saw that the tests fail with the latest version of WP CLI, which was released a couple of hours ago: https://github.com/wp-cli/wp-cli/releases/tag/v2.8.0.

I reran the tests for 2.x, and there they failed, too: https://github.com/timber/timber/actions/runs/5121366322/jobs/9244799025.

@nlemoine Do you have an idea why that might cause the tests to fail?

gchtr:
Let’s wait until tests start working again, and then this will be ready for review.

Levdbas (approved):

Levdbas:
I think we are good to go at merging this @gchtr

Issue #2766, opened 2023-06-08
[BUG] URLHelper::is_external() false-positive


mcaskill:
### Expected Behavior

Given a WordPress installation with a hostname `example.com` and external URL that contains the hostname as a substring (subdomain, path, query, fragment), for example: `https://facebook.com/example.com`.

Calling `URLHelper::is_external()` with the aforementioned external URL should return `true`.

### Actual behavior

Calling `URLHelper::is_external()` with an external returns `false` because `strstr()` is used to detect the current hostname.

### Steps to reproduce behavior

1. Call `URLHelper::is_external()` with an absolute URL that contains the current host as either:
	- subdomain: `example.com.github.io`
	- domain: `anotherexample.com`
	- path: `github.io/example.com`
	- query: `github.io?url=example.com`
	- fragment `github.io#example.com`

### Notes

* Timber v1: https://github.com/timber/timber/blob/1.22.1/lib/URLHelper.php#L327
* Timber v2: https://github.com/timber/timber/blob/2.0.0-beta.2/src/URLHelper.php#L442

This issue also affects the `URLHelper::is_local()` method.

### What version of Timber are you using?

v2 (issue present since v1)

### What version of WordPress are you using?

Irrelevant (v6.2)

### What version of PHP are you using?

irrelevant (PHP 8.1)

### How did you install Timber?

Installed or updated Timber through Composer

gchtr:
Fixed in #2767 and #2782.

PR #2767, opened 2023-06-08
2.x Fix URLHelper methods `is_local` and `is_external`

mcaskill:
Related:

- #2766

## Issue

The `is_local()` and `is_external()` methods use `strstr()` on the entire `$url` argument to indiscriminately search for the current host (see `get_host()` method). If ever the given `$url` includes the current host in a subset of the host or a non-host position, the aforementioned methods will return a false positive.

## Solution

Change the logic to match the `$url` argument's host against the current host, instead of searching the entire URL.

- Modified method `is_local()` to match the URL's hostname against the current host.
- Modified method `is_external()` to use `is_local()`.

## Impact

Fixes various false positives but could in itself create unexpected false negatives in existing projects using Timber?

## Usage Changes

No changes.

## Considerations

Open to considerations. 

## Testing

Tests have been updated but I am unable to run them locally due to a mysterious segmentation fault.


src/URLHelper.php
```diff
@@ -427,23 +428,20 @@ private static function is_internal_content($url)
     }
 
     /**
-     * Determines if URL is an external URL.
+     * Checks whether an URL is external.
      *
-     * True if `$url` is an external url or subdomain (http://cdn.example.org = true). False if
-     * relative or local true if it's a subdomain
+     * True if the URL is an external URL or subdomain (http://cdn.example.org = true). False if
+     * the URL is relative or local, true if it’s a subdomain.
```
Levdbas, line 434: We might want to enhance the docblock a bit here. We are listing subdomain twice here which is a bit confusing @mcaskill , are you able to clear that up as well?

Levdbas (commented):
Hi @mcaskill, thank you for your PR.

Other then the docblock that could use some clarification we are all good in my opinion!

mcaskill:
Hi @Levdbas, I've updated the docblocks for `is_local` and `is_external`.

I've also made fixed a merge conflict and improved handling of local URLs.

gchtr:
@mcaskill Thanks for the updated PR!

I added some more changes in https://github.com/timber/timber/pull/2782 that I want your opinion on.

In https://github.com/timber/timber/commit/8d06cf830efb727f288b11aac8a6d6bd5fa308cf, I added some more tests to make it more explicit what we check against.

With that, I found that passing a domain only without a scheme is always returning false. As a developer that uses that uses these functions, I would think that passing a domain only should work, too. So I added a check to catch that by adding a `//` scheme if the URL doesn’t contain one.

What do you think of that? Would it make sense to include that?

mcaskill:
> In 8d06cf8, I added some more tests to make it more explicit what we check against.
>
> With that, I found that passing a domain only without a scheme is always returning false. As a developer that uses that uses these functions, I would think that passing a domain only should work, too. So I added a check to catch that by adding a // scheme if the URL doesn’t contain one.

Good catch. I absolutely agree with these extra checks.

gchtr:
Perfect, thanks for the feedback 👍. This is now merged into 2.x.

PR #2774, opened 2023-07-11
Fix #2773 (send empty extension for image with no extension)

### LINKED ISSUE #2773: [BUG] Missing extension for image
### Expected Behavior

Hi, there,

I use the Eventbrite API (https://www.eventbrite.com/platform/api#/reference/event/retrieve/retrieve-an-event) and import events (title, content, cover image, ...) into WordPress.

Eventbrite gives us an image URL with no extension.
It's a URL that resizes the image on the fly
The image extension is indeed a jpeg, but Timber (in the ImageHelpers class) can't find it.

### Actual behavior

I've made a modification in the pull request that follows, so as to ignore the extension if it doesn't exist, and thus avoid displaying an error.

1. Retrieve event from API (eventbrite)
2. Import image with url
3. Display this image with timber

### Notes

_No response_

1.22

6.2.2

8.1

Spidlace:
Related:

- #2773 

## Issue
Display error if image extension does not exist

### DIFF
--- a/lib/ImageHelper.php
+++ b/lib/ImageHelper.php
@@ -437,7 +437,7 @@ class ImageHelper {
 		$parts = PathHelper::pathinfo($tmp);
 		$result['subdir'] = ($parts['dirname'] === '/') ? '' : $parts['dirname'];
 		$result['filename'] = $parts['filename'];
-		$result['extension'] = strtolower($parts['extension']);
+		$result['extension'] = (isset($parts['extension']) ? strtolower($parts['extension']) : '');
 		$result['basename'] = $parts['basename'];
 		return $result;
 	}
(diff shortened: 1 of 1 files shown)


Levdbas:
As discussed yesterday, this is a small fix that we can safely include in a point release. This PR is based on the 1.x branch, so we will have to port this to 2.x first and maybe merge this in 1.x before making a final release there.

Levdbas:
@timber/rangers , can you take a look at this PR if its ready for a merge?

gchtr:
@timber/rangers I ported this over to 2.x and added a test in https://github.com/timber/timber/pull/2934.

Levdbas:
Hi @Spidlace , thank you for your original PR. @gchtr just ported this to Timber 2 so this will ship in one of the next versions of Timber.

Issue #2772, opened 2023-06-26
[BUG] Deprecation Notice implicit conversion from float to int in ImageDimensions.php


joworeiter:
### Expected Behavior

No deprecation notice

### Actual behavior

Deprecated:  Implicit conversion from float to int loses precision in /timber/timber/src/ImageDimensions.php on line 151

### Steps to reproduce behavior

1. add an image with strange dimensons as float (eg. a svg)
2. load the page

### Notes

Here would be the fix, I just doesn't have timber dev setup to provide a pull request. 

```
 protected function get_dimension_loaded($dim = null): int
    {
        $dim = strtolower($dim);

        if ('h' === $dim || 'height' === $dim) {
            return (int) $this->dimensions[1];
        }

        return (int) $this->dimensions[0];
    }
```

### What version of Timber are you using?

Timber 2 Beta 2

### What version of WordPress are you using?

WordPress 6.2.2 (in Composer/Bedrock Setup)

### What version of PHP are you using?

PHP 8.1

### How did you install Timber?

Installed or updated Timber through Composer

gchtr:
Thanks for taking the time to open this issue 🙌.

This was fixed in #2775 by converting floats to rounded integers when getting the width and height from an SVG.

PR #2775, opened 2023-07-20
2.x Fix implicit conversion from float to int

gchtr:
Related:

- #2772

## Issue

We require dimensions to be int, but SVG might have float dimensions.

## Solution

Convert to int.

## Impact

No deprecation notice.

## Usage Changes

None.

## Considerations

Should we round the values before typecasting them to `int` using `round()`?

I think we also have to consider whether it might be important to get SVG dimensions as floats. For using the dimensions as `width` and `height` attributes in HTML, it has to be an integer value.

But maybe there’s a use case where you’ll want to access SVG dimensions as floats. In that case, we might somehow have to give access to `ImageDimensions::get_dimensions_svg()`, which is protected at the moment:

https://github.com/timber/timber/blob/a23d78178a3b043b888930186b7b384501d21855/src/ImageDimensions.php#L164

## Testing

No.

### DIFF
--- a/src/ImageDimensions.php
+++ b/src/ImageDimensions.php
@@ -121,13 +121,13 @@ class ImageDimensions
         if (file_exists($this->file_loc) && filesize($this->file_loc)) {
             if (ImageHelper::is_svg($this->file_loc)) {
                 $svg_size = $this->get_dimensions_svg($this->file_loc);
-                $this->dimensions = [$svg_size->width, $svg_size->height];
+                $this->dimensions = [(int) $svg_size->width, (int) $svg_size->height];
             } else {
                 list($width, $height) = getimagesize($this->file_loc);
 
                 $this->dimensions = [];
-                $this->dimensions[0] = $width;
-                $this->dimensions[1] = $height;
+                $this->dimensions[0] = (int) $width;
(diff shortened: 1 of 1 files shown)


Levdbas:
I think there is no need to disclose the original values via `get_dimensions_svg()` since having SVG widths and heights as floats are bad to begin with in my experience. I exerienced a lot of issues with for example hairlines when scaling a svg down with a large number after the decimal as width/height. And I think rounding the numbers is the way to go before typecasting them.

Levdbas:
It seems like [the tests](https://github.com/timber/timber/blob/129b9c6b9d1fb6eff9d7c2d8c1a33d9577da27b7/tests/test-external-image.php#L236) asserts that the SVG viewbox width of 530.91 is rounded down to 530, instead of up. In my opinion the current round function is what we would like.

We could change the testing value to 531 or the actual svg viewbox width 529.91, which gets then rounded up to 530?

But correct me if I am wrong @gchtr !

Levdbas (approved):

gchtr:
@Levdbas Thanks for the hint! I didn’t runt the tests locally, so I missed the rounded numbers issue.

PR #2776, opened 2023-07-21
2.x Fix MariaDB in automatic tests

gchtr:
## Issue

Tests don’t run after automatic update to MariaDB v11.

## Solution

As a first measure: downgrade to MariaDB v10.

Then, we can start to find out what went wrong. I read the Upgrade Guide for the 11 release (https://mariadb.com/kb/en/upgrading-from-mariadb-10-11-to-mariadb-11-0/), but I don’t know MariaDB or MySQL well enough to see what we would need to change.

@nlemoine Do you know what to do here?

## Impact

Fixes automatic tests.

## Usage Changes

None.

## Considerations

None.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,7 +23,7 @@ jobs:
 
     services:
       mysql:
-        image: mariadb:latest
+        image: mariadb:10
         env:
           MYSQL_USER: root
           MYSQL_ALLOW_EMPTY_PASSWORD: true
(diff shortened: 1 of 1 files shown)


Levdbas:
This is what we might need to fix this for MariaDB 11: See this [PR](https://github.com/dodona-edu/dodona/pull/4779/files).
I think we need to change the --health-cmd as well and maybe set those two constants as well? @gchtr

gchtr:
Cool, thanks for finding the solution. In the thread of that PR, I found a link to a blog post that contains even a simpler solution: https://mariadb.org/mariadb-server-docker-official-images-healthcheck-without-mysqladmin/.

Looks like the tests are not stuck on the MySQL health check anymore. So I’m merging this in.

Issue #2771, opened 2023-06-25
[BUG] toWebP() causes Uncaught TypeError: imageistruecolor(): Argument #1 ($image) must be of type GdImage, null given Error


joworeiter:
### Expected Behavior

Error should appear gracefully - not with fatal error

### Actual behavior

Fatal Error: 

[...]

PR #2780, opened 2023-07-25
add check to image create functions

Levdbas:
Related:

- #2771

## Issue
The imagecreatefrom functions should return false or an instance of GDimage. In the related issue null is returned, which is strange. But I think it is save to add an additional check to see whether we should postpone the image operation when the result of the function is null or false.

## Solution
Check if $input has a value before continuing.

## Impact
Better error handling.

## Usage Changes
None

## Considerations
There is a bit of overlap in these two classes. We might want to take a closer look at them in the future to make them more DRY.

## Testing
no.


gchtr:
Thanks for taking this on! 

I’m not sure about the best approach to handle this. Error handling in general is still an open issue for the future: https://github.com/timber/timber/issues/2210. Until we didn’t define how we should handle errors, I think we should silently return without adding a message to the error log.

src/Image/Operation/ToJpg.php
```diff
@@ -67,6 +67,10 @@ public function run($load_filename, $save_filename)
 
         $input = $imagecreate_function($load_filename);
 
+        if (!$input) {
```
nlemoine, line 70: With PHPStan being more strict as level increase, I changed my habits regarding those conditions and now check the expected value (this is also advised by the mighty @szepeviktor [here](https://github.com/szepeviktor/very-clean-code#-write-positive-conditions)).

```php
if (!$input instanceof GdImage) {
    return false;
}

```

nlemoine: > I am just a person.

I doubt that.

Levdbas: Hi @nlemoine,

Thanks for the suggestion. Just patched that.

Can you clear something up for me as well? From what I can see in the PHP docs, imagecreatefrompng() for example is already typed. Is the `instanceof` still needed in that case?

```php
imagecreatefrompng(string $filename): GdImage|false
```

szepeviktor: One can go three ways.
1. Type juggling to boolean `!$input` - the worst way 🔴 
2. Weak implicit check `$input === false` 🟠 
3. Strong explicit `!$input instanceof GdImage` 🟢

Levdbas: @nlemoine Seems PHP 7,4 is failing because the GdImage was only added since PHP 8.0. See changelog: https://www.php.net/manual/en/function.imagecreatefrompng.php. How to proceed?

szepeviktor: > How to proceed?

Oh noooooooooooo!
Then we need 🟠

nlemoine: Drop php 7.4? 😅

I'd say weak implicit check. This might be overkill to check for GdImage existence and adapt the condition.

szepeviktor (approved):

gchtr (approved):
Thank you all for thinking through the different solutions to this.

I wouldn’t drop support for PHP 7.4 for the 2.x release to provide an easier upgrade path.

But for the next major version, we can definitely drop it.

PR #2781, opened 2023-07-25
Prevent unneeded blog switching in multisite env.

### LINKED ISSUE #1180: \Timber\Site\__construct breaks image srcset
### Expected behavior

Should get srcset image urls like "/files/image_name.jpg".
### Actual behavior

Instead srcset image url is "wp-content/blogs.dir/blog_id/files/<image_name>.jpg". This prevents scrset from displaying responsive images.

Broken code 

**return switch_to_blog($blog_ids['old']);**

``` php
public function __construct( $site_name_or_id = null ) {
        if ( is_multisite() ) {
            $blog_ids = self::switch_to_blog($site_name_or_id);
            $this->init();
            $this->init_as_multisite($blog_ids['new']);
            return switch_to_blog($blog_ids['old']);
        } 
        $this->init();
        $this->init_as_singlesite();
}
```

Working code 

**return restore_current_blog();**

``` php
public function __construct( $site_name_or_id = null ) {
        if ( is_multisite() ) {
            $blog_ids = self::switch_to_blog($site_name_or_id);
            $this->init();
            $this->init_as_multisite($blog_ids['new']);
            return restore_current_blog();
        } 
        $this->init();
        $this->init_as_singlesite();
    }
```

WordPress 4.4.2, Timber 1.1.2

---

This function could use some documentation. Why is switching blogs necessary at all and why switch back to using `switch_to_blog()` again, instead of `restore_current_blog()`? With the current code, `$GLOBALS['_wp_switched_stack']` is not emptied, thus `ms_is_switched()` returns true, and `wp_upload_dir()` returns the wrong baseurl for srcset.

Levdbas:
To get some further experience with PRs I wanted to tackle an older issue.

Related:

- #1180

## Issue
switch_to_blog() was used when not needed in multisite environments, resulting in unneeded globals being set.

## Solution
Only use switch_to_blog() when the ID is different from the current id blog id. While at it, I changed some internal variables to reflect the new internal working and updated the docblock as well. I also added a fallback when get_blog_details() returns null instead of a site object.

## Impact
Less globals set in the form of $GLOBALS['_wp_switched_stack'] and $GLOBALS['switched']

## Testing
Current tests are covering it.

### DIFF
--- a/src/Site.php
+++ b/src/Site.php
@@ -188,17 +188,31 @@ class Site extends Core implements CoreInterface
     /**
      * Switches to the blog requested in the request
      *
-     * @param string|integer|null $site_name_or_id
+     * @param string|integer|null $blog_identifier The name or ID of the blog to switch to. If `null`, the current blog.
      * @return integer with the ID of the new blog
      */
-    protected static function switch_to_blog($site_name_or_id)
+    protected static function switch_to_blog($blog_identifier)
     {
-        if ($site_name_or_id === null) {
-            $site_name_or_id = \get_current_blog_id();
+        $current_id = \gget_current_blog_id();
+
+        if ($blog_identifier === null) {
+            $blog_identifier = $current_id;
         }
-        $info = \get_blog_details($site_name_or_id);
-        \switch_to_blog($info->blog_id);
-        return $info->blog_id;
+
+        $info = \gget_blog_details($blog_identifier);
+
+        if (false === $info) {
+            Helper::error_log('Timber\Site::switch_to_blog() could not find a blog with the identifier "' . $blog_identifier . '"');
+            return $current_id;
+        }
+
+        $blog_identifier = $info->blog_id;
+
+        if ($current_id !== $blog_identifier) {
+            \gswitch_to_blog($blog_identifier);
+        }
+
+        return $blog_identifier;
     }
 
     /**
--- a/src/Site.php
+++ b/src/Site.php
@@ -193,13 +193,13 @@ class Site extends Core implements CoreInterface
      */
     protected static function switch_to_blog($blog_identifier)
     {
-        $current_id = \gget_current_blog_id();
+        $current_id = \get_current_blog_id();
 
         if ($blog_identifier === null) {
             $blog_identifier = $current_id;
         }
 
-        $info = \gget_blog_details($blog_identifier);
+        $info = \get_blog_details($blog_identifier);
 
         if (false === $info) {
             Helper::error_log('Timber\Site::switch_to_blog() could not find a blog with the identifier "' . $blog_identifier . '"');
@@ -209,7 +209,7 @@ class Site extends Core implements CoreInterface
         $blog_identifier = $info->blog_id;
 
         if ($current_id !== $blog_identifier) {
-            \gswitch_to_blog($blog_identifier);
+            \switch_to_blog($blog_identifier);
         }
 
         return $blog_identifier;
--- a/src/Site.php
+++ b/src/Site.php
@@ -199,7 +199,7 @@ class Site extends Core implements CoreInterface
             $blog_identifier = $current_id;
         }
 
-        $info = \get_blog_details($blog_identifier);
+        $info = \get_blog_details($blog_identifier, false);
 
         if (false === $info) {
             return $current_id;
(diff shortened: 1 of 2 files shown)


src/Site.php
```diff
@@ -188,17 +188,30 @@ public function wp_object(): ?WP_Site
...
+        if (false === $info) {
+            return $current_id;
         }
-        $info = \get_blog_details($site_name_or_id);
-        \switch_to_blog($info->blog_id);
-        return $info->blog_id;
+
+        $blog_identifier = $info->blog_id;
+
+        if ($current_id !== $blog_identifier) {
```
gchtr, line 210: This check should be typecast, because `$blog_identifier` might be passed as a string instead of an integer.

```suggestion
        if ($current_id !== (int) $blog_identifier) {
```

src/Site.php
```diff
@@ -188,17 +188,30 @@ public function wp_object(): ?WP_Site
     /**
      * Switches to the blog requested in the request
      *
-     * @param string|integer|null $site_name_or_id
+     * @param string|integer|null $blog_identifier The name or ID of the blog to switch to. If `null`, the current blog.
      * @return integer with the ID of the new blog
      */
-    protected static function switch_to_blog($site_name_or_id)
+    protected static function switch_to_blog($blog_identifier)
```
gchtr, line 194:
```suggestion
    protected static function switch_to_blog($blog_identifier = null) : int
```

src/Site.php
```diff
@@ -188,17 +188,30 @@ public function wp_object(): ?WP_Site
...
-        \switch_to_blog($info->blog_id);
-        return $info->blog_id;
+
+        $blog_identifier = $info->blog_id;
+
+        if ($current_id !== $blog_identifier) {
+            \switch_to_blog($blog_identifier);
+        }
+
+        return $blog_identifier;
```
gchtr, line 214: Make sure we always return an integer.

```suggestion
        return (int) $blog_identifier;
```

gchtr (changes requested):
Good catch!

I only found some issues with strict types, otherwise this looks good to me.

src/Site.php
Levdbas: Since I am quite new to the whole typecasting thing, a small question.

Is it a possibility to typecast `$blog_identifier` at https://github.com/timber/timber/blob/43acfc764d6d96f60ab49fedf36ce784f788e832/src/Site.php#L208

instead of on line 210 and 214 as the return type. If so what would be preferred?

gchtr: Oh yeah, you could use:

```php
$blog_identifier = (int) $info->blog_id;
```

I think that would be better than using the typecasts later.

The basic problem is that when `$blog_identifier` is passed to the function, then it could be a string:

```php
$site->switch_to_blog( '3' );
```

And when we compare a string to a number, `===` will always return `false`. In the future, we could prevent that by requiring an integer to be passed, which could be `null` as well:

```
protected static function switch_to_blog( ?int $blog_identifier = null)
```

But I don’t want to add that breaking change just yet.

gchtr:
@Levdbas I’ve marked this as 2.x future, because it’s not essential to include this in the Release Candidate for Timber 2.0 ;).

Levdbas:
> @Levdbas I’ve marked this as 2.x future, because it’s not essential to include this in the Release Candidate for Timber 2.0 ;).

No problem! I think I managed to fix the PR though. Prior to this PR Timber was always switching to a blog, no matter what. After my initial changes Timber was **still** always changing because of the strict checking and without transforming the $blog_identifier in an INT first, still resulting in always switching in these tests and thus passing the tests.

After applying your suggestions, it finally passed the strict testing and resulting in not switching for the first time but resulting in testing errors.

What I did to make the tests work again:

Add `restore_current_blog();` after each switch blog function, just like it is recommended by WordPress. Adding these on rule [149](https://github.com/timber/timber/blob/3e45f8a78c0e6ef2e772ecabfcd3906423a4b0d7/tests/test-timber-multisite.php#L149C22-L149C22), 151, 153 was needed in the end to fix the test.

I think this is because `self::createSubDomainSite` switches to a particular blog as well. Since we ended up preventing unnecessary switching, the get_post() call was running on the wrong website at https://github.com/timber/timber/blob/3e45f8a78c0e6ef2e772ecabfcd3906423a4b0d7/tests/test-timber-multisite.php#L192 resulting in a wrong test result.

So by setting the `restore_current_blog();` we made sure each time we make sure that the starting blog is correct for making the query.

Let me know what your thoughts.

gchtr (approved):

PR #2786, opened 2023-08-01
fix: Allow overwrite of default avatar in comments.

### LINKED ISSUE #2468: Comment avatar functions ignore $args['default']
## Expected behavior

If you want to customize the default avatar in WordPress from a plugin or theme, normally you'd do so via one of the avatar filters like so:

```php
function example_custom_avatar_data($args, $id_or_email) {
  $args['default'] = 'https://placeimg.com/96/96/animals';
  return $args;
}

add_filter('pre_get_avatar_data', 'example_custom_avatar_data', 10, 2);
```

In which case any request for a Gravatar that does not exist will instead display our custom default image:

Screen Shot 2021-07-14 at 9 32 55 AM

This functionality is helpful in cases where we want to extend WordPress's default avatar functionality dynamically via a plugin.

## Actual behavior

In Timber, the following code in the context of a post comment…

```twig
<img src="{{comment.avatar}}" width="96" height="96" alt="">
```

…will never display the image path set via the filter using `$args['default']`. It will always fall back to `get_option('avatar_default')` unless `$default` is provided from the template.

Screen Shot 2021-07-14 at 9 37 12 AM

1. Activate the Timber starter theme.
2. Add the Twig snippet above to the theme's `templates/comment.twig`.
3. Add the PHP snippet above to the theme's `functions.php` or to a new plugin.
4. Navigate to `wp-admin/options-discussion.php` and scroll to the "Default avatar" section. All default avatars should display the image path provided in the PHP snippet.
5. Navigate to a post containing comments with a default image visible. Observe that the image displayed does not match what was defined in the filter.

- WordPress 5.7.2
- PHP 7.4.1
- Timber 1.18.2

- WP.org, synced locally via [Local](https://localwp.com/)

Levdbas:
Related:

- #2468

## Issue
Default avatar overrides could not be overwritten by the pre_get_avatar_data filter. 

## Solution
Set `default ` arg as well, while at it, add arguments more the WordPress way in `avatar_out`

## Impact
None, this only affects sites that override the default avatar.

## Usage Changes
none

## Considerations
The new default url is fed into Gravatar and then returned. Ideally you would like to check if an avatar returned by gravatar is a default one and if so, overwriting the resulting url with the default url, thus not resulting in a url like:

http://0.gravatar.com/avatar/a5f7937f4bc10cfe01abb42fe5504da9?s=92&d=https%3A%2F%2Fupload.wikimedia.org%2Fwikipedia%2Fcommons%2Fthumb%2F1%2F15%2FCat_August_2010-4.jpg%2F1920px-Cat_August_2010-4.jpg&r=X

where the d arguments holds the overwritten default. This would lead to less dependance on Gravatar as a whole but I did not found a good way to do that.

## Testing
no

### DIFF
--- a/src/Comment.php
+++ b/src/Comment.php
@@ -252,6 +252,10 @@ class Comment extends CoreEntity
             return $args['url'];
         }
 
+        if (isset($args['default'])) {
+            $default = $args['default'];
+        }
+
         $email_hash = '';
         if (!empty($email)) {
             $email_hash = \md5(\strtolower(\trim($email)));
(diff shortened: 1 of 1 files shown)


nlemoine (approved):
LGTM, thanks @Levdbas !

Issue #1313, opened 2017-01-31
Breaking change to TimberTheme's uri for child themes


adamtomat:
### Expected behavior
We have a child theme on-top of our own core theme, built on Timber. Normally we load assets from the child theme (e.g. theme specific styles), and sometimes need to load assets from the parent theme (e.g. core styles).

In `1.2.0`, in Twig I can do: `{{ site.theme.uri }}`, which returns the absolute path to the child theme `http://<domain>/app/themes/<child-theme-name>`.

I am also able to get the parent theme's uri with:

```twig
{% set parentUri = site.theme.uri %}

{% if site.theme.parent.uri %}
    {% set parentUri = site.theme.parent.uri %}
{% endif %}
```

In `1.2.1` I would expect the same behaviour, as there's no breaking changes.

### Actual behavior

However, in `1.2.1` when I do `{{ site.theme.uri }}`, it returns the absolute path to the *parent* theme, not the child theme. e.g: `http://<domain>/app/themes/<parent-theme-name>`.

It looks like [this commit is the one that changed behavior](https://github.com/timber/timber/pull/1281/files?diff=unified#diff-920c5ce40c6e0c76e2af998a5acef94f).

Primarily, this:

`$this->uri = get_stylesheet_directory_uri();`

was changed to:

`$this->uri = $this->theme->get_template_directory_uri();`

### Thoughts

Is this change the expected/intended behaviour? Where the uri always references the parent theme? 

Now, it looks like `uri` [isn't documented](http://timber.github.io/timber/#timber-theme) and therefore this change is fine as we should be using `{{ site.theme.link }}` anyway. If that's the case, is there an argument to make the uri protected?

### Steps to reproduce behavior
Create a child theme, and activate it. In the parent theme, in any twig view add `{{ site.theme.uri }}`. You should see different urls between `1.2.0` and `1.2.1`.

### What version of WordPress, PHP and Timber are you using?
WordPress 4.7.2, PHP 5.6, Timber 1.2.0

### How did you install Timber? (for example, from GitHub, Composer/Packagist, WP.org?)
Composer

xavivars:
I made the changes in the commit, so let me explain the reason why I did it. 

It turned out that the previous approach to get "parent" and "child" urls was not working when the theme was in a non standard location (by using `register_theme_directory`).

[...]

PR #2787, opened 2023-08-01
docs(cheatcheat): Explained theme path, link and uri helpers.

Levdbas:
Related:

- #1313

## Issue
In #2744 we updated the docblocks regarding these theme helpers, but the docs where still somewhat vague.

## Solution
Updated the cheatsheet in v2 as well with the same hints.

## Impact
Better understanding of the theme uri/path/link helpers straight from the docs.

## Usage Changes
none

## Considerations
none

## Testing
no


gchtr (approved):
Great, thanks for the update! 💪

Issue #2762, opened 2023-06-02
[BUG] symfony/cache-extra and timber 2, cache method unclear


Levdbas:
### Expected Behavior

Setting up the cache extention on v2 and cache mode to default (transient) will save cache to database.

### Actual behavior

I have no clue where the cache is coming from and how to flush it. :D

### Steps to reproduce behavior

1. Setup Timber cache via https://timber.github.io/docs/v2/guides/performance/#cache-parts-of-the-twig-file-and-data
2. Create cache in a template via 
```
{% cache "what_is_love" %}
Baby don't hurt me
{% endcache %}
```
3. see that cache is loaded on pageload
4. change contents between the cache tags and refresh page, see that old content is still loaded.
5. Check database for transients, none found.
6. Run wp timber clear-cache, no luck as well
...

### Notes

I tried debugging  #2483 to see if it still persists on v2, but got stuck here. So the results are obviously cached _somewhere_. But I have no clue where...

### What version of Timber are you using?

2.x-dev

### What version of WordPress are you using?

6.2

### What version of PHP are you using?

8.0.27

### How did you install Timber?

Installed or updated Timber through Composer

gchtr:
Cache files for Twig cache should be placed in `vendor/timber/timber/cache` by default. The can’t be saved in a transient, because Twig handles the whole caching for us. Probably it’s not clear from the documentation that the cache mode only applies to Timber’s caching function, but not to Twig cache.

You can update the cache location using the [`timber/twig/environment/options`](https://timber.github.io/docs/v2/hooks/filters/#timber/twig/environment/options) filter and set an absolute path for `cache`.

> `cache` *string* or false
>
> An absolute path where to store the compiled templates, or `false` to disable caching (which is the default).
>
> – via https://twig.symfony.com/doc/2.x/api.html#environment-options

But I’m not sure whether this works here. I’d be glad if you could quickly check that out.

In theory, the following functions should clear the Twig cache:

```php
Timber\Cache\Cleaner::clear_cache( 'twig' );
Timber\Cache\Cleaner::clear_cache_twig();
```

Or with the CLI:

```bash
wp timber clear-cache
wp timber clear-cache twig
```

The official way to clear Twig caches is to remove the files from the cache location. For this, you will have to know the location of the cache. `$twig->get_cache()` should return the location of the cache to remove it.

https://github.com/timber/timber/blob/a23d78178a3b043b888930186b7b384501d21855/src/Loader.php#L591

[...]

Issue #2441, opened 2021-04-19
[Docs] Improve Performance/Caching docs


wujekbogdan:
[Performance/Caching](https://timber.github.io/docs/guides/performance/) section of the docs isn't clear enough. It says that the default caching mode is `CACHE_TRANSIENT`, but there's no explanation on what does it actually mean. 

Does it mean that all queries are cached using transients? If so then it would be good to mention what is the default cache expiration and what are the implications of transient caching.

In addition to this, I would also include a paragraph explaining how to change the default caching mode using the `timber/cache/mode` filter.

Issue #2038, opened 2019-06-24
Minify html output 


raphaguasta:
Hello guys!

I wish to minify the html output of my website, to improve performance and loading time.

I didn't found anything in the docs to do this, can you help me?

Thanks! :)

jarednova:
@raphaguasta this is something that seems best for a caching plugin or another tool that's focused on performance optimizations. If this is something you're up for looking into, I'd love to add a section to the docs that explains some best practices to help others in the future!

palmiak:
I also agree this is more plugin territory, but if you prefer to do it with twig than there is https://github.com/nochso/html-compress-twig/. I didn't test it but I would love here is working correct.

bobhuf:
It won't compress everything (JS e.g.) but could use the spaceless twig filter or tag: https://twig.symfony.com/doc/2.x/tags/spaceless.html

danburzo:
I'm of the opinion you should not bother to minify your HTML — as long as the server GZIPs the request, you'll get _very little_ extra by minifying the HTML. Especially on a WordPress website, that would be my least concern. 

If you were to do it, I second the opinion this should happen at a level above Twig, possibly with a plugin dedicated to performance optimizations. Intuitively: at the Twig level, you're incurring more work for the library and you might end up with poorer performance; the problem with Twig extensions / filters / tags is you're looking to minify the _dynamic_ content as well, so even enabling the Twig cache won't (I think) preclude it from doing runtime minification of the content you place in your Twig templates.

gchtr:
> It won't compress everything (JS e.g.) but could use the spaceless twig filter or tag: [twig.symfony.com/doc/2.x/tags/spaceless.html](https://twig.symfony.com/doc/2.x/tags/spaceless.html)

That filter is specifically *not meant to minify your output*, as the documentation says itself:

> This tag is not meant to "optimize" the size of the generated HTML content but merely to avoid extra whitespace between HTML tags to avoid browser rendering quirks under some circumstances.
> 
> If you want to optimize the size of the generated HTML content, gzip compress the output instead.

@danburzo sums it up very well 👍.

I totally agree that we could add some more clarifications in the doc with a list of useful caching plugins and a hint about the spaceless tag in Twig and why you shouldn’t use it.

I assigned this to myself so I can eventually take care of it. If somebody else wants to take this over, feel free to comment here so that we know you’re working on this.

jl-a:
[...]

PR #2797, opened 2023-09-15
docs: Update v2 caching docs

Levdbas:
Related:

- #2762
- #2441
- #2038

## Issue
The caching docs needs some TLC. In this PR I will try to tackle these three tickets at once.

## Solution
Better documentation!

## Impact
Better understanding of caching in Timber.

## Usage Changes
None.

## Considerations
None.

## Testing
No.

### DIFF
--- a/docs/v2/guides/performance.md
+++ b/docs/v2/guides/performance.md
@@ -17,9 +17,9 @@ You can still use plugins like [W3 Total Cache](https://wordpress.org/plugins/w3
 
 With Timber you can cache the full Timber render/compile calls. When you do this, the whole template you render and its data will be cached. This results in faster page rendering by skipping queries and Twig compilations. But here’s the cool part: Timber hashes the fields in the view context. This means that **as soon as the data changes, the cache is automatically invalidated**. Yay!
 
-`Timber::render()` and `Timber::compile()` will use the default cache mode which by default is the transient cache mode.
+For Timber caching to take effect on your `Timber::render()` and `Timber::compile()` calls, you need to set the `$expires` argument. If the `$expires` argument is not set, Timber will not cache that particular template, even if the (global)cache mode is set.
 
-When rendering, use the `$expires` argument in [`Timber::render`](https://timber.github.io/docs/v2/reference/timber-timber/#render). For example:
+Example:
 
 ```php
 $context['posts'] = Timber::get_posts();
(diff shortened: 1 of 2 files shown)


docs/v2/guides/performance.md
```diff
@@ -15,6 +15,10 @@ You can still use plugins like [W3 Total Cache](https://wordpress.org/plugins/w3
 
 ## Cache the Entire Twig File and Data
 
+With Timber you can cache the full Timber render/compile calls. When you do this, the whole template you render and it's data will be cached. This results in faster page rendering by skipping queries and twig compilations. But here’s the cool part: Timber hashes the fields in the view context. This means that **as soon as the data changes, the cache is automatically invalidated**. Yay!
```
gchtr, line 18:
```suggestion
With Timber you can cache the full Timber render/compile calls. When you do this, the whole template you render and its data will be cached. This results in faster page rendering by skipping queries and Twig compilations. But here’s the cool part: Timber hashes the fields in the view context. This means that **as soon as the data changes, the cache is automatically invalidated**. Yay!
```

docs/v2/guides/performance.md
```diff
@@ -158,6 +105,69 @@ $loader = new Timber\Loader();
...
+If you want to use something meaningful for the cache key, you can also generate a cache key from the variables that you use. In the following example, we generate a cache key from `$posts` and then generate a key from it:
+
+```php
+$generator = new Timber\Cache\KeyGenerator();
+$key = $generator->generateKey($posts);
+```
+
+### Extra: TimberKeyGeneratorInterface
+
+Instead of hashing a whole object, you can specify the cache key in the object itself. If the object implements `TimberKeyGeneratorInterface`, it can pass a unique key through the method `get_cache_key`. That way a class can for example simply pass `'last_updated'` as the unique key.
```
gchtr, line 165:
```suggestion
Instead of hashing a whole object, you can specify the cache key in the object itself. If the object implements `TimberKeyGeneratorInterface`, it can pass a unique key through the method `get_cache_key()`. That way a class could for example pass a `'last_updated'` property as the unique key.
```

docs/v2/guides/performance.md
```diff
@@ -213,3 +223,6 @@ Timber::render('single.twig', $context, 600);
 // This reports the time diff by passing the $start time
 echo Timber\Helper::stop_timer($start);
 ```
+## Important notices
```
gchtr, line 226:
```suggestion
## Important notes
```

gchtr:
I see that you’re still working in this, but thanks a lot for taking this on!

Levdbas:
Finally took the time to finish the last part and tried to order the guide in a way that makes sense.

gchtr:
Thanks a lot @Levdbas. These updates surely help to understand what’s going on.

I want to take some more time to polish the performance docs. There are some sections in there that could just be deleted, like the TLDR; section and I saw some other details I want to improve.

I’m moving this to `2.x Future`.

Levdbas:
Should we just merge this for now @gchtr ? The change about the change at twig/cache-extra is then at least documented when people are going to migrate to 2.0. Let me know what you think.

gchtr:
> Should we just merge this for now @gchtr ? The change about the change at twig/cache-extra is then at least documented when people are going to migrate to 2.0. Let me know what you think.

Yeah, maybe that’s better. I created an issue in #2830 and I’m merging this.

gchtr (approved):

Issue #2741, opened 2023-05-17
Roadmap for Timber 2.0


gchtr:
This is an overview issue for the release of version 2.0 of Timber. If you find questions about the release of Timber 2.0, guide them here.

## Roadmap

The next steps for a 2.0 release are:

- [x] Release second beta [`2.0.0-beta.2`](https://github.com/timber/timber/releases/tag/2.0.0-beta.2) by the end of this week
- [x] Work through open pull requests and issues. Triage issues and pull requests that are not critical for the 2.0 release.
- [x] First release candidate [`2.0.0-rc.1`](https://github.com/timber/timber/releases/tag/2.0.0-rc.1) by June/July 2023
- [x] Release Timber 2.0 one to two months after developers could test out the release candidate.
	🚀 Check out the [Release notes](https://github.com/timber/timber/releases/tag/2.0.0).

The past has shown us that it’s difficult to predict an exact ETA. And with our limited resources, we don’t want to make promises we can’t keep. But let’s work towards a 2.0 release by being strict about what we really need for a 2.0 release and what we can postpone for later.

Help is welcome from everyone, also developers who are not maintainers. If you can provide insights, spot issues in the pull requests and comment on these, you’ll help us move the project forward.

## Pull requests

The following pull requests need to **be reviewed and resolved.**

- [x] https://github.com/timber/timber/pull/2669
- [x] https://github.com/timber/timber/pull/2730
- [x] https://github.com/timber/timber/pull/2660
- [x] https://github.com/timber/timber/pull/2667
- [x] https://github.com/timber/timber/pull/2668
- [x] https://github.com/timber/timber/pull/2749
- [x] https://github.com/timber/timber/pull/2750
- [x] https://github.com/timber/timber/pull/2751
- [x] https://github.com/timber/timber/pull/2700
- [x] #2753 

For each of these pull requests we should decide whether we can solve them in the future or whether we really need to solve them for a first 2.0 release. By "future", we mean that we can decide whether:

- The issue can be resolved in a minor version of Timber 2.x, so 2.1 or 2.2, etc.
- The issue will have to wait for Timber 3.x.

## Postponed issues and pull requests

- https://github.com/timber/timber/pull/2733 (postponed to `2.x Future`)
- https://github.com/timber/timber/issues/2453 (postponed to `2.x Future`)
- https://github.com/timber/timber/pull/2725 (postponed to `2.x Future`)
- https://github.com/timber/timber/issues/2762 (postponed to `2.x Future`)

[...]

PR #2798, opened 2023-09-15
V2 site overwrite magic __call method

Levdbas:
Related:

- [Overwrite __call method](https://github.com/timber/timber/issues/2741#issuecomment-1713435871_)
- https://github.com/timber/timber/pull/1700

## Issue
When calling a site option that does not have it's own method, like `site.home`. Timber would use the magic `__call` method from the Core class. This magic method by defaults calls the `meta()` method. Since the `meta()` method was deprecated in the site class (https://github.com/timber/timber/pull/1700) the magic `__call `method throws a deprecation notice.

## Solution
Overwrite the __call method in Site.php and call the appropriate `option()` method instead.

## Impact
Less questions on why deprecation notices are thrown.

## Usage Changes
None.

## Considerations
Sometimes it's hard to wrap your head around what are actual instantiated data in Timber and what is called via these magic methods. I always thought that `{{ site.home }}` was a valid method, but when digging deeper I found out that you should call` {{ site.link }}`

I would be in favor of deprecating these magic methods in the future as well so people (including myself) will learn the difference.

## Testing
No, should it be tested?

### DIFF
--- a/src/Site.php
+++ b/src/Site.php
@@ -174,6 +174,39 @@ class Site extends Core implements CoreInterface
     }
 
     /**
+     * Magic method dispatcher for site option fields, for convenience in Twig views.
+     *
+     * Called when explicitly invoking non-existent methods on the Site object. This method is not
+     * meant to be called directly.
+     *
+     * @example
+     * The following example will dynamically dispatch the magic __call() method with an argument
+     * of "users_can_register" #}
+     *
+     * ```twig
+     * {% if site.users_can_register %}
+     *   {# Show a notification and link to the register form #}
+     * {% endif %}
+     * @link https://secure.php.net/manual/en/language.oop5.overloading.php#object.call
+     * @link https://github.com/twigphp/Twig/issues/2
+     * @api
+     *
+     * @param string $option     The name of the method being called.
+     * @param array  $arguments Enumerated array containing the parameters passed to the function.
+     *                          Not used.
+     *
+     * @return mixed The value of the option field named `$field` if truthy, `false` otherwise.
+     */
+    public function __call($option, $arguments)
+    {
+        if (\method_exists($this, 'option') && $meta_value = $this->option($option)) {
+            return $meta_value;
+        }
+
+        return false;
+    }
+
+    /**
      * Gets the underlying WordPress Core object.
      *
      * @since 2.0.0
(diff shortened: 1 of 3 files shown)


Levdbas:
@gchtr , Directly created a PR for this. Should we add this to the 2.0 release?

jrathert:
> Sometimes it's hard to wrap your head around what are actual instantiated data in Timber and what is called via these magic methods. [...] I would be in favor of deprecating these magic methods in the future as well so people (including myself) will learn the difference.

I fully support that.

src/Site.php
```diff
@@ -173,6 +173,39 @@ public function __construct($site_name_or_id = null)
...
+     *
+     * @param string $option     The name of the method being called.
+     * @param array  $arguments Enumerated array containing the parameters passed to the function.
+     *                          Not used.
+     *
+     * @return mixed The value of the option field named `$field` if truthy, `false` otherwise.
+     */
+    public function __call($option, $arguments)
+    {
+        if (\method_exists($this, 'option') && $meta_value = $this->option($option)) {
```
nlemoine, line 202: We should probably return the value without evaluating it?
```php
return $this->option($option);
```

Levdbas: Ageed, fixed in https://github.com/timber/timber/pull/2798/commits/615917236bef098079189ab9fcf921b8e1eac0f9

gchtr:
> @gchtr , Directly created a PR for this. Should we add this to the 2.0 release?

Thanks for the PR. Yes, because this actually fixes a bug I’d say we should include it in the 2.0 release.



> ## Testing
> No, should it be tested?

@Levdbas I think it would be good if we could add a test for this. Do you need help with that?

Levdbas:
> > @gchtr , Directly created a PR for this. Should we add this to the 2.0 release?
> 
> Thanks for the PR. Yes, because this actually fixes a bug I’d say we should include it in the 2.0 release.
> 
> > ## Testing
> > No, should it be tested?
> 
> @Levdbas I think it would be good if we could add a test for this. Do you need help with that?

Alright I created my first test and used the timber compile_string method for that. removing the __call overwrite in the site class causes the test to fail, so I think we are good here. Any feedback on how to improve this test @gchtr ?

nlemoine (approved):
Thank you @Levdbas! This looks good to me and congrats on writing test 🙂

gchtr (approved):
This all looks fine, including the test 👍. Thanks, @Levdbas

PR #2800, opened 2023-09-16
Add documentation and plugin notice about the end of the plugin version

Levdbas:
Related:

- #2794

## Work in progress
This is a first version of this PR where I would like to have input from the core team: @gchtr , @nlemoine and @jarednova . Could you have a look at the wording of the notices in the .txt, .md file? Is that what we want to communicate? Please change/complete my suggested texts and notifications where necessary so that we can release this.

**Example of the plugin notice as currently commited**
afbeelding

### TODO's

- [x] Fill the links in the code changes
- [x] test the upgrade guide with a minimal theme and latest WordPress version
- [x] [Release a plugin version with the new code changes.](https://github.com/timber/timber/issues/2820)

## Issue
We are not going to release a plugin version of Timber anymore. We think that a lot of people will be stuck on the plugin version of Timber 1. With a install base of 50.000+ installations we are talking about a lot of people.

## Solution
Let's inform Timber users about the upcoming end of the Timber plugin and help them with extra guides on how to switch to the Composer based (1.x) version.

## Impact
Minimal on the codebase. This PR adds an inline plugin notification as the only code change.

## Usage Changes
See the whole guide :)

## Considerations

- I made a docs/getting-started/switch-to-composer.md guide. Is this the proper location?
- See the plugin notification I currently created. Is this the way we want to notify the plugin users about this change
- Is the switch-to-composer guide good enough?

## Testing
No

PR #2805, opened 2023-09-22
docs: updated plugin support part

Levdbas:
Related:

- https://github.com/timber/timber/pull/2800

## Issue
In the v2 guide there is no link/information about how to switch from the plugin version to the Composer-based installation.

## Solution
After we merged #2800 and deployed the new docs we can merge this PR to refer in the 2.0 guide to the _Switch from the plugin to Composer_ guide.

## Impact
Better information when you end up directly in the 2.0 upgrade guide.

## Usage Changes
no.

## Considerations
Cross-reference this 2.0 guide as the next step in the guide created in #2800? 

## Testing
no.


gchtr (approved):
Very good!

Issue #2802, opened 2023-09-21
[BUG] Timber\Post->terms() throws error if 'fields' parameter has value other than 'all'


herrschuessler:
### Expected Behavior

I'm running the example code found  at https://timber.github.io/docs/v2/getting-started/a-post-archive/#an-example-related-posts
```PHP
$context = Timber::context();

$post = $context['post'];

$context['related_posts'] = Timber::get_posts([
    'post_type' => 'post',
    'posts_per_page' => 3,
    'no_found_rows' => true,
    'orderby' => 'date',
    'order' => 'DESC',
    'post__not_in' => [$post->ID],
    'category__in' => $post->terms([
        'taxonomy' => 'category',
        'fields' => 'ids',
    ]),
]);
```
It should get a post collection posts from the same category.

### Actual behavior

```
Fatal error: Uncaught Error: Timber\Factory\TermFactory::build(): Argument #1 ($term) must be of type WP_Term, int given
in /timber/src/Factory/TermFactory.php on line 136
```

The issues seems to come from the 'fields' parameter in the category query. Setting the param to any other value than 'all' will result in the same error.

### Steps to reproduce behavior

Use Timber 2 rc.1 and run the above code

### What version of Timber are you using?

Timber 2 rc.1

### What version of WordPress are you using?

WordPress 6.3.1

### What version of PHP are you using?

PHP 8.1

### How did you install Timber?

Installed or updated Timber through Composer

jrathert:
Hm, as far as I understand the code, the problem is actually caused by this function in `TermFactory.php`:

```
protected function from_wp_term_query(WP_Term_Query $query): iterable
{
    return \array_map([$this, 'build'], $query->get_terms());
}

```
`$query->get_terms()` (actually: `WP_Term_Query::get_terms()` might - depending on the `fields` argument inside it - return very different things (`WP_Term[]|int[]|string[]|string`), but the code expects it returns a `ẀP_term[]` in all cases (which it only does, if `fields` is not set or set to `all` or  `all_with_object_id`). So I guess this requires quite some work here (and I do not see how the whole code/approach could work if someone sets `fields` to, e.g., `count`)?

See: https://developer.wordpress.org/reference/classes/wp_term_query/get_terms/

gchtr:
Resolved in #2806.

PR #2806, opened 2023-09-24
fix: consider 'fields' value when returning terms from query

jrathert:
Related:

- Ticket #2802 

## Issue
A `fields` argument other than `all`or `all_with_object_id` lead to a fatal error, as `TermFactory::from_wp_term_query()` did expect the result of a query to always be of type `WP_term[]` 

## Solution
In `TermFactory::from_wp_term_query()` call `TermFactory::build()` only if `fields` is `all` or `all_with_object_id`, otherwise, return plain result of `WP_Term_Query::get_terms()` (i.e., `int[]|string[]|string`)

## Considerations
One could - instead of examining the `fields` argument - check if the object returned from `WP_Term_Query::get_terms()` actually is a `WP_Term[]`. That would be future proof (if there are `fields` values added for which the query also returns an array of `WP_Term`) - but might have some impact on performance, so I decided to solve it this way.

## Testing
No specific

### DIFF
--- a/src/Factory/TermFactory.php
+++ b/src/Factory/TermFactory.php
@@ -60,9 +60,16 @@ class TermFactory
         return $this->build($wp_term);
     }
 
-    protected function from_wp_term_query(WP_Term_Query $query): iterable
+    protected function from_wp_term_query(WP_Term_Query $query)
     {
-        return \array_map([$this, 'build'], $query->get_terms());
+        $terms = $query->get_terms();
+
+        $fields = $query->query_vars['fields'];
+        if ('all' === $fields || 'all_with_object_id' === $fields) {
+            return \array_map([$this, 'build'], $terms);
+        }
+
+        return $terms;
     }
 
     protected function from_term_object(object $obj): CoreInterface
--- a/tests/test-term-factory.php
+++ b/tests/test-term-factory.php
@@ -310,6 +310,135 @@ class TestTermFactory extends Timber_UnitTestCase
         $this->assertTrue(MyTerm::class === get_class($res[1]));
     }
 
+    public function testFromTermQueryWithFields()
+    {
+        $term_ids = $this->factory->term->create_many(4, [
+            'taxonomy' => 'post_tag',
+        ]);
+
+        $post_id = $this->factory->post->create();
+        wp_set_object_terms(
+            $post_id,
+            $term_ids[0],
+            'post_tag'
+        );
+        wp_set_object_terms(
+            $this->factory->post->create(),
+            [$term_ids[1], $term_ids[2]],
+            'post_tag'
+        );
+
+        $termFactory = new TermFactory();
+
+        // all: array of used terms as Timber\Term object
+        $termQuery = new WP_Term_Query([
+            'taxonomy' => 'post_tag',
+            'fields' => 'all',
+        ]);
+        $terms = $termFactory->from($termQuery);
+        $this->assertCount(3, $terms);
+        foreach ($terms as $term) {
+            $this->assertInstanceOf(Term::class, $term);
+        }
+
+        // all_with_object_id: all terms used in a specific object as Timber\Term object
+        $termQuery = new WP_Term_Query([
+            'taxonomy' => 'post_tag',
+            'fields' => 'all_with_object_id',
+            'object_ids' => $post_id,
+        ]);
+        $terms = $termFactory->from($termQuery);
+        $this->assertCount(1, $terms);
+        $this->assertInstanceOf(Term::class, $terms[0]);
+        $this->assertSame($terms[0]->id, $term_ids[0]);
+
+        // register class map and repeat previous tests
+        $my_class_map = function (array $map) {
+            return array_merge($map, [
+                'post_tag' => MyTerm::class,
+            ]);
+        };
+        $this->add_filter_temporarily('timber/term/classmap', $my_class_map);
+
+        // all_with_object_id: all terms used in a specific object as MyTerm object
+        $termQuery = new WP_Term_Query([
+            'taxonomy' => 'post_tag',
+            'fields' => 'all',
+        ]);
+        $terms = $termFactory->from($termQuery);
+        $this->assertCount(3, $terms);
+        foreach ($terms as $term) {
+            $this->assertInstanceOf(MyTerm::class, $term);
+        }
+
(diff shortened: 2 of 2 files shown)


src/Factory/TermFactory.php
```diff
@@ -60,9 +60,16 @@ protected function from_id(int $id): ?Term
         return $this->build($wp_term);
     }
 
-    protected function from_wp_term_query(WP_Term_Query $query): iterable
+    protected function from_wp_term_query(WP_Term_Query $query)
```
gchtr, line 63: @nlemoine Would the function not always return an `iterable` result, even if it’s an array of ids?

src/Factory/TermFactory.php
```diff
@@ -60,9 +60,16 @@ protected function from_id(int $id): ?Term
...
-    protected function from_wp_term_query(WP_Term_Query $query): iterable
+    protected function from_wp_term_query(WP_Term_Query $query) 
     {
-        return \array_map([$this, 'build'], $query->get_terms());
+        $terms = $query->get_terms();
+        $_fields = $query->query_vars['fields'];
+        if ('all' === $_fields || 'all_with_object_id' === $_fields) {
+            return \array_map([$this, 'build'], $terms);
+        }
+        else {
```
gchtr, line 70: We don’t need an else statement here but could directly return `$terms`.

gchtr (commented):
This feels like sensible approach to me.

I think we would also need tests for this. @jrathert Do you have testing set up and can you add tests? Or would you rather have us do it?

@nlemoine I’d be glad about your opinion on this.

jrathert:
> @jrathert Do you have testing set up and can you add tests? 

I did not have it when I wrote the code, but now I have ;-). Can add some in the the next days.

src/Factory/TermFactory.php
jrathert: It might as well return a string (if `'fields' == 'count'`), which is not an iterable. So we need to remove it.

src/Factory/TermFactory.php
jrathert: Done

jrathert:
Did a few optimizaitons for the assertions (assertSame, not assertEquals etc)

src/Factory/TermFactory.php
nlemoine: This is a tricky situation.

Should `Timber::get_terms()` exactly match `get_terms` behavior?

Since we can't use union types yet (`int|iterable`), I would say: `Timber::get_terms()` should be used for getting objects or ids. If you need count, use `get_terms`. There's actually no benefit in using `Timber::get_terms()` for getting anything else than objects (event IDs) and keeping return types seems more important to me.

src/Factory/TermFactory.php
```diff
@@ -60,9 +60,16 @@ protected function from_id(int $id): ?Term
         return $this->build($wp_term);
     }
 
-    protected function from_wp_term_query(WP_Term_Query $query): iterable
+    protected function from_wp_term_query(WP_Term_Query $query)
     {
-        return \array_map([$this, 'build'], $query->get_terms());
+        $terms = $query->get_terms();
+
+        $_fields = $query->query_vars['fields'];
```
nlemoine, line 67: For consistency, I would remove the `_` from  variable name.

nlemoine (changes requested):

nlemoine:
Thank you @jrathert! We juste need to take a decision on https://github.com/timber/timber/pull/2806#discussion_r1366766656

And it will be good to go.

src/Factory/TermFactory.php
jrathert: @nlemoine - I see your point. My fix does more than is needed to fix the bug, and probably than is reasonable. 

On the other hand - if I pass a `WP_Term_Query` object to `TermFactory::from()` (even indirect via `Timber::get_terms()`) - what would you expect to happen? If the query is about the number of terms (i.e., `fields == count`) we should return exactly that. 
Or, to phrase it differently, what should we do if it is set like that and we actually need to return an `iterable`: throw an error? return an empty array?

Also note the method actually calling this protected, typed method is `TimberFactory::from()` - which does not have a return type and also simply ignores the return types of all methods it is calling (just passing on the results).

nlemoine: You're right @jrathert, the problem lies elsewhere. Maybe in the fact that we allow a `WP_Term_Query` to be passed or maybe we should throw something here, after all, it's a **term** factory, not a count term factory :) You should expect an iterable here.

Let's drop the return type then.

jrathert: The problem lies within the weird behavior/return types of `WP_Term_Query::get_terms()` I'd say. I am a bit new to PHP and WordPress, but I find it astonishing how pragmatic the language can be used - and actually is within WordPress.

nlemoine: PHP ≠ WordPress :) WordPress kind of lives outside regular PHP world.

jrathert:
Removed the unnecessary '`_` - should be ready to merge now.

gchtr:
I’m continuing the discussion we started in https://github.com/timber/timber/pull/2806#discussion_r1366766656 here.

 Very good that the issue with count popped up here.

> I would say: `Timber::get_terms()` should be used for getting objects or ids. If you need count, use `get_terms`. There's actually no benefit in using `Timber::get_terms()` for getting anything else than objects (event IDs) and keeping return types seems more important to me.

I agree with that.

> Maybe in the fact that we allow a `WP_Term_Query` to be passed or maybe we should throw something when `count` is met, after all, it's a **term** factory, not a count term factory :) You should expect an iterable here.

We could use [`Timber\Helper::doing_it_wrong()`](https://github.com/timber/timber/blob/38a99405ab322eb5d79a37d77eb149a65778ea8f/src/Helper.php#L318) here if `count` is found and advise to use `get_terms()` instead.

This could mean that we could keep `iterable`, right? Even though `TermFactory::from()` doesn’t have return types.

General error handling is still something we need to figure it in the future (https://github.com/timber/timber/issues/2210).

gchtr:
I just talked to @nlemoine. We going to drop `iterable` and will probably come back to this once we can use union types, which is when we drop support for PHP 7.4 in a future version.

gchtr (approved):

gchtr:
Some tests only fail because of an issue with testing against WordPress 6.4. For now, we can ignore these tests. I’m merging this in.

PR #2808, opened 2023-09-25
Prevent division by zero errors

mgussekloo:
Prevent division by zero errors for rare cases when images are not correctly uploaded to Wordpress.

## Issue
I had an error with images that reported a height of 0, and that led to a division by zero error white screen of death.

## Solution
Use max() to prevent division by zero.

## Impact
This shouldn't have much impact on anything.

### DIFF
--- a/lib/Image/Operation/Resize.php
+++ b/lib/Image/Operation/Resize.php
@@ -99,7 +99,7 @@ class Resize extends ImageOperation {
 		$current_size = $image->get_size();
 		$src_w = $current_size['width'];
 		$src_h = $current_size['height'];
-		$src_ratio = $src_w / $src_h;
+		$src_ratio = $src_w / max(1, $src_h);
 		if ( !$h ) {
 			$h = round($w / $src_ratio);
 		}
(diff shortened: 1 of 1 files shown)


Levdbas:
As discussed yesterday. We will port this to 2.x and maybe include this in a final release for 1.x

Levdbas (approved):

gchtr:
Hm... unless I’m missing something, if we use `max(1, $height)`, wouldn’t this cause the function to return a height that is the same as the width? So we would just assume that image is square.

Also, the following check would never run, because `$h` would never be falsey (https://github.com/timber/timber/pull/2808/files#diff-acfb97f3800030360cbd45aeebe5145dd35198d5763bb51dbc2e9993799f621eR103-R105):

```php
if ( !$h ) {
	$h = round($w / $src_ratio);
}
```

If no height can be read from the image, I think we shouldn’t crop it at all instead. We should probably add a bailout instead and a check after `get_target_sizes()` is used to prevent the cropping.

This looks like a more extensive fix and I’d rather not fix this in 1.x, but concentrate on 2.x.

Levdbas:
Hi @mgussekloo , thank you for your PR but we decided to not ship it in the last 1.x version of Timber for now. It probably needs more work than your suggested changes and we want to focus fully on 2.x now. Feel free to contribute to any outstanding issue for 2.x

PR #2823, opened 2023-10-20
Workflow: fix path to guide

Levdbas:
URL to the switch to composer guide is pointing to the wrong place in the bugreport template. Let's fix that!

We must merge this in master and v2.


nlemoine (approved):
Thanks @Levdbas !

gchtr:
@Levdbas I just merged this both in to `master` and `2.x`.

PR #2826, opened 2023-10-20
ci: Update GitHub Actions workflows

nlemoine:
## Issue

- Split each job into its own file
- Filter paths so Actions runs only when necessary (no need to run composer lint when it has not been modified)
- Add cs2pr to provide PR author direct feedback
- Add dependabot for GA
- Bump easy-coding-standards to ^12

## Solution

We should have much faster/accurate CI tests and cleaner UI view.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,7 +14,7 @@ trim_trailing_whitespace = true
 indent_style = spaces
 indent_size = 4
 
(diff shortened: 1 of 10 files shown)


gchtr (approved):
Wonderful, thanks @nlemoine 🎉

Issue #2813, opened 2023-09-30
2.x - 🥺 `final __constructor` really takes the jelly out of my doughnut


mindfullsilence:
Please don't do this to the end users (me). I'm in the midst of using v2rc1 and came across this restriction - which seriously limits my ability to provide services to the class unless I write a bunch of setter methods or create an init function that I have to call every time the class is instantiated (both options aren't great). This removes any possibility of DI, autowiring, and I'm sure a lot of other things I can't think of right now. 

In addition, the logic for this decision seems flawed: the `build()` method _also_ shares the same concern you're discussing [in pull 2660](https://github.com/timber/timber/pull/2660) about the constructor - it's just not throwing a warning because PHPStan only looks at the constructor for this potential issue. If I override the `build()` method with a different signature, it will break just as many things as overriding the `__constructor()` would. Marking both the `build` and `__constructor` as final would basically make initializing the class with additional setup code incredibly messy, if not impossible - but it doesn't make sense to mark one without marking the other for this PHPStan warning. 

How do you deal with this PHPStan warning, then? Simply don't define a constructor in the `Timber\Post` class. The library is already structured exactly how it should be to allow for this. By not defining it, you allow it to be defined by child classes, and the PHPStan warning goes away. 

To protect against the issue PHPStan is warning about, but for the `build` method, define an Interface based on what PostFactory needs. Also protect yourself from improperly written child classes by moving some of the import logic around and marking the `import()` method as `final`.

```php
// Timber\PostInterface.php
interface PostInterface {
  
    public static function build(WP_Post $post) : PostInterface;

    public function import(WP_Post $post) : void;

}


// in PostFactory.php
class PostFactory {
    //...

    // declare the interface as the return type, rather than the concrete instance
    protected function build(WP_Post $post) : PostInterface
    {
        $class = $this->get_post_class($post);
       
        if ( ! is_a($class, Timber\PostInterface::class) ) {
             throw new Exception("$class must implement Timber\PostInterface");
        }

        $instance = $class::build($post);
        $instance->import($post);

        return $instance;
    }

    //...
}

 // in Post.php

class Post extends CoreEntity implements DatedInterface, Setupable, PostInterface {
    //...

    // new build() method does not call import, since it's now called from PostFactory
    // build() is only responsible for returning an instance
    public static function build(WP_Post $wp_post) : PostInterface {
        $post = new static();

        // No longer call import from here
        // $post->import($data);

        return $post;
    }

    // mark import() as final
[...]

PR #2827, opened 2023-10-24
2.x Revert final constructors

gchtr:
Related:

- #2813
- #2660

## Issue

The final constructors prevent advanced use cases as described in #2813. We don’t want to make it so hard to work around it. But we first need to figure out a proper solution for this.

## Solution

Revert that pull request for 2.x and figure out a better solution in a future release.

## Impact

Less trouble to incorporate advanced coding patterns with Timber.

## Usage Changes

None.

## Considerations

None.

## Testing

No.

### DIFF
--- a/docs/v2/upgrade-guides/2.0.md
+++ b/docs/v2/upgrade-guides/2.0.md
@@ -19,7 +19,7 @@ We upgraded Timber to work with more modern PHP and functionalities that only we
 
 ## No more plugin support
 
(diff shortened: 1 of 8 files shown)


nlemoine (approved):

jarednova (approved):
LGTM! I had filed a ticket years ago w/ WordPress core about the `final`s on Post objects. So similar thinking! (though "takes the jelly out of my doughnut is a much more vibrant way to express it")

PR #2838, opened 2023-11-09
fix: Fix failing tests for WP 6.4

gchtr:
## Issue

With the release of WordPress 6.4, we’re running into lots of errors.

## Solution

Turns out that WordPress [still always calls `the_block_template_skip_link()`](https://github.com/WordPress/WordPress/blob/6.4/wp-includes/default-filters.php#L720), which always causes [a `_deprecated_function()` notice](https://github.com/WordPress/WordPress/blob/6.4/wp-includes/deprecated.php#L6138). My solution is to add `@expectedDeprecated` calls.

I also excluded the `docs/` folder from the PHPStan checks.

@nlemoine Can we also ignore the docs for the **PHP Lint** workflow? We’re getting errors there, too: https://github.com/timber/timber/actions/runs/6808609995/job/18513339121?pr=1617.

## Impact

Less noise in tests.

## Usage Changes

None.

## Considerations

Maybe.

## Testing

Yes.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -8,6 +8,7 @@ parameters:
         - vendor/php-stubs/acf-pro-stubs/acf-pro-stubs.php
     excludePaths:
         - tests/*
+        - docs/*
     ignoreErrors:
         - '#Instantiated class Twig\\CacheExtension\\CacheStrategy\\GenerationalCacheStrategy not found.#'
         - '#Instantiated class Twig\\CacheExtension\\Extension not found.#'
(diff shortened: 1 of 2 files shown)


gchtr:
@nlemoine Or wait, now it seems to run fine: https://github.com/timber/timber/actions/runs/6808992662/job/18514424700?pr=2838.

nlemoine (approved):
Yes, it's fine like this. Thanks for tackling this last hour issue :)

gchtr:
And now, of course tests fail for lower WordPress versions, because the deprecation is not there. But I think that’s fine for the moment.

PR #2840, opened 2023-11-09
[Docs] Use recommended method to get meta in ACF Integration sample code

xavivars:
## Issue
In ACF Integration docs, sample code directly accesses to meta fields as properties of the `post` object itself.

## Solution
Use the recommended approach (`meta` method) to access such fields.

## Impact
Better code samples

### DIFF
--- a/docs/v2/integrations/advanced-custom-fields.md
+++ b/docs/v2/integrations/advanced-custom-fields.md
@@ -77,8 +77,8 @@ This is where we’ll start in PHP.
 ```php
 $post = Timber::get_post();
 
-if (isset($post->hero_image) && strlen($post->hero_image)) {
-    $post->hero_image = Timber::get_image($post->hero_image);
+if (isset($post->meta('hero_image')) && strlen($post->meta('hero_image'))) {
+    $post->meta('hero_image') = Timber::get_image($post->meta('hero_image'));
 }
 
 $context = Timber::context([
@@ -93,7 +93,7 @@ Timber::render('single.twig', $context);
 You can now use all the above functions to transform your custom images in the same way, the format will be:
 
 ```twig
-<img src="{{ post.hero_image.src | resize(500, 300) }}" />
+<img src="{{ post.meta('hero_image').src | resize(500, 300) }}" />
 ```
 
 ## Gallery Field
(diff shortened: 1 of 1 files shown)


docs/v2/integrations/advanced-custom-fields.md
```diff
@@ -77,8 +77,8 @@ This is where we’ll start in PHP.
 ```php
 $post = Timber::get_post();
 
-if (isset($post->hero_image) && strlen($post->hero_image)) {
-    $post->hero_image = Timber::get_image($post->hero_image);
+if (isset($post->meta('hero_image')) && strlen($post->meta('hero_image'))) {
+    $post->meta('hero_image') = Timber::get_image($post->meta('hero_image'));
```
Levdbas:
```suggestion
    $post->hero_image = Timber::get_image($post->meta('hero_image'));
```
This will have to stay the same since we want to make $this->hero_image available directly in  the timber context.

Levdbas (commented):
Thank you for this @xavivars . I have suggested one change. We will wait until one of the other reviewers can take a look at your suggested changes as well.

docs/v2/integrations/advanced-custom-fields.md
xavivars: True, I changed it too naively

docs/v2/integrations/advanced-custom-fields.md
```diff
@@ -77,8 +77,8 @@ This is where we’ll start in PHP.
 ```php
 $post = Timber::get_post();
 
-if (isset($post->hero_image) && strlen($post->hero_image)) {
-    $post->hero_image = Timber::get_image($post->hero_image);
+if (isset($post->meta('hero_image')) && strlen($post->meta('hero_image'))) {
```
gchtr: I wonder whether we even need to set `$post->hero_image`. It’s not necessary, right? Because later, we call `post.meta('hero_image')` anyway. We could just delete the whole if call, which would make the whole example easier to understand. Instead, we could place the if call in Twig:

```twig
{% if post.meta('hero_image') %}
	<img src="{{ post.meta('hero_image').src | resize(500, 300) }}" />
{% endif %}
```

In addition to that, setting object properties dynamically isn’t something we should promote. If you wanted to use `post.hero_image` directly, the recommended way to do this would be to create a custom class that extends `Timber\Post` and then define a `hero_image` property or a `hero_image()` method there.

xavivars: I think that would make a lot of sense. I just "blindly" updated the example to use `meta` instead of relying on the magic call, but didn't really think about it

gchtr: @xavivars Good :). Do you want to update the PR or do you want me to do it?

xavivars: Please, go ahead. I completely forgot about this 😅

Levdbas (approved):

PR #2852, opened 2023-11-23
chore: update docs link

Levdbas:
## Issue
Docs links in readme is pointing to master

## Solution
point to v2 docs


gchtr:
This was fixed in https://github.com/timber/timber/pull/2853, right?

Levdbas:
You are right :D

PR #2853, opened 2023-11-23
docs: v2 readme refer to v1

Levdbas:
Related:

- https://github.com/timber/timber/issues/2741
- https://github.com/timber/timber/pull/2854

## Issue
People might be looking for the v1 documentation and code. 

## Solution
Let's help them a bit. While at it, update links to point to the proper location and update http:// links to https://

## Impact
Better understanding of the previous branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ Once Timber is installed, it gives any WordPress theme the ability to take advan
 
 * [Timber Documentation](https://timber.github.io/docs/)
 * [Timber 2.0 Upgrade Guide](https://timber.github.io/docs/v2/upgrade-guides/2.0/)
-* [Twig Reference](http://twig.sensiolabs.org/doc/templates.html)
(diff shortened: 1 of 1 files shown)


gchtr:
When looking at https://github.com/timber/timber/tree/readme-refer-to-v1, I think that now there’s heading level mixup. The **Installation** and **What now** sections are actually subheadings of **Documentation**. We could solve it by moving **Timber 1.x** down before **Timber’s mission**.



What do you think?

Levdbas:
> When looking at https://github.com/timber/timber/tree/readme-refer-to-v1, I think that now there’s heading level mixup. The **Installation** and **What now** sections are actually subheadings of **Documentation**. We could solve it by moving **Timber 1.x** down before **Timber’s mission**.
> 
> 
> 
> What do you think?

Totally agree, I moved it even a bit further down, just before links.

gchtr (approved):
Looking good 😎! I’ll let you merge this :)

PR #2855, opened 2023-11-29
2.x Fix information about updated Twig filters

### LINKED ISSUE #2851: [Docs] Contradictory instructions for `timber/twig` filter
If a reader is searching for a method to add a custom filter to Twig, they may come across the [documentation for `timber/twig/filters`](https://timber.github.io/docs/v2/hooks/filters/#timber/twig/filters) in the Filter Hooks section there is no mention here of any deprecation.

Or they might be working through the Timber 2.0 migration guide, which says that [you shouldn't use that filter any more](https://timber.github.io/docs/v2/upgrade-guides/2.0/#updated-timber/twig-filters). It also directs readers to "Check out the new [Extending Twig Guide](https://timber.github.io/docs/v2/guides/extending-timber/) to learn more." There is a bug here: it links to the `Extending Timber` page instead of the `Extending Twig` guide.

If the reader manages to find the `Extending Twig` guide after all (it is linked from the `Extending Timber` page as well), they will find the following under [`Changing filters`](https://timber.github.io/docs/v2/guides/extending-twig/#changing-filters):

> You can replace a filter with your own function or even remove a filter by updating the array items in $filters.

> `add_filter('timber/twig/filters', function ($filters) {`

Which contradicts the information in the upgrade guide.

Then just below that is the following, under [Adding functionality with the Twig Environment filter](https://timber.github.io/docs/v2/guides/extending-twig/#adding-functionality-with-the-twig-environment-filter):

> You can still extend the directly with the timber/twig filter. The only disadvantage of this is that once functions or filters are added to Twig, the can’t be removed anymore.

> Here’s the same functions and filters that we add above. But instead of using timber/twig/functions and timber/twig/filters, we use timber/twig directly.

This suggests that while using `timber/twig` is an option, it's slightly worse than using `timber/twig/filters`.

I would submit a pull request myself to correct these inconsistencies, but I honestly don't know what the correct information should be.

gchtr:
Related:

- #2851

## Issue

The info about the updated usage of some Twig filters might be confusing and one of the links was wrong.

## Solution

Improve Upgrade Guide.

## Usage Changes

None.

## Testing

No.

### DIFF
--- a/docs/v2/guides/extending-twig.md
+++ b/docs/v2/guides/extending-twig.md
@@ -5,7 +5,11 @@ order: "1650"
 
 ## Twig Environment
 
-Timber abstracts some of that functionality for you, so you can add functions or filters more easily.
+Twig has a `\Twig\Environment` class that you can [use to extend Twig](https://twig.symfony.com/doc/3.x/advanced.html). One of the most common ways this is used is to add [functions](https://twig.symfony.com/doc/3.x/advanced.html#functions) and [filters](https://twig.symfony.com/doc/3.x/advanced.html#filters) to Twig. But it could also be used to add globals, extensions, and more.
+
+Timber creates a `\Twig\Environment` object for you. You can use the `timber/twig` filter to access that object and extend Twig through it.
+
+Timber further abstracts some of that functionality for you, so you can add functions or filters more easily.
 
 ## Functions
 
--- a/docs/v2/upgrade-guides/2.0.md
+++ b/docs/v2/upgrade-guides/2.0.md
@@ -677,18 +677,26 @@ The following functions are now deprecated:
 
 ### Updated timber/twig filters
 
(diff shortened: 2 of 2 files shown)


Levdbas:
As from the linked ticket:

> That's an improvement for sure, but I'm still confused by what it means to "extend the Twig environment" that doesn't include adding functions or filters. The examples shown on the [Extending Twig Guide](https://timber.github.io/docs/v2/guides/extending-twig/#adding-functionality-with-the-twig-environment-filter) are all adding filters and functions. How is that different from adding filters and functions with the other hooks?
> 
> Putting my question more succinctly:
> If I want to add a custom filter to Twig in Timber, should I use timber/twig with $twig->addFunction or timber/twig/functions with $functions['function_name'] = ?

@gchtr , I am not sure about this either, but if you let me know I can just finish this PR.

Levdbas (approved):

Levdbas:
Thanks for clearing this up @gchtr !

PR #2863, opened 2023-12-06
feat: allow pagination object to be generated using $prefs only

### LINKED ISSUE #2635: Allow pagination object to be generated using $prefs only
I'd like to use [Timber\Pagination::get_pagination( $prefs = array() )](https://github.com/timber/timber/blob/8d9bff23d2677652469badb75dcb607449a18204/lib/Pagination.php#L23) to generate a pagination object using `$prefs` only. Since `$prefs` [overrides](https://github.com/timber/timber/blob/8d9bff23d2677652469badb75dcb607449a18204/lib/Pagination.php#L68) `$args` derived from `$wp_query`, this mostly works right now. However the "prev" link is not generated because it refers to the `$paged` query variable when [checking if the prev link should be removed](https://github.com/timber/timber/blob/8d9bff23d2677652469badb75dcb607449a18204/lib/Pagination.php#L93).

If you replace `$paged` with `$this->current` when [checking if the prev link should be removed](https://github.com/timber/timber/blob/8d9bff23d2677652469badb75dcb607449a18204/lib/Pagination.php#L93), then I can set `$current` in the `$prefs` array and the "prev" link will be generated as expected.

Calling `set_query_var('paged', $current)` before calling [Timber\Pagination::get_pagination()](https://github.com/timber/timber/blob/8d9bff23d2677652469badb75dcb607449a18204/lib/Pagination.php#L23) will work around this issue.

## Additional context
I'm rendering search results returned by a [SearchWP](https://searchwp.com/) supplemental search engine, which doesn't use `$wp_query`. If I can build a Timber pagination object from a SearchWP response, then I can use the same search results template for both native WP and SearchWP search engines.

Thanks for your hard work. Timber is awesome! :slightly_smiling_face:

ghost:
Related:

#2635 

## Issue
Allow pagination object to be generated using $prefs only. Currently there's a dependency on wp_query via the $paged wp_query param when checking if the previous pagination link should be rendered.

## Solution
Use $current instead of $paged when checking if previous pagination link should be generated. $current is initialized using $paged but then overridden with the corresponding value in $prefs. Now we can use $prefs to wholly configure a pagination object.

## Impact
No impact. Backwards compatible.

## Usage Changes
None.

## Considerations
None.

## Testing
No tests needed.


Levdbas (dismissed):
Thanks @phasdev , looks good to me!

Levdbas:
Just merged 2.0 back into this branch and there is a static analysis error @phasdev . @nlemoine , thoughts?

nlemoine:
PHPStan is right, the `$curent`variable might not be defined. @phasdev can you fix this?

ghost:
@nlemoine OK fixed.

Levdbas (approved):

Issue #2875, opened 2024-01-05
[BUG] `towebp` filter always returns new filename, even when conversion was unsuccessful


ghost:
### Expected Behavior

When using the `towebp` filter, the original filename should be returned if the file could not be converted.

### Actual behavior

The filename is returned with `towebp` after it.

### Steps to reproduce behavior

```php
{{ file | towebp }}
```

If file is `white-castle.tif`, this still returns `white-castle.webp`.

### Notes

- I just submitted a PR that also touches on the `towebp` filter, but that shouldn't affect solving this issue: https://github.com/timber/timber/pull/2876
- If this ticket is approved, I think it should be added to https://github.com/timber/timber/issues/2866

### What version of Timber are you using?

2.0.0

### How did you install Timber?

None

Levdbas:
Hi @expedition-robin-martijn , thank you so much for your ticket and PRs you created! I will add this to the Image handling ticket and we will take a look at your pull requests.

Issue #2850, opened 2023-11-21
[BUG] towebp filter not fool-proof against pseudo-duplicate filenames (image00001.jpeg,image00001.jpg)


raadegmbh:
### Expected Behavior

When uploading an image and using it in a post, users expect a webp representation for the correct source image to be generated by the towebp filter, and the correct image to be therefore displayed at the end of the chain.

### Actual behavior

Instead, the towebp filter overwrites competing pseudo-duplicate images:

- **image0001.jpg** becomes image0001-scaled.webp
- **image0001.jpeg** **_also_** becomes image0001-scaled.webp 

### Steps to reproduce behavior

The problem is contained in the [Image/Operation/ToWebp.php](https://github.com/timber/timber/blob/2.x/src/Image/Operation/ToWebp.php) file, as well as its 1.x equivalent (it seems to have been copied over from 1.x in 2.x).

A janky hotfix is to alter the [filename function in line 31ff:](https://github.com/timber/timber/blob/ee467c650e51d042a04d9107f35239cbc566d4d7/src/Image/Operation/ToWebp.php#L31C1-L35C6) to include the original source extension:

   ```
 public function filename($src_filename, $src_extension = 'webp')
    {
        $new_name = $src_filename . $src_extension . '.webp';
        return $new_name;
    }
```

which immediately ameliorates the issue.

### What version of Timber are you using?

Timber 2.x and 1.x

### What version of WordPress are you using?

Any version

### What version of PHP are you using?

Any version

### How did you install Timber?

Installed or updated Timber through Composer

CedricProCab:
I noticed the same issue 1 hour ago with two files with the same name but with different extensions : 
.i.e : velo 3.jpg and velo 3.png

The final name for the generated webp is velo-3.webp, so the second generated webp has overridden the first generated one

By doing the hotfix suggested by @raadegmbh, it will regenerate a new webp file with the original extension included, but there will be a lot of useless images in the uploads dir folders (the first generated webp without the extension included), so a clean of the uploads dir folders has to be done

Issue #2866, opened 2023-12-15
Big ticket about Image handling


Levdbas:
As discussed recently with the Timber core team, we have come to the conclusion that a lot of tickets/feature requests are about how we handle image in Timber. This ticket will serve as an archive for multiple tickets/pr's that are still based on the 1.x branch but might be helpful for development moving forward. Also some more recent issues and tickets.

Bottom line is that there are quite some issues with generation/deleting images and some requests regarding image manipulations via filters.

**Image modification via filters**
- https://github.com/timber/timber/pull/1916 
- https://github.com/timber/timber/pull/1942
- https://github.com/timber/timber/issues/1435
- https://github.com/timber/timber/issues/2285

**Images are not deleted**
- https://github.com/timber/timber/pull/2556
- https://github.com/timber/timber/issues/2535

**Image transformations error handling and edge cases**
- https://github.com/timber/timber/pull/2808
- https://github.com/timber/timber/pull/2774 and its issue https://github.com/timber/timber/issues/2773
- https://github.com/timber/timber/pull/2865
- https://github.com/timber/timber/pull/2725
- https://github.com/timber/timber/issues/2537
- https://github.com/timber/timber/issues/2850
- https://github.com/timber/timber/pull/2876 (multiple files with same name with different extension)
- https://github.com/timber/timber/issues/2875 Unsuccessful conversion dont throw false in some situations.

**Image transformations feature requests**
- https://github.com/timber/timber/issues/2518
- https://github.com/timber/timber/issues/2596

**Image & Attachement classes**
- https://github.com/timber/timber/issues/2785

**Sideloading**
- https://github.com/timber/timber/issues/1150

nlemoine:
Thank you @Levdbas! I'll post this issue I wrote some years ago and never posted. While reading it again, I feel that it's still relevant and can provide a starter base to discuss future Timber image handling.

----

It's been a long time that I wanted to share my thoughts on image manipulation.

Indeed, images handling has changed over the last years. Developers work sometimes very hard to win a few bytes on other assets (js, css, etc.) while images have an incredible reduction weight and performance gain potential.

### The ideal Twig image manipulation API

Here's my insights on an image manipulation API. 

- manipulations are defined in templates, at the image level. Having image sizes defined in a [some config file](https://github.com/liip/LiipImagineBundle) is tedious and becomes a nightmare at scale. The solution I'll come with will still be compatible with a global config approach.
- basic resize features: width/height/crop/fit
[...]

PR #2876, opened 2024-01-05
fix: fix overwriting files with webp filter

ghost:
Related:

- #2850

## Issue
When using the `webp` filter, the new filename is `filename.webp`. As a result, `filename.jpg` and `filename.png` would both create `filename.webp`, thus overwriting each other.

## Solution
I now include the original file extension in the new file. I've chosen to keep the original format, and simply add `.webp` behind it. This makes the original structure as clear as possible. Adding a dot is perfectly legitimate on every filesystem.

## Impact
After merging this, new files will be created in instances where the `webp` filter is already used. There's no way around that, however.

## Usage Changes
N/A

## Considerations
When updating the tests, I noticed that there is a bug in the current way this filter works, for which I created a new issue: https://github.com/timber/timber/issues/2875

## Testing
I updated the existing tests.

### DIFF
--- a/src/Image/Operation/ToWebp.php
+++ b/src/Image/Operation/ToWebp.php
@@ -30,7 +30,7 @@ class ToWebp extends ImageOperation
      */
     public function filename($src_filename, $src_extension = 'webp')
     {
-        $new_name = $src_filename . '.webp';
+        $new_name = $src_filename . '.' . $src_extension . '.webp';
         return $new_name;
     }
 
--- a/tests/test-timber-image-towebp.php
+++ b/tests/test-timber-image-towebp.php
@@ -25,7 +25,7 @@ class TestTimberImageToWEBP extends Timber_UnitTestCase
         $str = Timber::compile_string('{{file|towebp}}', [
             'file' => $filename,
         ]);
-        $renamed = str_replace('.png', '.webp', $filename);
+        $renamed = $filename . '.webp';
         $this->assertFileExists($renamed);
         $this->assertGreaterThan(1000, filesize($renamed));
         $this->assertEquals('image/png', mime_content_type($filename));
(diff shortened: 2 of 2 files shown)


gchtr:
We definitely need to discern the file types in the file names. Thanks for taking this on @expedition-robin-martijn!

As with all the PR involving file name changes for images, I wonder how we should handle this properly. Because this will **cause a lot of files to be regenerated**. It could cause lots of max execution errors and could quickly fill up space on some hosts and leave the uploads folder with lots of unused files.

Could we somehow add a migration path so that developers have a way to fix this on their sites?

- Provide a way to delete all deprecated (or even all) WebP files using the WP CLI.
- But then again, not all hosts allow you to run WP CLI. So maybe we could add integrations for popular [image regeneration plugins](https://wordpress.org/plugins/search/regenerate+thumbnails/) that could handle the migration as well.
- Maybe save generated WebP files to a different folder, as envisioned in https://github.com/timber/timber/issues/2866#:~:text=300%2C%20600%2C%20900%5D-,generated%20images%20are%20stored%20in%20a%20different%20directory%20than%20the%20source%20directory%20(uploads).%20I%20think%20it%27s%20important%20to%20clearly%20distinguish%20source%20images%20and%20manipulated%20images.%20Thus%20they%20can%20safely%20deleted%20and%20regenerated%20if%20something%20changed%20in%20the%20API.,-fluent%20API%2C%20that

I wonder whether we even should put this change behind a filter hook. The filter could be applied for all new Timber installations. For existing installations, we could maybe work with a flag that is saved in the `wp_options` database and is deleted once the change is through. Depending on the flag, we could show a notification in the admin dashboard.

I’m just thinking out loud here and am happy about better ideas.

Levdbas:
I have a feeling that we have to take the deep dive her as suggested by @nlemoine in https://github.com/timber/timber/issues/2866 and separate generated files at some point from the original media library. The tasks involved here would be so substantial we could dedicate a whole release to it. Now we just have to find the time to actually make such a big improvement.

So in hindsight, I think this PR can give us some ideas for handling filenames in that future release, but I think we cannot merge this as is. But thank you so much @expedition-robin-martijn , for taking the time to help with shaping the full solution.

ghost:
This completely makes sense!

Issue #2507, opened 2021-10-03
Query params are transformed to lowercase in URLHelper::get_params()


peterkracik:
Hi,

I am using `URLHelper::get_params()` to get parameters, but they are returned in lowercase. 
In my case, I would need to have them case-sensitive. My URL contains an ID similar to this one `8chr5ULa` which is then used for query and external API and the id in lowercase throws an error. 
I saw you call the `strtolower` and I was wondering if there is a reason for it, or it could be removed.

version: 1.18
but I can see, that the behaviour is the same on v2.

https://github.com/timber/timber/blob/2f6955077dfc264c116ef38433e181119523e318/lib/URLHelper.php#L536

kevin-quach:
I second this request as I use this for routing a deeplink which is case sensitive.

Levdbas:
Pr is merged and will be included in next 2.x release.

PR #2877, opened 2024-01-05
fix: unnecessary lowercasing parameters

ghost:
Related:

- #2507

## Issue
When using `URLHelper::get_params()`, everything is normalised to lowercase. I understand why you would do that with URLs, but in this specific instance it is about parameters. I think the author of the issue raises a valid point that case sensitive parameters are valid. The RFC only speaks of normalising the protocol and scheme: https://www.rfc-editor.org/rfc/rfc3986#page-11

## Solution
I removed the `strtolower`. (And fixed a small typo earlier in the file)

## Impact
If some code is relying on this method also turning everything into lowercase for them, that would break.

## Usage Changes
Developers should put `strtolower` around the output if they need that.

## Considerations
I realise it is a breaking change, and therefore not ideal. I am however convinced that the original solution is the actual breaking thing here.

## Testing
All existing tests were already lowercase.

### DIFF
--- a/src/URLHelper.php
+++ b/src/URLHelper.php
@@ -40,7 +40,7 @@ class URLHelper
     /**
      * Check to see if the URL begins with the string in question
      * Because it's a URL we don't care about protocol (HTTP vs HTTPS)
-     * Or case (so it's cAsE iNsEnSeTiVe)
+     * Or case (so it's cAsE iNsEnSiTiVe)
      *
      * @api
      * @return boolean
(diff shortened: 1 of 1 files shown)


Levdbas (approved):
Thank you for taking care of this and the typo along the way @expedition-robin-martijn

nlemoine (approved):
Still have to figure out what this URLHelper was made for ^^

In the meantime, this is legitimate 👍

Issue #2604, opened 2022-07-01
add page slug to transient name


Colir:
## Is your feature request related to a problem? Please describe.
I would like to be able to delete a specific page's cache (transient). 
Actually the transient name of a timber cached page is 
"timberloader_[HASH KEY BASE ON DATA]". 

It can be really to hard to retrieve the correct hash key.

## Describe the solution you’d like
Add the post slug to the transient name like "timberloader__[post_slug]_[HASH KEY BASE ON DATA]"

## Describe alternatives you’ve considered
It could be done by adding a filter to the CONSTANT "CACHEGROUP" ( like adding a filter in the set_cache method when generating the var "trans_key")

Thank you !

Colir:
Hello,
after re-dive in this difficulty to find the correct transient loader, i think the best solution would be to add a filter on

`$trans_key = substr($group.'_'.$key, 0, self::TRANS_KEY_LEN);`

In this way, we could be able to add any variable we want in the cache key (slug, ID, wathever...)

Thank you

gchtr:
Small note about how this was fixed: With the next release of Timber, you’ll be able to use the `timber/cache/transient_key` filter to update the transient key.

```php
add_filter( 'timber/cache/transient_key', function( $trans_key, $key, $group, $cache_mode ) {
    return $trans_key . '_my_suffix';
}, 10, 4 );
```

This way, you can use a different transient key when certain conditions are met. This should make it possible to delete the transient for a specific page.

PR #2878, opened 2024-01-05
feature: add  filter to cache methods

ghost:
Related:

- https://github.com/timber/timber/issues/2604

## Issue
The transient keys used for caching are hard to determine, making it hard to work with them as well.

## Solution
This PR adds a filter as proposed in the issue, which allows developers to modify the transient key. It gives a maximum of 4 variables to work with:

```php
/*
 * @param string $trans_key The transient key.
 * @param string $key The cache key.
 * @param string $group The cache group.
 * @param string $cache_mode The cache mode.
*/
```

## Impact
As this is a new filter, it does not affect existing code.

## Usage Changes
I added the parameters in a docblock as well. Although I am not sure if I did that the right way, so assistance would be appreciated here.

## Considerations
This is the only way I see to help developers such as those in the original issue, without modifying existing transient keys.

## Testing
I added a test to see if the filter works properly, which passes.

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -661,7 +661,7 @@ class Loader
          * Filters the transient key used for caching.
          *
          * @api
-         * @since 2.0.0
+         * @since 2.1.0
          * @example
          * ```
          * add_filter( 'timber/cache/transient_key', function( $trans_key, $key, $group, $cache_mode ) {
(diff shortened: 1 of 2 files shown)


Levdbas (dismissed):
Thank you @expedition-robin-martijn , this looks good to me. Thank you for taking the time for adding the filter docblocks!

nlemoine (approved):
All good, thank you @expedition-robin-martijn !

Issue #2843, opened 2023-11-13
[BUG] LocationManager returns wrong locations in child themes


Levdbas:
### Expected Behavior

In 1.x child themes could overwrite twig templates. This functionality is now broken since 2.0:

**Expected array of  `Timber\LocationManager::get_locations` with child theme active**
```php
array(4) {
  [0]=>
  string(58) "redacted/themes/child-theme/"
  [1]=>
  string(74) "redacted/themes/child-theme/resources/views/"
  [2]=>
  string(52) "redacted/themes/parent-theme/"
  [3]=>
  string(68) "redacted/themes/parent-theme/resources/views/"
}
```

### Actual behavior

**actual array of  `Timber\LocationManager::get_locations` with child theme active**
```php
array(7) {
  [0]=>
  string(48) "redacted/themes/parent-theme"
  [1]=>
  string(54) "redacted/themes/child-theme"
  [2]=>
  string(0) ""
  [3]=>
  string(64) "redacted/themes/parent-theme/resources/views"
  [4]=>
  string(70) "redacted/themes/child-theme/resources/views"
  [5]=>
  string(64) "redacted/themes/parent-theme/resources/views"
  [6]=>
  string(64) "redacted/themes/parent-theme/resources/views"
}
```

Since `redacted/themes/parent-theme/resources/views` comes earlier then` redacted/themes/child-theme/resources/views` the parent theme has precedence over the child theme.

### Steps to reproduce behavior

1. make minimal parent/child setup where views live in subdir resources.
2. set timber location via `Timber::$dirname = 'resources/views';`
3. activate child theme and overwrite index.twig for example
4. make sure you can spot the difference between the two twig templates.

### Notes

I debugged it to the following rules:

```php
		$locs = array();
		$locs = array_merge($locs, self::get_locations_user());
		$locs = array_merge($locs, self::get_locations_caller($caller));
		//remove themes from caller (this is what missing in v2)
		$locs = array_diff($locs, self::get_locations_theme());
		$locs = array_merge($locs, self::get_locations_theme());
		$locs = array_merge($locs, self::get_locations_caller($caller));
		$locs = array_unique($locs);
		//now make sure theres a trailing slash on everything
		$locs = array_map('trailingslashit', $locs);
		$locs = apply_filters('timber_locations', $locs);
		$locs = apply_filters('timber/locations', $locs);
```
In v2 there is the following code, which has not the same effect.

```php
        //remove themes from caller
[...]

PR #2884, opened 2024-01-15
Fix: LocationManager returns wrong locations in child themes

pratik-londhe4:
Related:

- #2843

## Issue

- Child theme not able to override templates of the parent theme due to wrong locations array.
## Solution

- This solution was proposed in #2843 by @palttamas but I could not find PR related to it so creating this

## Usage Changes

- None

### DIFF
--- a/src/LocationManager.php
+++ b/src/LocationManager.php
@@ -15,6 +15,7 @@ class LocationManager
         $locs = \array_merge_recursive($locs, self::get_locations_user());
         $locs = \array_merge_recursive($locs, self::get_locations_caller($caller));
         //remove themes from caller
+		$locs = \array_diff($locs, self::get_locations_theme());
         $locs = \array_merge_recursive($locs, self::get_locations_theme());
         $locs = \array_merge_recursive($locs, self::get_locations_caller($caller));
         $locs = \array_merge_recursive($locs, self::get_locations_open_basedir());


Levdbas:
Hi @pratik-londhe4 , thank you for taking a shot at this. Though your fix seems different from what @palttamas suggested in  https://github.com/timber/timber/issues/2843#issuecomment-1846757344 ?

```php
--$locs = \array_merge_recursive($locs, self::get_locations_theme());
--$locs = \array_merge_recursive($locs, self::get_locations_caller($caller));
++$locs = array_diff($locs, self::get_locations_theme());
++$locs = array_merge_recursive($locs, self::get_locations_theme());
```

pratik-londhe4:
hi, @levdbas. 
This `$locs = array_merge_recursive($locs, self::get_locations_theme());` is already present in the get_locations, so adding it again is unnecessary.

I have tested these changes in a child theme setup.

Levdbas:
Hi @pratik-londhe4 , unfortunately the tests fail. Are you willing to take another look at this?

palttamas:
> Hi @pratik-londhe4 , unfortunately the tests fail. Are you willing to take another look at this?

Hey, just chipping in. I did try to fix the issue, tried couple of variations but as @Levdbas is saying, the tests fail, at which point I quickly lost interest, since we resolved this thing internally with a filter anyway.

pratik-londhe4:
@Levdbas , sure I will take look at failing tests and find solution for it.

okadots:
I was having this same issue and adding `$locs = array_diff($locs, self::get_locations_theme());` as described above fixed the problem, but we are also seeing the `Array to string conversion` error noted [here](https://github.com/timber/timber/issues/2843#issuecomment-1844948922)

We are upgrading from the old Timber WP plugin, where we were not having this issue. I went back and looked at the old plugin code and that `array_diff` line is present.

UPDATE:
In our case, the parent theme is never used on it's own - it is used as a framework that can be customized in the child theme. I simply changed the order of the arrays in the array merge so that the current theme (child) locations come first. _This has not been thoroughly tested on a site that only uses the parent theme and this may not be the final fix_

`$locs = \array_merge_recursive($locs, self::get_locations_theme());`  becomes => 
`$locs = \array_merge_recursive(self::get_locations_theme(), $locs);`

Levdbas:
Hi @pratik-londhe4 , I fixed this up in  #2925. Everyone, thank you for your input on this.

PR #2886, opened 2024-01-20
Fix phpstan tests

Levdbas:
## Issue
Lately some tests fail. This PR focusses on getting the phpstan tests running again while also tackeling two other phpstan errors I got locally.

## Solution
Include a file to discover symbols, changing a property name and ignore an ACF function error.

## Impact
PHPSTAN tests that run again.

## Usage Changes
no

## Considerations
I got 20 more errors when running phpstan, mostly the unsafe use of new static(), we might want to ignore that as well? Other errors are also some other missing symbols. Let me know if we want to look into those as well.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -4,6 +4,7 @@ parameters:
     paths:
         - src/
     scanFiles:
+        - tests/phpstan/bootstrap.php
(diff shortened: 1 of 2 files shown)


nlemoine (approved):
Thank you @Levdbas, this looks good to me 👍

PR #2892, opened 2024-01-26
2.x-update-cs

nlemoine:
Bump ESC to lastest version (12).

Note it changed quite a lot a files and we might want to add this commit in the .git-blame-ignore-revs file.

It also removed to full namespace from the docblocks, I don't if this cause issues when generating API reference? @gchtr


nlemoine:
Sorry, completely forgot I had some pending changes locally and messed up with git. See #2893

PR #2893, opened 2024-01-26
feat: update ECS config and apply standards

nlemoine:
Bump ECS config.

Note it changed quite a lot a files and we might want to add this commit in the .git-blame-ignore-revs file.

It also removed to full namespace from the docblocks, I don't if this cause issues when generating API reference? @gchtr

### DIFF
--- a/src/AccessesPostsLazily.php
+++ b/src/AccessesPostsLazily.php
@@ -3,7 +3,6 @@
 namespace Timber;
 
 use ReturnTypeWillChange;
-
 use Timber\Factory\PostFactory;
 use WP_Post;
 
@@ -25,7 +24,7 @@ trait AccessesPostsLazily
     /**
      * PostFactory instance used internally to instantiate Posts.
      *
-     * @var \Timber\Factory\PostFactory
+     * @var PostFactory
      */
     private $factory;
 
@@ -67,7 +66,7 @@ trait AccessesPostsLazily
      *   // No additional overhead here.
      * }
      * ```
-     * @return \Timber\PostCollectionInterface The realized PostQuery.
+     * @return PostCollectionInterface The realized PostQuery.
      */
     public function realize(): self
     {
--- a/src/MenuItem.php
+++ b/src/MenuItem.php
@@ -90,7 +90,7 @@ class MenuItem extends CoreEntity
      *
      * @since 1.12.0
      * @see \Timber\MenuItem::menu()
-     * @var \Timber\Menu The `Timber\Menu` object the menu item is associated with.
+     * @var Menu The `Menu` object the menu item is associated with.
      */
     protected $menu;
 
@@ -119,8 +119,8 @@ class MenuItem extends CoreEntity
     /**
      * @internal
      * @param array|object $data The data this MenuItem is wrapping
-     * @param \Timber\Menu $menu The `Timber\Menu` object the menu item is associated with.
-     * @return \Timber\MenuItem a new MenuItem instance
+     * @param Menu $menu The `Menu` object the menu item is associated with.
+     * @return MenuItem a new MenuItem instance
      */
     public static function build($data, ?Menu $menu = null): self
     {
@@ -130,7 +130,7 @@ class MenuItem extends CoreEntity
     /**
      * @internal
      * @param WP_Post $data
-     * @param \Timber\Menu $menu The `Timber\Menu` object the menu item is associated with.
+     * @param Menu $menu The `Menu` object the menu item is associated with.
      */
     protected function __construct(WP_Post $data, $menu = null)
     {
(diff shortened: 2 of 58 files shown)


Levdbas:
I see we are no longer prefixing the docblock return types with the full namespace. Could this have a negative impact in the symbol discovery of IDE's?

nlemoine:
No, since they're referenced in imports: https://github.com/timber/timber/blob/77bdbdb0640c3057fd6489f43d08b266e39e474c/src/Factory/PagesMenuFactory.php#L5

I just wonder if the doc generation tool is able to compute the FQCN or is it parsing the doc block as is, example:
Capture d’écran 2024-01-26 à 12 34 09

Levdbas:
I've just tested the doc generation in combination of your branch and the docs still seem to compile the correct paths :)

src/MenuItem.php
```diff
@@ -90,7 +90,7 @@ class MenuItem extends CoreEntity
      *
      * @since 1.12.0
      * @see \Timber\MenuItem::menu()
-     * @var \Timber\Menu The `Timber\Menu` object the menu item is associated with.
+     * @var Menu The `Timber\Menu` object the menu item is associated with.
```
Levdbas:
```suggestion
     * @var Menu The `Menu` object the menu item is associated with.
```

src/MenuItem.php
```diff
@@ -119,8 +119,8 @@ class MenuItem extends CoreEntity
     /**
      * @internal
      * @param array|object $data The data this MenuItem is wrapping
-     * @param \Timber\Menu $menu The `Timber\Menu` object the menu item is associated with.
-     * @return \Timber\MenuItem a new MenuItem instance
+     * @param Menu $menu The `Timber\Menu` object the menu item is associated with.
```
Levdbas:
```suggestion
     * @param Menu $menu The `Menu` object the menu item is associated with.
```

src/MenuItem.php
```diff
@@ -130,7 +130,7 @@ public static function build($data, ?Menu $menu = null): self
     /**
      * @internal
      * @param WP_Post $data
-     * @param \Timber\Menu $menu The `Timber\Menu` object the menu item is associated with.
+     * @param Menu $menu The `Timber\Menu` object the menu item is associated with.
```
Levdbas:
```suggestion
     * @param Menu $menu The `Menu` object the menu item is associated with.
```

src/MenuItem.php
```diff
@@ -479,7 +479,7 @@ public function target()
      *
      * @api
      * @since 1.12.0
-     * @return \Timber\Menu The `Timber\Menu` object the menu item is associated with.
+     * @return Menu The `Timber\Menu` object the menu item is associated with.
```
Levdbas:
```suggestion
     * @return Menu The `Menu` object the menu item is associated with.
```

src/AccessesPostsLazily.php
```diff
@@ -67,7 +66,7 @@ trait AccessesPostsLazily
      *   // No additional overhead here.
      * }
      * ```
-     * @return \Timber\PostCollectionInterface The realized PostQuery.
+     * @return PostCollectionInterface The realized PostQuery.
```
Levdbas, line 69: This throws a static analysis error. Is this something we must take care of? Not sure if this was something that we already discussed in an issue.

Levdbas (commented):
I suggested some docblock updates while at it and there is one return type issue that we might already discussed in a ticket. Just checking. Other than that, looks good to me.

src/AccessesPostsLazily.php
Levdbas: @gchtr , mentioned this over here: https://github.com/timber/timber/pull/2668

nlemoine: Haven't figured it out yet but this shouldn't prevent us from merging the PR. We don't require PHPStan to pass (yet). :)

Levdbas:
@gchtr , can you take a final look at this?

gchtr:
> It also removed to full namespace from the docblocks, I don't if this cause issues when generating API reference? @gchtr

It’s not a problem, because the reflections still work and generate fully qualified names, apparently.

It only **doesn’t work** for hook documentation. There, the types have to be generated from the DocBlock. As you can see in the following screenshot, we could have `Post` or `Environment`, but wouldn’t know whether this comes from `Timber` or `Twig`.



These hook DocBlocks are parsed by looping over the tokens in https://github.com/timber/teak/blob/a5af1504cadb8e5d2277069165caf948d1d0c012/lib/Compiler/HookReference.php#L62-L70. I guess I could somehow also derive the types from the variables that are present before a hook appears. I’ll have to check this out. I created an issue for this in https://github.com/timber/teak/issues/3.

For now, I think it’s a tradeoff we can live with, because it only affects some few filters in https://timber.github.io/docs/v2/hooks/filters/.

gchtr (approved):

nlemoine:
Thanks @gchtr for the details about doc generation. There should be a way to grab the FQCN, let me know if I can help.

PR #2894, opened 2024-01-26
feat: bump php-stubs/acf-pro-stubs to ^6.0

nlemoine:
Let's use latest acf pro stubs version


Levdbas (approved):
Ran this locally without issue!

PR #2895, opened 2024-01-26
fix: tests failing since Twig 3.8.0

nlemoine:
Twig recently changed the way exceptions are thrown during template rendering. Depending on the version, different types of execption can be thrown.

See: https://github.com/twigphp/Twig/commit/85bf01b4abd4b4ee6f6d1aca19af74189c939d69

### DIFF
--- a/tests/test-timber-meta.php
+++ b/tests/test-timber-meta.php
@@ -2,6 +2,8 @@
 
 use Timber\Integration\AcfIntegration;
 use Timber\Timber;
+use Twig\Environment;
+use Twig\Error\RuntimeError;
 
 /**
  * Class TestTimberMeta
@@ -618,7 +620,17 @@ class TestTimberMeta extends Timber_UnitTestCase
      */
     public function testPostMetaDirectAccessMethodWithRequiredParametersConflict()
     {
-        $this->expectException(\ArgumentCountError::class);
+        /**
+         * Twig 3.8 changed the way some exception are handled and a different exception is thrown.
+         *
+         * @see https://github.com/twigphp/Twig/commit/85bf01b4abd4b4ee6f6d1aca19af74189c939d69
+         */
+        if (version_compare(Environment::VERSION, '3.8.0', '>=')) {
+            $this->expectException(RuntimeError::class);
+        } else {
+            $this->expectException(\ArgumentCountError::class);
+        }
+
         $post_id = $this->factory->post->create();
 
         update_post_meta($post_id, 'public_method_with_args', 'I am a meta value');
@@ -641,7 +653,17 @@ class TestTimberMeta extends Timber_UnitTestCase
      */
     public function testTermMetaDirectAccessMethodWithRequiredParametersConflict()
     {
-        $this->expectException(\ArgumentCountError::class);
+        /**
+         * Twig 3.8 changed the way some exceptions are handled and a different exception is thrown.
+         *
+         * @see https://github.com/twigphp/Twig/commit/85bf01b4abd4b4ee6f6d1aca19af74189c939d69
+         */
+        if (version_compare(Environment::VERSION, '3.8.0', '>=')) {
+            $this->expectException(RuntimeError::class);
+        } else {
+            $this->expectException(\ArgumentCountError::class);
+        }
+
         $term_id = $this->factory->term->create();
 
         update_term_meta($term_id, 'public_method_with_args', 'I am a meta value');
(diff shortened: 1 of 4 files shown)


tests/test-timber-meta.php
```diff
@@ -641,7 +653,17 @@ public function testPostMetaDirectAccessMethodWithRequiredParametersConflict()
      */
     public function testTermMetaDirectAccessMethodWithRequiredParametersConflict()
     {
-        $this->expectException(\ArgumentCountError::class);
+        /**
+         * Twig 3.8 changed the way some exception are handled and a different exception is thrown.
```
Levdbas:
```suggestion
         * Twig 3.8 changed the way some exceptions are handled and a different exception is thrown.
```

tests/test-timber-meta.php
```diff
@@ -666,7 +688,16 @@ public function testTermMetaDirectAccessMethodWithRequiredParametersConflict()
      */
     public function testUserMetaDirectAccessMethodWithRequiredParametersConflict()
     {
-        $this->expectException(\ArgumentCountError::class);
+        /**
+         * Twig 3.8 changed the way some exception are handled and a different exception is thrown.
```
Levdbas:
```suggestion
         * Twig 3.8 changed the way some exceptions are handled and a different exception is thrown.
```

tests/test-timber-meta.php
```diff
@@ -693,7 +724,16 @@ public function testUserMetaDirectAccessMethodWithRequiredParametersConflict()
      */
     public function testCommentMetaDirectAccessMethodWithRequiredParametersConflict()
     {
-        $this->expectException(\ArgumentCountError::class);
+        /**
+         * Twig 3.8 changed the way some exception are handled and a different exception is thrown.
```
Levdbas:
```suggestion
         * Twig 3.8 changed the way some exceptions are handled and a different exception is thrown.
```

Levdbas (changes requested):
just minor documentation changes. Bloody good job! I was already looking into this as well what caused these failed tests.

nlemoine:
Just banged my head for more than one hour on that bug 🤯

I don't really understand why we are testing this. Checking if those cases (calling `{{ object.method_expecting_argument }}`) throw exceptions is in Twig scope, not ours.

I feel (already noticed before) we are sometimes testing Twig twice. With the extra burden of supporting a wider range.

Levdbas:
> Just banged my head for more than one hour on that bug 🤯
> 
> I don't really understand why we are testing this. Checking if those cases (calling `{{ object.method_expecting_argument }}`) throw exceptions is in Twig scope, not ours.
> 
> I feel there (already noticed before) we are sometimes testing Twig twice. With the extra burden of supporting a wider range.

By lowering the amount of WP and PHP versions we test, hopefully the amount of these kind of tests deprecation will decrease as well.

nlemoine:
> By lowering the amount of WP and PHP versions we test, hopefully the amount of these kind of tests deprecation will decrease as well.

On some other cases, probably. 

On that particular one, unless bumping Twig to 3.8 (which we might not want), the problem is that we test things that [shouldn't require tests from our side](https://github.com/timber/timber/blob/664ea625504a0d781ac2efeb5e2b8a39c5ac3e70/tests/test-timber-meta.php#L639-L640) (+ all the other similar tests).

People using Twig should be aware of [how Twig handles variables](https://twig.symfony.com/doc/3.x/templates.html#rendering-templates:~:text=throws%20an%20exception.-,Implementation,-For%20convenience%27s%20sake).

nlemoine:
And it's all green again. Merging this to make all other PR pass (or not but not polluted with our own failing tests).

gchtr:
> On that particular one, unless bumping Twig to 3.8 (which we might not want), the problem is that we test things that [shouldn't require tests from our side](https://github.com/timber/timber/blob/664ea625504a0d781ac2efeb5e2b8a39c5ac3e70/tests/test-timber-meta.php#L639-L640) (+ all the other similar tests).

I agree. I guess the idea behind these tests were to check that our custom magic methods work as intended all the way through. I agree with you that we shouldn’t test for argument count errors, though.

> People using Twig should be aware of [how Twig handles variables](https://twig.symfony.com/doc/3.x/templates.html#rendering-templates:~:text=throws%20an%20exception.-,Implementation,-For%20convenience%27s%20sake).

In my experience, this is something that Timber developers sometimes are not aware of.

nlemoine:
> In my experience, this is something that Timber developers sometimes are not aware of.

Maybe, but is it in our scope? Timber users are in fact also using Twig 🙂

Further more, adding those tests won't bring that behavior knowledge to our end users.

gchtr:
> Maybe, but is it in our scope? Timber users are in fact also using Twig 🙂

It isn’t a problem big enough that we should act.

> Further more, adding those tests won't bring that behavior knowledge to our end users.

No, I didn’t think these are connected at all. If we should handle the problem in any way, it should be with more extensive documentation, but not with tests.

PR #2897, opened 2024-01-27
feat: add filter for sideloaded images basename

### LINKED ISSUE #2508: Naming configuration for side loaded images
Hello everyone, this project is amazing. Thanks for all the hard work!

I'm using Timber in combination with the [WP Offload Media](https://wordpress.org/plugins/amazon-s3-and-cloudfront/) plugin. When using one of Timber's image modification filters (e.g. `resize()`) it will cause Timber to download the image to the server's "uploads" directory in order to perform the modification. The name of the side loaded image is the `md5()` hash based on the image's external URL. The resulting URLs for modified images are not very SEO friendly:

```
/wp-content/uploads/2021/10/e16806f9aae4f1b0ec61251ea02847b8-250x250-c-default.png
```

In comparison, here is the URL of the same resized image without using the WP Offload Media plugin:

```
/wp-content/uploads/2021/10/SEO-Friendly-Title-250x250-c-default.png
```

I'm aware that the side loaded file name has to contain a hash based on the origin URL because it is likely, that the external storage location does not follow the same path structure as WordPress's "uploads" directory (`<year>/<month>`). But I don't see a reason why it should not be possible to override the default naming scheme of side loaded images. This could easily be done without breaking changes by introducing a filter.

This is how Timber currently determines the name of any side loaded image (`$file` being the external URL):

https://github.com/timber/timber/blob/e339378d81578c5587dc5c11a1f7d44c5fa29792/lib/ImageHelper.php#L347-L359

We just have to add one line to the function:

```diff
public static function get_sideloaded_file_loc( $file ) {
   $upload = wp_upload_dir();
   $dir = $upload['path'];
   $filename = $file;
   $file = parse_url($file);
   $path_parts = PathHelper::pathinfo($file['path']);
   $basename = md5($filename);
+  $basename = apply_filters('timber/ImageHelper/sideloaded_file/basename', $basename, $path_parts);
   $ext = 'jpg';
   if ( isset($path_parts['extension']) ) {
      $ext = $path_parts['extension'];
   }
   return $dir.'/'.$basename.'.'.$ext;
}
```

Then we'd be able to define the naming scheme inside the theme configuration:

```php
add_filter('timber/ImageHelper/sideloaded_file/basename', function ($basename, $path_parts) {
  return $path_parts['filename'] . '-' . substr($basename, 0, 6);
}, 10, 2);
```

Resulting in SEO friendly URLs like this one:

```
/wp-content/uploads/2021/10/seo-friendly-title-e16806-250x250-c-default.png
```

I'm happy to provide a PR if this proposal gets approved.

Levdbas:
Related:

- https://github.com/timber/timber/issues/2508

## Issue
Basename of sideloaded images could not be configured,

## Solution
Introduce filter as described in the original ticket to handle this.

## Impact
More control over sideload images filenames

## Considerations
When looking for a proper name for the filter I found an odd named filter. We might want to take a look at this one as well and provide consistent name? https://github.com/timber/timber/blob/e4ff72f451e11b05887179086e4bb5a82d799184/src/ImageHelper.php#L515

## Testing
no

### DIFF
--- a/src/ImageHelper.php
+++ b/src/ImageHelper.php
@@ -408,7 +408,7 @@ class ImageHelper
          * @param string $basename Current basename for the sideloaded file.
          * @param array $path_parts Array with path info for the sideloaded file.
          */
-        $basename = \apply_filters('timber/image_helper/sideload_image/basename', $basename, $path_parts);
+        $basename = \apply_filters('timber/sideload_image/basename', $basename, $path_parts);
 
         $ext = 'jpg';
         if (isset($path_parts['extension'])) {
--- a/src/ImageHelper.php
+++ b/src/ImageHelper.php
@@ -400,7 +400,7 @@ class ImageHelper
          * @example
          * ```php
          * // Change the basename used for sideloaded images.
-         * add_filter( 'timber/image_helper/sideload_image/basename', function ($basename, $path_parts) {
+         * add_filter( 'timber/sideload_image/basename', function ($basename, $path_parts) {
          *     return $path_parts['filename'] . '-' . substr($basename, 0, 6);
          * }, 10, 2)
          * ```
--- a/src/ImageHelper.php
+++ b/src/ImageHelper.php
@@ -393,6 +393,23 @@ class ImageHelper
         $file = \parse_url($file);
         $path_parts = PathHelper::pathinfo($file['path']);
         $basename = \md5($filename);
+
+        /**
+         * Filters basename for sideloaded files.
+         * @since 2.0.1
+         * @example
+         * ```php
+         * // Change the basename used for sideloaded images.
+         * add_filter( 'timber/image_helper/sideload_image/basename', function ($basename, $path_parts) {
+         *     return $path_parts['filename'] . '-' . substr($basename, 0, 6);
+         * }, 10, 2)
+         * ```
+         *
+         * @param string $basename Current basename for the sideloaded file.
+         * @param array $path_parts Array with path info for the sideloaded file.
+         */
+        $basename = \apply_filters('timber/image_helper/sideload_image/basename', $basename, $path_parts);
+
         $ext = 'jpg';
         if (isset($path_parts['extension'])) {
             $ext = $path_parts['extension'];


src/ImageHelper.php
```diff
@@ -393,6 +393,23 @@ public static function get_sideloaded_file_loc($file)
...
+         * // Change the basename used for sideloaded images.
+         * add_filter( 'timber/image_helper/sideload_image/basename', function ($basename, $path_parts) {
+         *     return $path_parts['filename'] . '-' . substr($basename, 0, 6);
+         * }, 10, 2)
+         * ```
+         *
+         * @param string $basename Current basename for the sideloaded file.
+         * @param array $path_parts Array with path info for the sideloaded file.
+         */
+        $basename = \apply_filters('timber/image_helper/sideload_image/basename', $basename, $path_parts);
```
gchtr: I would just remove the `image_helper` part in the path. In case we reorganize classes and this filter moves to a different location, it would still make sense.

```suggestion
        $basename = \apply_filters('timber/sideload_image/basename', $basename, $path_parts);
```

gchtr (changes requested):

src/ImageHelper.php
```diff
@@ -393,6 +393,23 @@ public static function get_sideloaded_file_loc($file)
...
         $path_parts = PathHelper::pathinfo($file['path']);
         $basename = \md5($filename);
+
+        /**
+         * Filters basename for sideloaded files.
+         * @since 2.0.1
+         * @example
+         * ```php
+         * // Change the basename used for sideloaded images.
+         * add_filter( 'timber/image_helper/sideload_image/basename', function ($basename, $path_parts) {
```
gchtr: Damn, forgot this one :)

```suggestion
         * add_filter( 'timber/image_helper/sideload_image/basename', function ($basename, $path_parts) {
```

gchtr (changes requested):
Thanks for the update!

> When looking for a proper name for the filter I found an odd named filter. We might want to take a look at this one as well and provide consistent name?
> 
> https://github.com/timber/timber/blob/e4ff72f451e11b05887179086e4bb5a82d799184/src/ImageHelper.php#L515

This would now be consistent with `timber/sideload_image/basename`, right? Do you think still think `timber/sideload_image/subdir` is an odd name?

src/ImageHelper.php
Levdbas: Nice catch! Fixed in https://github.com/timber/timber/pull/2897/commits/141eeb10495881cfe56e8140886ae56f308e73d2

gchtr (approved):

PR #2898, opened 2024-01-27
docs: Document filters and some methods

### LINKED ISSUE #1504: Document hooks used in Timber

Levdbas:
Related:

- #1504

## Issue
Some filters were not documented. Lets fix that.

## Solution
Document filters!

## Impact
Better understanding of the filters available in Timber.

## Usage Changes
none.

## Considerations
I did not solve all the todos. In Loader.php, a class still pretty vague for me, has still 4 undocumented actions/filters. Since this a pretty important Class I would like to leave that to someone with good knowledge of the innerworkings of the twig loader/compiler.

Please have a good look at my provided docs/examples. Sometimes it was pretty hard to come up with a good example for a filter. So please feel free to enhance these as well.

## Testing
No.


src/Helper.php
```diff
@@ -71,30 +71,34 @@ public static function transient($slug, $callback, $transient_time = 0, $lock_ti
...
+     * @param integer     $lock_timeout      How long (in seconds) to lock the transient to prevent race conditions
+     * @param boolean     $force             Force callback to be executed when transient is locked
      * @param boolean     $enable_transients Force callback to be executed when transient is locked
      */
     protected static function handle_transient_locking($slug, $callback, $transient_time, $lock_timeout, $force, $enable_transients)
     {
         if ($enable_transients && self::_is_transient_locked($slug)) {
             /**
-             * Filters …
+             * Whether to force a locked transients to be regenerated.
```
gchtr: When documenting filters, the DocBlock always needs to start with "Filters …", so that the script that generates the documentation can pick it up.

```suggestion
             * Filters whether to force a locked transients to be regenerated.
```

gchtr (changes requested):
Whoaaaa, thanks @Levdbas 🔥!

Just one thing you need to be careful of: When documenting filters, the DocBlock always needs to start with "Filters …", so that the script that generates the documentation can pick it up.

There’s a couple of instances where you start with "Whether". We need to change these before we can merge this.

Levdbas:
Just fixed the last issues as well @gchtr :)

gchtr (approved):

PR #2899, opened 2024-01-27
chore: remove Lando config

nlemoine:
I don't think we need this anymore.

There are more alternatives today (wp-env, etc.) if we need to spin up a dev environment.


Levdbas (approved):
Sounds good to me!

PR #2900, opened 2024-01-27
ci: add PHP 8.2/8.3 and update test matrix

nlemoine:
Time to update those tests and run them against more recent versions.

PHP 8.3 should not be experimental IMHO.

We can add it we'll get rid of WP 5.3 (2.1 ?)

### DIFF
--- a/.github/workflows/php-unit-tests.yml
+++ b/.github/workflows/php-unit-tests.yml
@@ -46,7 +46,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: ['7.4', '8.0', '8.1']
+        php: ['7.4', '8.0', '8.1', '8.2']
         wp: ['latest', '5.3']
         multisite: ['0', '1']
         dependency-version: ['highest', 'lowest']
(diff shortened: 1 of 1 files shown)


Levdbas:
Hi @nlemoine , nice work on updating the test matrix! 

> PHP 8.3 should not be experimental IMHO.

PHP8.3 is still labeled as experimental by WP. Should we follow that?

> We can add it we'll get rid of WP 5.3 (2.1 ?)

What do you mean by this?

nlemoine:
> PHP8.3 is still labeled as experimental by WP. Should we follow that?

If you take a look at [the compatibility table](https://make.wordpress.org/core/handbook/references/php-compatibility-and-wordpress-versions/), even PHP 8.2 is still marked as beta 😅

However, we should make sure our code behaves with more recent/latest versions, even if WP doesn't.

> What do you mean by this?

Sorry, I meant to write "We can add it **when** we'll get rid of WP 5.3 (2.1 ?)".

It refers to the whole PHP/WP compatibility policy we all agreed to introduce soon. 2.1 seems fair to me. Maybe we should make that public (pinned issue? readme? both? elsewhere?) already so users can anticipate that policy change.

Levdbas:
> > PHP8.3 is still labeled as experimental by WP. Should we follow that?
> 
> If you take a look at [the compatibility table](https://make.wordpress.org/core/handbook/references/php-compatibility-and-wordpress-versions/), even PHP 8.2 is still marked as beta 😅
> 
> However, we should make sure our code behaves with more recent/latest versions, even if WP doesn't.
> 
> > What do you mean by this?
> 
> Sorry, I meant to write "We can add it **when** we'll get rid of WP 5.3 (2.1 ?)".
> 
> It refers to the whole PHP/WP compatibility policy we all agreed to introduce soon. 2.1 seems fair to me. Maybe we should make that public (pinned issue? readme? both? elsewhere?) already so users can anticipate that policy change.

Thanks for clearing that up! Yes, I think a pinned post will be a good idea to notify of that upcoming change.  And we always have this check that we can bump to a specific required version: https://github.com/timber/timber/blob/5f38a1f8da5a9611fe43957d4e835810616da95f/src/Admin.php#L13

nlemoine:
I'm merging this a bit fast because I have another big PR related to tests that's coming.

> And we always have this check that we can bump to a specific required version

Oh I forgot about this, maybe we should put the WP version in a constant on the Timber class.

gchtr:
> Oh I forgot about this, maybe we should put the WP version in a constant on the Timber class.

👍

PR #2905, opened 2024-01-29
Fix adding classes in `MenuItem`

romainmenke:
Related:

N/A

## Issue

In Timber V1 `MenuItem` would have classes with their own ID e.g. `menu-item-123`.
In Timber V2 this was missing.

## Solution

Upon investigation I found that `add_class` had an inverted boolean check for `in_array`.
Correcting this fixes the issue.

## Impact

This changes improves backwards compat with V1.

## Usage Changes

N/A

## Considerations

N/A

## Testing

I assume that tests are missing or incorrect for this feature because it is a boolean flip.
Might be best if someone increased coverage.

### DIFF
--- a/src/MenuItem.php
+++ b/src/MenuItem.php
@@ -180,7 +180,7 @@ class MenuItem extends CoreEntity
     public function add_class(string $class_name)
     {
         // Class name is already there
-        if (!\in_array($class_name, $this->classes, true)) {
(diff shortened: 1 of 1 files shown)


Levdbas (approved):
Good catch @romainmenke

nlemoine (approved):

Issue #2908, opened 2024-01-30
Filter $output after render and before it is cached


jl-a:
I use the `timber/output` filter to modify output after it has been rendered. I can see that this filter is applied to Timber's output after the output has been cached. Therefore the filter will be run every time, regardless if the original output is cached or not.

This is great for filters that dynamically change content; however for deterministic filters it's a bit inefficient. In my specific case I use `timber/output` with a HTML minification library to clean up whitespace; obviously if the un-minified output is already being cached it seems inefficient to not also cache the minified render.

Could an additional filter for `$output` be added immediately after the template is rendered, and before the cache is set? Perhaps something like `timber/output/pre-cache`

Levdbas:
Hi @jl-a , thank you for your issue and welcome at Timber. Please fill in the full issue report next time so we know what we're working with. But in this case you seem pretty sure where a changes has to be made. Would you like to create a Pull Request to add this filter?

PR #2910, opened 2024-01-30
Add filter to $output before it is cached

jl-a:
Related:
- #2908 

## Issue
When filtering $output with `timber/output` the result is not cached. This is inefficient if the filter being applied is deterministic based on $output content, because the filter results could also be cached rather than being run every time.

## Solution
This pull request adds another filter, `timber/output/pre-cache`, immediately after the template is rendered and before the $output is cached. If this filter is used, then its results will be cached.

## Impact
This pull request is backwards compatible, and does not break any existing uses of `timber/output`. The `apply_filters` call may add a tiny bit of overhead to each render (not including the actual filter, if any are added) , but overall this could be considered negligible.

## Usage Changes
No changes to how Timber is used, other than an additional filter.

## Testing
No tests needed.

### DIFF
--- a/src/ImageHelper.php
+++ b/src/ImageHelper.php
@@ -396,7 +396,7 @@ class ImageHelper
 
         /**
          * Filters basename for sideloaded files.
-         * @since 2.0.1
+         * @since 2.1.0
          * @example
          * ```php
          * // Change the basename used for sideloaded images.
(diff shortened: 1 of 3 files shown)


nlemoine (approved):
Looks all good thanks @jl-a!

Issue #2859, opened 2023-12-01
Fatal error when get_posts with wrong ID in array


theamnesic:
### Expected Behavior

When using get_posts with array like this : `{% for item in get_posts( [ 1503, 543, 3220 ] ) %}`, Timber display posts and ignore if the ID doesn't exists.

### Actual behavior

Fatal error if an ID doesn't exists: 

`Fatal error: Uncaught Error: Call to a member function setup() on null in /Users/g/Sites/g/vendor/timber/timber/src/PostsIterator.php:49 Stack trace: #0 /Users/g/Sites/g/vendor/twig/twig/src/Environment.php(361) : eval()'d code(105): Timber\PostsIterator->current() #1`

### Steps to reproduce behavior

Loop with an array width `get_posts` with a wrong ID.

### What version of Timber are you using?

2.0.0

### What version of WordPress are you using?

6.4.1

### What version of PHP are you using?

8.1.13

### How did you install Timber?

Installed or updated Timber through Composer

Levdbas:
I tried this myself and can confirm this behavior. `PostsIterator` seems not to like null posts.

Levdbas:
Should be fixed by https://github.com/timber/timber/pull/2918

Issue #2917, opened 2024-02-08
[BUG] PostIterator breaks with post->setup(), if an element is null


rubas:
# Problem
Source: https://github.com/timber/timber/blob/2.x/src/PostsIterator.php#L49
```php
// Lazily instantiate a Timber\Post instance exactly once.

$post = $factory->from($wp_post);
$post->setup();
```

`$factory->from($wp_post);`  can return `null`, which breaks `$post->setup()`. 

This bug gets triggered, if you use a post collection with `null` in a loop.

# Solution
There needs to be a check for `null` or `$post?->setup()`.

## Real Example
![Screenshot 2024-02-08 at 15 37 37](https://github.com/timber/timber/assets/1042243/be99fcfa-bda4-4763-aad9-5dabdd35edbe)

### What version of Timber are you using?

2

### What version of WordPress are you using?

6.2

### What version of PHP are you using?

8.1

### How did you install Timber?

Installed or updated Timber through Composer

PR #2918, opened 2024-02-08
Fix PostIterator, if post is null

rubas:
Related: #2917 

## Issue
`$post` can be `null`, which breaks the code.

### DIFF
--- a/src/PostsIterator.php
+++ b/src/PostsIterator.php
@@ -12,10 +12,10 @@ use Timber\Factory\PostFactory;
 class PostsIterator extends ArrayIterator
 {
     /**
-     * @var Post The last post that was returned by the iterator. Used
+     * @var null|Post The last post that was returned by the iterator. Used
      *                   to skip the logic in `current()`.
      */
-    protected null|Post $last_post;
+    protected $last_post;
 
     /**
      * Prepares the state before working on a post.
@@ -46,7 +46,10 @@ class PostsIterator extends ArrayIterator
 
         // Lazily instantiate a Timber\Post instance exactly once.
         $post = $factory->from($wp_post);
-        $post?->setup();
+
+        if ($post instanceof Post) {
+          $post->setup();
+        }
 
         $this->last_post = $post;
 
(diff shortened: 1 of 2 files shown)


nlemoine:
Thank you @rubas for this PR. 

We still support PHP 7.4 so `$post?->setup();` syntax will not work. Can you adapt your PR to match 7.4 requirements?

rubas:
`protected null|Post $last_post;`  

The support for union type was only introduced in PHP 8.0.

rubas:
The failed coding standard test is outside of this PR. I'm happy to push a fix...


```php
1) tests/test-timber-cache.php

    ---------- begin diff ----------
@@ -329,7 +329,7 @@
         $clear = $loader->clear_cache_timber(Timber\Loader::CACHE_OBJECT);
         $this->assertTrue($clear);
         $works = true;
-      
+
         if (isset($wp_object_cache->cache[Timber\Loader::CACHEGROUP])
             && !empty($wp_object_cache->cache[Timber\Loader::CACHEGROUP])) {
             $works = false;
    ----------- end diff -----------


Applied checkers:

 * PhpCsFixer\Fixer\Whitespace\NoWhitespaceInBlankLineFixer
```

Levdbas:
Thank you @rubas  for taking this on. This might fix https://github.com/timber/timber/issues/2859 as well?

nlemoine:
for coding standards, you just need to run `composer run cs:fix`. But you can leave the failing test file or fix it. We'll take care of it later.

rubas:
> Thank you @rubas for taking this on. This might fix #2859 as well?

Correct. This is the same bug.

src/PostsIterator.php
```diff
@@ -12,10 +12,10 @@
 class PostsIterator extends ArrayIterator
 {
     /**
-     * @var Post The last post that was returned by the iterator. Used
+     * @var null|Post The last post that was returned by the iterator. Used
      *                   to skip the logic in `current()`.
      */
-    protected Post $last_post;
+    protected $last_post;
```
nlemoine: You still use a nullable type here.
```suggestion
    protected ?Post $last_post;
```

nlemoine (changes requested):
Thanks @rubas. This is almost good for me.

Can you add the nullable type to `$last_post` and a test case for this?

nlemoine (approved):
This is all good, thanks @rubas!

PR #2923, opened 2024-02-18
2873 bug cannot use timberpostfactory when extending post class as the method is private

### LINKED ISSUE #2873: [BUG] Cannot use Timber\Post::factory when extending Post class as the method is private
### Expected Behavior

When extending a Timber\Post, I should be able to access the factory() method. 

I'm trying to override the children() method which needs to have access to the factory.

### Actual behavior

Timber\Post::factory() is private and cannot be accessed by inhertitance.

1. Create a class that extends Timber\Post
2. Try to access the factory method
3. It returns false instead of a PostFactory object.

### Notes

_No response_

Timber 2.0

WordPress 6.4.2

PHP 8.1

Levdbas:
Related:

- #2873

## Issue
The Timber\Post::factory() is private and thus cannot be overwritten. So there was no way to alter the query to get the children.

## Solution
Added two new ways to alter the query:

1. added a filter to change the arguments
2. Add a new functionality to overload the children() parameter with query arguments.

I came up with the second approach later because that might be easier for most people. You could just change the arguments in your `children` call or make a separate children method in your custom class that calls the parent::children method with these additional arguments.

## Impact
No negative impact since loading the argument with a string or a non-indexed array still works.

## Usage Changes
No

## Considerations
Maybe one of these methods is enough. Not sure if adding the full argument list is a bit too much for this method. 

## Testing
Yes! First unit test I've ever written 👯

### DIFF
--- a/src/Post.php
+++ b/src/Post.php
@@ -994,11 +994,20 @@ class Post extends CoreEntity implements DatedInterface, Setupable
      *     {% endfor %}
      * {% endif %}
      * ```
-     * @param string|array $post_type _optional_ use to find children of a particular post type (attachment vs. page for example). You might want to restrict to certain types of children in case other stuff gets all mucked in there. You can use 'parent' to use the parent's post type or you can pass an array of post types.
+     * @param string|array $args _optional_ An array of arguments for the `get_children` function or a string/non-indexed array to use as the post type(s).
      * @return PostCollectionInterface
      */
-    public function children($post_type = 'any')
+    public function children($args = 'any')
     {
+        $post_type = 'any';
+        if (\is_string($args)) {
+            $post_type = $args;
+        } elseif (\array_values($args) === $args) {
+            $post_type = $args;
+        } elseif (\is_array($args)) {
+            $additional_args = $args;
+        }
+
         if ($post_type === 'parent') {
             $post_type = $this->post_type;
         }
@@ -1016,6 +1025,9 @@ class Post extends CoreEntity implements DatedInterface, Setupable
             $args['post_status'] = ['publish', 'inherit'];
         }
 
+        if (isset($additional_args)) {
+            $args = \array_merge($args, $additional_args);
+        }
         /**
          * Filters the arguments for the query used to get the children of a post.
          *
@@ -1028,7 +1040,7 @@ class Post extends CoreEntity implements DatedInterface, Setupable
          * add_filter( 'timber/post/children_args', function( $args, $post ) {
          *
          *     if ( $post->post_type === 'custom_post_type' ) {
-         *        $args['post_type'] = 'private';
+         *        $args['post_status'] = 'private';
          *     }
          *
          *     return $args;
@@ -1041,7 +1053,7 @@ class Post extends CoreEntity implements DatedInterface, Setupable
          * @param array        $arguments An array of arguments for the `get_children` function.
          * @param Post $post   The post object.
          */
-        $args = \apply_filters('timber/post/children/args', $args, $this);
+        $args = \apply_filters('timber/post/children_args', $args, $this);
 
         return $this->factory()->from(\get_children($args));
     }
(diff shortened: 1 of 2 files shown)


src/Post.php
```diff
@@ -994,23 +994,68 @@ public function category()
...
     {
+        $post_type = 'any';
+        if (\is_string($args)) {
+            $post_type = $args;
+        } elseif (\array_values($args) === $args) {
+            $post_type = $args;
+        } elseif (\is_array($args)) {
+            $additional_args = $args;
+        }
+
```
gchtr: Could we maybe rewrite this using `wp_parse_args()`?

```suggestion
		if (\is_string($args) || \array_values($args) === $args) {
			$args = [
				'post_type' => $args,
			];
		}

		$args = wp_parse_args($args, [
            'post_parent' => $this->ID,
            'post_type' => 'any',
            'posts_per_page' => -1,
            'orderby' => 'menu_order title',
            'order' => 'ASC',
            'post_status' => 'publish',
        ]);
```

Then, we wouldn’t need `$additional_args`.

src/Post.php
```diff
@@ -994,23 +994,68 @@ public function category()
...
-        $query = 'post_parent=' . $this->ID . '&post_type[]=' . $post_type . '&posts_per_page=-1&orderby=menu_order title&order=ASC&post_status[]=publish';
+
+        $args = [
+            'post_parent' => $this->ID,
+            'post_type' => $post_type,
+            'posts_per_page' => -1,
+            'orderby' => 'menu_order title',
+            'order' => 'ASC',
+            'post_status' => 'publish',
+        ];
```
gchtr:
```suggestion

```

src/Post.php
```diff
@@ -994,23 +994,68 @@ public function category()
...
+
         if ($this->post_status === 'publish') {
-            $query .= '&post_status[]=inherit';
+            $args['post_status'] = ['publish', 'inherit'];
+        }
+
+        if (isset($additional_args)) {
+            $args = \array_merge($args, $additional_args);
         }
```
gchtr:
```suggestion

```

gchtr (changes requested):
Nice one, @Levdbas! Thanks for this.

I only have a suggestion about the args handling.

src/Post.php
Levdbas: Fixed this with some extra sauce to also get the parent status working

Levdbas:
@gchtr , this is now fixed!

src/Post.php
```diff
@@ -994,23 +994,54 @@ public function category()
...
+        /**
+         * Filters the arguments for the query used to get the children of a post.
+         *
+         * This filter is used by the `Timber\Post::children()` method. It allows you to modify the
+         * arguments for the `get_children` function. This way you can change the query to get the
+         * children of a post.
+         *
+         * @example
+         * ```
+         * add_filter( 'timber/post/children_args', function( $args, $post ) {
```
gchtr, line 1026:
```suggestion
         * add_filter( 'timber/post/children/query_args', function( $args, $post ) {
```

src/Post.php
```diff
@@ -994,23 +994,54 @@ public function category()
...
+         * } );
+         * ```
+         *
+         * @see   \Timber\Post::children()
+         * @since 2.1.0
+         *
+         * @param array        $arguments An array of arguments for the `get_children` function.
+         * @param Post $post   The post object.
+         */
+        $args = \apply_filters('timber/post/children_args', $args, $this);
```
gchtr, line 1042:
```suggestion
        $args = \apply_filters('timber/post/children/query_args', $args, $this);
```

gchtr (approved):
Hey @Levdbas, thanks for the updates!

I have only one small change that I didn’t check before. I would rename the filter name to `timber/post/children/query_args`. This way, it’s more clear what args are being filtered.

Levdbas:
@gchtr , applied and merged :)

gchtr:
@Levdbas Great, thanks! 🎉

PR #2925, opened 2024-02-19
Fix child theme twig files not bing loaded

Levdbas:
Related:

- #2843

## Issue
When setting `Timber::$dirname` in your parent theme to a non-default location and overwriting a file in your child theme in the same, non-default, location, the template is not getting overwritten.

This is because the location logic has been changed around this commit: https://github.com/timber/timber/commit/e0659ebbd9bb3b5cc48441d6d47c1650f8e93927

## Solution
Since we are now working with an array of namesspaces and locations, the logic used to exclude the Parent theme from the caller, was not working anymore. I therefore introduced a new parameter in `LocationManager::get_locations_caller` to exclude the theme over there when the second parameter is true.

## Impact
This functionality works again!

## Usage Changes
None

## Considerations
The whole logic of building the locations seems a bit bloated to be and certainly not easy to debug. Maybe this is something we can improve over time.

## Testing
Alright, the nasty part.

I have introduced to new themes with the files in a non-standard location. I will make some annotations in the code itself why the test looks how it does. There is probably room for improvements which I am very much open for.


tests/test-timber-loader.php
```diff
@@ -302,4 +302,24 @@ public function testTwigLoadsFirstTemplateWhenMultipleLocationsWithSameNamespace
...
     }
+
+    public function testTwigLoadsFromNotStandardDirectoryInChildTheme()
+    {
+        $this->assertFileExists(WP_CONTENT_DIR . '/themes/timber-test-theme-child-non-standard/style.css');
+        switch_theme('timber-test-theme-child-non-standard');
+        $parent_theme_dir = get_template_directory();
+
+        // Load parent theme functions.php specifically from this directory to fake the caller location.
+        require_once $parent_theme_dir . '/functions.php';
```
Levdbas, line 313: When placing the 

`Timber::$dirname = ['resources/views'];` in the test directly the tests pass without the fix. This has probably something to do with the caller location not being the same as the template location. So by requiring the functions.php from the parent theme over here, I mimic the code that led to the bug in the first place. This might be a bit nasty though.

tests/test-timber-loader.php
```diff
@@ -302,4 +302,24 @@ public function testTwigLoadsFirstTemplateWhenMultipleLocationsWithSameNamespace
...
+
+        $child_theme = get_stylesheet_directory_uri();
+        $this->assertEquals(WP_CONTENT_URL . '/themes/timber-test-theme-child-non-standard', $child_theme);
+        $context = [];
+        $str = Timber::compile('single.twig', $context);
+        $this->assertEquals('I am single.twig', trim($str));
+        switch_theme('default');
+
+        // Reset the Timber::$dirname to the default value.
+        Timber::$dirname = 'views';
```
Levdbas, line 323: And I here reset the location back to the default one or else the https://github.com/timber/timber/blob/2.x/tests/test-timber-parent-child.php test (that runs after this one) failed.

nlemoine (approved):
> The whole logic of building the locations seems a bit bloated to be and certainly not easy to debug. Maybe this is something we can improve over time.

The whole caller/location thing is something I don't really understand because Twig already provides all the needed things for templates overrides.

Anyway, that's a bigger topic we'll discuss later.

Issue #1787, opened 2018-09-12
Better documentation for escaping in Timber


gchtr:
Following up on #1557, I’d like to add two todos:

- [ ] Update all examples in DocBlocks and in Markdown documentation files with the proper `|e` filters where needed.
- [ ] Explicitly write it into the description of the `@return` tag when a function is already escaped. Example: `@return type This is my description. Escaped.`

People will copy-paste code from the documentation. At least they should copy-paste proper code. To document in the reference whether a function is escaped would hopefully help as well when checking whether escapers should be used.

This is quite a big task. I think it doesn’t necessarily has to be in version 2.0, but should be approached soon. It would make for a good contribution opportunity. There’s a [list of already escaped functions](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/206b30cd1051f36199f139bb11664bb1b75026e4/WordPress/Sniff.php#L174) in the WordPress Coding Standards that could possibly help with this.

PR #2933, opened 2024-02-25
feat: Add WP escapers via Twig filters

### LINKED ISSUE #1848: Idea: change escaping functions to use Twig filters
Currently, Timber implements custom WordPress escapers in the following way:

```
{{ my_content|e('wp_kses_post') }}
```

This works great when autoescaping is turned off. However, when autoescaping is enabled (which in my opinion should always be the case, but that's not relevant right now), Twig applies the default escaping strategy after applying the custom escaper, which leads to double escaping and wrong output.

To fix this, the escaper needs to be piped through the raw filter:

```
{{ my_content|e('wp_kses_post')|raw }}
```

This works, but feels wrong. Unfortunately, from my research there seems to be no way to mark custom escaper output as safe/raw. However, [it is possible to mark filter output as safe](https://twig.symfony.com/doc/2.x/advanced.html#automatic-escaping).

This means that it would be possible to implement the escapers we have in this way:

```
{{ my_content|wp_kses_post }}
```

To me, this better communicates intent and as a bonus is easier to write. Implementation would be trivial, and current escapers could be kept in place for backwards compatiblity. What do you think?

Note: this was [almost implemented](https://github.com/twigphp/Twig/pull/2441) in Twig itself but the PR was closed due to inactivity.

Levdbas:
Related:

- #1848
- #1787

## Issue
Looking at the old issues I found this particular one. With escaping getting ever more important I wanted to take a shot at this.
Escaping currently in Timber is a bit cumbersome since you have to do things like this:

```twig
{{ my_content|e('esc_attr') }}
```
or even twig when setting the default Twig escaper already on in the environment. The |raw filter makes sure that the string is not double escaped.

```twig
{{ my_content|e('wp_kses_post')|raw }}
```

## Solution
Make escaping functions available via filters. 

- in `get_timber_escaper_filters` I add most of the default escapers that come with WP
- in `add_timber_escaper_filters` I have a simular function as `add_timber_filters` but add the following options by default:
```php
options' => [
  " 'is_safe' => ['html'],
],
```

## Impact
A better way of applying escapers

## Usage Changes
So instead of running `{{ my_content|e('esc_attr') }}` we can now run {{ my_content|esc_attr }} and it does not matter if the global escaping is on or off.

## Considerations
I did not spend much time on the docs at this time, but this is needed. We also might want to deprecate the old way of adding filters in the future, so this would be the recommended way.

If we want to implement this, we also need to take a look at  #1787 on updating the existing escaping examples in the PHP files.

## Testing
Yes. there were no escaper tests at this time so I added them and run some of them with global escaping on to test #1848

### DIFF
--- a/src/Twig.php
+++ b/src/Twig.php
@@ -29,8 +29,8 @@ class Twig
 
         \add_filter('timber/twig', [$self, 'add_timber_functions']);
         \add_filter('timber/twig', [$self, 'add_timber_filters']);
+        \add_filter('timber/twig', [$self, 'add_timber_escaper_filters']);
         \add_filter('timber/twig', [$self, 'add_timber_escapers']);
-
         \add_filter('timber/loader/twig', [$self, 'set_defaults']);
     }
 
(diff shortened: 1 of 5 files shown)


src/Twig.php
```diff
@@ -427,6 +427,73 @@ public function get_timber_filters()
...
+            'esc_html' => [
+                'callable' => 'esc_html',
+            ],
+            'esc_js' => [
+                'callable' => 'esc_js',
+            ],
+        ];
+
+        /**
+         * Filters the filters that are added to Twig.
```
gchtr:
```suggestion
         * Filters the escaping filters that are added to Twig.
```

gchtr (commented):
Oh that’s great, @Levdbas. That’s a good step towards making escaping easier.

> I did not spend much time on the docs at this time, but this is needed.

What ideas did you have to improve the docs, apart from what’s stated in #1787?

> We also might want to deprecate the old way of adding filters in the future, so this would be the recommended way.

Definitely something we could do.

Levdbas:
@gchtr , I have updated te escaping guide with new examples and removed a escaping function that was not recommended for being used on the front-end. About the deprecation of the existing escapers, do we want to do that right away?

We could do it like this for all escaper functions/filters

```php
        $esc_url = function (Environment $env, $string) {
            Helper::deprecated(
                "{{ url|e('esc_url') }}",
                "{{ url|esc_url }}",
                '2.1.0'
            );
            return \esc_url($string);
        };
```

[over here ](https://github.com/timber/timber/blob/6ebe3ef4eb4c23c05d6caf193d63cfdef972de38/src/Twig.php#L541-L548)

src/Twig.php
```diff
@@ -427,6 +427,70 @@ public function get_timber_filters()
...
 
+    /**
+     * Get Timber default filters
+     *
+     * @return array Default Timber filters
+     */
+    public function get_timber_escaper_filters()
+    {
+        $escaper_filters = [
+            /* image filters */
```
gchtr:
```suggestion
```

gchtr (dismissed):
This looks good to me!

> About the deprecation of the existing escapers, do we want to do that right away?

No, this seems like a breaking change to me that would be something for 3.0.

gchtr (approved):

PR #2934, opened 2024-02-26
2.x Fix #2773 (send empty extension for image with no extension)

### LINKED ISSUE #2773: [BUG] Missing extension for image

gchtr:
Related:

- #2773
- #2774

## Issue

The pull request in #2774 was still for the 1.x branch.

## Solution

This pull request ports the changes from #2774 over to the 2.x branch. And it adds a test to make sure this works.

## Impact

No errors for image URLs without extensions.

## Usage Changes

None.

## Considerations

When no image extension is given, Timber will just assume it’s a JPG. If you know it’s another image type, there’s no way to pass that. The question is: should you be able to pass that as a parameter?

## Testing

Yes.

### DIFF
--- a/src/ImageHelper.php
+++ b/src/ImageHelper.php
@@ -619,7 +619,7 @@ class ImageHelper
         $parts = PathHelper::pathinfo($tmp);
         $result['subdir'] = ($parts['dirname'] === '/') ? '' : $parts['dirname'];
         $result['filename'] = $parts['filename'];
-        $result['extension'] = \strtolower($parts['extension']);
+        $result['extension'] = (isset($parts['extension']) ? \strtolower($parts['extension']) : '');
         $result['basename'] = $parts['basename'];
 
         return $result;
--- a/src/ImageHelper.php
+++ b/src/ImageHelper.php
@@ -457,8 +457,8 @@ class ImageHelper
         }
         $tmp = \download_url($file);
         \preg_match('/[^\?]+\.(jpe?g|jpe|gif|png)\b/i', $file, $matches);
+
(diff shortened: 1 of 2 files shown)


Levdbas (approved):

Levdbas:
`When no image extension is given, Timber will just assume it’s a JPG. If you know it’s another image type, there’s no way to pass that. The question is: should you be able to pass that as a parameter?`

I would in favor of only introducing that when the need arises. There might be other things coming up with that as well that could be taken on in one go? And looking at the whole #2866 issue, who knows what will happen with sideloading in the future.

gchtr:
> I would in favor of only introducing that when the need arises. There might be other things coming up with that as well that could be taken on in one go? And looking at the whole #2866 issue, who knows what will happen with sideloading in the future.

Yep, let’s see what we come up with when we work on the big image overhaul, before we jump into extending the current API.

PR #2940, opened 2024-03-01
ACFIntegration: Empty Datepickerfield returns "" instead of null

rubas:
# Problem

If you have an empty `DatePicker` field in ACF:

- `$post->raw_meta('empty_date_field')` returns `null`
- `$post->meta('empty_date_field')` returns `""`

# Solution

We return `null`, if `!$value` instead of `$value`.

### Alternative

Maybe it's more in the spirit of the ACF to return `false`, but 😕 
`null` though allows you to chain easily `$post->meta('empty_date_field')?->format('d. F Y')`.

### DIFF
--- a/src/Integration/AcfIntegration.php
+++ b/src/Integration/AcfIntegration.php
@@ -153,7 +153,7 @@ class AcfIntegration implements IntegrationInterface
     public static function transform_date_picker($value, $id, $field)
     {
         if (!$value) {
-            return $value;
+            return null;
         }
         return new DateTimeImmutable(\acf_format_date($value, 'Y-m-d H:i:s'));
     }


gchtr:
A note for when we look at this more closely:

We made a change in #2904 to [not change return values in meta to `null`](https://github.com/timber/timber/pull/2904/files#diff-c9842c12f24fb44c49dd0c107eb1dd451210bc39efa587b5c43758179c110a97L195-L199), because it could lead to problems if we over-optimize. The failing tests were on of these problems.

It could very well be that with the change in #2904, the result for `raw_meta()` and `meta()` doesn’t look different for date fields.

PR #2948, opened 2024-03-13
cli: update ramsey/composer-install to v3

Levdbas:
Related:

- #2941

## Issue
Dependabot missed this older version.

## Solution
Update that as well to use the latest version to get rid of the node 16 errors.

## Impact
Better tests!

## Usage Changes
none


nlemoine (approved):
👍 Thank you @Levdbas !

PR #2950, opened 2024-03-13
fix: fix minor codestyle issue in loader.php to make easy-coding-stan…

Levdbas:
## Issue
There was one [nasty CS issue](https://github.com/timber/timber/actions/runs/8268553717/job/22621705987?pr=2949) that popped up in every PR I tried to create the last few weeks.

## Solution
Run CS:fix and make the code style gods happy.


nlemoine (approved):
Thanks @Levdbas :)

PR #2951, opened 2024-03-14
chore: Add script descriptions to composer file

Levdbas:
## Issue
We are getting more and more commands in our composer file. For first time collaborators its must sometimes be hard to find out what all these commands are meant for.

## Solution
Use [scripts-description](https://getcomposer.org/doc/articles/scripts.md#custom-descriptions)

## Impact
Better understanding of what tests do.

## Usage Changes
No

## Considerations
commands are being grouped when running `composer list`. But the command `cs` for example is not being shown here because its more at the top of the list.

image

We could even better group commands by prefixing them all. But maybe just adding a description in the composer file for all to read is a perfect start.


nlemoine:
Good addition, thank you @Levdbas 👍

nlemoine (approved):

PR #2957, opened 2024-03-25
ci: Use draft release for release please

gchtr:
When creating a new release with Release Please, it would be released as is.

I’d rather have it released as a draft first, so we can polish the notes a bit and make use of the GitHub feature that auto-generates release notes.

Once we get used to Release Please and feel more confident, we could switch this back.


nlemoine (approved):
All good 👍

PR #2960, opened 2024-04-02
Fix problem with empty acf taxonomy relationship field loading all terms instead of none.

rubas:
## Problem

The function `transform_taxonomy` will load **ALL** terms with `Timber::get_terms([])`, if there is no value stored in an ACF relationship field.

```
  public static function transform_taxonomy($value, $id, $field)
  {
    if ('select' === $field['field_type'] || 'radio' === $field['field_type']) {
      return Timber::get_term((int) $value);
    }

    return Timber::get_terms((array) $value);
  }
```

## Solution

Check for empty and return false. 

## Alternative

Return `[]` instead of false.

### DIFF
--- a/src/Integration/AcfIntegration.php
+++ b/src/Integration/AcfIntegration.php
@@ -200,6 +200,9 @@ class AcfIntegration implements IntegrationInterface
      */
     public static function transform_taxonomy($value, $id, $field)
     {
+        if (empty($value)) {
+            return false;
+        }
         if ($field['field_type'] === 'select' || $field['field_type'] === 'radio') {
             return Timber::get_term((int) $value);
         }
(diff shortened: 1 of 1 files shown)


Levdbas:
Hi @rubas , nice find!

Levdbas (approved):

rubas:
> It would be nice if you could create a test for this as well. Are you in for that?

Sorry, I was pretty busy (and holidays) lately. I haven't look at your test setup and it will be a few weeks (June?) before I find the time to do so.

I guess I'll do with the next PR ;-)

Levdbas:
> do with the next PR ;-)

I took a more in depth dive in your PR and I think we can do without tests I think. Or else we would need to make tests for all three transform methods to check what the results are with an empty field.

gchtr:
Thanks for the fix, @rubas! 🙌

PR #2962, opened 2024-04-09
docs: update deprecated Timber::get_context() in code examples (v2)

jasalt:
Searched also for "get_context" mentions in other v2 docs files but these should be the last (accidental) ones for that.


Levdbas:
Hi @jasalt and welcome at Timber.  Thank you a lot for this contribution! Good docs are the basis for starting with Timber so every contribution helps!

Levdbas (approved):

PR #2963, opened 2024-04-10
feat: remove functionality that disabled updates via the dashboard fo…

Levdbas:
## Issue
Updates via the admin dashboard were disabled for a while. Since we only going to ship (security)fixes in the future for the 1.x branch when the need occurs, we don't need this logic anymore.

## Solution
Remove logic and update the wording a bit. (screenshot was made while faking the version)

afbeelding

## Impact
Easier updates for folks that still run Timber plugin version from version 1.23.1 and 1.25.x 

## Usage Changes
No


nlemoine (dismissed):
Thanks @Levdbas !

gchtr (dismissed):

PR #2964, opened 2024-04-10
1.x Fix tests failing with WordPress 6.4

gchtr:


gchtr:
Looks like tests only really fail with prefer-lowest: https://github.com/timber/timber/actions/runs/8631578378/job/23660294443.

PR #2363, opened 2020-11-05
Fix typo

19h47:
Remove unnecessary quotes in line 321: 

```php
	 * <a href="{{ item.link }}" {{ item.is_external or item.is_target_blank ? 'target="_blank"' }}">
```

PR #2969, opened 2024-04-10
Lint only PHP files

szepeviktor:
#2363 revealed PHP linter runs on a Markdown document.

### DIFF
--- a/.github/workflows/php-lint.yml
+++ b/.github/workflows/php-lint.yml
@@ -39,7 +39,9 @@ jobs:
 
       - name: Get changed files
         id: changed-files
-        uses: tj-actions/changed-files@v42
+        uses: tj-actions/changed-files@v44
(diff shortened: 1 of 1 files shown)


nlemoine:
Thank you @szepeviktor !

Can you apply that change at workflow root? (path:)

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

szepeviktor:
```
      - '**.php'
      - '**composer.json'
```
These are all right. Composer config and PHP file changes **trigger** the workflow but a PR could contain several other files, e.g. Markdown files.

nlemoine:
Whoops, you're right, I commented too fast ^^

Thanks

nlemoine (approved):

PR #2970, opened 2024-04-10
feat: upgrade Timber requirements (PHP 8.1/WP 6.0/Twig 3.5)

nlemoine:
- PHP >=8.1
- WP >=6.0
- Twig >= 3.5
- Remove symfony/polyfill-php80
- Update workflow matrix accordingly

## Issue

As previously discussed, Timber >=2.1 will drop support for PHP <= 8.1 and WP <= 6.0.

[Packagist stats](https://packagist.org/packages/timber/timber/php-stats#2) shows pretty wide 8.1+ adoption and we should encourage users to move forward.

## Considerations

Twig <= 3.5 has security issues, so I bumped directly to Twig 3.5:

Capture d’écran 2024-04-10 à 21 28 06

### DIFF
--- a/.github/workflows/php-unit-tests.yml
+++ b/.github/workflows/php-unit-tests.yml
@@ -47,7 +47,7 @@ jobs:
       fail-fast: false
       matrix:
         php: ['8.1', '8.2']
-        wp: ['latest', '6.1']
+        wp: ['latest', '6.2']
         multisite: ['0', '1']
         dependency-version: ['highest', 'lowest']
         webp: [false]
--- a/.github/workflows/php-unit-tests.yml
+++ b/.github/workflows/php-unit-tests.yml
@@ -46,21 +46,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: ['7.4', '8.0', '8.1', '8.2']
-        wp: ['latest', '5.3']
+        php: ['8.1', '8.2']
+        wp: ['latest', '6.0']
         multisite: ['0', '1']
         dependency-version: ['highest', 'lowest']
         webp: [false]
         coverage: [false]
         extensions: ['gd']
         experimental: [false]
-        exclude:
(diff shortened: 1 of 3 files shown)


szepeviktor:
Half of the WordPress ecosystem (core, plugins, themes) runs on PHP 7.4.
https://wordpress.org/about/stats/#php_versions
Dropping it may be too early.

mikelittle:
> Half of the WordPress ecosystem (core, plugins, themes) runs on PHP 7.4. https://wordpress.org/about/stats/#php_versions Dropping it may be too early.

I disagree. Those who insist on sticking with an out of date version of PHP can stick with Timber 1.x

szepeviktor:
All right. One can upgrade/fix core, the theme, and self-created plugins but how about other plugins?

rmens:
This is the type of discussion that will keep the ecosystem on PHP 7.4 indefinitely.

nlemoine:
> This is the type of discussion that will keep the ecosystem on PHP 7.4 indefinitely.

💯 Agreed! I may also add:
- Timber 2 is a library and now composer only, this means it will probably target more "advanced" users. WordPress stats ≠ Timber stats. I'm running all my projects on PHP 8.1 without any problems since a year or so. Some of them are even running PHP 8.2 smoothly.
- If the plugin/theme/package you're using crashes your site or fills the screen/logs with notices/warnings in debug mode on PHP 8+, this is a bad quality signal (at least to me) and actually a good hint that you may want to find a better candidate
- Finally, supporting wide version ranges requires a lot of time and work, I often found myself fixing tests that were passing on some version and not another. It makes testing and developing more complicated. Considering Timber stats, I think it's not worth it and we can't afford that extra effort, I'm sure users will understand.

gchtr (approved):

szepeviktor (approved):

Levdbas (approved):

PR #2972, opened 2024-04-20
Fix for infinite loops in post.excerpt in specific cases

### LINKED ISSUE #2041: post.preview causes infinite loops
## Expected behavior
Twig Template loaded via ServerSideRender renders as wanted.

## Actual behavior
The backend is stuck in an infinite loop. It never reaches to the end of the request until it crashes.

Create a React component with <ServerSideRender>. Create a block and render any Twig template with {{ post.preview }} in it.

Removing the filter `do_blocks` around the the_content filter solves it. https://github.com/timber/timber/blob/master/lib/PostPreview.php#L290
```
remove_filter('the_content', 'do_blocks', 9);
$text = $this->post->content();
add_filter('the_content', 'do_blocks', 9);
```

@jarednova Any ideas how we could solve this? Should the post preview even apply the_content filter? I think it shouldn't as I had many troubles with that in the past.

WordPress 5.2.2
PHP 7.2
Timber 1.10.1

composer

Levdbas:
Related:

- #2041

## Issue
When you render a post preview in a dynamic block this will cause an infinite loop. 

## Solution
When fetching the content from the excerpt method, run the `excerpt_remove_blocks ` function inside the `content()` method.

## Impact
Post previews will probably render faster since now we are not parsing blocks anymore.

## Usage Changes
I added a third parameter to the  `content()` method but this is optional.

## Considerations
Nothing I can think about right now.

## Testing
Test and additional assets added to register a block, render the block via a render callback and then run a post.preview inside it. if you remove the suggested changes in Post.php and run the test it will cause a timeout which shows up in PHPunit as a segmentation fault.

### DIFF
--- a/src/Post.php
+++ b/src/Post.php
@@ -1279,7 +1279,7 @@ class Post extends CoreEntity implements DatedInterface, Setupable
          */
         $remove_blocks = (bool) \apply_filters('timber/post/content/remove_blocks', $remove_blocks);
 
-        if ($remove_blocks && \function_exists('excerpt_remove_blocks')) {
+        if ($remove_blocks) {
             $content = \excerpt_remove_blocks($content);
         }
 
--- a/src/Post.php
+++ b/src/Post.php
@@ -1215,26 +1215,26 @@ class Post extends CoreEntity implements DatedInterface, Setupable
      *
      * @param int $page Optional. The page to show if the content of the post is split into multiple
      *                  pages. Read more about this in the [Pagination Guide](https://timber.github.io/docs/v2/guides/pagination/#paged-content-within-a-post). Default `0`.
-     * @param int $num_words Optional. The number of words to show. Default `-1` (show all).
+     * @param int $len Optional. The number of words to show. Default `-1` (show all).
      * @param bool $remove_blocks Optional. Whether to remove blocks. Defaults to false. True when called from the $post->excerpt() method.
      * @return string The content of the post.
      */
-    public function content($page = 0, $num_words = -1, $remove_blocks = false)
+    public function content($page = 0, $len = -1, $remove_blocks = false)
     {
-        if ($rd = $this->get_revised_data_from_method('content', [$page, $num_words])) {
+        if ($rd = $this->get_revised_data_from_method('content', [$page, $len])) {
             return $rd;
         }
         if ($form = $this->maybe_show_password_form()) {
             return $form;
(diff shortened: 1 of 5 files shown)


src/Post.php
```diff
@@ -1215,25 +1215,26 @@ protected function get_revised_data_from_method($method, $args = false)
...
      * @param int $page Optional. The page to show if the content of the post is split into multiple
      *                  pages. Read more about this in the [Pagination Guide](https://timber.github.io/docs/v2/guides/pagination/#paged-content-within-a-post). Default `0`.
-     *
-     * @return string
+     * @param int $num_words Optional. The number of words to show. Default `-1` (show all).
+     * @param bool $remove_blocks Optional. Whether to remove blocks. Defaults to false. True when called from the $post->excerpt() method.
+     * @return string The content of the post.
      */
-    public function content($page = 0, $len = -1)
+    public function content($page = 0, $num_words = -1, $remove_blocks = false)
```
gchtr: Just a hint that if we rename parameters, as of PHP 8.0, it should be considered a breaking change, see https://php.watch/versions/8.0/named-parameters and the [Named Parameters in Public APIs](https://php.watch/versions/8.0/named-parameters#:~:text=Named%20Parameters%20in%20Public%20APIs) in particular.

Levdbas: Gotcha! Just reverted to the old name.

Levdbas:
@gchtr , I reverted the parameters to the old ones.

gchtr (approved):
This looks good to me 👍.

I’m not a huge fan of adding the `$remove_blocks` flag as an additional parameter, though. But I can’t think of anything better at the moment.

@nlemoine Anything to add here?

Levdbas:
> This looks good to me 👍.
> 
> I’m not a huge fan of adding the `$remove_blocks` flag as an additional parameter, though. But I can’t think of anything better at the moment.
> 
> @nlemoine Anything to add here?

I had the same feeling, Ideally you would want to know what the caller is of the `content` method is, so you can apply the remove block logic that way. But no idea how to solve that. Duplication the content method is also not the way to go in my opinion.

nlemoine:
>  But no idea how to solve that

Using `debug_backtrace()` ;)

Sorry, I didn't dig that topic much, I trust you guys on this 👍

nlemoine (approved):

PR #2977, opened 2024-04-26
feat: Test Rector to upgrade code for PHP 8.1

nlemoine:
Just giving Rector a try :)


Levdbas (dismissed):

gchtr:
Just to understand this correctly: We would use this once we move to PHP 8.1 as a minimum required version?

Levdbas (dismissed):

Levdbas:
> would use this once we move to PHP 8.1 as a minimum required version?

I think we could run this from time to time when we have a new release to see if we introduced new code that could be modernized, right @nlemoine ?

nlemoine:
> I think we could run this from time to time when we have a new release to see if we introduced new code that could be modernized, right @nlemoine ?

Actually, since there is a rector task for GrumPHP, we should run this at every commit, avoiding the need for "modernizing" the code base from time to time.

.github/workflows/php-rector.yml
```diff
@@ -0,0 +1,43 @@
...
+  push:
+    branches:
+      - master
+      - 2.x
+    paths:
+      - '**.php'
+      - '**composer.json'
+  pull_request:
+    branches:
+      - master
```
Levdbas, line 13:
```suggestion
```
Here as well

.github/workflows/php-rector.yml
```diff
@@ -0,0 +1,43 @@
+name: PHP rector
+
+on:
+  push:
+    branches:
+      - master
```
Levdbas, line 6:
```suggestion
```

I think we can remove master?

.github/workflows/php-rector.yml
nlemoine: I saw that too but it's in every workflow. I'll make a separate PR for this.

src/FunctionWrapper.php
```diff
@@ -35,11 +32,14 @@ public function __toString()
...
-     * @param array   $args
-     * @param bool    $return_output_buffer
+     * @param array $_args
+     * @param bool $_use_ob
      */
-    public function __construct($function, $args = [], $return_output_buffer = false)
-    {
+    public function __construct(
+        $function,
+        private $_args = [],
```
gchtr: Should we rename the properties to match the parameter names? Otherwise, this would also be a breaking change because PHP 8 has named parameter names.

nlemoine: I don't get how this changed. You're right! I renamed the parameters correctly.

gchtr (approved):

PR #2998, opened 2024-05-13
fix: fix regression where crops with the default crop setting would s…

### LINKED ISSUE #2986: [BUG] Resized images look wrong
### Expected Behavior

Resized images to not look squashed or stretched.

### Actual behavior

Resized images get squashed or stretched.

1. Install WP 6.5
2. Install the Timber Starter Theme
3. Make a post with this image as thumbnail as example https://www.pexels.com/nl-nl/foto/zwart-en-wit-zwart-wit-vrouw-mevrouw-19501709/
4. Publish the post
5. Look at the post and you will see that the image looks very squashed or stretched

Scherm­afbeelding 2024-05-03 om 22 35 12
Scherm­afbeelding 2024-05-03 om 22 34 54

### Notes

Also happens in a theme that's not the Timber starter theme. To rule out any user error, I tested with the starter theme.

Timber 2.1.0

WordPress 6.5.2

PHP 8.2

Levdbas:
Related:

- #2986

## Issue
In https://github.com/timber/timber/commit/18bf3586322a7bf829d5a246449a2c794f1e7862 a bug was introduced where the default crop parameter would lead to squeezing the image inside a confined space. instead of cropping into the image.

## Solution
Revert that if statement to the way it was in 1.x

## Impact
Working crops!

## Usage Changes
No.

## Considerations
[See ](https://github.com/timber/timber/issues/2866)

## Testing
no.

### DIFF
--- a/src/Image/Operation/Resize.php
+++ b/src/Image/Operation/Resize.php
@@ -210,12 +210,12 @@ class Resize extends ImageOperation
 
             $crop = $this->get_target_sizes($image);
             $image->crop(
-                $crop['x'],
-                $crop['y'],
-                $crop['src_w'],
-                $crop['src_h'],
-                $crop['target_w'],
-                $crop['target_h']
+                (int) $crop['x'],
+                (int) $crop['y'],
+                (int) $crop['src_w'],
+                (int) $crop['src_h'],
+                (int) $crop['target_w'],
+                (int) $crop['target_h']
             );
             $quality = \apply_filters('wp_editor_set_quality', 82, 'image/jpeg');
             $image->set_quality($quality);
--- a/src/Image/Operation/Resize.php
+++ b/src/Image/Operation/Resize.php
@@ -113,7 +113,7 @@ class Resize extends ImageOperation
             $w = \round($h * $src_ratio);
         }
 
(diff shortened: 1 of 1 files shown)


Levdbas:
@nlemoine , could you help me with this? The tests are false positives from what I can see from local tests. They just run fine. How to proceed?

Levdbas:
@timber/rangers , only one error remains in the testExternalImageWithExternalUrlAndNoImageExtension test. @gchtr , you introduced that test a few months ago. Could you shine your light on why path would be empty? Running the tests locally works just fine.

Edit: I see this PR runs into the same issues: https://github.com/timber/timber/actions/runs/9073988724/job/24932007649?pr=2999

gchtr:
@Levdbas I fixed it! Just joking. I had no idea why the error happened so I just re-ran the failing tests. It looks like there was some issue in the containers used for the tests.

I also ran into this issue on a site I’m working on, and I can confirm that this fixes the issue.

Just a note for anyone running into this and wondering why an image doesn’t look right after the Update: Images sizes need to be deleted and generated again for the fix to be applied to the image.

gchtr (approved):

PR #3001, opened 2024-05-15
fix: Update admin notice for minimum required WordPress version

gchtr:
Related:

- #2970

## Issue

In #2970, we update the minimum required WordPress version to 6.0.0. But we forgot a check done in the WordPress admin.

## Solution

Update the check.

## Impact

Better error messages.

## Usage Changes

None.

## Considerations

None.

## Testing

No.


gchtr:
The tests are failing because we test for the function to return `true` in https://github.com/timber/timber/blob/2.x/tests/test-timber-admin.php.

So I’m wondering, do we need that test? It doesn’t actually do a lot. If there’s a PHP error, it would show up in our tests. So I would just remove that test class.

src/Admin.php
```diff
@@ -9,13 +9,14 @@ class Admin
...
         global $wp_version;
-        if (\version_compare('5.3.0', $wp_version) === 1) {
-            // user is running something older that WordPress 5.3 show them an error
+        if (\version_compare('6.0.0', $wp_version) === 1) {
             $upgrade_url = \admin_url('update-core.php');
-            self::show_notice("<a href='https://github.com/timber/timber'>Timber 2.0</a> requires <strong>WordPress 5.3</strong> or greater, but you are running <strong>WordPress $wp_version</strong>. Please <a href='$upgrade_url'>upgrade WordPress</a> in order to use Timber 2.0.");
+
+            self::show_notice("<a href='https://github.com/timber/timber'>Timber 2.0</a> requires <strong>WordPress 5.3</strong> or greater, but you are running <strong>WordPress $wp_version</strong>. Please <a href='$upgrade_url'>update WordPress</a> in order to use Timber 2.0.");
         }
-        return true;
```
Levdbas, line 18: Why was the true removed here? Thats is why the tests fail atm.

src/Admin.php
```diff
@@ -9,13 +9,14 @@ class Admin
...
+        // Show notice if user is running something older than the required
+        // WordPress version.
         global $wp_version;
-        if (\version_compare('5.3.0', $wp_version) === 1) {
-            // user is running something older that WordPress 5.3 show them an error
+        if (\version_compare('6.0.0', $wp_version) === 1) {
             $upgrade_url = \admin_url('update-core.php');
-            self::show_notice("<a href='https://github.com/timber/timber'>Timber 2.0</a> requires <strong>WordPress 5.3</strong> or greater, but you are running <strong>WordPress $wp_version</strong>. Please <a href='$upgrade_url'>upgrade WordPress</a> in order to use Timber 2.0.");
+
+            self::show_notice("<a href='https://github.com/timber/timber'>Timber 2.0</a> requires <strong>WordPress 5.3</strong> or greater, but you are running <strong>WordPress $wp_version</strong>. Please <a href='$upgrade_url'>update WordPress</a> in order to use Timber 2.0.");
```
nlemoine: shouldn't it be 6.0?

Levdbas (approved):

src/Admin.php
gchtr: See https://github.com/timber/timber/pull/3001#issuecomment-2111667465

PR #3005, opened 2024-05-20
docs: Mark Helper::convert_wp_object as API method

### LINKED ISSUE #3004: [BUG] Missing Doc for `convert_wp_object` method
### Expected Behavior

Run into situation where i needed to convert the array of `WP_Post` obj to Array of `Timber\Post` obj and was looking for simple solution. Didn't find any on Timber official documentation under 2.x. tag.
That's why i started to dig inside source code and found method under Helper Class `convert_wp_object`

Was the documentation about that method deliberately left out?

Since I'm i prototype stage, the solution is pretty cheap:
```php
$query_args = array(
    's'              => $search_query,
    'paged'          => $paged,
    'posts_per_page' => $posts_per_page,
    'post_type'      => $post_types,
    'post_status'    => 'publish',
    'orderby'        => 'relevance',
);

$wp_query = new WP_Query($query_args);
$wp_posts = $wp_query->posts;

foreach($wp_posts as $wp_post) {
  $timber_posts[] = Timber\Helper::convert_wp_object($wp_post);
}
$context['posts'] = $timber_posts;
```
Is there a more elegant way to achieve that?
Keep in mind that im using Algolia with following option: `Use Algolia with the native WordPress search template`
which hijacks the wp search feature and returns the array of `WP_Post` obj.

### Actual behavior

Serch official documentation and find the solution to the problem. For the problem in question: a way to convert WP_Post obj to Timber\Post obj :)

1.
2.
3.
...

### Notes

_No response_

Timbe 2.2.

Wordpress 6.5

PHP 8.2

Levdbas:
Related:

- #3004

## Issue
`Timber\Helper::convert_wp_object` is not listed as an API function

## Solution
Mark as API function and rebuild docs.

## Impact
More functions documentated

## Usage Changes
No

## Considerations
When taking a look at the Helper methods, they are sparsely documented. We could improve on that.

When trying to commit, `Rector` was being a pain in the butt since it wanted to remove the mixed parameters from methods, including our API functions.

@nlemoine , could we remove that conversion like so? I can make a separate PR for that. This would make the life of of other contributors a lot easier.

image

## Testing
already tested

### DIFF
--- a/src/Helper.php
+++ b/src/Helper.php
@@ -282,7 +282,6 @@ class Helper
      * Output a value (string, array, object, etc.) to the error log
      *
      * @api
-     * @param mixed $error The error that you want to error_log().
      * @return void
      */
     public static function error_log(mixed $error)
@@ -707,7 +706,8 @@ class Helper
      *
      * If no match is found the function will return the initial argument.
      *
-     * @param mixed $obj WP Object
+     * @api
+     * @param mixed $obj WP Object to convert
      * @return mixed Instance of equivalent Timber object, or the argument if no match is found
      */
     public static function convert_wp_object(mixed $obj)


gchtr (approved):
The PR looks good, thanks!

I’ll let @nlemoine think about the Rector issue (once he’s back from summer holidays).

PR #3006, opened 2024-05-24
chore: add release checklist template for 2.x

Levdbas:
## Issue
With the last few releases we forgot to update version numbers in the codebase.

## Solution
Introduce a release checklist for 2.x so we do not forget these steps.

## Impact
Less confusion after releases.


nlemoine:
I think the `Timber::$version` can also be handled by Release Please, we just need to [mark it with comment tags](https://github.com/googleapis/release-please/blob/main/docs/customizing.md#updating-arbitrary-files):
```php
// x-release-please-start-version
public static $version = '2.0.0';
// x-release-please-end-version
```

PR #3010, opened 2024-06-07
ci: Add cs2pr for ECS which now supports checktyle format

nlemoine:
See: https://github.com/easy-coding-standard/easy-coding-standard/pull/204

We will now get nice PR feedback on coding standards :)


gchtr (approved):
🙌

PR #3013, opened 2024-06-12
Fix typo on Users guide docs

heybran:
A fix to update `Timber::get_user()` to `Timber::get_users()` located here: https://timber.github.io/docs/v2/guides/users/#querying-users


gchtr (approved):
Thanks for caring about these details! 🙌

Issue #3015, opened 2024-06-18
Timber 2 returns null when trying to use "get_image" on an AVIF image


arlg:
### Expected Behavior

`Timber::get_image($id)` should return image data. It insteads returns `NULL` when the image is an image/avif type.

### Actual behavior

Should return image data.

### Steps to reproduce behavior

Fill an AVIF image on back-end. Then try to get the image with get_image on front -> The function returns null.

**If i use the outdated method from Timber 1.0 "Image" (ex: Image(image_id)), it works.**

### What version of Timber are you using?

TImber 2.0

### What version of WordPress are you using?

WordPress 6.2

### What version of PHP are you using?

Php 8.2

### How did you install Timber?

Installed or updated Timber through Composer

Levdbas:
Hi @arlg ,

In postfactory there is an filterable array where you can add additional image formats.

You could add your own by adding the following filter:

```php
add_filter('timber/post/image_extensions', function ($extensions) {
	$extensions[] = 'avif';
	return $extensions;
});

```
We could add `avif` as a newly supported format I think, @timber/rangers , thoughts?

arlg:
Thanks for your answer, i did not know about this filters and it makes AVIF images work !

I also think that adding this by default in Timber is a good idea : WordPress supports AVIF format and browses now all supports it.

Levdbas:
Hi @arlg ,

If WP already supports AVIF I totally agree with you. I wasn't aware that the support is already there.
If you would like to take a shot at a PR to include AVIF formats, feel free to do so!

PR #3019, opened 2024-07-02
feat: Add support for avif image format #3015

arlg:
Related: #3015 

## Issue
Now that WordPress widely supports Avif Format Twig should add support to the avif image format.
Browsers also widely support this image format : https://caniuse.com/avif

## Solution
Adds avif support in the extension list.


gchtr (approved):
Looking good, thanks for the fix! 😎

Issue #2944, opened 2024-03-06
Add Twig renderBlock


JacobPrice:
## Is your feature request related to a problem? Please describe.
Currently, Timber's render method is designed to render entire templates. While this works well for most use cases, there are scenarios, especially in dynamic and hypermedia-driven applications (e.g., those using [HTMX](https://htmx.org/)), where rendering only a specific part of a template (a block or fragment) would be more efficient and effective. This is particularly relevant when only a small portion of the page needs to be updated in response to an action, avoiding the overhead of rendering and sending the entire template over the network.

## Describe the solution you’d like
I propose adding a new method, ``render_block``. This method would allow developers to render specific blocks within a Twig template, without needing to render the entire template.

A high-level overview of this would look something like this:
```php
    /**
     * Renders a specific block within a Twig template.
     * 
     * @param string $file Path to the Twig template.
     * @param string $block_name Name of the block to render.
     * @param array $data Data to be passed to the block for rendering.
     * @return string|false Rendered content of the block or false on failure.
     */
    public function render_block(string $file, string $block_name, array $data)
    {
        try {
            $twig = $this->get_twig();
            $template = $twig->load($file);

            if (!$template->hasBlock($block_name)) {
                // Handle the error as needed (log, throw exception, etc.)
                return false;
            }

            $output = $template->renderBlock($block_name, $data);

            return $output;
        }
        catch (\Exception $e) {
            // Handle exceptions as needed
            return false;
        }
    }
```
An additional argument ``$via_block_render`` could be added to ``Timber::compile()``  :
```php
public static function compile($filenames, $data = [], $expires = false, $cache_mode = Loader::CACHE_USE_DEFAULT, $via_render = false, $via_block_render = null)
```

Lastly, a conditional would be added to the output within the compile method:

```php
            if($via_block_render){
                $output = $loader->render_block($file, $via_block_render, $data);
            } else {
                $output = $loader->render($file, $data, $expires, $cache_mode);
            }
```

To make this accessible via the Timber class an additional method could be added called render_block:
```php
    /**
     * Renders a Twig block from a Twig file.
     *
     * Passes data to a Twig file and echoes the output of a specific block.
     *
     * @api
     * @example
     * ```php
     * $context = Timber::context();
     *
     * Timber::render_block( 'index.twig', 'content', $context );
     * ```
     * @param array|string   $filenames      Name or full path of the Twig file to render. If this is an array of file
     *                                       names or paths, Timber will render the first file that exists.
[...]

PR #3023, opened 2024-07-16
Add render_twig_block method

JacobPrice:
Related:

- #2944

## Issue
Twig has a built-in way to render blocks from a twig file instead of the entire file. Timber does not have this implemented.
[Twig for Developers - Documentation - Twig PHP](https://twig.symfony.com/doc/3.x/api.html)

## Solution
Add two new methods that leverage pre-existing methods.
- Timber::render_block
- Timber::compile_block

By adding an additional parameter called ``$block_name`` to the render method of the Loader class and to the compile function of the Timber class we can use the current compile function.

If a ``$block_name`` is passed and exists within the template provided, then twigs renderBlock is used.

## Impact
The primary goal of this change is to address a feature request as mentioned in #2944. 

This article highlights some of the general ideas that inspired this feature addition: [</> htmx ~ Template Fragments](https://htmx.org/essays/template-fragments/)

## Considerations

Potentially a seperate compile method could be used, but I am unsure of where this would be applicable past the current filters that are already in place for the current compile method.

It may be worth throwing an error instead of continuing if the block  is not found found in a template or returning false.

As discussed in #2944 it would be a good idea to add some documentation.

The two primary uses that I have found and would be benefiting from are in hypermedia systems (htmx) uses and for reducing files. 

#### Examples - 
HTMX/AJAX/FETCH:
Rendering post block in pagination instead of splitting this into a seperate file. 

File Reduction:
Instead of something like toast-success.twig, toast-fail.twig, toast-info.twig all of these could live in a toast.twig file and render each block from there. 

## Testing

Not included, but would be caught by other unit tests.


src/Loader.php
```diff
@@ -175,7 +176,12 @@ public function render($file, $data = null, $expires = false, $cache_mode = self
             );
 
             $template = $twig->load($file);
-            $output = $template->render($data);
+
+            if ($block_name && $template->hasBlock($block_name)) {
```
mrfsrf: Just came across this PR and wanted to let you know that `hasBlock` has been Deprecated since v3.12.  First thing that came to mind is to create helper function that parses the template source code and traverses the AST to achieve similar functionality.
https://github.com/twigphp/Twig/blob/4d19472d4ac1838e0b1f0e029ce1fa4040eb34ea/CHANGELOG#L8

Levdbas: Thank you for your review @mrfsrf . @JacobPrice , could you please take this in consideration as well?

mrfsrf: Hi all! I think i was to quick to jump the gun. Quick search trough Twig source, @JacobPrice  was/is using method from \TemplateWrapper class and not from the Parser class.

Levdbas:
Hi @JacobPrice ,

Love the addition. We discussed this during our 2.3.0 release meeting and we would like to add some guides on how to use this in the wild as well. This is a pretty big feature that could use some guidence on how/when to use this.

Maybe you and @jasalt can work on a guide on how to use this functionality together? Jarkko came up with some examples of using a similar use case via https://github.com/timber/timber/discussions/3044 and maybe that can be used in conjunction with your code?

jasalt:
Thank's for considering @levdbas. Looking at the render_block implementation, I could give it a go if @JacobPrice can provide some rough examples or instructions, relating to some common use case, how would you see that?

I could use it in some of my projects and supply tutorial/doc improvements, edit suggestions, or code examples that I might see generally useful. Practically I'm mostly interested in having some general HTMX user interface patterns working like endless scroll, active search, pop up carts for web store themes and such with Timber themes, and maybe also plugins.

I added an update just now on some further experiments I had with the same technique I mentioned at https://github.com/timber/timber/discussions/3044, related to multi-language production site project from which I extracted that initial example from while it was in progress. Quite happy already how it works with that plugin already but sure I would be happy to improve.

JacobPrice:
@Levdbas I will work on putting something together for this. It's a feature that would largely be a preference, as it allows for less code separation, which in some cases may be beneficial.

Another thing that may be worth adding would be a render_block twig function. That would allow for this function to be used inside of a twig file, which sounds odd, but in practice seems to feel nice. 

This is a simplistic example:
```twig
<!DOCTYPE html>
<html lang="en">
<body>
    {{ render_block(toast_type, 'toasts.twig', {message: toast_message}) }}
    <main>
    {# ... Content ... #}
    </main>
</body>
</html>
```

```twig
{# toasts.twig #}
{% set common_classes = 'top-0 left-0 w-full p-4' %}

{% block error %}
<div class="bg-red-500 text-red-50 font-bold {{ common_classes }}" role="alert">{{ message }}</div>
{% endblock %}

{% block warning %}
<div class="bg-yellow-500 text-yellow-50 {{ common_classes }}" role="alert">{{ message }}</div>
{% endblock %}

{% block success %}
<div class="bg-green-500 text-green-50 {{ common_classes }}" role="alert">{{ message }}</div>
{% endblock %}

{% block info %}
<div class="bg-blue-500 text-blue-50 {{ common_classes }}" role="alert">{{ message }}</div>
{% endblock %}
```

Levdbas:
Ah Gotcha @JacobPrice , this looks great. But with the changes made a render_block function in twig files is not available yet, right?

Things that, in my opinion, are still needed in this PR:

- [ ] Unit tests to test this new functionality, if you need help with that, let me know!
- [ ] further documentation inside the function doc block on how to use this. You can use [this example](https://github.com/timber/timber/blob/5434dde5889f174bf1d36c0686b94b180d93fa5d/src/Timber.php#L1260) to further enhance the documentation by including examples in multiple languages
- [ ] check if the function also works as intended inside twig files as you show in your example

gchtr:
Thanks for all your work on this so far. This is a great feature!

In addition to the points that @Levdbas listed, I would like to propose we switch the naming from "block" to "twig_block".

I feel like having methods named `render_block` or `compile_block` could lead to confusion with the methods that WordPress provides for the block editor. There’s a WordPress function called [`render_block()`](https://developer.wordpress.org/reference/functions/render_block/).

If people see code like the following in Twig, they might think it’s an implementation that renders a WordPress block, if they don’t know about Twig blocks yet.

```twig
{{ render_block(toast_type, 'toasts.twig', {message: toast_message}) }}
```

Yes, there’s no functionality to handle WordPress blocks through Twig templates in Timber, but there might be in the future.

I would like it much better if we could add the disambiguation and use `render_twig_block` instead:

```twig
{{ render_twig_block(toast_type, 'toasts.twig', {message: toast_message}) }}
```

I would use:

- `render_twig_block` instead of `render_block`
- `compile_twig_block` instead of `compile_block`
- Each time we write about Twig blocks, we should call them "Twig blocks" instead of just "blocks".

Levdbas:
Hi @JacobPrice , best wishes! We would like to include this function if you can take on some or all of the things mentioned in the two comments before this one. If you need any help with this, let me know!

Levdbas:
Hey @JacobPrice , @gchtr @jasalt ,

I found some time to iterate on this. I renamed the functions, added docblocks and added tests. I've also added the `render_twig_block` twig function. @gchtr , I had to write some custom loader logic here for reasons I don't completely understand. Can you take a look at this?

And everyone else, please take a good look at the changes as well. Preferably I would still like to add a guide on how to use this in the "wild". Is somebody to do that while also testing my code?

jasalt:
@Levdbas sure, I’ll experiment soon-ish and write something about it. Glad to see this moving forward.

Levdbas:
Added some documentation as well now in https://github.com/timber/timber/pull/3023/commits/f0f7236408d8bf420da269cf1038ccec7f924f5a. I haven't had the time to test the docs though. Will do that later or if somebody else is willing to do so, that would be nice!

gchtr (commented):
My gut feeling says that adding a `$block_name` parameter to the `Timber::compile()` and `Timber::render()` function is not the right solution for this. I feel that this would somehow bite us in the future when we need to incorporate other functionality.

Ideally, we would extract the logic from `render()` and `compile()` into a separate method that can be reused by `render_twig_block()` and `compile_twig_block()`.

And maybe we need to use a different loader that extends `Timber\Loader` instead of adding the logic and more parameters to the existing loader. Maybe we’ll have a TwigLoader and a TwigBlockLoader using a decorator pattern?

I’m not sure yet about the best way to do this.

jasalt:
Thinking about the final form of this might be wise.

Just noted that upcoming Django 6 includes this sort of functionality from django-template-partials to Django core, this could give some reference also.

- https://docs.djangoproject.com/en/dev/releases/6.0/#template-partials (release docs)
- https://code.djangoproject.com/search?q=template+partials&noquickjump=1&changeset=on&milestone=on&ticket=on&wiki=on (related issues)
- https://github.com/django/django/commit/5e06b970956af4854e970e74990cb971ba31c96b

jasalt:
Evaluating tools for a new Django project and with the Django 6 it looks like the trick there is that partial name is appended to template name that is passed to the same `render` function:

The template partial in `authors.html`:
```
{% partialdef user-info %}
    {# ... #}
{% endpartialdef user-info %}

... other template content continues ...
```
Rendering it with context:
```
render(request, "authors.html#user-info", {"user": user})
```

- https://docs.djangoproject.com/en/dev/ref/templates/language/#accessing-partials-directly

Levdbas:
Hey @gchtr , I found some time to continue working on this:

- Refactor to isolate block rendering via a dedicated loader:
- Added TwigBlockLoader that extends Loader and only overrides:
  - get_cache_key() to include the block name.
  - render_twig_template() to render a specific block (with graceful fallback to full template when the block doesn’t exist).
  - Avoided adding block-related parameters to Loader::render and Timber::compile/Timber::render.
- Updated the Twig function to delegate to Timber::compile_twig_block and pass LocationManager::get_locations() as the caller

nlemoine:
@Levdbas I rebased the branch onto 2.x and refactored the test to fit the new testing infrastructure. You should be good to keep going.

Levdbas:
> @Levdbas I rebased the branch onto 2.x and refactored the test to fit the new testing infrastructure. You should be good to keep going.

I don't have any other more changes in mind at this point, but I can take a look if I can cover these last few lines with unit tests as well @nlemoine

Levdbas:
Hey @JacobPrice , we saw you closed this PR but we still want to merge this in for 2.4. Would you mind reopening this PR. If not, we will open a new PR with your code if you don't mind.

PR #3034, opened 2024-09-01
docs: Fix transform_value filter with `__return_true`

Jasonlgrd:
Thanks for this package 🙏

I created this PR to just fix very little error in doc. I see it when I went copy/paste code from doc so I want fix that.

## Issue

The filter value contains too many `_`
https://developer.wordpress.org/reference/functions/__return_true/

## Solution

I have removed the underscore.

## Impact

Nothing.

## Usage Changes

Simplify copy/paste from doc.


gchtr (approved):
Great catch, thanks!

And thanks for taking the time to make a PR for this 🙌.

PR #3042, opened 2024-10-01
[DOCS] small typo in Update a-post-archive.md

SvenJuergens:
hi, it's only a small typo.
the UL is not closed after the loop.

best regards and many thanks for this Plugin :)


gchtr (approved):
Perfect, thanks for taking the time to fix this! 🙌

PR #3043, opened 2024-10-02
docs: Add `\Twig\` to TwigFilter and TwigFunction examples

jasalt:
Previous format was giving Class "TwigFunction" not found so I think this is how it should be written in docs, following starter-theme example on adding filters which I got working (https://github.com/timber/starter-theme/blob/c116aea47525f2412d54430088c81fa93b0f045b/src/StarterSite.php#L135).


gchtr (approved):
Very good, thanks! 🙌

PR #3045, opened 2024-10-07
test: Replace missing svn in GitHub Actions when installing WordPress

gchtr:
Tests have all failed recently. Apparently, GitHub has removed support for `svn` in GitHub Actions (https://github.blog/news-insights/product-news/sunsetting-subversion-support/).

This PR replaces the `svn` commands in **bin/install-wp-tests.sh** with other commands.

The only thing that doesn’t work now is when we use `latest` or `trunk` for the WordPress version. That’s why some tests are still failing.

Any idea on how to solve that without throwing a lot of code at the bash script?

### DIFF
--- a/.github/workflows/php-unit-tests.yml
+++ b/.github/workflows/php-unit-tests.yml
@@ -47,7 +47,7 @@ jobs:
       fail-fast: false
       matrix:
         php: ['8.1', '8.2']
-        wp: ['latest', '6.2']
+        wp: ['6.6.2', '6.2']
         multisite: ['0', '1']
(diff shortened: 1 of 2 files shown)


gchtr:
Ha! I missed we have `$LATEST_VERSION` in the script already! Fantastic, thanks!

Levdbas (approved):
Think we are good!

Levdbas:
> Ha! I missed we have `$LATEST_VERSION` in the script already! Fantastic, thanks!

Also set the depth to 1 for the clone process, really sped up the process.

PR #3052, opened 2024-10-14
test: update external image provider to a more reliable external service

Levdbas:
## Issue
External image test results in server errors. 

## Solution
Use another provider to fix these failing tests: https://github.com/timber/timber/actions/runs/11327013914/job/31497171429

## Impact
Better running tests

## Usage Changes
none


gchtr (approved):
This seems to be an endless story 🙈, but if it works for now, then fine!

PR #3061, opened 2024-11-15
docs: Add wp-i18n-twig to Internationalization docs

nlemoine:
Add our WP-CLI package to the docs :)

### DIFF
--- a/docs/v2/guides/internationalization.md
+++ b/docs/v2/guides/internationalization.md
@@ -25,13 +25,17 @@ The functions `_e()` and `_ex()` are also supported, but you probably won’t ne
 **WordPress:**
 
 ```html
-<p class="entry-meta"><?php _e( 'Posted on', 'my-text-domain' ) ?> [...]</p>
+<p class="entry-meta">
+    <?php _e( 'Posted on', 'my-text-domain' ) ?> [...]
+</p>
 ```
 
 **Timber:**
 
 ```twig
-<p class="entry-meta">{{ __('Posted on', 'my-text-domain') }} [...]</p>
+<p class="entry-meta">
+    {{ __('Posted on', 'my-text-domain') }} [...]
+</p>
 ```
 
 ### sprintf notation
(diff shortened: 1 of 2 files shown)


docs/v2/guides/internationalization.md
```diff
@@ -41,26 +45,41 @@ You can use sprintf-type placeholders, using the `format` filter:
...
 
 ## Generating localization files
 
-To generate `.pot`, `.po` and `.mo` files, you need a tool that supports parsing Twig files to detect all your translations. While there are a lot of tools that can parse PHP files, the solution that works best for Twig files is [Poedit](https://poedit.net/).
+### wp-i18n-twig (recommended)
 
-### Generating l10n files with Poedit 2
+The Timber team provides a [dedicated WP-CLI package](https://github.com/timber/wp-i18n-twig) that takes care of extracting translations from Twig files.
+
+This main benefit over other tools is that it perfectly integrates with the existing `wp i18n` command, making `.pot` file generation simple and easy.
```
Levdbas:
```suggestion
The main benefit over other tools is that it perfectly integrates with the existing `wp i18n make-pot` command, making `.pot` file generation simple and easy.
```

docs/v2/guides/internationalization.md
```diff
@@ -41,26 +45,41 @@ You can use sprintf-type placeholders, using the `format` filter:
...
+
+Note that the package isn't Timber specific, it will work with any Twig implentation making use of WordPress translation functions.
```
Levdbas:
```suggestion
Note that the package isn't Timber specific, it will work with any Twig implementation that also provides WordPress translation functions.
```

docs/v2/guides/internationalization.md
```diff
@@ -41,26 +45,41 @@ You can use sprintf-type placeholders, using the `format` filter:
...
+
+### Alternatives
+
+You can aslo generate `.pot`, `.po` and `.mo` files, with [Poedit](https://poedit.net/).
```
Levdbas:
```suggestion
You can also generate `.pot`, `.po` and `.mo` files, with [Poedit](https://poedit.net/).
```

Levdbas (commented):
Some minor fixes @nlemoine !

nlemoine:
Thanks for your fixes @Levdbas! I guess I wrote this too late 😅 (and you're a far better english writer than I ;))

nlemoine:
Just wondering, maybe we could remove those old extraction methods mentioned in the docs (python parser, Twig Gettext Extractor)? Those are a bit outdated and brings more confusion. It may be simpler to directly point out the current best option. We all use `wp i18n make-pot` for a while and it works pretty good so far (far better than my previous gulp based workflow!).

gchtr:
> Just wondering, maybe we could remove those old extraction methods mentioned in the docs (python parser, Twig Gettext Extractor)?

Oh yes, I’m all for this. I could never manage to get the python parser or the gettext extractor working. Poedit Pro was always the only thing that worked for me.

I would keep the solution with Poedit 2 in there, because it’s an approach that has worked really well over the years. But I would kick out the other solutions.

nlemoine:
@Levdbas What do you think?

Levdbas:
All in for some housekeeping here!

nlemoine:
@timber/rangers I removed all the outdated parts and updated the docs for v1 too since nothing really changed.

Let me know if this needs more changes or feel free to edit if needed :)

Levdbas (approved):

PR #3064, opened 2024-11-20
Fix: Use correct deprecation_info for Twig callables

### LINKED ISSUE #3069: Deprecation of use of the "deprecated", "deprecating_package", and "alternative" options
After updating my composer packages, I'm hitting deprecation issue: `Since twig/twig 3.15: Using the "deprecated", "deprecating_package", and "alternative" options is deprecated, pass a "deprecation_info" one instead.`

Firing from `.../vendor/symfony/deprecation-contracts/function.php` line ~25

_Using:_
Timber 2.3.0
Twig v3.16.0
Wordpress 6.7.1
PHP 8.2.24

mrfsrf:
## Issue

Twig v.3.15. introduced some nice features. Testing it with Timber v.2.x. works fine. However, there is this annoying deprecation warning:
```
'twig/twig', '3.15', 'Using the "deprecated", "deprecating_package", and "alternative" options is deprecated, pass a "deprecation_info" one instead.')
```

`$options['deprecated']` is deprecated since Twig 3.15. 
From Changelog:
[ * Improve the way one can deprecate a Twig callable (use `deprecation_info` instead of the other callable options)
](https://github.com/twigphp/Twig/blob/709fc6abcef276d40581be663c120307ccf0f2ce/CHANGELOG#L41) 
https://github.com/twigphp/Twig/blob/709fc6abcef276d40581be663c120307ccf0f2ce/src/AbstractTwigCallable.php#L46

## Solution

updates Timber to use the new deprecation_info option introduced in Twig 3.15 for handling deprecated features.
Question:
Why was the deprecated option used to signal Twig in the first place when Timber already provides the Helper::deprecated() function to handle deprecation warnings? This seems redundant, as Helper::deprecated() outputs its own warning messages. Or am I missing something?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,7 +44,7 @@
   },
   "require": {
     "php": "^8.1",
-    "twig/twig": "^3.5"
+    "twig/twig": "^3.16"
   },
   "require-dev": {
     "ergebnis/composer-normalize": "^2.28",
--- a/src/Twig.php
+++ b/src/Twig.php
@@ -14,6 +14,7 @@ use Twig\Extension\EscaperExtension;
 use Twig\Runtime\EscaperRuntime;
 use Twig\TwigFilter;
 use Twig\TwigFunction;
+use Twig\DeprecatedCallableInfo;
 
 /**
  * Class Twig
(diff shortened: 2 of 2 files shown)


Levdbas:
Hi @mrfsrf ,

Thank you for this, since we are currently testing for as low as Twig 3.5.0 the tests are failing. @timber/rangers , should we bump Twig to this latest version for an upcoming release or do we want fallbacks in this PR to also support lower versions? Bumping would be the way to go in my opinion.

gchtr:
I agree to bump the required Twig version. I don’t see a reason why we should keep backwards compatibility for a change like this.

nlemoine:
> Bumping would be the way to go in my opinion.

Same here!

Levdbas:
@gchtr , @nlemoine , added a commit that bumped the minimal version of Twig.

gchtr (approved):

PR #3070, opened 2024-12-03
3063 timberget menu0 returns alphabetically first menu instead of nothing

### LINKED ISSUE #3063: Timber::get_menu(0) returns alphabetically first menu instead of nothing
To reproduce on a fresh site:
1. Add a menu, which will have an ID that is greater than 0.
2. Use `Timber::get_menu(0)`. You'll get the menu.
3. Add another menu, making sure the slug/name comes after the first one alphabetically.
4. Use `Timber::get_menu(0)` again. You'll get the first menu again.

<details>

<summary>How I found this, if it matters</summary>

While resolving the issue I had that prompted #3062, I switched to using [`get_nav_menu_locations`](https://developer.wordpress.org/reference/functions/get_nav_menu_locations/) to add all menus into the global context. In doing that, I realized that menu locations with no assigned menu are assigned an ID of `0` when returned from that function.

(That function feels better than using `get_registered_nav_menus` [as suggested in the docs](https://timber.github.io/docs/v2/guides/menus/#set-up-all-menus-globally) since I don't have to run a function to see if a menu exists, I can just check for `0`? 🤷‍♂️)

</details>

Levdbas:
Related:

- #3063

## Issue
Menus are retrievable by `0`. It returns the alphabetically first menu. This is not what we want. 👯 

## Solution
In the case of using the get_menu method in combination with a  `0`, return `null`.

While at it, update the documentation as well with a new example.

## Impact
Less strange edge cases and a better example.

## Usage Changes
no

## Considerations
no

## Testing
no, should we?

### DIFF
--- a/docs/v2/guides/menus.md
+++ b/docs/v2/guides/menus.md
@@ -166,14 +166,9 @@ add_filter('timber/context', 'add_to_context');
 function add_to_context($context)
 {
     // Set all nav menus in context.
-    foreach (array_keys(get_registered_nav_menus()) as $location) {
-        // Set all nav menus in context.
-        foreach (array_keys(get_registered_nav_menus()) as $location) {
-            $menu = Timber::get_menu($location);
-            $context[$location] = $menu;
-        }
-
-        return $context;
+    foreach (array_keys(get_nav_menu_locations()) as $location) {
+        $menu = Timber::get_menu($location);
+        $context[$location] = $menu;
     }
 
     return $context;
--- a/docs/v2/guides/menus.md
+++ b/docs/v2/guides/menus.md
@@ -167,14 +167,13 @@ function add_to_context($context)
 {
     // Set all nav menus in context.
     foreach (array_keys(get_registered_nav_menus()) as $location) {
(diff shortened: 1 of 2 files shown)


docs/v2/guides/menus.md
```diff
@@ -167,14 +167,13 @@ function add_to_context($context)
 {
     // Set all nav menus in context.
     foreach (array_keys(get_registered_nav_menus()) as $location) {
-        // Bail out if menu has no location.
-        if (!has_nav_menu($location)) {
-            continue;
+        // Set all nav menus in context.
+        foreach (array_keys(get_registered_nav_menus()) as $location) {
```
gchtr: There’s something wrong here. We now have two nested foreach loops.

gchtr (changes requested):

docs/v2/guides/menus.md
Levdbas: Whoops, good spot @gchtr , last commit now does has the correct/tested code.

gchtr (approved):

PR #3071, opened 2024-12-03
test: Fix avatar test

Levdbas:
Fix pesky avatar test by not checking the whole URL but only the last part of the string by using `assertStringEndsWith`. Should be good enough then 👍


gchtr (approved):
Good idea!

PR #3073, opened 2024-12-15
fix: Fix bug with Attachment::path() method

### LINKED ISSUE #3072: Timber\Attachment::path() returns incorrect value
### Expected Behavior

`Timber\Attachment::path()` returns `/wp-content/uploads/file.png`

### Actual behavior

`Timber\Attachment::path()` returns `wp-contenthttps://example.com/wp-content/uploads/file.png`

```php
<?php

var_dump(\Timber\Timber::get_post(get_posts( [ 'post_type' => 'attachment', 'numberposts' => 1 ] )[0])->path())
```

### Notes

_No response_

Timber 2.3.0

Wordpress 6.7.1

PHP 8.2

Levdbas:
Related:

- #3072

Solution was provided by @mrfsrf

## Issue
When `'Timber\Attachment::path()`; is requested, the Path ends up being a mix of a path and a URL.

## Solution
If we are working with a URL, the convert the path first to a filesystem path before getting the relative path.

## Impact
A method that works again.

## Usage Changes
No.

## Considerations
Do we need to run the is_url check first since we are probably working with a URL anyway?

## Testing
Yes, test added.

### DIFF
--- a/src/Attachment.php
+++ b/src/Attachment.php
@@ -125,10 +125,7 @@ class Attachment extends Post
     public function path(): string
     {
         $src = $this->src();
-        // If src is a URL, convert it to a filesystem path first.
-        if (URLHelper::is_url($src)) {
-            $src = URLHelper::url_to_file_system($src);
-        }
+        $src = URLHelper::url_to_file_system($src);
 
         return URLHelper::get_rel_path($src);
     }
--- a/src/Attachment.php
+++ b/src/Attachment.php
@@ -124,7 +124,13 @@ class Attachment extends Post
      */
     public function path(): string
     {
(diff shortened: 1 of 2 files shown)


src/Attachment.php
```diff
@@ -124,7 +124,13 @@ public function link()
      */
     public function path(): string
     {
-        return URLHelper::get_rel_path($this->src());
+        $src = $this->src();
+        // If src is a URL, convert it to a filesystem path first.
+        if (URLHelper::is_url($src)) {
```
gchtr: @Levdbas As you mentioned, this check is probably not needed, as `$src` should already be a URL.

gchtr (approved):

PR #3083, opened 2025-01-31
docs: incorrect anchor from upgrade 2.0 guide to get_posts

johannpinson:
Hello,

This is a simple doc fix.

## Issue
A link inside the upgrade v2 docs is incorrect, the anchor targets to the `get_post` instead of the `get_posts` function.

## Solution
Link updated inside the markdown doc.


Levdbas (approved):
Thanks for catching this @johannpinson !

Issue #2882, opened 2024-01-14
[BUG] Resized Webp images generated by Timber/ImageHelper::towebp and Timber/ImageHelper::resize filter are not deleted along original image


montagmorgens:
### Expected Behavior

When I delete an image from the WP media library, I expect that all derivative image files that were created by Timber image filters are deleted as well.

### Actual behavior

Images that were created by a combination of `resize` and `towebp` filters are not deleted from the file system.

### Steps to reproduce behavior

1. Create image in Twig: `image.src|towebp|resize(100,100)`
2. Delete the image from the WP media library

### What version of Timber are you using?

Timber 2.0

### What version of WordPress are you using?

WordPress 6.4.2

### What version of PHP are you using?

PHP 8.1

### How did you install Timber?

Installed or updated Timber through Composer

ghost:
This shouldn't be implemented before this is implemented, or the wrong images might be deleted: https://github.com/timber/timber/issues/2850

marcoluzi:
+1 from me for this issue. We had a user, that deleted an image, changed something on it and uploaded again with the same name. Obviously, in the frontend it displayed the old deleted image, since the thumbnail was still lingering around and was not regenerated.

PR #3086, opened 2025-02-11
Opt-In Advanced Deletion for Generated Image Variants (tojpg/towebp & scaled)

marcoluzi:
Related:

- #2882
- #2866
- #2556

## Issue
Some generated image sizes (especially those produced via the Timber filters `tojpg` and `towebp`, as well as files generated from scaled originals) are not being properly deleted when the original image is removed. This leaves behind stray files (e.g. `example-768x958-c-center-towebp.webp`) in the uploads folder.

## Solution
This PR refactors the deletion logic in `ImageHelper::delete_generated_files()` by integrating an advanced, broad glob-based deletion mode into the existing `process_delete_generated_files()` method. A new optional parameter `$advanced` has been added to that method. When the developer opts in via the filter:
```php
add_filter('timber/image/advanced_file_names', '__return_true');
```
The method will ignore the original fixed regex patterns and instead delete any file that starts with the original filename followed by a hyphen (e.g. matching the glob pattern example-*.jpg or example-*.webp). This also covers generated files from the unsclaed variant (when the original filename contains -scaled). The original deletion logic is preserved if the filter is not enabled.

## Impact
- **Backwards Compatibility:**
By default, if the filter is not enabled, the existing deletion logic remains unchanged.

- **Performance:**
The advanced mode uses glob patterns which are efficient for typical uploads directories. However, on directories with a very large number of files, there could be a slight performance impact.

## Usage Changes
To enable the advanced deletion logic (which covers all generated sizes including both JPG and WEBP variants and handles -scaled filenames), add the following filter in your theme’s or plugin’s initialization:
```php
add_filter('timber/image/advanced_file_names', '__return_true');
```
The conversion operations still append the -tojpg or -towebp suffix to generated filenames to avoid collisions with original files.

## Considerations
- The -towebp and -tojpg suffixes are still maintained. Even though the advanced deletion mode uses a glob pattern that deletes any file starting with the original base name and a hyphen, these prefixes help differentiate generated images from the originals and prevent accidental overwriting.

- Future improvements might consider a complete redesign of both naming and deletion logic to handle more edge cases. For now, this opt-in advanced mode provides a balance between backward compatibility and improved cleanup.

## Testing
Manual Testing Scenario (Pseudo-code):

1. Upload an image named example.jpg.
2. Generate various image sizes using Timber filters (e.g., resize, tojpg, towebp) so that files like:
        - example-150x150-c-default-towebp.webp
        - example-150x150-c-default.jpg
        - example-scaled.jpg
        - … and others are created.
3. Delete the original image.
4. Without the filter enabled, verify that some generated files may remain.
5. With the filter enabled via `add_filter('timber/image/advanced_file_names', '__return_true');` verify that all generated variants (files matching example-*.jpg and example-*.webp) are properly deleted.

No new unit tests are provided in this PR; manual testing has confirmed the intended behavior.

### DIFF
--- a/src/ImageHelper.php
+++ b/src/ImageHelper.php
@@ -321,25 +321,40 @@ class ImageHelper
             $local_file = URLHelper::url_to_file_system($local_file);
         }
         $info = PathHelper::pathinfo($local_file);
-        $dir = $info['dirname'];
+        $original_dir = $info['dirname'];
         $ext = $info['extension'];
         $filename = $info['filename'];
-        self::process_delete_generated_files($filename, $ext, $dir, '-[0-9999999]*', '-[0-9]*x[0-9]*-c-[a-z]*.');
-        self::process_delete_generated_files($filename, $ext, $dir, '-lbox-[0-9999999]*', '-lbox-[0-9]*x[0-9]*-[a-zA-Z0-9]*.');
-        self::process_delete_generated_files($filename, 'jpg', $dir, '-tojpg.*');
-        self::process_delete_generated_files($filename, 'jpg', $dir, '-tojpg-[0-9999999]*');
+
+        // Re-analyze the URL to see if a generated folder is configured.
+        $au = self::analyze_url($local_file);
+        $generated_folder = \apply_filters('timber/image/generated_folder', false, $au);
+        if ($generated_folder && $au['base'] == self::BASE_UPLOADS) {
+            // If the source has a year/month subfolder, preserve that structure.
+            if (\preg_match('#^/\d{4}/\d{2}$#', (string) $au['subdir'])) {
+                $dest_subdir = '/' . \trim($generated_folder, '/') . $au['subdir'];
+            } else {
+                $dest_subdir = '/' . \trim($generated_folder, '/');
+            }
+            $upload_dir = \wp_upload_dir();
+            $dest_dir = $upload_dir['basedir'] . $dest_subdir;
+        } else {
+            $dest_dir = $original_dir;
+        }
+
+        // Now run deletion logic in the destination directory.
+        self::process_delete_generated_files($filename, $ext, $dest_dir, '-[0-9999999]*', '-[0-9]*x[0-9]*-c-[a-z]*.');
+        self::process_delete_generated_files($filename, $ext, $dest_dir, '-lbox-[0-9999999]*', '-lbox-[0-9]*x[0-9]*-[a-zA-Z0-9]*.');
+        self::process_delete_generated_files($filename, 'jpg', $dest_dir, '-tojpg.*');
+        self::process_delete_generated_files($filename, 'jpg', $dest_dir, '-tojpg-[0-9999999]*');
 
         $advanced = \apply_filters('timber/image/advanced_file_names', false);
-
         if ($advanced) {
-            self::process_delete_generated_files_advanced($filename, 'jpg', $dir);
-            self::process_delete_generated_files_advanced($filename, 'webp', $dir);
-
-            // Delete the original generated files if it is a scaled image.
+            self::process_delete_generated_files_advanced($filename, 'jpg', $dest_dir);
+            self::process_delete_generated_files_advanced($filename, 'webp', $dest_dir);
             $filename_without_scaled = \str_replace('-scaled', '', $filename);
             if ($filename_without_scaled !== $filename) {
-                self::process_delete_generated_files_advanced($filename_without_scaled, 'jpg', $dir);
-                self::process_delete_generated_files_advanced($filename_without_scaled, 'webp', $dir);
+                self::process_delete_generated_files_advanced($filename_without_scaled, 'jpg', $dest_dir);
+                self::process_delete_generated_files_advanced($filename_without_scaled, 'webp', $dest_dir);
             }
         }
     }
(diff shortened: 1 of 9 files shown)


marcoluzi:
@gchtr As discussed in #2556, I have created a new PR, branched off from the current state of the 2.x branch. I'm not yet very experienced with contributing to open source projects or writing tests, so I'd gladly welcome any advice. :)

src/ImageHelper.php
```diff
@@ -328,19 +328,30 @@ public static function delete_generated_files($local_file)
...
+        $advanced = \apply_filters('timber/image/advanced_file_names', false);
+
+        if ($advanced) {
+            self::process_delete_generated_files_advanced($filename, 'jpg', $dir);
+            self::process_delete_generated_files_advanced($filename, 'webp', $dir);
+
+            // Delete the original generated files if it is a scaled image.
+            $filename_without_scaled = \str_replace('-scaled', '', $filename);
+            if ($filename_without_scaled !== $filename) {
+                self::process_delete_generated_files_advanced($filename_without_scaled, 'jpg', $dir);
```
gchtr: There’s an issue with this logic here: It will delete any generated image sizes, even sizes generated by WordPress itself as well as scaled images.

To check that this doesn’t happen, I added two new tests in https://github.com/timber/timber/pull/3086/files#diff-e1de6e719b8d6c994eacc2c8740cd69d7978d5a7131af6550a2fb3066a0d3434.

I don’t know yet how we best solve this, but I guess we will have to work with some kind of more advanced patterns that we check.

The hard part here is probably that file names can be different depending on the order of image filters are run in.

gchtr (commented):
Thanks for this PR, @marcoluzi! Your first PR, woohoo 🎉.

And thanks for the very detailed description. That surely helps when we continue working on this.

> I'm not yet very experienced with contributing to open source projects or writing tests, so I'd gladly welcome any advice. :)

We all start out like that, so no worries if you can’t provide any tests. These things take time to get into.

I updated the PR to use separate functions for the image deletion in [`7397ea3` (#3086)](https://github.com/timber/timber/pull/3086/commits/7397ea34659c8efbde6570504ccd66ae543cb9dd).

And I started adding some tests in [`2919ba6` (#3086)](https://github.com/timber/timber/pull/3086/commits/2919ba6133470798f8fb49d6841e6514fbcfac21). More tests might be needed that try different combinations of image filters.

We need to decide though how far we want to go with this, as you mentioned yourself:

> Future improvements might consider a complete redesign of both naming and deletion logic to handle more edge cases. For now, this opt-in advanced mode provides a balance between backward compatibility and improved cleanup.

src/ImageHelper.php
marcoluzi: Maybe I am not deep enough into the WordPress Image Logic. When we delete an image from the Media Library, don't we want all files linked to that Media to be deleted aswell? Weather it is generated by WordPress or Timber?

gchtr: @marcoluzi In general, we would want that yes. And I initially thought this is a good idea. But we would also take some assumptions, which could lead to issues that I found while reviewing in more detail.

Timber hooks into `wp_generate_attachment_metadata` to delete generated image files whenever image metadata is generated. This can be the case if you run plugin like [Regenerate Thumbnails](https://wordpress.org/plugins/regenerate-thumbnails/) or [TinyPNG](https://wordpress.org/plugins/tiny-compress-images/), which don’t delete the original image, but do something else with it.

So, what happens with the logic in this PR is:

1. You upload an image into the media library
2. WordPress generates image sizes
3. Timber deletes these image sizes right away, because `wp_generate_attachment_metadata` is also called when uploading an image, leaving only the original image

In addition to that, when a scaled version is generated, it would also delete the `-scaled.jpg` version of the image, even if that is still referenced in the database.

Another case I can think of: If you upload **pizza.jpg** and then upload **pizza-2.jpg** and **pizza-3.jpg**, these files would be deleted as well, even if they are different attachments, right?

I think we should use a naming pattern that we can target directly when deleting generated sizes and not delete images too broadly.

marcoluzi: @gchtr Could we include a `-timber-` Prefix in the file name or something like that to only target timber images in the deletion process? 🤔

svenvonarx: Imho a timber related suffix right after the original filename could solve these struggles with other plugins or for example the case of `pizza.jpg`, `pizza-1.jpg` etc.. Then you could also simplify the matching patterns for the images that have to be deleted.

gchtr: I would be careful to include `-timber-` in any generated name, because image file names are relevant for SEO. And it’s hard to assess the consequences of that. I would always choose more generic names.

In https://github.com/timber/timber/issues/2866, the idea popped up to store generated images in a separate folder than the folder with the original images:

> generated images are stored in a different directory than the source directory (uploads). I think it's important to clearly distinguish source images and manipulated images. Thus they can safely deleted and regenerated if something changed in the API.

This could mean we would use a folder like `wp-content/uploads/t-processed/`, with matching year/month subfolders if that option is selected. We could also provide a way to define the name of that folder.

Because we will probably follow that direction anyway, we might have to build the logic for that first now.

@Levdbas @nlemoine  What do you think?

svenvonarx: Otherwise I could also see that the filenames of the generated files are stored in the database as post meta, but if it can be avoided with a folder solution that would be great too.

Levdbas: A sub-folder in `uploads` that holds all the generated files sounds like the way to go. Do we want to put that behind a separate filter toggle?

gchtr: @marcoluzi Thanks for the update in [`5da5173` (#3086)](https://github.com/timber/timber/pull/3086/commits/5da51734b54315c6f5b1bd7c8b8d96a211a02d38). I’ll have a look at it soon to write the tests for this.

> Do we want to put that behind a separate filter toggle?

Yes, I would use a separate filter to define the folder name for this. And I see the logic for this is already there.

marcoluzi: @gchtr No need to check yet. It's just a WIP, where I am trying some things. I will re-request the review once i got a decent solution.

gchtr: Great, thanks!

marcoluzi:
Quick update from our side: we don’t currently have the resources to continue working on this, and as we’re gradually moving away from Timber, this isn’t a priority for us anymore.

I’ll close this PR for now, but please feel free to reopen it and continue the work if you’d like.

Issue #3080, opened 2025-01-09
MenuItem::is_external returns false positive for local paths


amustill:
### Expected Behavior

`MenuItem::is_external` should return `false` when using local paths in a custom link, eg. `/about/`.

### Actual behavior

`MenuItem::is_external` returns `true`.

### Steps to reproduce behavior

1. Create a menu and add a custom link using a local path as the URL, eg. `/about/`.
2. Check if the URL is external using the `is_external` function: `{{ dump(item.is_external) }}`

### What version of Timber are you using?

Timber 2.x

### How did you install Timber?

Installed or updated Timber through Composer

Levdbas:
Hi @amustill , thank you for your report. This seems to be related to this change: https://github.com/timber/timber/commit/c3afa18cde1d511085d814675a221e3334c2eede#diff-2484f91d742ea05a569b75d47cd670b0b6a547f6485de421558865178ac85df3L394

@nlemoine , it seems you removed the `type` method back then but I couldn't find a replacement. Any idea on what the best way to fix this is? Re-adding that method and update the check again in `is_external`?

amustill:
Thanks for the quick response @Levdbas.

It doesn't appear to be related to that specific change as the type check is still there and rightly checks the link type. The URL does not need to be tested if it's a Page, Category, etc. as these are always internal URLs, so this type check returns `false` if it's not a custom link.

What I should have noted in the bug report that it's  actually`URLHelper::is_external` that performs the check on the URL. However, does it make sense to update this method or the one in the `MenuItem` class?

Updating `URLHelper` could have an undesired impact elsewhere and you could argue that this bug is more related to the menu functionality rather than URL functionality.

amustill:
Sorry, I accidentally closed the issue with that last comment.

Levdbas:
Hi @amustill ,

Sory for the late reply. I think we can use `URLHelper::is_url` to check if its a URL we are checking. that method returns false on relative URLs. Are you open to do a PR for that?

```php
public function is_external()
    {
    if ($this->type !== 'custom') {
        return false;
    }
    
    // adds addition check for relative urls/non-urls
    if(false === URLHelper::is_url($this->link())){
        return false;
    }
    
    return URLHelper::is_external($this->link());
    }
```

PR #3089, opened 2025-02-22
fix: Fix MenuItem::is_external() returning false positives for relative URLs

amustill:
Related: #3080

## Issue
`MenuItem::is_external` currently returns `true` for relative paths in a custom link, eg. `/about/`.

## Solution
An additional check inside `MenuItem::is_external` has been added to return false for links failing a `URLHelper::is_url` check as proposed by @Levdbas [here](https://github.com/timber/timber/issues/3080#issuecomment-2663340923).

### DIFF
--- a/src/MenuItem.php
+++ b/src/MenuItem.php
@@ -415,6 +415,12 @@ class MenuItem extends CoreEntity implements Stringable
         if ($this->type !== 'custom') {
             return false;
         }
+
+        // Additional check for relative/non-URLs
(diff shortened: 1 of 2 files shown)


Levdbas:
Thanks @amustill , added some tests to make sure this keeps working in the future!

Levdbas:
@nlemoine , @gchtr , can you take a look at this PR?

gchtr (approved):
Looks good to me!

PR #3099, opened 2025-05-06
Fixing doc references to post.modified

sdunham:
## Issue
The v2 docs reference a `post.modified` method which doesn't seem to exist. Looks like it should be `post.modified_date` instead.

## Solution
Updated docs to use the correct method name.

## Impact
More accurate docs!

### DIFF
--- a/docs/v2/guides/date-time.md
+++ b/docs/v2/guides/date-time.md
@@ -181,11 +181,11 @@ The date a post was published is accessible through `{{ post.date }}`.
 {{ post.date }}
 ```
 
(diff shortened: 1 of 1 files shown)


gchtr:
Thanks a lot for taking the time to fix even small things like these, @sdunham! Very much appreciated!

I found some more small issues and fixed them in #3100, together with your commit. That’s why I’m going to close this PR.

sdunham:
@gchtr just checking, did the docs site ever get updated after #3100 was merged? I'm still seeing references to `.modified` on this page https://timber.github.io/docs/v2/guides/date-time/

PR #3115, opened 2025-06-05
test: Install svn to revert replacing svn with other methods when it’s missing in GitHub Actions

gchtr:
We replaced installing WordPress through SVN because SVN was removed from GitHub Actions. The easier way is to just install svn again.

This will make it easier to update install-wp-tests.sh in the future.


Levdbas (approved):
LGTM!

PR #3117, opened 2025-06-05
fix: Prevent PostsIterator from resetting the post global in the admin

gchtr:
## Issue

When calling `Timber::get_posts()` in the admin, Timber will call `$post->setup()` for all posts, which will cause the `$post` global to be overwritten. This intended when looping over posts in the frontend, but it shouldn’t be done in the WordPress admin. Because there it can mess up your posts.

For example, you could open a post in the block editor and the content of a different post is loaded in there.

## Solution

Add a check to only run the `setup()` in the frontend.

## Impact

Fewer side-effects.

## Usage Changes

None.

## Considerations

The `is_admin()` check might not be sufficient for all cases. But for the moment, it fixes the most obvious problem when looping through posts in the admin.

## Testing

Not yet.

### DIFF
--- a/src/PostsIterator.php
+++ b/src/PostsIterator.php
@@ -20,8 +20,9 @@ class PostsIterator extends ArrayIterator
     /**
      * Prepares the state before working on a post.
      *
-     * Calls the `setup()` function of the current post to setup post data. Before starting the
-     * loop, it will call the 'loop_start' hook to improve compatibility with WordPress.
+     * Calls the `setup()` function of the current post to setup post data if
+     * we’re in the frontend. Before starting the loop, it will call the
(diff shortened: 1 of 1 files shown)


Levdbas:
Tested this on our custom ACF block setup and this still allows queries to be run in the backend and displaying results.

Levdbas (approved):

gchtr:
> Tested this on our custom ACF block setup and this still allows queries to be run in the backend and displaying results.

Make sense. This change is essentially only about Timber not touching any globals when querying posts in the admin. Accessing data from posts still works as expected. In the admin, there’s no loop we need to handle, but still some globals we shouldn’t touch.

PR #3121, opened 2025-06-23
fix: Fix incorrect ACF user filter

NielsdeBlaauw:
## Issue

After removing the user field format filter earlier in the method, it was incorrectly recreated with the `taxonomy_field_type`. 

## Solution

This fix recreates the correct filter for the `format_value` method for user fields.

## Impact

Fixes breaking issue with ACF user fields being interpreted as a taxonomy field.

## Usage Changes

None

## Considerations

None


Levdbas:
Hi @NielsdeBlaauw , and welcome! Thank you for this and for catching this issue!

Levdbas (approved):
LGTM!

Issue #3124, opened 2025-07-04
[BUG] Timber/ImageHelper throws Error when sideloaded image url is_wp_error


herrschuessler:
### Expected Behavior

When sideloading a non-existent image url e.g. a 404 url, ImageHelper should not throw an Error.

### Actual behavior

When sideloading a non-existent image url e.g. a 404 url, ImageHelper throws a ValueError from `file_get_contents()` because the function gets called with an empty value. Since PHP 8.0, `file_get_contents()` cannot be called on an empty string.

### Steps to reproduce behavior

`Timber::get_external_image( 'NON_EXISTENT_URL' )`;

### Notes

The error should be caught in https://github.com/timber/timber/blob/6f0a6dd0bc24cee4468f16e374c590401be13492/src/ImageHelper.php#L469-L476 or a check should be made if `$file_array['tmp_name']` is not empty.

### What version of Timber are you using?

Timber 2.3.2

### What version of WordPress are you using?

WordPress 6.8.1

### What version of PHP are you using?

PHP 8.4

### How did you install Timber?

Installed or updated Timber through Composer

Levdbas:
Hi @herrschuessler ,

Good catch, are you willing to do a PR for this? If you need help with the tests, let me know!

herrschuessler:
@Levdbas I'd be happy to. My proposed solution would be to have the function `sideload_image()` return an empty string in Line 466 https://github.com/timber/timber/blob/6f0a6dd0bc24cee4468f16e374c590401be13492/src/ImageHelper.php#L465-L467 rather than `false`because the value might get passed through `ImageHelper::analyze_url()` and other functions afterwards which all require a string.

PR #3125, opened 2025-07-11
fix: Prevent ValueError for faulty or empty sideloaded image URLs in …

herrschuessler:
Related:
- #3124 

## Issue
When sideloading a non-existent image url e.g. a 404 url, ImageHelper throws a ValueError from `file_get_contents()` because the function gets called with an empty value. Since PHP 8.0, `file_get_contents()` cannot be called on an empty string.

## Solution
Return an empty string if there's a problem with the iamge URL before instead of passing it to `file_get_contents()`

## Impact
None

## Considerations
Force `string` as return type in function signature as other typed functions in code base rely on a string as return type of this function.

### DIFF
--- a/tests/test-external-image.php
+++ b/tests/test-external-image.php
@@ -61,7 +61,7 @@ class TestExternalImage extends TimberAttachment_UnitTestCase
     {
         $image =  Timber\ImageHelper::sideload_image('invalid-url');
 
-        $this->assertEmpty($image, 'Sideloading an invalid URL should return null');
+        $this->assertSame($image, 'Sideloading an invalid URL should return an empty string');
     }
 
     public function delete_existing_sideloaded_image($file)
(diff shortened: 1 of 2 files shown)


Levdbas:
Thanks @herrschuessler , I added a quick test to make sure you code works and keeps working 👍

Levdbas (dismissed):

tests/test-external-image.php
```diff
@@ -57,6 +57,13 @@ public static function copy_image_to_stylesheet($target_dir = '', $filename = 'c
         return $dest;
     }
 
+    public function testFailingToSideloadImage()
+    {
+        $image =  Timber\ImageHelper::sideload_image('invalid-url');
+
+        $this->assertEmpty($image, 'Sideloading an invalid URL should return null');
```
gchtr:
```suggestion
        $this->assertSame($image, 'Sideloading an invalid URL should return an empty string');
```

gchtr (changes requested):

gchtr (approved):

PR #3128, opened 2025-07-22
ci: Add Conductor workflow

gchtr:
This adds the Conductor Github Action Workflow according to https://packagist.com/docs/conductor/getting-started.

Do we have to update the PHP version for the "Install PHP" step?

### DIFF
--- a/.github/workflows/conductor.yml
+++ b/.github/workflows/conductor.yml
@@ -1,0 +1,40 @@
+# See the Conductor setup guide at https://packagist.com/docs/conductor/getting-started
+
+on:
+    repository_dispatch:
+        types:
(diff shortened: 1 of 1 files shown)


Levdbas:
Looking good @gchtr !

Over [here](https://github.com/timber/timber/blob/2.x/.github/actions/setup/action.yml) we use 8.1, so maybe use that over here as well? As discussed, we can further bump up the PHP version if needed in a new major release.

Levdbas (dismissed):

gchtr:
Alright, that makes sense. I updated it to use 8.1 👍

Levdbas (approved):

PR #3131, opened 2025-08-05
fix: Fix deprecation notice in `trim_words` method when `null` is passed.

### LINKED ISSUE #3129: [BUG] prevent deprecation notice from truncate filter
### Expected Behavior

No deprecation notice.

### Actual behavior

Running into this deprecation notice:

`strip_tags(): Passing null to parameter #1 ($string) of type string is deprecated | wp-content/themes/theme-name/vendor/timber/timber/src/TextHelper.php:68`

Pass the truncate filter past a null valued variable.  Quick and easy test: `{{ null|truncate(25) }}`

### Notes

Just avoid passing in a null value into `strip_tags()` in `TextHelper.php` line ~68.  Maybe just return a blank string at the beginning of the `trim_words()` method if `empty($text)`.

Timber 2.3.2

Wordpress 6.8.2

8.2.29

Levdbas:
Related:

- #3129

## Issue
As said in #3129 , passing null to `TextHelper::trip_words ` or `truncate`, causes a deprecation message.

## Solution
Short-circuit the method early when text is null.

## Testing
yes

### DIFF
--- a/src/TextHelper.php
+++ b/src/TextHelper.php
@@ -43,7 +43,7 @@ class TextHelper
      */
     public static function trim_words($text, $num_words = 55, $more = null, $allowed_tags = 'p a span b i br blockquote')
     {
-        if (null === $text) {
+        if (empty($text)) {
             return '';
         }
 
--- a/src/TextHelper.php
+++ b/src/TextHelper.php
@@ -43,6 +43,10 @@ class TextHelper
      */
     public static function trim_words($text, $num_words = 55, $more = null, $allowed_tags = 'p a span b i br blockquote')
     {
+        if (null === $text) {
+            return '';
+        }
+
         if (null === $more) {
             $more = \__('&hellip;');
         }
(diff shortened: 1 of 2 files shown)


src/TextHelper.php
```diff
@@ -43,6 +43,10 @@ public static function trim_characters($text, $num_chars = 60, $more = '&hellip;
      */
     public static function trim_words($text, $num_words = 55, $more = null, $allowed_tags = 'p a span b i br blockquote')
     {
+        if (null === $text) {
```
gchtr: We could also return here when the text is empty, right?

Levdbas: as in 

`if(empty($text)) {`

@gchtr ? That would short circuit the function a lot earlier in some cases. Which sounds like a good idea.

gchtr: Yeah, exactly like that.

My only question is: does this change the logic so that `$more` will not be displayed, when it was before, even with empty text? And is this an acceptable change? In my opinion, this would be okay.

In the future, this function will probably get proper parameter types as well (`trim_words(?string $text, ...)`, so a check for `empty()` would catch empty text and null values, while other values would be discarded earlier with an error.

gchtr (approved):

gchtr:
It’s weird. The PHP coding standards action has an error, but when I run it locally with PHP 8.1, I see no errors to fix. The action seems to be using PHP 8.1 as well, so I can’t say what’s wrong.

But I think this doesn’t block us from merging this PR.

PR #3156, opened 2025-10-27
fix(docs): Fix classname for Timber\Helper in performance docs

n3storm:
## Issue
Old import TimberHelper 

## Solution
Current v2 import Import\Helper

## Impact
Just example at documentation changed


gchtr (approved):
Thanks for taking the time to fix even such a small detail 🙌.

PR #3158, opened 2025-11-20
Add ancestors method to Timber\Post

### LINKED ISSUE #3140: Add ancestors method to Timber\Post
It would be really nice if there was a `parents` method in Timber post, similar to `children`. It's fairly easy to fetch them using `get_post_ancestors`, and it makes it really easy to make a breadcrumb trail in the same way:

```php
// extending Timber\Post
public function parents() {
	if (!$this->post_parent) {
		return false;
	}

	$parents = array_reverse(get_post_ancestors($this->ID));
	if (!empty($parents)) {
		return Timber::get_posts($parents);
	}

	return false;
}
```
```twig
{% if post.parents is not empty %}
	<ul>
		{% for parent in post.parents %}
			<li>
				<a href="{{ parent.link }}">
					{{ parent.title }}
				</a>
			</li>
		{% endfor %}
	</ul>
{% endif %}

### PR DESCRIPTION
**Pull request by cbirdsong**
Related:

- #3140

### DIFF
--- a/src/Post.php
+++ b/src/Post.php
@@ -91,7 +91,6 @@ class Post extends CoreEntity implements DatedInterface, Setupable, Stringable
      */
     protected $_prev = [];
 
-
     /**
      * @var array Stores the results of the ancestors of the post as Timber\Posts
      */
@@ -903,7 +902,7 @@ class Post extends CoreEntity implements DatedInterface, Setupable, Stringable
      */
     public function ancestors()
     {
-        if (null !== $this->_ancestors) {
+        if (isset($this->_ancestors)) {
             return $this->_ancestors;
         }
 
@@ -2017,7 +2016,7 @@ class Post extends CoreEntity implements DatedInterface, Setupable, Stringable
      */
     private function partition_tax_queries(array $query, array $taxonomies): array
     {
-        return \array_map(fn(string $tax): array => \array_merge($query, [
+        return \array_map(fn (string $tax): array => \array_merge($query, [
             'taxonomy' => [$tax],
         ]), $taxonomies);
     }
--- a/src/Post.php
+++ b/src/Post.php
@@ -91,6 +91,12 @@ class Post extends CoreEntity implements DatedInterface, Setupable, Stringable
      */
     protected $_prev = [];
 
+
+    /**
+     * @var array Stores the results of the ancestors of the post as Timber\Posts
+     */
+    protected $_ancestors = [];
+
     /**
      * @var string Stores the CSS classes for the post (ex: "post post-type-book post-123")
      */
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Levdbas** [comment]
Hey @cbirdsong , I've added tests to your PR. I think we are ready for review on this one 👍

**Comment by cbirdsong** [comment]
Thanks! I've been slammed and haven't had time to circle back to this.

File: src/Post.php [hunk 1]
```diff
@@ -878,6 +885,34 @@ public function get_method_values(): array
...
+     *     {% for ancestor in post.ancestors %}
+     *         <a href="{{ ancestor.link }}">{{ ancestor.title }}</a>
+     *     {% endfor %}
+     * {% endif %}
+     * ```
+     * @return PostCollectionInterface
+     */
+    public function ancestors()
+    {
+        if (!empty($this->_ancestors)) {
```
nlemoine: I would leave the `_ancestors` uninitialized. And only check if it's set.

For whatever reason, if `get_post_ancestors()` returns an empty array, you'll loose memoization and `get_post_ancestors` will be called again.

Note that the function does not execute any complex query. Ancestors posts will be stored in the object cache and subsequent calls should be negligible.

src/Post.php
```diff
@@ -878,6 +885,34 @@ public function get_method_values(): array
...
+            return $this->_ancestors;
+        }
+
+        $ancestors = \array_reverse(\get_post_ancestors($this->ID));
+        $this->_ancestors = $this->factory()->from($ancestors);
+
+        return $this->_ancestors;
```
nlemoine, line 912: If the prop is uninitialized, you could even shorten the syntax to a single line:

```php
return $this->_ancestors ??= $this->factory()->from(\array_reverse(\get_post_ancestors($this->ID)));
````

src/Post.php
Levdbas: @nlemoine , gotcha, uninitialized the property and did a null check. The short syntax in my opinion is harder to read and debug if needed so i'd like to keep it this way. Lemme know what you think :)

nlemoine: Sure about the syntax. It's a matter of preference :)

The only thing I would change is the null check. null comparison check ≠ `isset` exactly. `isset` will check if the prop has been initialized, comparing will assume the prop as is and throw if not initialized. Works for now but would break if we add a type to the `_ancestors` in the future (or now). See [this playground](https://php-play.dev/?c=DwfgDgFmBQ0MYBsCGBnFACAygewLYFMAXCASwDsBzdAb2nXvTACdtD842ATdJJppAJ7oAJAH0kZOPhSFsTFAG46DZfTABXAEYIScdADN1kwiWxkek6bPkAKAJSqajhuhL70NkmiI3hxLwC0AHziljJyKHZ2Ti4uAPRxru42ZOoICOgAhAC82SL%2BKMGhUuHy0bSxlehMROpM5n6khSESJdaKzgwAvrBV1bX1%2BU1FrVYR6HkA2gDkAB64qNMAukouPT3Qwhh5ZPgA7lh4RKSU9koAbryinOq4YL7No6Uo9nYKQA&v=8.3&f=html) and switch comment.

Levdbas: Gotcha @nlemoine! Fixed that as well :)

nlemoine (approved):
LGTM, thank you @Levdbas!

PR #3163, opened 2025-12-08
fix: Fix timezone bug when transforming ACF date fields

gchtr:
## Issue

While working with ACF date fields, I figured they were returned as `DateTimeImmutable` objects without a timezone applied when I enabled transforms for all values using the `timber/meta/transform_value` filter.

In the past, I spent a lot of time to figure out how to correctly work with dates and times Timber. I realized that we always have to pass the WordPress timezone when creating DateTime objects, if the date string does not include a timezone string yet. See https://timber.github.io/docs/v2/guides/date-time/#create-a-date-from-a-date-string.

ACF always saves values without a timezone applied, so I think we should pass `wp_timezone()` when creating the DateTime object.

## Solution

Pass `wp_timezone()` when creating the `DateTimeImmutable` object.

## Impact

More robust time handling.

## Usage Changes

None.

## Considerations

None.

## Testing

Updated the tests to make sure that the `DateTimeImmutable` uses the WordPress timezone.

### DIFF
--- a/src/Integration/AcfIntegration.php
+++ b/src/Integration/AcfIntegration.php
@@ -156,7 +156,7 @@ class AcfIntegration implements IntegrationInterface
         if (!$value) {
             return $value;
         }
-        return new DateTimeImmutable(\acf_format_date($value, 'Y-m-d H:i:s'));
+        return new DateTimeImmutable(\acf_format_date($value, 'Y-m-d H:i:s'), \wp_timezone());
     }
 
     /**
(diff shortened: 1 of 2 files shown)


nlemoine (approved):
Nice fix! Indeed, ACF has a weak storage format for dates (`mysql` for datetime picker and `YYYYMMDD` for date picker 🤷 )

Thank you @gchtr 🙂

Levdbas (approved):

PR #3176, opened 2026-01-05
test: Enhance test suite idempotency

nlemoine:
## Issue

While working on Timber test refactoring, I stumbled upon some inconsistencies. Those came from the lack of idempotency: e.g. one test change state (permalink structure, date format, server global, etc.) and creates side effects on subsequent tests, making them fragile/inconsistent.

I learned that PHPUnit can execute tests in a random order. I enabled that option and indeed, tests were randomly failing (sometimes a dozen errors, sometimes more, sometimes less). 

Ideally, a test class should always pass, wether it runs in isolated mode (`composer run test -- --filter=PaginationTest`) or inside the full test suite. 

## Solution

To improve idempotency:
- Enable `executionOrder="random"`
- **Always** ensure state is cleaned up after changing global data (I introduced some useful PHP attributes to make that easier)

## Impact

Improved/stronger tests.

## Usage Changes

Check available attributes in ./tests/Support/Attributes and see their usage inside codebase.


gchtr:
I ran into the issue occasionally when running only a subset of tests would fail, or vice versa, when I created new tests, they would fail when running all tests, but not when running a single test.

I love the new PHP attributes! They make this so much easier.

Thanks a lot for this! 🎉

PR #3179, opened 2026-01-06
fix: fix empty $params returning timber object

### LINKED ISSUE #3169: Passing an empty array to Timber::get_posts() returns current page
### Expected behavior

When given an array containing nothing, `Timber::get_posts` should return nothing.

### Actual behavior

[The docs say](https://timber.github.io/docs/v2/guides/posts/#the-fallback-case-postarrayobject) that if you are dealing with a plugin you can just pass an array of posts into `Timber::get_posts`, like so:

```php
$wp_posts = fancy_plugin_get_custom_posts(); // -> an array of WP_Posts
$timber_posts = Timber::get_posts($wp_posts); // -> Timber\PostArrayObject
```

If the plugin function finds no posts and returns an empty array instead of `false`, then `Timber::get_posts` ends up returning a collection containing the current post.

1. Do `Timber::get_posts(array())`
2. Get the current post back instead of nothing.

Levdbas:
Related:

- #3169

## Issue
See ticket

## Solution
Add two additional checks for array types.

## Impact
Better API usage

## Usage Changes
no

## Considerations
Are there better ways to check for empty arrays?

## Testing
Yes

### DIFF
--- a/tests/Factory/PostFactoryTest.php
+++ b/tests/Factory/PostFactoryTest.php
@@ -369,4 +369,26 @@ class PostFactoryTest extends TimberIntegrationTestCase
 
         $this->assertTrue(Post::class === $res[0]::class);
     }
+
+    public function testFromEmptyArray()
+    {
+        // Create a post to ensure there's a "current" post in the context
+        $post_id = static::factory()->post->create([
+            'post_type' => 'post',
+            'post_title' => 'Current Post',
+        ]);
+
+        // Set up the global post to simulate being on a page/post
+        global $post;
+        $post = \get_post($post_id);
+        \setup_postdata($post);
+
+        $postFactory = new PostFactory();
+        $result = $postFactory->from([]);
+
+        // Passing an empty array should return null, not the current post
+        $this->assertNull($result);
+
+        \wp_reset_postdata();
+    }
 }
--- a/tests/Factory/PostFactoryTest.php
+++ b/tests/Factory/PostFactoryTest.php
@@ -379,16 +379,12 @@ class PostFactoryTest extends TimberIntegrationTestCase
         ]);
 
         // Set up the global post to simulate being on a page/post
-        global $post;
-        $post = \get_post($post_id);
-        \setup_postdata($post);
+        $this->get(\get_permalink($post_id));
 
         $postFactory = new PostFactory();
         $result = $postFactory->from([]);
 
         // Passing an empty array should return null, not the current post
         $this->assertNull($result);
-
-        \wp_reset_postdata();
     }
 }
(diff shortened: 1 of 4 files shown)


gchtr (dismissed):
This looks good to me!

> Are there better ways to check for empty arrays?

We could use `[] === $array` or `count($array) === 0)`, which is a more strict comparison than `empty()`, which does a loose type comparison. But I guess with the `is_array()` check in front of it, `empty()` is fine as well.

tests/Factory/PostFactoryTest.php
```diff
@@ -369,4 +369,26 @@ public function testDeprecatedPostClassMapFilter()
...
+
+    public function testFromEmptyArray()
+    {
+        // Create a post to ensure there's a "current" post in the context
+        $post_id = static::factory()->post->create([
+            'post_type' => 'post',
+            'post_title' => 'Current Post',
+        ]);
+
+        // Set up the global post to simulate being on a page/post
```
nlemoine, line 381: Use Mantle TestKit methods ;)

```php
$this->get(get_permalink($post_id));
```
This will take care of simulating a server request (globals, wp_query, etc.)

tests/Factory/PostFactoryTest.php
```diff
@@ -369,4 +369,26 @@ public function testDeprecatedPostClassMapFilter()
...
+        $post = \get_post($post_id);
+        \setup_postdata($post);
+
+        $postFactory = new PostFactory();
+        $result = $postFactory->from([]);
+
+        // Passing an empty array should return null, not the current post
+        $this->assertNull($result);
+
+        \wp_reset_postdata();
```
nlemoine: Not needed.

nlemoine (changes requested):
@Levdbas Just a few changes to the test to make use of our new testing framework. 

Agreed with @gchtr, [moving away from `empty`](https://localheinz.com/articles/2023/05/10/avoiding-empty-in-php/) is, in most cases, a good thing.

Levdbas:
@gchtr , @nlemoine , thanks for the feedback! I've made the necessary changes.

nlemoine:
Hey @Levdbas,

I took the freedom to change a few things, I hope that's fine for you.

When there are changes upstream, try to use "rebase" instead of "merge":
Capture d’écran 2026-01-09 à 11 13 47

This helps maintaining a clean linear history.

You still had one test failing. `ancestors()` used to return a `PostCollectionInterface` but now empty array is handled, [the post factory returns `null`](https://github.com/timber/timber/blob/6ae23b1a24c1ec79a6db0e9ed6ae9776e5a2dad3/src/Factory/PostFactory.php#L48), which isn't a countable type.

To avoid breaking existing code, I chose to still return a(n) (empty) `PostCollectionInterface` (`PostArrayObject`), because users might still iterate on the results or do count checks.

Let me know if it looks good to you.

nlemoine (approved):

PR #3181, opened 2026-01-08
test: Fix issue with image test that sets a constant

gchtr:
## Issue
There’s an issue with an image test that sets the `UPLOADS` constant. It affects other image tests when tests run in random order.

## Solution
Replace constant.


nlemoine (dismissed):
I didn't run into this but looks like a good improvement 👍 

I don't even know why that test is ran in isolation, maybe it could be merged into the main test suite?

gchtr:
> I don't even know why that test is ran in isolation, maybe it could be merged into the main test suite?

The old comment of that test function says it:

>  /*
>   * This test HAS to be in a separate file, otherwise the UPLOADS const bleeds
>   * in the other tests.
>   */

But it’s true, it doesn’t need to be in a separate file. I changed that in [`366af4d` (#3181)](https://github.com/timber/timber/pull/3181/commits/366af4d57ac7a3547b2df88cf42d196801ccc960).

nlemoine:
Thanks @gchtr 🙏 

You might need to update CI files/composer.json. You can look for the "isolated" keyword to remove leftovers.

gchtr:
> You might need to update CI files/composer.json. You can look for the "isolated" keyword to remove leftovers.

Oh, I didn’t realize there was more involved to that test file. I just removed the leftovers. Even better if we have one less testsuite.

nlemoine (approved):

PR #2997, opened 2024-05-13
fix: Fix deprecation notice since twig 3.10 to now use EscaperRuntime…

Levdbas:
Related:

- #2993

## Issue
Since Twig 3.10 the old methods of adding additional escapers is deprecated.

## Solution
Use the new escaping extension instead with a fallback to use the old escapers. Since we support twig from version 3.5 and onward.

I took the if/else setup from here: https://github.com/symfony/ux/blob/796dd587827f2bb849064c33cc4e053c39e0a0e9/src/TwigComponent/src/Twig/TwigEnvironmentConfigurator.php
## Impact
No more deprecation notices.

## Usage Changes
No.

## Considerations
No.

## Testing
Yes. I added a test to test one of the old escaper methods so we can test that functionality as long as we support that.

PR #3193, opened 2026-02-09
chore: remove deprecated escaper functions and streamline escaper setup

Levdbas:
Related:

- https://github.com/timber/timber/pull/2997

Since we now have a minimum version of Twig 3.19 its safe to cleanup twig workaround dating back almost two years ago.

## Impact
Cleaner code

## Usage Changes
none

## Considerations
We should probably do more of these cleanups.

## Testing
no, tests stay the same

### DIFF
--- a/src/Twig.php
+++ b/src/Twig.php
@@ -412,7 +412,7 @@ class Twig
             ],
 
             'relative' => [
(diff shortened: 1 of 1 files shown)


gchtr (approved):
Nice!

Issue #3188, opened 2026-02-05
`get_url_by_slug` not accessible in template


mrleblanc101:
## Is your feature request related to a problem? Please describe.

I can use `<a href="{{ get_post(123).link }}">` in the template, but not `<a href="{{ get_post_by_slug('contact-us').link }}">`, why not ?

## Describe the solution you’d like

Add `get_post_by_slug` to the global context.

## Describe alternatives you’ve considered

Adding it manually, use id, or return the URL from the controller.

Levdbas:
Hi @mrleblanc101 , We do not have this method available via twig, but you can easily do it yourself this way:

```php
add_filter(
	'timber/twig/functions',
	function ($functions) {

		$additional_functions = [
			'get_post_by'       => [
				'callable' => [\Timber\Timber::class, 'get_post_by'],
			],
		];
		return array_merge($functions, $additional_functions);
	}
);
```

```twig
<pre>
      {{ dump(get_post_by('slug','my-slug')) }}
  </pre>
```

mrleblanc101:
@Levdbas Thanks, is there any good reason why it's not available by default by `get_post` is ? It seems to me more useful to get a post/page by slug then by ID

Levdbas:
Because slugs can change, IDs won't. That's why we advice not querying by slug but always by a full Query or id. That's also why it's probably not available inside Twig files because it's bad practice in our opinion.

mrleblanc101:
@Levdbas Where can I get a list of only fonctions exposed to Twig ?
[On this page](https://timber.github.io/docs/v2/reference/timber-timber/#get_post_by), it list `get_post_by`

Levdbas:
Hi @mrleblanc101 , these are the functions/methods available via PHP which does not mean they are available via Twig. I think that will be something handy for the docs.

mrleblanc101:
@Levdbas yep, that's what i figured out. So currently the function exposed to Twig are not documented ?

Levdbas:
I started working on twig on the following PR: #3194 @mrleblanc101 , if you want, you can review my suggestions over there

PR #3194, opened 2026-02-09
docs: Expand documentation for available Twig functions and filters

Levdbas:
Related:

- https://github.com/timber/timber/issues/3188

## Issue
Without looking into the codebase, you wouldn't have a full picture of filters/functions that are available via twig.

## Solution
I restructured the two guides to give a full picture of the functions/filters that are currently available in the codebase and restructured the functions guide to showcase the three ways to work with functions.

## Impact
For beginners a better understanding whats directly available via twig.

## Usage Changes
none.

### DIFF
--- a/docs/v2/guides/extending-twig.md
+++ b/docs/v2/guides/extending-twig.md
@@ -21,8 +21,8 @@ By default, you’ll have to use `{{ fn('function_name') }}` to call a function 
 
 ```php
 add_filter('timber/twig/functions', function ($functions) {
-    $functions['edit_post_link'] = [
-        'callable' => 'edit_post_link',
+    $functions['wp_get_shortlink'] = [
+        'callable' => 'wp_get_shortlink',
     ];
 
     return $functions;
@@ -31,20 +31,18 @@ add_filter('timber/twig/functions', function ($functions) {
 
 The `$functions` variable is an array of functions that Timber already adds by default.
 
-In the example above, we add a `edit_post_link` function by defining an array with a `callable` key that contains the name of the PHP function we want to call. In this case: [`edit_post_link()`](https://developer.wordpress.org/reference/functions/edit_post_link/).
+In the example above, we add a `wp_get_shortlink` function by defining an array with a `callable` key that contains the name of the PHP function we want to call. In this case: [`wp_get_shortlink()`](https://developer.wordpress.org/reference/functions/wp_get_shortlink/).
 
 In Twig, we can then use it like this:
 
 **single.twig**
 
 ```twig
-{# Calls edit_post_link using default arguments #}
-<div class="admin-tools">{{ edit_post_link() }}</div>
+{# Get shortlink for the current post #}
+<a href="{{ wp_get_shortlink() }}">Short URL</a>
 
-{# Calls edit_post_link with all defaults, except for second argument #}
-<div class="admin-tools">
-    {{ edit_post_link(null, '<span class="edit-my-post-type-link">') }}
-</div>
+{# Get shortlink for a specific post #}
+<a href="{{ wp_get_shortlink(post.id) }}">Short URL</a>
 ```
 
 ### Functions that Timber provides
--- a/docs/v2/guides/extending-twig.md
+++ b/docs/v2/guides/extending-twig.md
@@ -51,7 +51,7 @@ In Twig, we can then use it like this:
 
 Timber already comes with a list of functions it adds by default. You can read more about these functions in the [Functions Guide](https://timber.github.io/docs/v2/guides/functions/).
 
-If you want to check out the functions that Timber provides, you can do a debug dump using `var_dump()`.
+You can also var dump all filters directly by adding the following code to your `functions.php` file:
 
 **functions.php**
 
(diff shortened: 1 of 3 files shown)


mrleblanc101 (approved):

docs/v2/guides/functions.md
```diff
@@ -3,12 +3,61 @@ title: "Functions"
...
+- `get_term()` - Retrieve a single term
+- `get_terms()` - Retrieve multiple terms
+- `get_user()` - Retrieve a single user
+- `get_users()` - Retrieve multiple users
+- `get_comment()` - Retrieve a single comment
+- `get_comments()` - Retrieve multiple comments
+
+### Utility Functions
+
+- `action()` - Execute WordPress actions
```
gchtr:
```suggestion
- `action()` - Execute WordPress actions, see [Twig – WordPress Actions](https://timber.github.io/docs/v2/guides/twig/#wordpress-actions)
```

docs/v2/guides/twig-filters.md
```diff
@@ -5,8 +5,64 @@ order: "220"
...
+These filters help protect against security vulnerabilities:
+
+- [`esc_url`](#esc_url) - Escape URLs for safe use in HTML attributes
+- [`esc_attr`](#esc_attr) - Escape HTML attributes
+- [`esc_html`](#esc_html) - Escape HTML content
+- [`esc_js`](#esc_js) - Escape JavaScript strings
+- [`wp_kses`](#wp_kses) - Strip unwanted HTML tags with allowed list
+- [`wp_kses_post`](#wp_kses_post) - Strip unwanted HTML tags (allows post content tags)
+
 Twig offers a variety of [filters](https://twig.symfony.com/doc/filters/index.html) to transform text and other information into the desired output. In addition, Timber has added some valuable custom filters for your WordPress theme:
```
gchtr: Can we move this paragraph right after the `## General filters` heading? I think we should mention first that Twig has a lot of filters, and then that Timber also adds some more.

docs/v2/guides/twig-filters.md
```diff
@@ -233,3 +297,196 @@ Contributions made by {{ contributors|list(',', '&') }}
...
+```twig
+<picture>
+    <source srcset="{{ image|towebp }}" type="image/webp">
+    <img src="{{ image }}" alt="{{ image.alt }}">
+</picture>
+```
+
+## `date`
+
+Formats a date using WordPress date formats. For more details, see [Twig's date filter documentation](https://twig.symfony.com/doc/3.x/filters/date.html).
```
gchtr:
```suggestion
Formats a date using WordPress date formats, while considering the timezone and date format settings you set in your WordPress settings. For more details, see [Twig's date filter documentation](https://twig.symfony.com/doc/3.x/filters/date.html).
```

docs/v2/guides/extending-twig.md
```diff
@@ -31,7 +31,7 @@ add_filter('timber/twig/functions', function ($functions) {
 
 The `$functions` variable is an array of functions that Timber already adds by default.
 
-In the example above, we add a `edit_post_link` function by defining  an array with a `callable` key that contains the name of the PHP function we want to call. In this case: [`edit_post_link()`](https://developer.wordpress.org/reference/functions/edit_post_link/).
+In the example above, we add a `edit_post_link` function by defining an array with a `callable` key that contains the name of the PHP function we want to call. In this case: [`edit_post_link()`](https://developer.wordpress.org/reference/functions/edit_post_link/).
```
gchtr: In the meantime, we added `Timber\Post::edit_link()`, so using `edit_post_link()` is actually not the best example here anymore.

https://github.com/timber/timber/blob/d6c4d6191dda8c3ae70dd74242c7b8fafa365ac0/src/Post.php#L1655

As an alternative, we might use `wp_get_shortlink()` or `get_post_embed_url()`?

docs/v2/guides/functions.md
```diff
@@ -41,7 +90,7 @@ What if you need to pass arguments to a function? Easy, add them as additional a
 </div>
 ```
 
-Nice! Any gotchas? Unfortunately yes. While the above example will totally work in a single.twig file it will not in The Loop. Why? Single.twig/single.php retain the context of the current post. A function like `edit_post_link` will try to guess the ID of the post you want to edit from the current post in The Loop. the same function requires some modification in a file like **archive.twig** or **index.twig**. There, you will need to explicitly pass the post ID:
+**Important note about context:** While this works in a `single.twig` file that retains the context of the current post, it may not work the same way in The Loop (like in `archive.twig` or `index.twig`). Functions like `edit_post_link` try to guess the post ID from the current post context. In archive or index templates, you need to explicitly pass the post ID:
```
gchtr: Same as above: we might use a different example than `edit_post_link()`, because it’s available as `Timber\Post::edit_link()` already.

gchtr (changes requested):
@Levdbas That’s a great idea to list the filters here, because not all of us are going into Timber’s source code all the time to check how things work!

I found some small things I would improve.

Levdbas:
Thanks @gchtr. I've committed your suggestions and made some additional changes as well

gchtr (approved):
@Levdbas Great, LGTM!

PR #3198, opened 2026-02-13
feat: Add missing class properties to core entities

### LINKED ISSUE #3196: Timber isn't playing nicely with static analysers
On a taxonomy/term page I need to access the slug of a term.

I use `$term = Timber::get_term();` and get the correct `\Timber\Term` element back, so far so good. When I look at it, it does have the slug with `$term->slug` and I can access it at runtime as well. Alas, the static analyzer in PHPStorm is very unhappy with me, because it can't find `slug` and PHPStan complains as well.

Are there any plans to handle this situation? Or is there a "Timber way of doing things"? If so, I'd be happy to help a bit with a PR or two as well.

Levdbas:
Related:

- #3196

## Issue
We have quite a few public properties that are missing on our Timber object. 

## Solution
I've added the missing public properties for Comments, Post, Site,Term, Menu, Menuitem and User

## Impact
Better static analysis.

## Usage Changes
No

## Considerations
Any other Timber class that we need to do the same for?

## Testing
no

### DIFF
--- a/src/User.php
+++ b/src/User.php
@@ -141,7 +141,7 @@ class User extends CoreEntity implements Stringable
     /**
      * All capabilities the user has, including individual and role based.
      *
-     * @var bool[] Array of key/value pairs where keys represent a capability name
+     * @var array<string, bool> Array of key/value pairs where keys represent a capability name
      *             and boolean values represent whether the user has that capability.
      */
     public $allcaps;
--- a/src/User.php
+++ b/src/User.php
@@ -126,7 +126,7 @@ class User extends CoreEntity implements Stringable
     /**
      * Capabilities that the individual user has been granted outside of those inherited from their role.
      *
-     * @var bool[] Array of key/value pairs where keys represent a capability name
+     * @var array<string, bool> Array of key/value pairs where keys represent a capability name
      *             and boolean values represent whether the user has that capability.
      */
     public $caps;
--- a/src/User.php
+++ b/src/User.php
@@ -73,6 +73,8 @@ class User extends CoreEntity implements Stringable
     public $id;
 
     /**
+     * The user's ID.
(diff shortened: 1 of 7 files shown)


src/User.php
```diff
@@ -86,6 +98,61 @@ class User extends CoreEntity implements Stringable
...
+
+    /**
+     * @var string
+     */
+    public $display_name;
+
+    /**
+     * Capabilities that the individual user has been granted outside of those inherited from their role.
+     *
+     * @var bool[] Array of key/value pairs where keys represent a capability name
```
nlemoine:
```suggestion
     * @var array<string, bool> Array of key/value pairs where keys represent a capability name
```

src/User.php
```diff
@@ -86,6 +98,61 @@ class User extends CoreEntity implements Stringable
...
+     * User metadata option name.
+     *
+     * @var string
+     */
+    public $cap_key;
+
+    /**
+     * All capabilities the user has, including individual and role based.
+     *
+     * @var bool[] Array of key/value pairs where keys represent a capability name
```
nlemoine:
```suggestion
     * @var array<string, bool> Array of key/value pairs where keys represent a capability name
```

nlemoine (changes requested):
Just a few tweaks to provide better accuracy on types. Otherwise, all good! 

Thanks @Levdbas!

gchtr (approved):
Nice, thanks for going through all the core entity classes.

Just an obvious hint for anyone reading: We don’t add PHP types yet, because we will handle that in a next major version of Timber, because it might mean introducing breaking changes. The properties together with the comments already provide enough type hinting in IDEs for the moment.

nlemoine (approved):

PR #3210, opened 2026-03-10
feat: Add ACF option field retrieval methods with transformations

### LINKED ISSUE #3204: Automatically transform field from option page.

If I have an ACF Gallery field on an option page which return an array of data:

Image

But the same field inside a flexible content will result in a completely different data structure because it's transformed when accessed using `post.meta('blocks')` and using `add_filter('timber/meta/transform_value', '__return_true')`

Image

I'd like a way to retreive option page field, and transform the data automatically just like with flexible content

Levdbas:
Related:

- #3204

## Issue
There is no easy way to apply transforms on option fields.

## Solution
Introduce some code changes to make this happen. There are two new methods:

```php
use Timber\Integration\ACF;

/**
 * Filters global context.
 *
 * @param array $context An array of existing context variables.
 * @return mixed
 */
function global_timber_context($context)
{
    $context['options'] = ACF::get_options(); // get all option fields, transformed if possible.
    $context['footer_gallery'] = ACF::get_option('footer_gallery'); // get a single option, transformed if possble.

    return $context;
}
add_filter('timber/context', 'global_timber_context');
```

## Impact
Better integration with ACF.

## Usage Changes
It's now easier to use option fields since they are already transformed.

## Considerations
Please have a good look at my proposed changes and see if this is the way to go @timber/rangers . As @mrleblanc101 stated, maybe we want to have the `add_filter('timber/meta/transform_value', '__return_true')` to also have effect? But I have a feeling this would over complicate this feature. Use the already Documented way to add ACF option fields the default way, or use one of new helper methods to get (a) transformed option(s).

## Testing
First test is included, an additional test for testing get_options with multiple option fields still needs to be implemented.

### DIFF
--- a/docs/v2/integrations/advanced-custom-fields.md
+++ b/docs/v2/integrations/advanced-custom-fields.md
@@ -21,6 +21,7 @@ $meta = $post->meta('my_acf_field');
 ```
 
 ### Transform values to Timber/PHP objects
+
 Timber by default returns all field values as is based on the return type set in your ACF field setting.
 
 But sometimes you might want to transform values directly into Timber/PHP objects. For example, if you have a relationship field, you might want to transform the values directly into `Timber\Post` objects.
@@ -28,6 +29,7 @@ But sometimes you might want to transform values directly into Timber/PHP object
 You can do so using the `timber/meta/transform_value` filter:
 
 **functions.php**
+
 ```php
 add_filter('timber/meta/transform_value', '__return_true');
 ```
@@ -50,20 +52,17 @@ You can also use both the filter and parameter options at the same time to globa
 
 The values of the following field types will be transformed into Timber/PHP objects when using transforms:
 
-| Field type | Returns  |
-|---------|---------|
-| File    | `Timber\Attachment`         |
-| Image    | `Timber\Image`         |
-| Gallery    | array of `Timber\Post` objects         |
-| Date picker     |  `DateTimeImmutable`       |
-| Date time picker     |  `DateTimeImmutable`       |
-| Post object     | array of `Timber\Post` objects         |
-| Relationship     | array of `Timber\Post` objects         |
-| Taxonomy     | array of `Timber\Term` objects         |
-| User     | array of `Timber\User` objects         |
-
-
-
+| Field type       | Returns                        |
+| ---------------- | ------------------------------ |
+| File             | `Timber\Attachment`            |
+| Image            | `Timber\Image`                 |
+| Gallery          | array of `Timber\Post` objects |
+| Date picker      | `DateTimeImmutable`            |
+| Date time picker | `DateTimeImmutable`            |
+| Post object      | array of `Timber\Post` objects |
+| Relationship     | array of `Timber\Post` objects |
+| Taxonomy         | array of `Timber\Term` objects |
+| User             | array of `Timber\User` objects |
 
 ### Unformatted values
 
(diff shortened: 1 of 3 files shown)


gchtr (commented):
I’m still a bit torn between adding a new public method to `AcfIntegration`. I think we never had this case before, but it also kinda makes sense to place it in the integration class, because it’s very tied to ACF.

If we go this way, we should make sure that this method is somehow documented as well. I would have to check if it should appear in the class reference. Maybe it would be enough if we documented it in the [ACF guide](https://timber.github.io/docs/v2/integrations/advanced-custom-fields/).

In your code example in the PR description, you use `Timber\Integration\ACF` instead of `Timber\Integration\AcfIntegration` as a class name.

mrleblanc101:
@gchtr I'm pretty surprised you've never faced this request before, seems pretty straightforward to me that getting fields for anywhere should always return the data the same way (TimberPost instead of WP_Post).

gchtr:
@mrleblanc101 My note is not about your use case, which makes total sense to me. What I mean is that we’ve never had a case where we suggested a public method in an integration class in Timber as a public API.

mrleblanc101:
@gchtr Thank you, I understand now 👌

Levdbas:
@gchtr , I've added documentation. Also tried to add tests for the get_fields() helper but that only returned an empty array in the tests. No idea why that would be the case though. Maybe because of option pages or something that needs to be registered but those aren't available in the free version. Either way, that helper works locally.

nlemoine:
If we're about to tackle the way retrieving option work, we should keep the same approach we have for metas: a class that handles getting options (core) and a filter that integrations can hook into. 

That way of grabbing option should not be vendor locked. If one's move from ACF to XXX, she/he would have to change various function calls. Not great from a maintainability POV.

I can see two approaches here.

### The global `options` key

```php

class Options {
    public function __get($name)
    {
         return apply_filters('timber/options', get_option($name));
    }
}

$twig->addGlobal('options', new Options());
```

Because options can be used a bit anywhere, add a global so you don't have to explicitly pass it as context. It's always available. Users can now access their options using:

```twig
{% if options.can_register %}
    {# do something #}
{% endif %}
```

Integrations can hook into `timber/options` and override result if needed:

```php
add_filter('timber/options', function($name) {
    return get_field($name, 'option');
});
```

👎 Kinda breaking change, if someone used `options` as a context, errors/conflicts might pop. To mitigate that, we could still introduce it behind a flag and let users that need it directly implement their own key or pass it as context in their own code. 

### The global `app` key

Long time ago (I don't remember where: issue or Slack/live discussion), I mentioned that Timber tends to "pollute" the global context a bit too much IMHO: 

```php
[
     'site' => ...,
     'user' => ...,
     'post' => ...,
     'posts' => ...,
     'term' => ...,
     'search_query' => ...,
     ...
]
```

Instead, we could have a single global entry for all things global in WP. Just like Symfony has [`AppVariable`](https://github.com/symfony/twig-bridge/blob/8.1/AppVariable.php). A single, universal key to access (lazily) some app state/service.

```php
class App extends \ArrayObject
{
    private Options $options;
    public function __construct()
    {
         $this->options = new Options();
         // add as many objects we need
    }

    public function options(): Options
    {
        return $this->options;
    }
}

$twig->addGlobal('app', new App());
```

In Twig:

```twig
{% if app.options.can_register %}
    {# do something #}
{% endif %}
```

- 👍 No context pollution
- 👍 Data/services on demand
- 👍 Your context focus on your app specific context and stops caring about globals

This way, we use that option feature as an opportunity for future proof context handling.

----

Disclaimer: code here is mostly pseudo code to pitch the idea. Not real/ready implementation :) 

Thoughts @timber/rangers?

Levdbas:
I like this `app` idea @nlemoine , sound like a good way to restructure data inside Timber. Moving `Site`, `Theme`, `User` and potentially a new 'Options'  object under this global variable makes a lot of sense. 

```twig
{{app.user.nicename}}
{{app.site.title}}
{{app.theme.author}}
```

Adding a mechanism for developers to also boot their extended objects over here instead of only the default ones is also a good developer experience on how to extend the global context in a performant way.

Regarding the options in particular, you have the site options and "extended" options that are provided via a plugin like ACF for example. That I always found a bit odd.

mrleblanc101:
I like it 👏  
In our case we were already injecting ACF options pages as `site.options` and menus as `site.menus`.
So changing from `site.optoons` to `app.options` could be a good idea to clean things up.
I think it would be nice to also think about adding menus easily to the context.

```php
$context['site']->options = get_fields('option'); // <-- not transformed
$context['site']->menus = $menus;
```

Also I don't think you should be affraid to make breaking change to the context, seems like something that can be very easily migrated if it's well documented.

gchtr:
I can remember well when you talked about this, @nlemoine and I still like the idea of having an `app` global that can load things when requested.

It would also resolve my concerns about adding public methods to an integration class.

Because this would be a bigger breaking change, I would move this to a Timber 3.x milestone.

mrleblanc101:
What's the plan for 3.0 if there is one ? 👀 
I'd love to know when (approx) we can expect this 😄

gchtr:
> What's the plan for 3.0 if there is one ? 👀
> I'd love to know when (approx) we can expect this 😄

We don’t give out timelines anymore because it was always difficult to meet them in the past. New problems popping up or things we didn’t consider before can come in the way of timelines. And as this is a community-led project, we don’t hold anyone accountable for the time they put into the project – or don’t.

The closest thing to a roadmap is the https://github.com/timber/timber/milestone/15, though, depending on the urgency of things we need to fix or release, that might change as well. We have some things planned that we should probably add to the milestone though ;).

The best way to speed this up would be to create a full-fledged PR that is always up to date, so it’s easy to test and merge.

Issue #3211, opened 2026-03-10
Ability to get terms from PostQuery


cbirdsong:
I often want to creating a set of taxonomy filters to use alongside a query, but using `get_terms()` with the taxonomies attached to the post type returns _all_ of the terms in the taxonomy even if none of the posts inside the query have that term attached.

My current workaround looks like this: 

```php
// Get some posts 
$posts = Timber::get_posts([ "post_type" => "projects", "category_name" => "featured", "posts_per_page" => "-1" ]);

// Make an array to hold the query's active taxonomies
$query_taxonomies = [];

// Loop through all the posts in the query
foreach ($posts as $post) {
	// Then loop through their terms
	foreach ($post->terms() as $term) {
		// Add a key to the array of taxonomies if it doesn't have one
		if (!array_key_exists($term->taxonomy, $query_taxonomies)) {
			$query_taxonomies[$term->taxonomy] = [];
		}

		// Add the term's ID to the corresponding taxonomy array if it isn't already included
		if (!in_array($term->id, $query_taxonomies[$term->taxonomy])) {
			$query_taxonomies[$term->taxonomy][] = $term->id;
		}
	}
}

// Loop through the query's active taxonomies, use the array of IDs to get an array of term objects, and add those to the Timber context
foreach ($query_taxonomies as $filter_key => $filter_term_ids) {
	if (!empty($filter_term_ids)) {
		$terms = Timber::get_terms([
			"taxonomy" => $filter_key,
			"include" => $filter_term_ids,
		]);
		$context["filters"][$filter_key] = $terms;
	}
}
```

I'm sure there is a more performant and resilient way to handle this, so it would be nice if Timber's `PostQuery` had a `terms` method that returns the terms from its posts.

Levdbas:
Hi @cbirdsong ,

I started to iterate on this because this sounds like a nice feature. However, now [this issue](https://github.com/timber/timber/issues/2211) is more relevant then ever because we first need the option to split out the results on the get_term level before we can start working on this feature. I opened a new pr for that in #3213

PR #3213, opened 2026-03-16
feat: Add merge/group option for `Timber::get_terms()`

### LINKED ISSUE #2211: Feature: Timber::get_terms() merge option

There is a `merge` option available in [`Post::terms()`](https://github.com/timber/timber/blob/2203/post_terms_api/lib/Post.php#L683) that lets the caller specify whether they'd like all terms in one flat array, or broken out by taxonomy:

```php
// merge = true (default): one flat array
// these are equivalent:
$post->terms();
$post->terms( [], ['merge' => true] );
// -> [ (category 123), (category 456), ... , (tag 789), ... ]

// merge = false: sub-arrays by taxonomy
$post->terms( [], ['merge' => false] );
// -> [
//   'category' => [ (category 123), (category 456), ... ],
//   'post_tag' => [ (tag 789), ... ],
// ]
```

This is nice but there's no reason it needs to exist solely in `Post`. We can push it down into `Timber::get_terms()` without breaking API changes.

Same as above, but at a higher level of abstraction in `Timber::get_terms()`:

```php
// merge = true (default): one flat array
Timber::get_terms( [], ['merge' => true] );
// -> [ (category 123), (category 456), ... , (tag 789), ... ]

// merge = false: sub-arrays by taxonomy
Timber::get_terms( [], ['merge' => false] );
// -> [
//   'category' => [ (category 123), (category 456), ... ],
//   'post_tag' => [ (tag 789), ... ],
// ]
```

`Post::terms()` would simply forward the `merge` option; calling code would not change.

...Not doing this?

## Additional context

Relevant PRs: #2203  #2208

Levdbas:
Related:

- #2211
- #3211

## Issue

When working on the `PostQuery::terms()` feature suggested by @cbirdsong , I ran into code code duplication between `Post` and `PostQuery` for splitting term results by taxonomy. Then I remembered the very old Feature request in #2211 and decided to give that a shot.

## Solution

Moved the taxonomy splitting logic into `TermFactory` and exposed it through the existing but unused `$options` parameter in `Timber::get_terms()`.

**Changes:**
1. **TermFactory Enhancement** - Added `from_with_options()` method to handle the `merge` option and `partition_tax_queries()` helper. I am not really a fan of this. Maybe @gchtr or @nlemoine can lend some eyes on this?
2. **Timber API Update** - Updated `get_terms()` to pass options through to TermFactory
3. **Post Simplification** - Removed `partition_tax_queries()` and delegated to `Timber::get_terms()`

## Impact

- Centralizes term taxonomy splitting logic in one place (TermFactory)
- Makes the feature discoverable through `Timber::get_terms()` public API
- Maintains full backwards compatibility - existing code continues to work unchanged

**(Backwards) Compatibility:**
All existing `Post::terms()` and maybe future `PostQuery::terms()` calls work identically

## Usage Changes

Users can now use the `merge` option directly with `Timber::get_terms()`:

```php
// Get terms grouped by taxonomy
$terms_by_tax = Timber::get_terms([
    'taxonomy' => ['category', 'post_tag'],
    'object_ids' => [123, 456],
], ['merge' => false]);

// Returns:
// [
//     'category' => [Term, Term, ...],
//     'post_tag' => [Term, Term, ...],
// ]
```

This was previously only available through `Post::terms()` and `PostQuery::terms()` methods, which continue to work as before:

```php
// Still works exactly the same
$terms_by_tax = $post->terms(['category', 'post_tag'], ['merge' => false]);
$terms_by_tax = $posts->terms(['category', 'post_tag'], ['merge' => false]);
```

## Considerations

**Future enhancements:**
The `merge: false` option name is functional but somewhat unintuitive (you set it to `false` to get grouping). Consider adding a `group_by_taxonomy: true` alias in the future for better developer experience

## Testing

**Unit tests included**

Added two new tests in `TermTest.php`:
- `testGetTermsGroupedByTaxonomy()` - Verifies `merge: false` correctly groups terms by taxonomy
- `testGetTermsMergedByDefault()` - Verifies default `merge: true` behavior returns flat array


gchtr (changes requested):
@Levdbas I saw that you updated the implementation and removed the `from_with_options()` method again in favor of an `$options` parameter in the `from()` method. This looks clean to me now.

About the following consideration:

> **Future enhancements:**
The `merge: false` option name is functional but somewhat unintuitive (you set it to false to get grouping). Consider adding a `group_by_taxonomy: true` alias in the future for better developer experience

Now we have different terms for the merge parameter in various places: merge, partition, group, split. I think it would be better if we only used one term, and I like the suggestion of using `group` instead of `merge`. What this would mean:

- Rename `partition` methods and comments to `group` in this PR
- Leave the name of the `$merge` parameter, because that would be a breaking change
- Starting using `group` in comments, variable names, methods, where we can.

I think this might help to understand this feature better.

Levdbas:
Yep, thats done @gchtr

src/Timber.php
```diff
@@ -726,9 +741,14 @@ public static function get_attachment_by(string $field_or_ident, string $ident =
      *                              and accept any valid parameters to that constructor.
      *                              Default `null`, which will get terms from all queryable
      *                              taxonomies.
-     * @param array        $options Optional. None are currently supported. Default empty array.
+     * @param array        $options {
+     *     Optional. An array of options for the function.
+     *
+     *     @type bool $merge Whether the resulting array is spit per taxonomy `false` or merged into a single flat array `true`. Default `true`.
```
gchtr:
```suggestion
     *     @type bool $merge Whether the resulting array is grouped by taxonomy (`false`) or merged into a single flat array (`true`). Default `true`.
```

gchtr (changes requested):
One small thing and then we’re good to go from my side!

gchtr (approved):

PR #3218, opened 2026-03-24
Fix typehints

cnizzardini:
## Issue

Type hints are wrong. I would prefer to just remove them altogether, but it seems to be a theme of this package so I corrected them.

## Solution

Returns `int|null` instead of just `int` (see corresponding methods in ImageDimensions).

## Impact

If developers rely on these type hints they get unexpected behavior.

## Considerations

Don't specify `@return` unless you have to.

## Testing

N/A


Levdbas (approved):

Levdbas:
Hi and welcome @cnizzardini and thank you for the contribution 👍

PR #3220, opened 2026-03-26
fix: Fix bug with transient cache expiration

### LINKED ISSUE #3219: [BUG] transient cache expiration
### Expected Behavior

Calling a cached page via transients should not reset its expiration date

### Actual behavior

Calling a cached page via transients does not change its key (so the content remains exactly the same), but resets its expiration date

1.Calling `Timber::render($templates, $context, 3600);`
2.Look at the created transient via the plugin 'Transients Manager'
ex : timberloader_ec1528d45fa9065c73c77e51cd16537d - expiration : 59 minutes 47 seconds
3.Reload the frontpage
4. In transient manager, the transient key is the same, but the expiration is reset to 59 minutes 59 seconds

### Notes

So as i understand, calling a cached page during the life of is transient set the page cached as it was "Persistent" because the transient never expire

Timber 2.3.3

WordPress 6.9.4

php 8.1

Levdbas:
Related:

- #3219

## Issue

When `Timber::render()` (or `Timber::compile()`) is called with a cache duration, the rendered output is stored as a WordPress transient. On every subsequent request — even when the result is served from the cache — the transient was being overwritten with `set_transient()`, which silently **reset the expiration timer** back to the full TTL. This means a frequently visited page could remain cached indefinitely, never expiring.

## Solution

Track whether the output came from a cache hit in `Loader::render()`. A new `$cache_hit` boolean flag is set to `true` when `get_cache()` returns a valid (non-false, non-null) value. The `set_cache()` call at the end of `render()` is now guarded by `!$cache_hit`, so the transient is only written on a cache miss and is never touched (or refreshed) on a cache hit.

```php
// Before — set_cache() is always called, resetting the TTL on every request:
if (false !== $output && false !== $expires && null !== $key) {
    $this->set_cache(...);
}

// After — set_cache() is only called when there was no cached value:
if (!$cache_hit && false !== $output && false !== $expires && null !== $key) {
    $this->set_cache(...);
}
```

## Impact

- **Correctness**: Transient expiration now behaves as documented, the TTL (stored in epoch) counts from the moment the cache entry was *created*, not from the last page view.
- **Performance**: A small positive improvement on cache hits, as one redundant `set_transient()` DB write per request is eliminated.

## Usage Changes

None.

## Testing

A regression test has been added to `tests/Cache/CacheTest.php` (`testTransientExpirationNotResetOnCacheHit`):

1. Render a template with a 600 s TTL — records the transient timeout value from the DB.
2. Sleep 2 seconds.
3. Render the same template again (cache hit).
4. Assert the epoch time stored in the transient remains the same (+600s from initial hit) and not +600 seconds from second hit.

Without the fix the timeout advances by ~2 seconds on every request. With the fix the assertion passes.

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -109,7 +109,7 @@ class Loader
             $encoded = \json_encode($data);
 
             /**
-             * The encoding might fail, e.g. when an object has a recursion. In that case, we'd rather not cache the
+             * The encoding might fail, e.g. when an object has a recursion. In that case, we’d rather not cache the
              * data instead of possibly returning the wrong data.
              */
             if (false !== $encoded) {
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -103,17 +103,21 @@ class Loader
 
         $key = null;
         $output = false;
+        $cache_hit = false;
         if (false !== $expires) {
             \ksort($data);
             $encoded = \json_encode($data);
 
             /**
-             * The encoding might fail, e.g. when an object has a recursion. In that case, we’d rather not cache the
+             * The encoding might fail, e.g. when an object has a recursion. In that case, we'd rather not cache the
              * data instead of possibly returning the wrong data.
              */
             if (false !== $encoded) {
                 $key = \md5($file . $encoded);
                 $output = $this->get_cache($key, self::CACHEGROUP, $cache_mode);
+                if (false !== $output && null !== $output) {
+                    $cache_hit = true;
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/Loader.php
```diff
@@ -103,17 +103,21 @@ public function render($file, $data = null, $expires = false, $cache_mode = self
...
         $key = null;
         $output = false;
+        $cache_hit = false;
         if (false !== $expires) {
             \ksort($data);
             $encoded = \json_encode($data);
 
             /**
-             * The encoding might fail, e.g. when an object has a recursion. In that case, we’d rather not cache the
+             * The encoding might fail, e.g. when an object has a recursion. In that case, we'd rather not cache the
```
gchtr: Oh, one of my favorite nitpicks! See https://smartquotesforsmartpeople.com/.

```suggestion
             * The encoding might fail, e.g. when an object has a recursion. In that case, we’d rather not cache the
```

gchtr (dismissed):

PR #3223, opened 2026-03-30
fix: Fix context stacking issue with `Timber\Site::switch_to_blog()` in multisite environment

### LINKED ISSUE #3222: [BUG] Rendering twig template whilst being switched to blog, will restore to the main blog
### Expected Behavior

When I am switched to another blog ID, and then render a timber template, I expect after rendering the template to still be on the blog ID i'm switched to.

### Actual behavior

When I am switched to another blog ID, and then render a timber template, my code continues in the main blog ID, instead of the switched blog ID.

1. Set up a basic WordPress multisite with Timber installed. Make sure to add a second blog (in this example blog ID 2).

2. In the root of your twig templates directory, place a twig file named `test.twig` with some simple content in it.

```html
<p>Example content</p>
```

3. In functions.php, add the following code

```php
add_action( 'init', function() {
	if ( ! filter_input( INPUT_GET, 'test-mlp-acf-timber-issue' ) ) {
		return;
	}

	$blog_id_to_render_on = 2;
	switch_to_blog( $blog_id_to_render_on );

	$context = Timber::context();
	$html = Timber::compile(
		'test.twig',
		$context
	);
	echo $html;

	echo sprintf( 'Current blog ID: %d', get_current_blog_id() );
	echo sprintf( 'Expected blog ID: %d', $blog_id_to_render_on );

	restore_current_blog();
	wp_die();
} );
```

3. Visit `https://yourtestdomainhere.tld?test-mlp-acf-timber-issue=true`
4. See that after template rendering, the blog ID is suddenly 1 instead of 2.

### Notes

This can cause all kind of issues further in the process.
In our case, it resulted in a fatal error when MultilingualPress tried to copy post content to another blog, resulting in it attempting to render ACF block timber templates. But it can cause all types of issues, especially when you try to still do an action after rendering a template. For example, if you do a get_post after rendering to maybe render another template, that post cannot be found.

The actual cause of the issue, is the `Timber\Site::__construct` function.
This calls `Timber\Site::switch_to_blog`, which does **not always** run the `switch_to_blog` function.
Then `restore_current_blog()` is **always** called, which results in the blog being switched back to the main blog ID.

The solution is to just always call Wordpress's `switch_to_blog()` function, which does add the blog switch to the stack, but doesn't do any complex logic if the blog ID you're switching to is the same.

I will attempt to see if I can write a PR to fix it.

Timber 2.3.3

WordPress 6.9.4

PHP 8.2

menno-ll:
Related:

- #3222

## Issue

When rendering a template timber runs the constructor of the Site class.
This then only does a switch_to_blog if the blog ID is not the same, but always calls restore_current_blog.
This causes issues when twig templates get rendered when someone has already switched to another blog.
Code after the template render will then run on the main blog ID, instead of the blog ID the template was rendered on.

## Solution

To always run WordPress's `switch_to_blog()` function, even if the blog ID stays the same. This will make sure that the blog switch is added to the stack. WordPress doesn't perform complex logic if the blog ID stays the same, so this shouldn't impact performance.

## Impact

- It should have no impact on performance
- It should not have impact on the current codebase or backward compatibility, except for people who're already actively bypassing this issue in their own projects, instead of writing a bug report.

## Usage Changes

None.

## Considerations

I've also considered to not always run WordPress's `restore_current_blog()`, but determined that removing complex logic in `Site::switch_to_blog()` and instead relying on how WordPress handles it already is way less complex and less prone to bugs than writing a custom `Site::restore_current_blog()` function.

## Testing

Yes, unit test is included.

⚠️ The Unit test created cannot test the bugfix at the moment, as `Timber::get_sites()` returns two blogs with the same blog ID (1). Therefore testing this scenario is only possible if that will return two blogs with different ID's. I don't know how to solve that, so i'm hoping someone can assist with that.

### DIFF
--- a/src/Site.php
+++ b/src/Site.php
@@ -338,7 +338,7 @@ class Site extends Core implements CoreInterface
         }
 
         // Always call switch_to_blog to add the blog switch to the stack. WordPress doesn't perform complex logic if the blog ID stays the same.
-        \switch_to_blog((int)$blog_identifier);
+        \switch_to_blog((int) $blog_identifier);
 
         return (int) $blog_identifier;
     }
--- a/src/Site.php
+++ b/src/Site.php
@@ -337,17 +337,8 @@ class Site extends Core implements CoreInterface
             $blog_identifier = $current_id;
         }
 
-        $info = \get_blog_details($blog_identifier, false);
-
-        if (false === $info) {
-            return $current_id;
-        }
-
-        $blog_identifier = $info->blog_id;
-
-        if ((int) $current_id !== (int) $blog_identifier) {
-            \switch_to_blog($blog_identifier);
-        }
+        // Always call switch_to_blog to add the blog switch to the stack. WordPress doesn't perform complex logic if the blog ID stays the same.
+        \switch_to_blog((int)$blog_identifier);
 
         return (int) $blog_identifier;
     }
--- a/tests/TimberMultisiteTest.php
+++ b/tests/TimberMultisiteTest.php
@@ -245,10 +245,9 @@ class TimberMultisiteTest extends TimberIntegrationTestCase
 
         $sites = Timber::get_sites();
         foreach ($sites as $site) {
-            \switch_to_blog($site->blog_id);
+            \switch_to_blog((int) $site->blog_id);
             $ts = new Site();
-            $this->assertSame($site->blog_id, $ts->ID);
-            $this->assertSame($site->blog_id, \get_current_blog_id());
+            $this->assertSame((int) $site->blog_id, \get_current_blog_id());
             \restore_current_blog();
         }
     }
--- a/tests/TimberMultisiteTest.php
+++ b/tests/TimberMultisiteTest.php
@@ -232,6 +232,27 @@ class TimberMultisiteTest extends TimberIntegrationTestCase
         $this->assertStringStartsWith($site_2_upload_dir['baseurl'], $img_resized_src);
     }
 
+    /**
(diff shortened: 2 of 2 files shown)


Levdbas:
Thanks for the proposed fix @menno-ll . @nlemoine , are you able to share your thoughts on how to get this test up and running?

Levdbas:
Hi @menno-ll , I've extended your test case a bit by adding explicit site creation to really test multisite behavior. At the end we do another assertion to make sure the blog IDs match.

menno-ll:
Can confirm the fixed test case is working as intended.
- Undoing the fix will make the test fail
- Having the fix intact will make the test pass

Thanks for helping out @Levdbas !

gchtr:
When looking at why we initially added the logic to only switch in certain conditions, I see that we did that trying to prevent a bug with certain image sources.

But we also introduced the new bug by not considering context stacking and we should run `restore_current_blog()` only when we really switched.

See

- https://github.com/timber/timber/pull/2781
- https://github.com/timber/timber/issues/1180

Now I wonder: Can we make sure that we also provide a test for the initial issue in https://github.com/timber/timber/issues/1180?

gchtr:
> Now I wonder: Can we make sure that we also provide a test for the initial issue in #1180?

I looked at this more closely and in my view, there’s not enough info in the issue to create a real test case for this. I think we can leave it at what we have.

gchtr (approved):

PR #3229, opened 2026-04-01
docs: Fix wrong use of apply_filters() instead of add_filter() in Performance Guide

### LINKED ISSUE #3226: Docs: Performance guide uses apply_filters() instead of add_filter() for timber/cache/mode
## Description

The [Performance/Caching guide](https://timber.github.io/docs/v2/guides/performance/) shows the following code to set the global cache mode:

```php
apply_filters('timber/cache/mode', function () {
    return Timber\Loader::CACHE_OBJECT;
});
```

This is incorrect. `apply_filters()` **executes the filter immediately** and discards the result — it does not register a callback. The closure is never stored and `CACHE_OBJECT` mode is never activated.

## Correct usage

The [Hooks/Filters reference page](https://timber.github.io/docs/v2/hooks/filters/) correctly shows:

```php
add_filter( 'timber/cache/mode', function( $cache_mode ) {
    return Timber\Loader::CACHE_OBJECT;
} );
```

## Impact

Any project that copied the Performance guide example has **non-functional object cache mode**. Timber falls back to the default (`CACHE_TRANSIENT`), so sites with Redis/Memcached object cache backends are not utilizing them for Timber template output caching.

## Suggested fix

In the Performance/Caching guide, replace:
```php
apply_filters('timber/cache/mode', function () {
```
with:
```php
add_filter('timber/cache/mode', function () {
```

Levdbas:
Related:

- #3226

## Issue
Bad docs

## Solution
Good docs

## Impact
Better docs

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -120,8 +120,6 @@
   "scripts": {
     "analyze": "phpstan analyse --memory-limit=1G",
     "cs": "ecs check",
-    "cs:docs": "ecs check-markdown docs/v2/**/*.md --fix",
-    "cs:docs:fix": "ecs check-markdown docs/v2/**/*.md",
     "cs:fix": "ecs check --fix",
     "grump": "grumphp run",
     "grump:install": "grumphp git:init",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -120,8 +120,6 @@
   "scripts": {
     "analyze": "phpstan analyse --memory-limit=1G",
     "cs": "ecs check",
-    "cs:docs": "ecs check-markdown docs/v2/**/*.md --fix",
```
gchtr, line 123: @Levdbas Oh, what about these commands? Are they not working anymore?

Levdbas: Indeed see [here](https://github.com/easy-coding-standard/easy-coding-standard/pull/45)

gchtr: @Levdbas I see, good to know!

gchtr (approved):

PR #3232, opened 2026-04-13
fix: Improve performance for reading attachment image dimensions

### LINKED ISSUE #3162: Performance issue when using width or height from get_image
I've been noticing an issue when using `get_image` and then using the width and height properties of the resulting object.

It takes around 150ms to load those properties. I tried replacing the width and height attributes of images with hardcoded values and performance improved dramatically.

From looking through the code it seems like the width and height getters are hitting the filesystem to get the actual image dimensions instead of using the media metadata from the db.

Is this intentional? It seems to be causing a huge speed hit for pages with lots of images.

In the mean time I'm just hardcoding image dimensions since those values are immediately overridden by CSS anyway (and SEO tools complain if you don't have the attributes at all).

gchtr:
Related:

- #3162

## Issue

Timber reads image width and height for image attachments from the files directly by using `filesize()`. When we have attachment metadata available, we can also read from the metadata that’s available in the database.

While the issue in #3162 is also related to using a CDN, there is still a benefit to doing this for all environments.

## Solution

Provide an attachment ID to the `ImageDimensions` class when the image is an attachment image, and read from metadata if the relevant metadata exists for an image.

## Impact

- Better performance
- Fewer I/O operations and reduced server load
- If object cache is in use, image dimensions can be read even faster

## Usage Changes

None.

## Considerations

Design-wise, is it fine to add a new parameter to the `ImageDimensions` constructor, or should be do this differently? Maybe @nlemoine has thoughts.

## Testing

Yes.

### DIFF
--- a/src/ImageHelper.php
+++ b/src/ImageHelper.php
@@ -301,8 +301,7 @@ class ImageHelper
     public static function _delete_generated_if_image($post_id)
     {
         if (\wp_attachment_is_image($post_id)) {
-            $attachment = Timber::get_post($post_id);
-            /** @var Attachment $attachment */
+            $attachment = Timber::get_attachment($post_id);
             if ($file_loc = $attachment->file_loc()) {
                 ImageHelper::delete_generated_files($file_loc);
             }
--- a/src/Image.php
+++ b/src/Image.php
@@ -87,7 +87,7 @@ class Image extends Attachment implements ImageInterface
         $data = parent::get_info($data);
 
         if ($this->file_loc()) {
-            $data['image_dimensions'] = new ImageDimensions($this->file_loc());
+            $data['image_dimensions'] = new ImageDimensions($this->file_loc(), $data['ID']);
         }
 
         return $data;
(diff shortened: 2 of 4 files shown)


src/ImageHelper.php
```diff
@@ -301,8 +301,7 @@ public static function add_relative_upload_dir_key($arr)
     public static function _delete_generated_if_image($post_id)
     {
         if (\wp_attachment_is_image($post_id)) {
-            $attachment = Timber::get_post($post_id);
-            /** @var Attachment $attachment */
+            $attachment = Timber::get_attachment($post_id);
```
Levdbas, line 304: Haha how we never spotted this one 👍

Levdbas (approved):
LGTM @gchtr

nlemoine (approved):
Great addition! Thanks @gchtr !

src/ImageHelper.php
gchtr: :)

PR #3234, opened 2026-04-14
feat: Add PostQuery::terms() to get terms scoped to the current query

### LINKED ISSUE #3211: Ability to get terms from PostQuery

Levdbas:
Related:

- #3211

## Issue

When building taxonomy filters alongside a `PostQuery`, calling `Timber::get_terms()` with a taxonomy returns *all* terms in that taxonomy — not just the ones actually used by posts in the current query. Developers had to manually loop over every post, collect term IDs per taxonomy, and then fire a second `get_terms()` call per taxonomy to work around this.

## Solution

Add a `terms()` method to `PostQuery` that returns only the terms attached to the posts in the collection.

Internally it collects the IDs of all posts in the `PostQuery` and passes them as `object_ids` to `WP_Term_Query` via `Timber::get_terms()`. This narrows the result to only terms that are actually used.

The signature mirrors `Timber::get_terms()` and supports the same `merge` option (already introduced in #3213 / PR #3226) so results can be returned either as a flat merged list or grouped by taxonomy:

```php
$posts = Timber::get_posts([
    'post_type'     => 'projects',
    'category_name' => 'featured',
]);

// All terms from all taxonomies (merged flat list).
$all_terms = $posts->terms();

// Only categories used by these posts.
$categories = $posts->terms('category');

// Terms grouped by taxonomy.
$filters = $posts->terms(['category', 'post_tag'], ['merge' => false]);
```

```twig
{# Flat list #}
{% for category in posts.terms('category') %}
    <a href="{{ category.link }}">{{ category.name }}</a>
{% endfor %}

{# Grouped by taxonomy #}
{% set filters = posts.terms('all', {merge: false}) %}
{% for taxonomy, terms in filters %}
    <h3>{{ taxonomy }}</h3>
    {% for term in terms %}{{ term.name }}{% endfor %}
{% endfor %}
```

The grouping work is delegated entirely to `TermFactory::maybe_group_by_taxonomy()`, so `PostQuery` contains no duplicate logic.

## Impact

- **New API**: `PostQuery::terms()` is a new public method
- **No breaking changes**: existing code is unaffected.
- **Performance**: one `WP_Term_Query` call scoped to the post IDs already in memory, replacing the multi-query manual workaround.

## Usage Changes

`PostQuery` gains a `terms()` method. No existing method signatures are changed.

```php
// Before (manual workaround from the issue):
$query_taxonomies = [];
foreach ($posts as $post) {
    foreach ($post->terms() as $term) {
        $query_taxonomies[$term->taxonomy][$term->id] = $term->id;
    }
}
foreach ($query_taxonomies as $tax => $ids) {
    $context['filters'][$tax] = Timber::get_terms(['taxonomy' => $tax, 'include' => $ids]);
}

// After:
$context['filters'] = $posts->terms('all', ['merge' => false]);
```

## Testing

Eight integration tests are added to `PostQueryTest`:

| Test | What it verifies |
|---|---|
| `testPostQueryTermsBasic` | Returns only terms used by the queried posts across all taxonomies |
| `testPostQueryTermsSpecificTaxonomy` | Filters to a single taxonomy |
| `testPostQueryTermsMultipleTaxonomies` | Accepts an array of taxonomy names |
| `testPostQueryTermsGroupedByTaxonomy` | `merge => false` returns results keyed by taxonomy |
| `testPostQueryTermsWithEmptyQuery` | Returns `[]` when the query has no posts |
| `testPostQueryTermsWithCustomTaxonomy` | Works with custom registered taxonomies |
| `testPostQueryTermsUniqueness` | A term shared by many posts appears only once |
| `testPostQueryTermsWithQueryArgs` | Full `WP_Term_Query` args (e.g. `orderby`) pass through correctly |

All 27 `PostQueryTest` tests pass.

### DIFF
--- a/docs/v2/guides/terms.md
+++ b/docs/v2/guides/terms.md
@@ -162,6 +162,115 @@ Or you can use a for loop:
 {%- endfor %}
 ```
 
-We make use of the `loop` variable in Twig to either display an *and* or a comma.
+We make use of the `loop` variable in Twig to either display an _and_ or a comma.
 
-See how we end the opening tag of the for-loop with `-%}` and start the closing tag with `{%-`? These are [Whitespace Controls](https://twig.symfony.com/doc/2.x/templates.html#whitespace-control) and can come in quite handy. Here, we use them to remove all the superfluous markup whitespace we don’t need.
+See how we end the opening tag of the for-loop with `-%}` and start the closing tag with `{%-`? These are [Whitespace Controls](https://twig.symfony.com/doc/2.x/templates.html#whitespace-control) and can come in quite handy. Here, we use them to remove all the superfluous markup whitespace we don’t need.## Get terms from a post collection
+
+## Get terms from a post collection
+
+When working with a collection of posts (via `Timber::get_posts()`), you can get all the terms that are associated with those posts using the `terms()` method. This is useful for creating taxonomy filters, tag clouds, or displaying all terms used across a set of posts.
+
+```php
+$posts = Timber::get_posts([
+    'post_type' => 'article',
+    'posts_per_page' => 20,
+]);
+
+// Get all terms from all taxonomies used by these posts
+$all_terms = $posts->terms();
+
+// Get only categories
+$categories = $posts->terms('category');
+
+// Get multiple taxonomies, grouped by taxonomy name
+$terms_by_taxonomy = $posts->terms(['category', 'post_tag'], ['merge' => false]);
+```
+
+In Twig, you can use the same approach:
+
+```twig
+{# Get posts #}
+{% set posts = get_posts({
+    post_type: 'project',
+    posts_per_page: 10
+}) %}
+
+{# Display all categories used in these posts as filter links #}
+<div class="filters">
+    {% for category in posts.terms('category') %}
+        <a href="{{ category.link }}" class="filter-link">
+            {{ category.name }}
+        </a>
+    {% endfor %}
+</div>
+
+{# Or group terms by taxonomy #}
+{% set terms_by_tax = posts.terms('all', {merge: false}) %}
+{% for taxonomy, terms in terms_by_tax %}
+    <h3>{{ taxonomy|title }}</h3>
+    <ul>
+        {% for term in terms %}
+            <li><a href="{{ term.link }}">{{ term.name }}</a></li>
+        {% endfor %}
+    </ul>
+{% endfor %}
+```
+
+### Parameters
+
+The `terms()` method accepts two parameters:
+
+**`$query_args`** (string|array) – Optional. Default: `[]`
+
+- A taxonomy slug as a string: `'category'`
+- An array of taxonomy slugs: `['category', 'post_tag']`
+- A full `WP_Term_Query` arguments array with additional parameters
+
+**`$options`** (array) – Optional. Default: `[]`
+
+- **`merge`** (bool) – Whether to merge terms from all taxonomies into a single array (`true`, default) or return them grouped by taxonomy name (`false`)
+
+### Examples
+
+**Get terms from a specific query**
+
+```php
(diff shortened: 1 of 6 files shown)


gchtr (dismissed):
This looks good to me!

The only thing I wonder is if it’s enough to list it in the API docs only, or if we should mention it elsewhere, e.g. in the [Terms Guide](https://timber.github.io/docs/v2/guides/terms/).

I also wonder whether there could be confusion that it’s only available on `PostQuery` objects, but not on `PostArrayObject`. This means that you can’t use it on every instance where you do `Timber::get_posts()`. But I’m not sure yet how to document that properly.

src/PostArrayObject.php
```diff
@@ -37,6 +38,22 @@ public function pagination(array $options = [])
         return null;
     }
 
+    /**
+     * Get the post types to use when determining which taxonomies to query.
+     *
+     * @internal
+     * @return array Array of post type slugs.
+     */
+    protected function getPostTypesForTermQuery()
```
gchtr: This method name doesn’t match the coding standards.

docs/v2/guides/terms.md
```diff
@@ -162,6 +162,115 @@ Or you can use a for loop:
 {%- endfor %}
 ```
 
-We make use of the `loop` variable in Twig to either display an *and* or a comma.
+We make use of the `loop` variable in Twig to either display an _and_ or a comma.
 
-See how we end the opening tag of the for-loop with `-%}` and start the closing tag with `{%-`? These are [Whitespace Controls](https://twig.symfony.com/doc/2.x/templates.html#whitespace-control) and can come in quite handy. Here, we use them to remove all the superfluous markup whitespace we don’t need.
+See how we end the opening tag of the for-loop with `-%}` and start the closing tag with `{%-`? These are [Whitespace Controls](https://twig.symfony.com/doc/2.x/templates.html#whitespace-control) and can come in quite handy. Here, we use them to remove all the superfluous markup whitespace we don’t need.## Get terms from a post collection
+
+## Get terms from a post collection
```
gchtr, line 169: Very nice! This surely helps developers discovering the feature.

gchtr (changes requested):

gchtr (approved):
LGTM

PR #3235, opened 2026-04-15
fix: Fix bug when image resize with invalid input causes a fatal error

### LINKED ISSUE #3231: making image resize method fail gracefully
I faced an odd scenario this week where I was using the twig resize filter in what was supposed to be an image reference but was unexpectedly configured wrong.  What it was passing into the filter was a page URL, which the filter tried to process like it was an image, and it looks like it ended up bringing my entire site down.  When I fixed the image reference, the site returned to being stable and performing normally again.

Is there any way we might be able to have the filter detect that what's being processed is actually an image reference and fail gracefully if it isn't?  Or maybe that isn't doable and we would have to just add a timeout on the image load or something?

I will add that [the error that the failure logs](https://github.com/timber/timber/blob/2.x/src/Image/Operation/Resize.php#L228) was pretty unhelpful, and I found myself adding more details to the `Resize.php` file just to get the error logs to provide enough details to find the source of what was causing the problem.  I was seeing hundreds of lines in my error logs like this: `[ Timber ] Error loading /https://www.example.com`  I had to search the Timber codebase for where this was being spit out so I could add additional details so that my error log would record where the error was being generated just to be able to diagnose.

I believe these issues apply to `Retina.php`'s setup as well.

Levdbas:
Related:

- #3231

## Issue

User @columbian-chris reported that accidentally passing a page URL (instead of an image URL) to Timber's resize filter caused the entire site to crash. The image operation classes did not:

1. Validate that the input file was actually an image
2. Provide sufficient error logging to identify the problematic source
3. Fail gracefully when non-image files were passed

Error messages were minimal and unhelpful for debugging:
```
[ Timber ] Error loading /https://www.example.com
```

This forced developers to modify Timber's source code just to diagnose the issue.

## Solution

Centralized file validation in `ImageHelper::_operate()` before any image operation runs:

1. **Added validation methods to ImageHelper:**
   - `validate_image_file()` - Checks file existence and validates extension against known image types (jpg, jpeg, jpe, png, gif, webp, bmp, tiff, svg)
   - `get_operation_name()` - Maps operation classes to human-readable names for error messages

2. **Enhanced error logging:**
   - Operation context (resize, retina, letterbox, etc.)
   - Full file path that caused the error
   - File extension information
   - Helpful hints about common causes

3. **Graceful failure pattern:**
   - Invalid files are rejected early before operation setup
   - Returns original source URL instead of crashing
   - Prevents memory exhaustion from attempting to process non-image files

**Example improved error message:**
```
[Timber] Cannot resize: file does not appear to be a valid image (extension: .html): /var/www/html/wp-content/uploads/page.html
```

## Impact

**Positive impacts:**
- Prevents site crashes when non-image files are passed to image filters
- Significantly improves debugging experience with contextual error messages
- Better code architecture: validation centralized in `ImageHelper` where file path handling occurs
- Operation classes simplified - focus purely on image transformation, not file validation

**Code changes:**
- `ImageHelper.php` - Added validation methods and validation call in `_operate()`
- All Operation classes (`Resize`, `Retina`, `Letterbox`, `ToJpg`, `ToWebp`) - Removed redundant validation code, making them cleaner

**Performance:**
- Negligible impact - validation is minimal (file existence check + extension check)
- Actually improves performance by failing early instead of attempting to process invalid files

**Backward compatibility:**
- ✅ Fully backward compatible
- Valid image processing continues to work normally
- Invalid inputs that would have crashed now fail gracefully
- Return value behavior unchanged (returns `false`/original source on failure)
- Only error logging has been enhanced

## Usage Changes

None. This is an internal improvement that doesn't affect the public API. All Twig filters (`resize`, `retina`, `letterbox`, `tojpg`, `towebp`) continue to work exactly as before.

The only user-facing change is improved error messages in logs, which is a quality improvement.

## Considerations

Future improvements that could build on this work:

1. **MIME type validation:** Currently validates by extension only. Could add runtime MIME type checking using `finfo_file()` for additional safety, though this adds overhead.

2. **Configurable validation:** Could add a filter to allow developers to customize allowed extensions or disable validation if needed (though this seems unlikely to be needed).

3. **Image dimension validation:** Could add checks for reasonable image dimensions to prevent processing extremely large files that might cause memory issues.

4. **Centralized error handler:** Could create a dedicated error handling class for all image operations to further standardize error reporting across Timber.

For now, the current implementation strikes a good balance between safety, performance, and maintainability.

## Testing

**Unit tests included:** Yes - comprehensive test suite with 49 tests (61 assertions), all passing.

### New test file: `tests/Image/Operation/ValidationTest.php` (33 tests)

Tests validation behavior across all 5 image operations:

```php
// Validates that non-image files are rejected gracefully
#[DataProvider('invalidFileProvider')] // HTML, PDF, TXT, PHP, JSON
public function testResizeValidatesFileExtension(string $filename, string $content)

// Validates that missing files are handled gracefully  
public function testResizeHandlesMissingFiles()

// The exact scenario from issue #3231
public function testDoesNotCrashWithPageUrl()
{
    $html_file = $upload_dir['path'] . '/index.html';
    file_put_contents($html_file, '<html>...</html>');
    
    // This should NOT crash the site
    $result = ImageHelper::resize($html_file, 300, 200);
    
    // Should return the original file path (graceful failure)
    $this->assertEquals($html_file, $result);
}
```

### Enhanced test file: `tests/Image/Operation/ResizeTest.php` (4 new tests)

Additional resize-specific validation tests for non-image files, text files, missing files, and PDF files.

### Running the tests:

```bash
# Run all validation tests
vendor/bin/phpunit tests/Image/Operation/ValidationTest.php

# Run resize tests including new validation tests
vendor/bin/phpunit tests/Image/Operation/ResizeTest.php

# Run both test files
vendor/bin/phpunit tests/Image/Operation/ValidationTest.php tests/Image/Operation/ResizeTest.php
```

**Test Results:** ✅ 49 tests passing, 61 assertions

All tests verify:
- ✓ All operations reject HTML files gracefully
- ✓ All operations reject PDF files gracefully
- ✓ All operations reject text files gracefully
- ✓ All operations reject PHP files gracefully
- ✓ All operations reject JSON files gracefully
- ✓ All operations handle missing files gracefully
- ✓ Files without extensions fail gracefully
- ✓ The exact scenario from issue #3231 (page URL) doesn't crash
- ✓ Valid images continue to work correctly
- ✓ Error messages are logged with helpful context

### DIFF
--- a/src/ImageHelper.php
+++ b/src/ImageHelper.php
@@ -301,8 +301,7 @@ class ImageHelper
     public static function _delete_generated_if_image($post_id)
     {
         if (\wp_attachment_is_image($post_id)) {
-            $attachment = Timber::get_post($post_id);
-            /** @var Attachment $attachment */
+            $attachment = Timber::get_attachment($post_id);
             if ($file_loc = $attachment->file_loc()) {
                 ImageHelper::delete_generated_files($file_loc);
             }
--- a/src/ImageHelper.php
+++ b/src/ImageHelper.php
@@ -903,7 +903,7 @@ class ImageHelper
                 throw $e;
             }
             Helper::error_log(\sprintf(
-                'Timber image operation %s failed for %s: %s',
+                'Image operation %s failed for %s: %s',
                 $op::class,
                 $source_path,
                 $e->getMessage()
--- a/src/ImageHelper.php
+++ b/src/ImageHelper.php
@@ -876,7 +876,7 @@ class ImageHelper
         $destination_path = \apply_filters('timber/image/new_path', $destination_path);
 
         if (!\file_exists($source_path)) {
-            Helper::error_log('Timber image operation: source file does not exist: ' . $source_path);
+            Helper::error_log('Image operation: source file does not exist: ' . $source_path);
             return $src;
         }
 
--- a/src/ImageHelper.php
+++ b/src/ImageHelper.php
@@ -301,7 +301,8 @@ class ImageHelper
     public static function _delete_generated_if_image($post_id)
     {
         if (\wp_attachment_is_image($post_id)) {
-            $attachment = Timber::get_attachment($post_id);
+            $attachment = Timber::get_post($post_id);
+            /** @var Attachment $attachment */
             if ($file_loc = $attachment->file_loc()) {
                 ImageHelper::delete_generated_files($file_loc);
             }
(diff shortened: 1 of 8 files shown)


nlemoine (commented):
Thanks @Levdbas for taking care of this!

I think validating against extension isn't safe enough.  What if I try to load `/page.jpg` that contains `<html>...</html>`?

It's not theoretical. When an attachement has been deleted, WP will hit PHP and trigger a 404 for URLs such as `http://example.org/wp-content/uploads/2026/04/image.jpg` and thus return HTML.

I would personally just wrap the offending code part in a `try {} catch {}` block. Throwing when `WP_DEBUG`, triggering a warming otherwise so user can catch it in their logs.

src/ImageHelper.php
```diff
@@ -301,7 +301,8 @@ public static function add_relative_upload_dir_key($arr)
     public static function _delete_generated_if_image($post_id)
     {
         if (\wp_attachment_is_image($post_id)) {
-            $attachment = Timber::get_attachment($post_id);
+            $attachment = Timber::get_post($post_id);
```
gchtr: @Levdbas Why did this change come back in? :)

src/ImageHelper.php
```diff
@@ -874,8 +875,13 @@ private static function _operate($src, $op, $force = false)
          */
         $destination_path = \apply_filters('timber/image/new_path', $destination_path);
 
+        if (!\file_exists($source_path)) {
+            Helper::error_log('Timber image operation: source file does not exist: ' . $source_path);
```
Levdbas:
```suggestion
            Helper::error_log('Image operation: source file does not exist: ' . $source_path);
```

[Timber] is already outputted to the error log by Helper::error_log

src/ImageHelper.php
```diff
@@ -885,15 +891,26 @@ private static function _operate($src, $op, $force = false)
...
-        } else {
-            // in case of error, we return source file itself
-            return $src;
-        }
+        } catch (\Throwable $e) {
+            if (\defined('WP_DEBUG') && WP_DEBUG) {
+                throw $e;
+            }
+            Helper::error_log(\sprintf(
+                'Timber image operation %s failed for %s: %s',
```
Levdbas:
```suggestion
                'Image operation %s failed for %s: %s',
```

Levdbas (commented):
Two minor things @nlemoine  :)

nlemoine:
Fixed! :) Thank you @Levdbas !

src/ImageHelper.php
Levdbas: No idea, but its reverted 🗡️

gchtr (approved):
This looks good to me 👍

Issue #334, opened 2014-09-16
Post Title & ACF Deliciousness Colliding


Nessworthy:
Just encountered a small issue where an existing post had ACF / postmeta field labelled `title`.
Using `{{ post.title }}` ends up returning the postmeta field, instead of the expected post title (filtered).

Is this intentional? Of course, the fix was to just remove the stray data (in this case we didn't need it any more). Just something that others may want to watch out for though.

PR #3249, opened 2026-05-18
fix: Fix regression bug when Timber would access empty properties in Twig

### LINKED ISSUE #3245: [BUG] term.description returns empty strings
### Expected Behavior

Using `{term.description}` on a twig template is always empty, because Timber exposes a `description()` method.

At least that's what I got from https://github.com/timber/timber/blob/2.x/src/Core.php#L144

### Actual behavior

`{term.description}` returns the description, instead of an empty string

### Notes

_No response_

2.4

6.9

8.2

gchtr:
Related:

- #3245 

## Issue
 
By introducing some new properties in https://github.com/timber/timber/pull/3198, we introduced a regression bug where `{{ term.description }}` or `{{ post.comment_count }}` would return empty instead, because the object property was accessed instead of the method with the same name.

## Solution

I looked at the issue more closely, and I found that setting the property to protected fixes the issue. However, the property would still be empty, even though a value would be available on the `WP_Term` object. Even if the property were protected, this would be wrong in my opinion.

The issue is the following line:

https://github.com/timber/timber/blob/43c4e4964b11d2ce6020b14bc77759dfda71de79/src/Core.php#L144

It means that if we have a method defined, we won’t import the property. This logic was introduced 12 years ago in https://github.com/timber/timber/issues/334.

While this was a solution that worked for a long time, I think it’s also quite unexpected that a property that you want to import is not actually set and will always stay null.

We can set the properties that have a corresponding public method to protected to fix the regression.

We actually already use the same approach in `MenuItem`:

https://github.com/timber/timber/blob/43c4e4964b11d2ce6020b14bc77759dfda71de79/src/MenuItem.php#L46-L50

In this class, we had to set the `title` property manually, because it’s not imported:

https://github.com/timber/timber/blob/43c4e4964b11d2ce6020b14bc77759dfda71de79/src/MenuItem.php#L293-L297

That’s why I also changed the logic in `\Timber\Core::import()` to always import properties that exist and also have a method defined.

This will allow us to also remove the custom import fix from `MenuItem`.

I only found the issue to be present for `Term::$description`, `Post::$comment_count` and `MenuItem::$title`, but not for any other property.

## Impact

Fixes regression bug.

## Usage Changes

None.

## Considerations

We might think about removing the `method_exists()` logic entirely from the `import()` method. With dynamic properties going away more and more in modern PHP, this would be a direction that would also help static analysis.

## Testing

Yes, I added two tests that specifically check for the methods to be accessed in Twig instead of the properties.

### DIFF
--- a/src/Core.php
+++ b/src/Core.php
@@ -3,6 +3,7 @@
 namespace Timber;
 
 use AllowDynamicProperties;
+use Error;
 
 /**
  * Class Core
@@ -143,9 +144,14 @@ abstract class Core
                     $this->$key = $value;
                 } elseif (\method_exists($this, $key) && \property_exists($this, $key)) {
                     // While we don’t import properties where a method exists by default, this makes
-                    // sure that properties are imported where they are defined. This reduces
-                    // unexpected side effects.
+                    // sure that properties are imported when they are defined. This reduces
+                    // unexpected side effects. The try-catch is here to not trigger a fatal error
+                    // when we try to set a private property.
+                    try {
                     $this->$key = $value;
+                    } catch (Error $e) {
+                        \trigger_error($e->getMessage(), \E_USER_WARNING);
+                    }
                 } elseif (!empty($key) && !\method_exists($this, $key)) {
                     if ($only_declared_properties) {
                         if (\property_exists($this, $key)) {
--- a/src/Core.php
+++ b/src/Core.php
@@ -141,6 +141,11 @@ abstract class Core
                 }
                 if (!empty($key) && $force) {
                     $this->$key = $value;
+                } elseif (\method_exists($this, $key) && \property_exists($this, $key)) {
+                    // While we don’t import properties where a method exists by default, this makes
+                    // sure that properties are imported where they are defined. This reduces
+                    // unexpected side effects.
+                    $this->$key = $value;
                 } elseif (!empty($key) && !\method_exists($this, $key)) {
                     if ($only_declared_properties) {
                         if (\property_exists($this, $key)) {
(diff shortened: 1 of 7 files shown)


Levdbas (dismissed):
LGTM

src/Core.php
```diff
@@ -141,6 +141,11 @@ public function import($info, $force = false, $only_declared_properties = false)
                 }
                 if (!empty($key) && $force) {
                     $this->$key = $value;
+                } elseif (\method_exists($this, $key) && \property_exists($this, $key)) {
```
nlemoine, line 145: I think there's a [tiny catch](https://php-play.dev/?c=DwfgDgFmBQDGA2BDAzsgBABQPbIC5oG9o0S0wAnASwDdFcBTNAEl0t3noG5jSKsHYDACZoAZgFcAdoMpZJaVu3oAKAJQEAvtC1MwOfAF40k%2BgHdM%2Btd1rkA%2BkPEBbMMsf1cELENv0AHpTxkZV19ABo0AHJFDgjVcL4wenJcAE8ff0DgvTxwqLYY1VVOIA&v=8.5&f=html).

Method/property visibility is assumed to be public here. But those methods will also return `true` for protected/private.

```php
class MyPost extends Timber\Post {
    private $internal_id;
    public function internal_id(): string { /* ... */ }
}

add_filter('timber/post/import_data', fn($d) => $d + ['internal_id' => 'x']);
```

That example would trigger a fatal error. Might be an edge case but still worth addressing with a try/catch around property assignement?

gchtr: Added an update in [`290f79b` (this PR)](https://github.com/timber/timber/pull/3249/commits/290f79b54f1c170d7c6be476968c21a3038e8f72).

gchtr:
I added a try-catch to prevent fatal errors when Timber would try to set a private property. In that case, a user warning will be issued instead.

nlemoine (approved):
All good, thanks @gchtr !

PR #3253, opened 2026-05-21
chore: bump minimum WordPress version to 6.4

nlemoine:
Update the CI matrix lowest-deps job from WP 6.3 to 6.4, and drop the now-dead version conditionals in TimberWPFunctionsTest (the the_block_template_skip_link deprecation is always emitted on 6.4+) and the WP 6.3 in_the_loop workaround in TimberContextTest.

## Issue

Bump WP minimum tested version as per our policy. We should probably display it somewhere BTW. 

## Considerations

Just wonder if we should change version here?
https://github.com/timber/timber/blob/35b869a7f1029d0b54e10066eb0da04c9c54e46f/src/Admin.php#L14-L22

But the minimum WP version is only the minimum version we test ≠ version Timber actually supports.

### DIFF
--- a/.github/workflows/php-unit-tests.yml
+++ b/.github/workflows/php-unit-tests.yml
@@ -71,7 +71,7 @@ jobs:
         include:
           # Oldest supported WP with lowest deps
           - php: '8.2'
-            wp: '6.3'
+            wp: '6.4'
             dependency-version: 'lowest'
             experimental: false
             coverage: false
--- a/.github/workflows/php-unit-tests.yml
+++ b/.github/workflows/php-unit-tests.yml
@@ -71,7 +71,7 @@ jobs:
         include:
           # Oldest supported WP with lowest deps
           - php: '8.2'
-            wp: '6.4'
+            wp: '6.8'
             dependency-version: 'lowest'
             experimental: false
             coverage: false
--- a/src/Admin.php
+++ b/src/Admin.php
@@ -7,33 +7,24 @@ namespace Timber;
  */
 class Admin
 {
-    public static function init()
+    public static function init(?string $wp_version = null): void
     {
-        global $wp_version;
+        // wp_get_wp_version() exists since WP 6.7; fall back to the global for older versions
+        // (which is exactly the range this notice targets).
+        $wp_version ??= \function_exists('wp_get_wp_version') ? \wp_get_wp_version() : (string) ($GLOBALS['wp_version'] ?? '');
 
-        $minimum_version = '6.0.0';
-
-        // Show notice if user is running something older than the required
-        // WordPress version.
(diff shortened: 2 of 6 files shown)


gchtr:
> Bump WP minimum tested version as per our policy. We should probably display it somewhere BTW.

Yeah, good point. Probably on the [main README](https://github.com/timber/timber) and on the [docs homepage](https://timber.github.io/docs/v2/) would be a good start. Until now, we only have it in https://timber.github.io/docs/v2/upgrade-guides/2.0/#new-requirements.

> Just wonder if we should change version here?

I guess it would be good to bump the version against the tested version. We could adapt the text to the following:

```php
"Your installed version of <a href='https://github.com/timber/timber'>Timber</a> is only tested against <strong>WordPress $minimum_version</strong> or greater, but you are running <strong>WordPress $wp_version</strong>. Please <a href='$upgrade_url'>update WordPress</a> to make sure Timber runs fine."
```

gchtr (dismissed):

nlemoine:
Thanks for your review @gchtr! 

> I guess it would be good to bump the version against the tested version. We could adapt the text to the following:

Agreed! I'll make the changes now. Seems like the right PR to do that?

gchtr:
> Seems like the right PR to do that?

@nlemoine Yes, I think we can do it all in this PR

nlemoine:
@timber/rangers Since our policy isn't live yet and before I make the relevant changes, I think we could review the range and maybe narrow it down a bit? If I remember well, we discussed something like current minus 6?

Which today would give: 
- current: 7
- tested: 6.4

That's pretty wide IMHO. 

- Timber 2 isn't a plugin anymore, more developer oriented means projects using Timber are probably up to date. Or at least not 6 versions behind current.
- AI age, updating isn't the pain it used to be.
- Even Yoast (massive WP users audience ≠ developers) [requires 6.8](https://github.com/Yoast/wordpress-seo/blob/290a547256e1bf22509e581a8771979272dbdc25/wp-seo.php#L19)
- Less tests headhaches, backward compat issues, etc.

Would "current minus 3" (6.8 as of today) be ok for you? It would also be aligned with the PHP version policy. (8.5 -> 8.2).

Levdbas:
Hey @nlemoine , sounds like a plan and I like the fact that the PHP version is also aligned then

nlemoine:
Typical code that could be removed: https://github.com/timber/timber/pull/3253/changes#diff-dd7574c63cefd65c36ce6bf3cbd5f6e0a5ac664696e1b8546dad8994faff407bR14

gchtr:
@nlemoine We once said current version minus 5, but current minus 3 sounds good to me as well 👍

nlemoine:
Thank you both for your feedback! It's now done.

PR #3274, opened 2026-06-12
ci: surface public API changes on pull requests

nlemoine:
Add composer/api-surface-check, which statically diffs the public API surface of changed PHP files and posts a summary comment on PRs. Split into a permissionless pull_request trigger and a workflow_run comment workflow so fork PRs are analyzed without granting them write access.


nlemoine:
Could provide useful insights when reviewing PRs. BC can be easily missed. From https://seld.be/notes/surfacing-public-api-changes/

We can give it a try and see if that fits.

gchtr:
@nlemoine Awesome, thanks!

PR #3283, opened 2026-08-21
chore: Fix Rector ruleset to use Composer based set

gchtr:
The latest version of Rector caused a fatal error with the referenced set list. In 2.6.2, Rector introduced composer-based sets, so it can read the required sets from composer.json, see https://github.com/rectorphp/rector/releases/tag/2.6.2.

Along with the change in **rector.php**, I also ran Rector to fix outstanding issues.

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -18,8 +18,7 @@ return RectorConfig::configure()
         php82: true,
     )
     ->withSets([
-        PHPUnitSetList::PHPUNIT_100,
-        PHPUnitSetList::PHPUNIT_110,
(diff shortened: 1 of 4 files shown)


nlemoine (commented):
Thank you @gchtr! There's one test that fails because it asserts an exact upload name. The test seems correct, it just needs to assert a filename that's not duplicate aware.

I've fixed many tests like this involving filename. We should probably find a better way to deal with those assertions in the future.

gchtr:
@nlemoine Yes, I figured that was the case. Finding a solution for these issues would be nice, but is a different issue we will tackle later.

PR #3284, opened 2026-09-02
feat: Add `Timber::get_taxonomy()` and `Timber::get_taxonomies()`

### LINKED ISSUE #3282: `Timber::get_taxonomy` and `Timber::get_taxonomies`

Hi,
Am I missing something obvious ?
Timber has `get_term` and `get_terms`, which returns a term or the terms of a taxonomy.
But why doesn't it have `get_taxonomy` and `get_taxonomies`, which would return the whole taxonomy object including details such as the `labels` defined in `register_taxonomy` ? 

I need to display the `all_items` in the template, and wanted to use `<taxonomy>.labels.all_items`, so I ended up using `{{ fn('get_post_type_label', term.taxonomy) }}` instead. This looks backward to me, and I don't like using `{{ fn() }}` if I can avoid it.

Ex: 
```php
Timber::get_taxonomies([ 'post_type' => 'recette' ])
```

Levdbas:
## Issue

Timber has `get_term()` and `get_terms()`, but no way to get the taxonomy itself. That means everything you configure in `register_taxonomy()` – most importantly the `labels` – is out of reach from a template. To display something as simple as “All Genres”, you currently have to drop down to `{{ fn('get_taxonomy', 'genre').labels.all_items }}`, which is exactly the kind of thing the object API exists to avoid.

The workaround that gets suggested is to assemble the data by hand in PHP:

```php
$taxonomy = get_taxonomy('gamme');

$context['filters'][] = [
    'name' => $taxonomy->name,
    'labels' => $taxonomy->labels,
    'terms' => Timber::get_terms(['taxonomy' => 'gamme']),
];
```

Related:

- #3282

## Solution

Adds a `Timber\Taxonomy` object and the two API methods to get it.

`Timber\Taxonomy` extends `Timber\Core` and is built the same way `Timber\Term` is – a protected constructor plus `static build(WP_Taxonomy)` – so it can’t be instantiated directly and always goes through the factory. `init()` imports the `WP_Taxonomy` object, which puts everything from `register_taxonomy()` (`labels`, `hierarchical`, `rewrite`, `show_in_rest`, …) directly on the object. On top of that it adds:

- `terms($query_args = [], $options = [])` – delegates to `Timber::get_terms()`, so you get real `Timber\Term` objects and the Term Class Map keeps applying. Terms are only queried when this is called, and the unfiltered result is memoized.
- `title()` – the human-readable label. Worth having because `name` on a taxonomy is the registration name (`post_tag`), while `name` on a `Timber\Term` is the label. `title()` removes that trap and matches `Post::title()` and `Term::title()`.
- `default_term()` – the term registered through the `default_term` argument of `register_taxonomy()`, resolved from the `default_term_{$taxonomy}` option and returned as a `Timber\Term`.
- `can_edit()` and `edit_link()` – the term overview in the admin, guarded by the taxonomy’s own `manage_terms` capability. Same pair as on `Post` and `Term`.
- `post_types()` – the post types the taxonomy is registered for, as `Timber\PostType` objects.
- `wp_object()` and `__toString()` (returns the taxonomy name), matching `Timber\Term`.

`Timber\Factory\TaxonomyFactory` mirrors `TermFactory`. Its `from()` accepts a taxonomy name, a list of names, an arguments array, a `WP_Taxonomy`, or an existing `Timber\Taxonomy` (idempotent). It introduces the `timber/taxonomy/classmap` and `timber/taxonomy/class` filters, with the same array-or-callable shape as the other class maps.

`Timber::get_taxonomy()` falls back to the taxonomy of the currently queried term when called without arguments, the same way `Timber::get_term()` falls back to the queried term. `Timber::get_taxonomies()` defaults to public taxonomies and returns an array keyed by taxonomy name. Both are registered as Twig functions.

Two implementation details worth calling out for review:

- **`object_type`.** `WP_Taxonomy::$object_type` (the post types the taxonomy is attached to) collides with `Timber\Core::$object_type`. `Timber\Taxonomy` extends `Core` rather than `CoreEntity` – taxonomies have no meta – so `Core::$object_type` is only used for meta lookups and deprecation notices, neither of which apply here. WordPress’ meaning wins, which is what a template author would expect. There’s a test pinning this.
- **The `post_type` argument.** WordPress’ own `object_type` argument to `get_taxonomies()` compares with `wp_filter_object_list()` and therefore requires an _exact_ match of the registered post type array, so `['object_type' => ['recipe']]` will not find a taxonomy registered for `['post', 'recipe']`. The `post_type` alias is resolved through `get_object_taxonomies()` and then intersected with the result of the remaining arguments, so it behaves the way you’d expect and composes with the other arguments.- **`default_term`.** Because a `default_term()` method exists and no matching property is declared, `Core::import()` skips the raw `default_term` registration array from `WP_Taxonomy` and `{{ taxonomy.default_term }}` resolves to the method instead. That’s deliberate – the resolved `Timber\Term` is more useful than the registration arguments – and follows the same approach as the protected `Term::$description`.

## Impact

Additive only. No existing class, method, filter or behaviour changes, so there is no backwards compatibility concern.

Performance impact is limited by design: terms are lazy, so `Timber::get_taxonomies()` on a site with many taxonomies does not fire a term query per taxonomy. Building a `Taxonomy` is just `get_taxonomy()` plus a property import, with no database access.

## Usage Changes

```php
// Get a taxonomy by name.
$taxonomy = Timber::get_taxonomy('genre');

// Use the taxonomy of the currently queried term archive.
$taxonomy = Timber::get_taxonomy();

// All public taxonomies, keyed by name.
$taxonomies = Timber::get_taxonomies();

// A list of taxonomy names.
$taxonomies = Timber::get_taxonomies(['category', 'genre']);

// Any get_taxonomies() argument, plus a post_type alias.
$taxonomies = Timber::get_taxonomies([
    'post_type' => 'recipe',
    'hierarchical' => true,
]);
```

```twig
{% set genre = get_taxonomy('genre') %}

<h1>{{ genre.title }}</h1>

{% for term in genre.terms %}
    <a href="{{ term.link }}">{{ term.title }}</a>
{% endfor %}

{% if genre.can_edit %}
    <a href="{{ genre.edit_link }}">Manage genres</a>
{% endif %}
```

New filters: `timber/taxonomy/classmap` and `timber/taxonomy/class`.

New Twig functions: `get_taxonomy()` and `get_taxonomies()`.

Docs: a new **Taxonomies** guide, a **The Taxonomy Class Map** section in the Class Maps guide, and a cross-link from the Terms guide.

## Considerations

- `Timber::get_taxonomy_by()` is deliberately not included. Unlike terms and users, taxonomies are only ever addressed by name, so there is no second field to look up by.
- `Timber\Term::$taxonomy` is still a string rather than a `Timber\Taxonomy` object. Changing it would be a breaking change and would risk queries in unexpected places, so it stayed as is. `Timber::get_taxonomy(term.taxonomy)` covers the gap.
- While writing the laziness test I noticed each Timber term query fires `pre_get_terms` twice: `WP_Term_Query::__construct()` already runs the query, and `TermFactory::from_wp_term_query()` then calls `get_terms()` on the same instance again. WordPress’ object cache absorbs the second fetch so it isn’t a duplicate database hit, and it predates this PR, but it may be worth a separate look.

## Testing

Yes – 34 tests, all included.

`tests/TaxonomyTest.php` covers getting a taxonomy by name and from a `WP_Taxonomy`, `null` for an unregistered name, `__toString()`, the `object_type`/`post_types()` behaviour described above, `terms()` with and without query arguments, and all four `get_taxonomies()` input shapes including `post_type` on its own and combined with another argument.

The added API methods are covered too: `title()` in PHP and Twig, `default_term()` against a taxonomy registered with a `default_term` as well as one without, and `can_edit()`/`edit_link()` for both an administrator and a subscriber.

Laziness is asserted rather than assumed. `testTermsAreLazy` counts `pre_get_terms` and checks that getting a taxonomy runs no term query at all, that the first `terms()` call does, that a second unfiltered call is served from the memoized result, and that passing arguments bypasses it.

Class map behaviour is covered from both directions. `tests/Factory/TaxonomyFactoryTest.php` tests the array form, the callable form and the `timber/taxonomy/class` filter, plus invalid input. `tests/TaxonomyTest.php` then does it end to end with a `Genre extends Taxonomy` fixture that adds `top_level_terms()` and `term_count()`, asserting that the class map applies through both `get_taxonomy()` and `get_taxonomies()`, that unmapped taxonomies still get the base class, that the methods return the right values against a real parent/child term set, and that they render from Twig:

```twig
{% set genre = get_taxonomy('genre') %}
{{ genre.title }} ({{ genre.term_count }}): {{ genre.top_level_terms|join(', ') }}
{# => Genres (3): Ambient, Jazz #}
```

The full suite (1314 tests), PHPStan and ECS all pass.

### DIFF
--- a/src/Factory/TaxonomyFactory.php
+++ b/src/Factory/TaxonomyFactory.php
@@ -36,7 +36,7 @@ class TaxonomyFactory
         }
 
         // A flat list of taxonomy names.
-        if (!empty($params) && $this->is_numeric_array($params) && $this->is_array_of_strings($params)) {
+        if (!empty($params) && \array_is_list($params) && $this->is_array_of_strings($params)) {
             return $this->from_names($params);
         }
 
@@ -181,19 +181,6 @@ class TaxonomyFactory
         $class = $this->get_taxonomy_class($taxonomy);
 
         return $class::build($taxonomy);
-    }
-
-    protected function is_numeric_array($arr)
-    {
-        if (!\is_array($arr)) {
-            return false;
-        }
-        foreach (\array_keys($arr) as $k) {
-            if (!\is_int($k)) {
-                return false;
-            }
-        }
-        return true;
     }
 
     protected function is_array_of_strings($arr)
--- a/src/Factory/TaxonomyFactory.php
+++ b/src/Factory/TaxonomyFactory.php
@@ -1,0 +1,211 @@
+<?php
+
+namespace Timber\Factory;
+
+use InvalidArgumentException;
+use Timber\CoreInterface;
+use Timber\Taxonomy;
+use WP_Taxonomy;
+
+/**
+ * Internal API class for instantiating Taxonomies
+ */
+class TaxonomyFactory
+{
+    /**
+     * @param mixed $params A taxonomy name, an array of taxonomy names, an array of arguments for
+     *                      `get_taxonomies()`, a `WP_Taxonomy` object or a `Timber\Taxonomy`.
+     *
+     * @return Taxonomy|Taxonomy[]|null
+     */
+    public function from($params)
+    {
+        if (\is_object($params)) {
+            return $this->from_taxonomy_object($params);
+        }
+
+        if (\is_string($params)) {
+            return $this->from_name($params);
+        }
+
+        if (!\is_array($params)) {
+            throw new InvalidArgumentException(\sprintf(
+                'Expected a taxonomy name, an array or an instance of Timber\CoreInterface or WP_Taxonomy, got %s',
+                \get_debug_type($params)
+            ));
+        }
+
+        // A flat list of taxonomy names.
+        if (!empty($params) && $this->is_numeric_array($params) && $this->is_array_of_strings($params)) {
+            return $this->from_names($params);
+        }
+
+        return $this->from_query_args($params);
+    }
+
+    protected function from_name(string $name): ?Taxonomy
+    {
+        $wp_taxonomy = \get_taxonomy($name);
+
+        if (!$wp_taxonomy) {
+            return null;
+        }
+
+        return $this->build($wp_taxonomy);
+    }
+
+    /**
+     * @param string[] $names
+     * @return Taxonomy[] Keyed by taxonomy name. Names that aren’t registered are skipped.
+     */
+    protected function from_names(array $names): array
+    {
+        $taxonomies = [];
+
+        foreach ($names as $name) {
+            $taxonomy = $this->from_name($name);
+
+            if ($taxonomy) {
+                $taxonomies[$name] = $taxonomy;
+            }
+        }
+
+        return $taxonomies;
+    }
+
+    /**
+     * @return Taxonomy[] Keyed by taxonomy name.
+     */
+    protected function from_query_args(array $args): array
+    {
+        // WP’s own `object_type` argument requires an exact match of the registered post types,
+        // so the `post_type` alias is resolved separately.
+        $post_types = $args['post_type'] ?? null;
+        unset($args['post_type']);
+
+        $names = \get_taxonomies($args);
+
+        if (null !== $post_types) {
+            $names = \array_intersect($names, $this->taxonomy_names_for_post_types((array) $post_types));
+        }
+
+        return $this->from_names(\array_values($names));
+    }
+
+    /**
+     * @param string[] $post_types
+     * @return string[]
+     */
+    protected function taxonomy_names_for_post_types(array $post_types): array
+    {
+        $names = [];
+
+        foreach ($post_types as $post_type) {
+            $names = \array_merge($names, \get_object_taxonomies($post_type));
+        }
+
+        return \array_unique($names);
+    }
+
+    protected function from_taxonomy_object(object $obj): CoreInterface
+    {
+        if ($obj instanceof CoreInterface) {
+            // We already have a Timber Core object of some kind.
+            return $obj;
+        }
+
+        if ($obj instanceof WP_Taxonomy) {
+            return $this->build($obj);
+        }
+
+        throw new InvalidArgumentException(\sprintf(
+            'Expected an instance of Timber\CoreInterface or WP_Taxonomy, got %s',
+            $obj::class
+        ));
+    }
+
+    protected function get_taxonomy_class(WP_Taxonomy $taxonomy): string
+    {
+        /**
+         * Filters the class(es) used for different taxonomies.
+         *
+         * @since 2.6.0
+         * @example
+         * ```
+         * add_filter( 'timber/taxonomy/classmap', function( $classmap ) {
+         *     $custom_classmap = [
+         *         'genre' => GenreTaxonomy::class,
+         *     ];
+         *
+         *     return array_merge( $classmap, $custom_classmap );
+         * } );
+         * ```
+         *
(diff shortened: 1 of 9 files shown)


src/Factory/TaxonomyFactory.php
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    protected function build(WP_Taxonomy $taxonomy): CoreInterface
+    {
+        $class = $this->get_taxonomy_class($taxonomy);
+
+        return $class::build($taxonomy);
+    }
+
+    protected function is_numeric_array($arr)
```
nlemoine: We should be able to use https://www.php.net/manual/en/function.array-is-list.php?

src/Taxonomy.php
```diff
@@ -0,0 +1,270 @@
...
+    {
+        if (empty($query_args) && empty($options) && $this->terms_cache !== null) {
+            return $this->terms_cache;
+        }
+
+        $terms = Timber::get_terms(\array_merge($query_args, [
+            'taxonomy' => $this->name,
+        ]), $options);
+
+        if (empty($query_args) && empty($options)) {
```
nlemoine: WP already caches terms internally. And it's not invalidated, as opposed to core runtime (or persistent depending on your object cache config) cache.

src/Taxonomy.php
```diff
@@ -0,0 +1,270 @@
...
+     *     {{ post_type.labels.name }}
+     * {% endfor %}
+     * ```
+     *
+     * @return PostType[]
+     */
+    public function post_types(): array
+    {
+        return \array_map(
+            static fn ($post_type) => new PostType($post_type),
```
nlemoine, line 249: Not related to this PR but I checked `PostType`. If the post type is unknown or unregistered (which depends on call timing: usually available at `init`), the class still build the `PostType` object but empty, which might let you think you have a valid `PostType` whereas you're just getting an empty shell.

Separate issue but might worth using:
```php
return \array_map(
    static fn ($post_type) => new PostType($post_type),
    \array_values(\array_filter($this->wp_object?->object_type ?? [], 'post_type_exists'))
);
```

tests/TaxonomyTest.php
```diff
@@ -0,0 +1,366 @@
...
+        ]);
+
+        $this->assertEquals(['Blues', 'Ambient'], \array_map(strval(...), $terms));
+    }
+
+    public function testTermsIsMemoized()
+    {
+        $taxonomy = Timber::get_taxonomy('genre');
+
+        $this->assertSame($taxonomy->terms(), $taxonomy->terms());
```
nlemoine: The test asserts empty arrays `assertSame([], [])`

tests/TaxonomyTest.php
```diff
@@ -0,0 +1,366 @@
...
+    }
+
+    public function testTermsAreLazy()
+    {
+        $queries = 0;
+
+        $this->add_filter_temporarily('pre_get_terms', function () use (&$queries) {
```
nlemoine: Assert that `WP_Term_Query` isn't constructed, not that an actual SQL query didn't happen.

nlemoine (changes requested):
@Levdbas Thanks for working on this. I think a few changes are still needed. Happy to help if you need.

PR #3286, opened 2026-09-03
fix: prevent ToJpg filename collisions between same-basename images

si-kui-a:
Related:

- #2850 (same root cause, sibling `ToJpg` class — see below)

**Update (2026-09-09):** see #3285's update note — same history applies here (originally an unconditional rename, reworked to be filter-gated after @gchtr flagged both PRs as effectively the same shape as #2556). Sections below rewritten to match the current diff.

## Issue

While fixing #2850 for `ToWebp` (see #3285), I found the exact same bug pattern in the sibling `ToJpg` operation: `ToJpg::filename()` also ignores its `$src_extension` parameter, always returning `"$src_filename.jpg"`. Two source images sharing a basename but differing only in extension (e.g. `flag.png` vs `flag.gif`) resolve to the same destination path. `ImageHelper::_operate()`'s destination-already-exists cache-hit branch then returns the first-converted image's jpg content for every subsequent same-basename source instead of converting its own.

No separate GitHub issue exists for this one — it was found via code review while fixing #2850, not from an independent user report. Happy to have this linked/tracked however maintainers prefer.

## Solution

Applied the identical fix as #3285: fold the source extension into the generated filename (`flag-png.jpg`, `flag-gif.jpg`, ...), gated behind the same `timber/image/collision_safe_filenames` filter — off by default.

Two cases keep the bare name regardless of the filter, same as #3285:

- **Already-jpg source** — preserves the existing self-collision-as-no-op behavior, covered by the unmodified `testJPGtoJPG`.
- **Extensionless source** — matches the sibling `Resize::filename()`'s existing handling of a falsy `$src_extension` (see #3285's Solution section for the full reasoning); avoids producing a stray `name-.jpg`.

## Impact

Same as #3285: off by default, no forced regeneration for sites that don't opt in. Sites that enable the filter get the new naming for every non-jpg source going forward. Existing already-converted `name.jpg` files are untouched either way.

## Usage Changes

No change for sites that don't opt in. Sites that enable `timber/image/collision_safe_filenames`: generated `tojpg` filenames for non-jpg sources now include the source extension. No API/signature changes.

## Considerations

Same as #3285: same filter name (chosen before finding @gchtr's `timber/image/advanced_file_names` suggestion on #2556 — happy to rename), and same `-` separator choice for consistency between the two sibling operations.

## Testing

Updated the four existing filename-based assertions in `tests/Image/Operation/ToJpgTest.php` affected by the naming change, reverted for the default (filter-off) case, added a test pinning the default-off collision as intentional, and added `testCollidingBasenamesProduceDistinctJpgWhenFilterEnabled()`, mirroring `ToWebp`'s regression test. Added `testFilenameKeepsBareNameForExtensionlessSource()` mirroring #3285's, for the same reason (direct call, not through the full pipeline, since an extensionless source can't reach `wp_check_filetype`-based format detection in `ToJpg::run()`).

**Disclosure**: same update as #3285 — earlier revisions were hand-traced only; since then, actually ran this branch's tests in the same PHP 8.2 + GD(WebP) + Mantle/SQLite container. `ToJpgTest` alone: 9/9 passed, 23 assertions. Full `image` testsuite: 178 passed / 3 skipped (Imagick-only, not installed here) / 0 failed. CI will still run the full matrix on top of this.

Co-authored with Claude Code (Anthropic); reviewed by @si-kui-a before submission.

### DIFF
--- a/src/Image/Operation/ToJpg.php
+++ b/src/Image/Operation/ToJpg.php
@@ -22,13 +22,21 @@ class ToJpg extends ImageOperation
 
     /**
      * @param   string    $src_filename     the basename of the file (ex: my-awesome-pic)
-     * @param   string    $src_extension    ignored
-     * @return  string    the final filename to be used (ex: my-awesome-pic.jpg)
+     * @param   string    $src_extension    the source file's extension (ex: png), used to keep
+     *                                      pseudo-duplicate source files (ex: pic.png and
+     *                                      pic.gif) from colliding on the same output name
+     * @return  string    the final filename to be used (ex: my-awesome-pic-png.jpg)
      */
     public function filename($src_filename, $src_extension = 'jpg')
     {
-        $new_name = $src_filename . '.jpg';
-        return $new_name;
+        // A source that's already jpg keeps the bare name: ToJpg is then converting it to
+        // itself, and _operate()'s destination-already-exists check treats that as a no-op,
+        // which is the existing, desired behavior for already-jpg sources.
+        if ($src_extension === 'jpg') {
+            return $src_filename . '.jpg';
+        }
+
+        return $src_filename . '-' . $src_extension . '.jpg';
     }
 
     /**
--- a/src/Image/Operation/ToJpg.php
+++ b/src/Image/Operation/ToJpg.php
@@ -22,17 +22,44 @@ class ToJpg extends ImageOperation
 
     /**
      * @param   string    $src_filename     the basename of the file (ex: my-awesome-pic)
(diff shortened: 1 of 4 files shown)


gchtr:
This is a duplicate of https://github.com/timber/timber/pull/2556.

si-kui-a:
Same as #3285 (copying here since GitHub doesn't thread these together) — the version you saw here had the same #2556 shape. Reworked the same way; PR description above updated to match.

Would appreciate a re-look whenever convenient.

PR #3288, opened 2026-09-14
chore(cs): Apply codestyle fixes to make contributing easier

Levdbas:
## Description
Ran our codestyle tooling (ECS/Rector) across the codebase again to bring everything back in line with our current style rules.

Keeping the codebase consistently formatted reduces noise in future diffs and makes it easier for contributors to open clean PRs without unrelated style changes getting mixed in.

Also fixed the failing compile tests introduced by Twig 3.28 by normalizing null template data to an empty array before rendering.

## Changes

- Applied codestyle fixes across `src/` and `tests/`
- Normalized null data in `Timber::compile()` and `Timber::compile_twig_block()` for Twig 3.28 compatibility
- Verified `testCompileAbsolutePath`, `testCompileNull`, and `testCompileTwigBlockWithNullData` pass under PHP 8.3
- Fixed the `PHP coding standards` GitHub Action: ECS removed its `checkstyle` output format (it's a fixer, not a static analyzer), which broke the `cs2pr` piping and caused the workflow to fail with an XML parse error. The workflow now runs `ecs check` directly.


.github/workflows/php-coding-standards.yml
```diff
@@ -55,4 +53,4 @@ jobs:
         if: steps.changed-files.outputs.all_changed_files != ''
         env:
           CHANGED_FILES: ${{ steps.changed-files.outputs.all_changed_files }}
-        run: ./vendor/bin/ecs check --output-format=checkstyle $CHANGED_FILES | cs2pr
```
nlemoine, line 58: @Levdbas Why the cs2pr removal?

Levdbas: Leaving it there causes the following error:

`Run ./vendor/bin/ecs check $CHANGED_FILES | cs2pr
Error: Start tag expected, '<' not found on line 1, column 2
`

PR #3289, opened 2026-09-14
docs: Update docs about supported Timber versions and update Security Guidelines

gchtr:
It’s time to drop support for Timber v1 and reflect that in the relevant places.

This PR also updates the Security Guidelines to help developers and agents decide whether a security report qualifies as one we need to fix.


SECURITY.md
```diff
@@ -1,17 +1,58 @@
...
 We will make a best effort to meet the following response targets for security reports:
 
-- Time to first response (from report submit) - 5 business days
-- Time to triage (from report submit) - 10 business days
-- Time to fix (from triage) - 15 business days
+- Time to first response (from report submit) - 7 business days
+- Time to triage (from report submit) - 15 business days
+- Time to fix (from triage) - 30 business days
+
+If we close a report as out of scope, we will say which part of the scope section applies and why. You are welcome to push back if you think we got it wrong.
```
Levdbas, line 58:
```suggestion
If we determine that a security report is out of scope, we will clearly identify the applicable part of the scope and explain why the report falls outside it.

If you believe we have incorrectly classified a report as out of scope, we welcome you to provide additional context or challenge our assessment. We will review the information provided and reconsider our decision where appropriate.
```

Levdbas (changes requested):

SECURITY.md
nlemoine: Thanks @Levdbas!

Levdbas (approved):

PR #3290, opened 2026-09-14
fix: Use `XMLReader` to read SVG dimensions instead of `simplexml_load_file()`

gchtr:
## Issue

- Switching to `XMLReader` can improve performance, especially for bigger SVG files.
- On a libxml build where external entity loading is enabled, an attacker-influenced SVG — e.g. one passed to `Timber::get_external_image()` — could use an external entity to disclose local files or drive SSRF. On the supported PHP range (^8.2, libxml >= 2.9) entity substitution is off unless `LIBXML_NOENT` is passed, so the issue is latent rather than exploitable on a default build.

## Solution

- Replace the parse with an `XMLReader` pull parser that stops at the root element, using `LIBXML_NONET | LIBXML_NOERROR | LIBXML_NOWARNING` and never passing `LIBXML_NOENT` or `LIBXML_DTDLOAD`. External entities are therefore neither substituted nor loaded, and network fetches are blocked.
- Add `ext-xmlreader` to composer.json require

## Impact

Reading only the root element also means the body of the document is never parsed or built into a tree. On a 9.6 MB SVG, reading the dimensions drops from ~61 ms to ~0.3 ms in tests.

## Usage Changes

- A root element that is not an `<svg>` now yields no dimensions. Previously
  any XML root carrying `width`/`height` was sized as if it were an SVG.
- Malformed SVGs no longer emit PHP warnings containing the full server path.
- A truncated or corrupt body no longer prevents the dimensions being read,
  provided the corruption falls beyond the parser’s first input buffer.

Namespace-prefixed roots (`<svg:svg>`) are matched on local name, so they keep
working.

## Considerations

`¯\_(ツ)_/¯`

## Testing

Included.


Levdbas (approved):

nlemoine (approved):
Thanks @gchtr for the PR! I added a couple of tests and made the dimensions parsing a bit more tolerant (case, comma separated values in `viewBox`).

I'll let you review my changes if you want :)

